PR #8, opened 2020-01-08
changes NSH command line editing so ctrl-K deletes from char pos to EOL like GNU readline

adamfeuer:
### Summary

- changes NSH command line editing so ctrl-K deletes from char pos to EOL like GNU readline
- now you can delete entire line with `ctrl-A` `ctrl-K`

### Impact

- None if `CONFIG_NSH_CLE` is not set
- if `CONFIG_NSH_CLE=y` is set, `ctrl-K` deletes from cursor position to end of line, instead of the old behavior, cursor position + 1 to end of line. 
- The new behavior is more like GNU Readline

### Testing

- Manual, on the SAMA5D36-Xplained board.

### How To Verify (beyond provided automated tests)

- `CONFIG_NSH_CLE=y`in `.config`
- make clean; make
- connect to the NSH via a serial console
- You should see the new command line editing behavior

### DIFF
--- a/system/cle/cle.c
+++ b/system/cle/cle.c
@@ -1076,7 +1076,7 @@ static int cle_editloop(FAR struct cle_s *priv)
 
         case KEY_DELEOL:  /* Delete to the end of the line */
           {
-            priv->nchars = (priv->nchars > 0 ? priv->curpos + 1 : 0);
+            priv->nchars = (priv->nchars > 0 ? priv->curpos : 0);
           }
           break;
 


patacongo:
Doesn't this behavior depend on the nature of the cursor?  If it is a vertical bar "between" characters wouldn't that be different if it is a reversed character, for example?  The reversed character is at the current position.  You would think that the reversed character should not be deleted, i.e., that the current position + 1 to EOL should be deleted.  The real Bash verical bar appears to lie at the left side of the current position, so this behavior is correct, you would want do delete the current character and to the EOL.  I 'm not sure, but it does seem that should be different with different cursors.

adamfeuer:
@patacongo Thanks for merging this. I didn't think about the bar cursor case, but the behavior with a bar cursor using bash seems the same... I'm on a Mac, running XTerm2, ssh'd into a VMWare Ubuntu 19.10 box. I switched the cursor to a bar using this:

`echo -e -n "\x1b[\x36 q" # changes to steady bar`

`ctrl-A` puts the bar at the beginning of the line, before the first character, so it works the same way bash does with an inverse box cursor... I tested in bash and in nsh from picocom.

PR #9, opened 2020-01-09
fix passwd.h:103:30: warning: passwd_verify with no effect [-Wunused-…

xiaoxiang781216:
…value]


jerpelea:
please check 
https://github.com/apache/incubator-nuttx-apps/blob/master/fsutils/passwd/passwd_lock.c#L86

PR #14, opened 2020-01-14
.asf.yaml:  Enable Issues in the apps/ repository.

patacongo:


btashton:
Do we really want two issue trackers? Splitting the project management board across two repos seems even more of an issue. What about the testing repo will it also have an issue tracker? I would be much more in favor of having a notice in the description that they should file issues against the os project.

patacongo:
Yes, most definitely.  I do NOT want security violations against crappy little applications in apps/ being reported against the OS.  The is a disservice to the years of effort that have but put into it.  A badly done afternoon's work develop a bad application MUST not reflect on the core OS.  We cannot have that.

It is very important (at least to me) that we separate the issues base upon what they are reported against.

patacongo:
We will also need to eventually enable issues in incubator-nuttx-testing and incubator-nuttx-website too.

btashton:
I would have liked to just have a single issue tracker and just attach labels for OS/arm OS/net Website etc... But looks like we are not doing that. What about the project feature can we at least have one of those? I cannot imagine being able to use that effectively in 5 different places. You can link issues from different repos to a single project board.

patacongo:
I really, really, really dislike that idea.  I do not want apps/ issues in the OS.  That is a really repulsive concept to me and I have to resist that idea with all means available to me.

justinmclean:
The PPMC as a whole should discuss this and come to some consensus on it. I suggest you start a thread on the mailing list.

patacongo:
There already is a thread.  But is never came to a consensus.  Some wanted to use Jira, some wanted to use Github issues.

I don't really care, but I don't want all issues reported against the OS.  That is wrong.  Especially after this BS discussion about security in a pre-Apache code release.  I do not want those reports associated with the core OS.  I am really not comfortable with that at all.

patacongo:
"There already is a thread."  There were the threads called "Bug Tracking" started on January 3 and again as "Issue Reporting" sometime before January 9

justinmclean:
I'm aware of those threads. Why is action being taken if consensus hasn't been reached? Consensus sometimes can be hard and takes time, but I don't think there's a huge rush here. How many app issues have been raised to date in the main repo? I'd also suggest the PPMC focus on dealing with security issues as per the documented process.

patacongo:
Let's keep everything as it is now so that no application issues are reported against the OS in the mean time.

I am not the right person to deal with the security nonsense.  I don't have sufficient patience for that.

PR #18, opened 2020-01-16
examples/sx127x_demo/sx127x_demo.c: RX/TX only if RX/TX enabled

raiden00pl:


examples/sx127x_demo/sx127x_demo.c
```diff
@@ -541,7 +544,9 @@ int main(int argc, FAR char *argv[])
 
               break;
             }
+#endif
```
jerpelea: you can use a single ifdef block

PR #24, opened 2020-01-20
Add install dependency on chardev BIN to fix parallel build break

liuguo09:
chardev bin may not availabe before make install in parallel build,
so add install dependency on chardev BIN here.

Change-Id: If28451ceeeed0a6463544d8c342871cecda5a057


jerpelea:
@masayuki2009

PR #45, opened 2020-02-03
Improve telnetd

xiaoxiang781216:


netutils/telnetd/telnetd_daemon.c
```diff
@@ -183,14 +181,12 @@ static int telnetd_daemon(int argc, FAR char *argv[])
 
   /* Set socket to reuse address */
 
-#ifdef CONFIG_NET_HAVE_REUSEADDR
```
jerpelea: the same config option appears also in 
netlib_listenon.c
ftpd.c

patacongo: Those are just hooks. The NuttX stack does not support SO_REUSEADDR.  So checking if an undefine-able configuration is defined is equivalent to #if 0

I am not sure why it is removed here.  SO_REUSEADDR is not implemented.  Other than adding some code size, I suppose it does not matter.

xiaoxiang781216: @jerpelea will add new patch to remove them.
@patacongo we may plan to implement SO_REUSEADDR since the code size increasement is small, we mayn't add option to complicate the user.

jerpelea: @xiaoxiang781216 thanks

patacongo: @xiaoxiang781216 I don't have any problem with small increases in code size.  I think these kinds of changes are the natural consequence of the NuttX roadmap.. which is to become a fully viable, fully compatable POSIX OS.

We do have to be careful that the code side does not grow out of control.  Perhaps we should extend the build test to monitor the size of each build so that we can see changes over time.  Just add 'size nuttx' after each build.

We must be balanced.  We cannot permit excessive bloat to the OS while we should support modest growth for better standardization and more world class features

A consequence, I believe, is that eventually NuttX will not be able to run on some of the very tiny platforms.  That will impact a few people.  But I would not recommend NuttX to people using hardware with very minimum memory configurations.  I don't see a way to support both preferences.

PR #56, opened 2020-02-07
Ensure all source code end with one and only one newline

xiaoxiang781216:
by this command:
git ls-files -z | while IFS= read -rd '' f; do tail -c1 < "$f" | read -r _ || echo >> "$f"; done


davids5:
@xiaoxiang781216 is this safe on all types of files?

examples/cromfs/cromfs/testdir2/TheThreeLittlePigs.txt
```diff
@@ -14,4 +14,4 @@ The little pig saw the wolf climb up on the roof and lit a roaring fire in the f
```
patacongo: This is just a test file, but in this case, the trailing newline is for visual spacing since the story of the 3 little pigs is displayed.  It really needs the trailing newline since all of the text is double spaced and without the final newline, the text will run together on the display.

examples/elf/tests/helloxx/.gitignore
```diff
@@ -2,4 +2,4 @@ hello++1
 hello++2
 hello++3
 hello++4
-hello++5
\ No newline at end of file
+hello++5
```
patacongo: It is really impossible to see.  Is there a newline after hello++5 or not?  Where there two before and now only one?  Or is there no newline now.  That would be bad.

patacongo:
"is this safe on all types of files?"  It is inappropriate on the .txt file that needs the newline for visual spacing.

Other source files will depend on if there is 0 or 1 newlines at the end.  Looking at the diffs, I cannot tell if there is 0 or 1 newlines at the end now.

I would think that most apps should work with a single newline.  Many applications are even tolerant of missing newlines at the end altogether.  kconfig-frontends is not.  Kconfig files with no newline at the end causes errors.

examples/elf/tests/helloxx/.gitignore
xiaoxiang781216: No, you can see there is a red dot at the end of line before change which mean no newline there:
image

xiaoxiang781216:
> "is this safe on all types of files?" It is inappropriate on the .txt file that needs the newline for visual spacing.
> 

Ok, I will remove the test file from PR.

> Other source files will depend on if there is 0 or 1 newlines at the end. Looking at the diffs, I cannot tell if there is 0 or 1 newlines at the end now.
> 

The command can ensure this:
git ls-files -z | while IFS= read -rd '' f; do tail -c1 < "$f" | read -r _ || echo >> "$f"; done
Here has the detailed explanation for each step:
https://unix.stackexchange.com/questions/31947/how-to-add-a-newline-to-the-end-of-a-file
Actually, it's good thing to ensure one and only one newline at the end of file for consistency and avoid git complain "No newline at end of file":
image


> I would think that most apps should work with a single newline. Many applications are even tolerant of missing newlines at the end altogether. kconfig-frontends is not. Kconfig files with no newline at the end causes errors.

xiaoxiang781216:
> @xiaoxiang781216 is this safe on all types of files?

@davids5 I remove all binary file from the commit manually after running the command. If some text file isn't suitable to have the new line at the end of file, I can remove it from the commit.

xiaoxiang781216:
@patacongo should we reconsider this patch again? the consistence is most important thing for coding sandard:
1.Ensure all source code end with one and only one newline
2.Enusre all source code don't end with any newline
Mixing both is the most bad thing.

davids5:
Have you considered just enabling the pre commit hook that comes with git that checks for the sort of thing?

xiaoxiang781216:
> Have you considered just enabling the pre commit hook that comes with git that checks for the sort of thing?

Yes, we should improve the tool. Before that we have do this manually, but it's better than we do nothing. The only important thing for coding standard is consistent.

xiaoxiang781216:
@patacongo since the samiliar patch is alrady merged in nuttx repo:
https://github.com/apache/incubator-nuttx/pull/226
Could you reopen this PR, so I can remove TheThreeLittlePigs.txt from patch?

PR #62, opened 2020-02-10
examples: elf: Fix warnings when compiled for RISC-V 64bit

masayuki2009:
### Summary

- Fix warnings such as "warning: cast from pointer to integer of different size"

### Impact

- This PR only affects apps/examples/elf.

### Testing

- Tested with Sony Spresense (armv7-m 32bit) and K210 (RISC-V RV64GC)

### DIFF
--- a/examples/elf/tests/pthread/pthread.c
+++ b/examples/elf/tests/pthread/pthread.c
@@ -82,16 +82,16 @@ enum exit_values_e
 
 void *child_start_routine(void *arg)
 {
-  printf("CHILD: started with arg=%d\n", (int)arg);
+  printf("CHILD: started with arg=%d\n", (intptr_t)arg);
 
   if (arg != CHILD_ARG)
     {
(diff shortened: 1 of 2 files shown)


examples/elf/tests/pthread/pthread.c
```diff
@@ -82,16 +82,16 @@ enum exit_values_e
 
 void *child_start_routine(void *arg)
 {
-  printf("CHILD: started with arg=%d\n", (int)arg);
+  printf("CHILD: started with arg=%d\n", (intptr_t)arg);
```
patacongo: This will cause warnings in some environments where there is type checking (like sim).  It can also cause parateter mismatches in other cases.

For example, on 64-bit simulation, %d will expect an int argument and sizeof(int) == 32-bits.  But intprt_t will be 64-bits so the printf will be wrong and, in certain cases, can lead to a crash.

My suggesting is to use the more complex cast (int)((intptr_t)arg).  That will both avoid the warning and assure that the value provided to printf is a proper integer type.

PR #64, opened 2020-02-10
apps/examples/tcpblaster:  Fix compile errors in loopback mode.

patacongo:


adamfeuer:
@patacongo This works for me.

PR #66, opened 2020-02-13
Remove unused BUILTIN_PROXY_STACKSIZE config

yamt:
It has been unused since the following change:
"All argv types should be char * const * not const char **"


xiaoxiang781216:
@yamt could you provide a patch for nuttx to remove all CONFIG_BUILTIN_PROXY_STACKSIZE in nuttx repo?

yamt:
@xiaoxiang781216
is there an easy way to update a bunch of defconfigs?
i mean easier than awk/sed/vi/etc

xiaoxiang781216:
run this command:
tools/refresh.sh --silent all
but some windows specific config need to bypass.

yamt:
thank you. i'll try it later.

yamt:
done https://github.com/apache/incubator-nuttx/pull/267

PR #69, opened 2020-02-14
mksymtab.sh: Avoid using find -executable

yamt:
As it seems GNU-find only.  macOS doesn't have it.

-perm is in POSIX and should be more widely available.

### DIFF
--- a/examples/elf/tests/mksymtab.sh
+++ b/examples/elf/tests/mksymtab.sh
@@ -22,7 +22,7 @@ fi
 # Extract all of the undefined symbols from the ELF files and create a
 # list of sorted, unique undefined variable names.
 
-varlist=`find ${dir} -executable -type f | xargs nm | fgrep ' U ' | sed -e "s/^[ ]*//g" | cut -d' ' -f2 | sort | uniq`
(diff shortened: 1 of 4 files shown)


examples/elf/tests/mksymtab.sh
```diff
@@ -22,7 +22,7 @@ fi
 # Extract all of the undefined symbols from the ELF files and create a
 # list of sorted, unique undefined variable names.
 
-varlist=`find ${dir} -executable -type f | xargs nm | fgrep ' U ' | sed -e "s/^[ ]*//g" | cut -d' ' -f2 | sort | uniq`
```
xiaoxiang781216: Can we merge the similar script into one:
apps/tools/mksymtab.sh

yamt: maybe we can. but i'm not going to do it today.
it seems each of them generates slightly different code to avoid conflicting symbols.

xiaoxiang781216: Maybe we can modify the code to use the same name, I don't see any benefit that each project have the different symbol table name.

patacongo: I am merging this.  I presume that the suggested enhancements would be a different PR.

PR #75, opened 2020-02-15
testing/fstest/fstest_main.c: Eliminate a warning found during the build

Ouss4:
test.

### DIFF
--- a/testing/fstest/fstest_main.c
+++ b/testing/fstest/fstest_main.c
@@ -1084,7 +1084,8 @@ int main(int argc, FAR char *argv[])
 
       /* Perform garbage collection, integrity checks */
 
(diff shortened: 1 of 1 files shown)


testing/fstest/fstest_main.c
```diff
@@ -1084,7 +1084,7 @@ int main(int argc, FAR char *argv[])
 
       /* Perform garbage collection, integrity checks */
 
-      fstest_gc(buf.f_bfree);
+      (void)fstest_gc(buf.f_bfree);
```
patacongo: The use of (void) to ignore return values was recently removed in some recent PRs.  How do we know that this void will not be removed in the future, re-introducing the warning.   An alternative would be to make ftest_gc return void instead of int.  The return value is never used.

PR #79, opened 2020-02-20
build: Use flock to serialize ar steps as workaround

xiaoxiang781216:
Use flock to serialize ar steps to avoid parallel build break
sometimes.

Change-Id: I3a916b3b27422cdb6d718f7d554361565457fa08

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.arlock
 .depend
 Make.dep
 *.o
--- a/Application.mk
+++ b/Application.mk
@@ -130,9 +130,9 @@ $(CXXOBJS): %$(SUFFIX)$(OBJEXT): %$(CXXEXT)
 
 .built: $(OBJS)
 ifeq ($(WINTOOL),y)
-	$(call ARCHIVE, "${shell cygpath -w $(BIN)}", $(OBJS))
+	$(call ARLOCK, "${shell cygpath -w $(BIN)}", $(OBJS))
 else
-	$(call ARCHIVE, $(BIN), $(OBJS))
+	$(call ARLOCK, $(BIN), $(OBJS))
 endif
 	$(Q) touch $@
 
--- a/Make.defs
+++ b/Make.defs
@@ -52,7 +52,7 @@ CLEANDIRS := $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Makefile))
 CONFIGURED_APPS :=
 
 define Add_Application
-  include $(1)Make.defs
+	include $(1)Make.defs
 endef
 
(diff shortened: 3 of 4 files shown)


xiaoxiang781216:
A general version of:
https://github.com/apache/incubator-nuttx/pull/102
Before we get a better solution, this patch can fix the parallel build which is very important for the upcoming Apache Jenkins and Github Action.

xiaoxiang781216:
@masayuki2009 please associate your github account with apache account:
https://whimsy.apache.org/roster/ppmc/nuttx
so I can add you to reviewer list.

masayuki2009:
> @masayuki2009 please associate your github account with apache account:
> https://whimsy.apache.org/roster/ppmc/nuttx
> so I can add you to reviewer list.

@xiaoxiang781216 I'll setup it later.

masayuki2009:
@xiaoxiang781216 I've just added my github account info to the apache.

yamt:
macOS doesn't have flock.
how about using shlock instead? macOS has it.

xiaoxiang781216:
> macOS doesn't have flock.
> how about using shlock instead? macOS has it.

But shlock doesn't exist on cygwin and msys2, even Ubuntu need install manually with apt-get install:(.

xiaoxiang781216:
How about https://github.com/discoteq/flock:
brew tap discoteq/discoteq
brew install flock

masayuki2009:
@xiaoxiang781216 It looks good to me.

patacongo:
> 
> 
> How about https://github.com/discoteq/flock:
> brew tap discoteq/discoteq
> brew install flock

This would need to be documented so macOS users would have to install flock?  I will merge this PR when the documentation updates are in place.  This should be covered in the top-level README.txt file.

xiaoxiang781216:
@patacongo here is the patch updating README.txt:
https://github.com/apache/incubator-nuttx/pull/337

masayuki2009:
Hi, @xiaoxiang781216,

I found that this solution still has a problem.
Because .arlock files are created in each application directory.

I think the lock file should be in the same directory as '*.a' file, 
because we want to protect '*.a' file access.

Or as I tried before, the lock file could be at the top directory. (rather giant lock)

xiaoxiang781216:
> Hi, @xiaoxiang781216,
> 
> I found that this solution still has a problem.
> Because .arlock files are created in each application directory.
> 
> I think the lock file should be in the same directory as '_.a' file, because we want to protect '_.a' file access.
> 
> Or as I tried before, the lock file could be at the top directory. (rather giant lock)

Yes, I will send a new PR fix this issue.

xiaoxiang781216:
@masayuki2009 see this:
https://github.com/apache/incubator-nuttx-apps/pull/80

PR #81, opened 2020-02-21
do ARLOCK things a bit differently

yamt:


yamt:
i only tested on macOS

Make.defs
```diff
@@ -99,9 +99,13 @@ define REGISTER
...
 
-define ARLOCK
-	$(Q) flock .arlock $(call ARCHIVE, $1, $(2))
-endef
+# Workaround for concurrent execution to the same library (libapps)
+ifeq ($(CONFIG_HOST_MACOS),y)
+# macOS doesn't have flock. Use shlock instead.
+AR := $(APPDIR)/tools/flock.sh $(APPDIR)/.arlock $(AR)
+else
+AR := flock $(APPDIR)/.arlock $(AR)
```
xiaoxiang781216: But I think this change still hit "@ can't find", I mention here:
https://github.com/apache/incubator-nuttx/pull/342
since the symbol @ in side the original AR still pass to bash, but I can't understand why macOS no problem.

yamt: AR doesn't have @. ARCHIVE does.

xiaoxiang781216: But we can't bypass ARCHIVE and invoke AR directly, you can see there are mutiple ARCHIVE defintion in nuttx to fight or workaround the toolchain difference.

yamt: i checked some of ARCHIVE but they all seem calling AR one way or another.
you are right this doesn't work if ARCHIVE does something completely different.
i suspect it isn't the case though.

xiaoxiang781216: Here is the case:
https://github.com/xiaoxiang781216/nuttx/blob/song-u1/arch/ceva/src/song/Toolchain.defs
You can see to support CEVA toolchain we have to do thing hard.

yamt: ok, i'm now convinced. thank you for the pointer.

PR #85, opened 2020-02-21
Fix usrsocktest errors

masayuki2009:
### Summary

- This PR fixes a part of usrsocktest errors. (i.e. errors in NoBlockRecv and BlockRecv)
- Actually, the test cases checked sockaddr_in values which contain padding field, however, the field should be ignored because it contains random values.

### Impact

- This PR affects a usrsocktest application only.

### Limitations / TODO

- There still remain errors and need more investigation.

### Testing

- I tested this PR with spresense:wifi which I added usrsocktest application.

### DIFF
--- a/examples/usrsocktest/usrsocktest_block_recv.c
+++ b/examples/usrsocktest/usrsocktest_block_recv.c
@@ -265,7 +265,8 @@ static void NoBlockConnect(FAR struct usrsocktest_daemon_conf_s *dconf)
   TEST_ASSERT_EQUAL(1, ret);
   TEST_ASSERT_EQUAL_UINT8_ARRAY("a", data, 1);
   TEST_ASSERT_EQUAL(sizeof(remoteaddr), addrlen);
-  TEST_ASSERT_EQUAL_UINT8_ARRAY(&remoteaddr, &addr, addrlen);
+  TEST_ASSERT_EQUAL_UINT8_ARRAY(&remoteaddr, &addr,
+                                addrlen - sizeof(addr.sin_zero));
   TEST_ASSERT_EQUAL(1, usrsocktest_daemon_get_num_connected_sockets());
   TEST_ASSERT_EQUAL(1, usrsocktest_daemon_get_num_active_sockets());
   TEST_ASSERT_EQUAL(6, usrsocktest_daemon_get_recv_bytes());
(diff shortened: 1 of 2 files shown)


xiaoxiang781216:
@masayuki2009 how about remove Make.defs patch from this PR? so I can merge it.

masayuki2009:
@xiaoxiang781216 sorry, I added the file by mistake.

masayuki2009:
@xiaoxiang781216 Sorry, I've just pushed with -f

xiaoxiang781216:
@masayuki2009 no problem, I have merged it.

lukegluke:
It fix fails in NoBlockRecv and BlockRecv, yes, thanks!

By the way, when I was investigating this I encountered on this sentence in [stackoverflow](https://stackoverflow.com/questions/15608707/why-is-zero-padding-needed-in-sockaddr-in):

> Its required by specification to clear sin_zero

I'm too far from an expert in net stack, and don't know about any specification, but maybe (just a thought) it would be more correct to nulling padding somewhere inside the net stack?

masayuki2009:
@lukegluke 

>By the way, when I was investigating this I encountered on this sentence in stackoverflow:

Thanks for the comment. I'll check the article.

xiaoxiang781216:
But I think the requirement is the application should zero out the input addr before pass to socket API to avoid the undefined behaviour? addr is an output argument in our case.

lukegluke:
> But I think the requirement is the application should zero out the input addr before pass to socket API to avoid the undefined behaviour? addr is an output argument in our case.

It was the first thing I did trying fix it - clear input addr in usrsocktest before pass to socket API, but netstack insert random values in sin_zero itself anyway.

masayuki2009:
@lukegluke 

>It was the first thing I did trying fix it - clear input addr in usrsocktest before pass to socket API, >but netstack insert random values in sin_zero itself anyway.

Actually recvfrom_request() in usrsocktest_daemon.c sets the return values,
so we need to modify the above function as well, if we do not modify
network stack code.

masayuki2009:
@lukegluke 

>I'm too far from an expert in net stack, and don't know about any specification, 
>but maybe (just a thought) it would be more correct to nulling padding somewhere 
>inside the net stack?

As far as I checked  the Linux kernel source code, you are right.
I think zero padding should be implemented in net/socket/recvfrom.c,
if the from address family is AF_INET and fromlen is larger than sockaddr_in size.

Anyway, I'll try to implement it and then should be reviewed by @patacongo.

xiaoxiang781216:
If so other similar function accept, getsockname getpeername... need the similar change too

masayuki2009:
@xiaoxiang781216 

>If so other similar function accept, getsockname getpeername... need the similar change too

Thanks for the comments. I understand.

masayuki2009:
I've just sent a new PR. https://github.com/apache/incubator-nuttx/pull/370

PR #87, opened 2020-02-22
build: Use flock to serialize ar steps as workaround

xiaoxiang781216:
Use flock to serialize ar steps to avoid parallel build break
sometimes.

Change-Id: I3a916b3b27422cdb6d718f7d554361565457fa08


xiaoxiang781216:
We need apply this patch again since we revert this patch:
#86

PR #95, opened 2020-02-25
Avoid calling ARCHIVE in multiple places (again)

yamt:


Application.mk
```diff
@@ -163,6 +157,9 @@ endif
 
 install:: $(PROGLIST)
 
+show-objs:
```
xiaoxiang781216: need add object file into libapps for module case

.gitignore
```diff
@@ -15,8 +15,8 @@ cscope.out
 /external
 /.context
 /.config
+/.depend
```
xiaoxiang781216: dup to line 1

Directory.mk
```diff
@@ -54,6 +54,7 @@ $(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),context)))
 $(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),depend)))
 $(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),clean)))
 $(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),distclean)))
+$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),show-objs)))
```
xiaoxiang781216: change to showobjs?

Makefile
```diff
@@ -55,6 +55,11 @@ endif
 SYMTABSRC = symtab_apps.c
 SYMTABOBJ = $(SYMTABSRC:.c=$(OBJEXT))
 
+# Turn off "Entering directory" messages as they mess up the output of
+# show-objs target.
+MAKEFLAGS += --no-print-directory
```
xiaoxiang781216: show we add --no-print-directory to line 75 and remove this line to avoid the change to th default behaviour.

.gitignore
yamt: oops.
it's one of reasons one should not make changes without mentioning it in the commit message. :-)

Application.mk
yamt: maybe i misunderstood how module thing works.
only MAINOBJ turns into a binary and others (OBJS) go into libapps?
how can it work with KERNEL build?

xiaoxiang781216: for module build, this is two step process:
1.All source files(except main) compile and archieve into libapps like builtin.
2.Main soruce file compile and link with LDLIBS to form the final image.
The difference from kernel mode and other mode is LDLIBS content:
1.Kernel mode: LDLIBS has all userspace library(libc, libproxy, libapps...), so the binary is totally self contain.
2.Other mode: LDLIBS just contain libapps, so the binary will link with the base image to resovle the symbol at in runtime.
But, I see your patch remove libapps from LDLIBS:
https://github.com/apache/incubator-nuttx-apps/pull/83/files

yamt: thank you for explanation.

yamt: are there some sort of circular dependency?
libapps -> SYMTABSRC -> install -> linking with LDLIBS, including libapps

xiaoxiang781216: The right sequence is:
1.Compile all source files(except main) and archieve into libapps.a
  Note: this is done step by step, with this patch we do it in one step
2.Compile main source file and link with libapps.a to get the executable
  Note:
  a.For non kernel build, the executable still contian the unresolved symbols outside libapps.a(.e.g. printf, open...).
  b.For kernel build, since LDLIBS contain all userspace library, the executable is totally self contained.
  c.NuttX support the explicit dynamic loading, but not implicit dynamic loading now, which mean that the executable must be self contained in kernel mode, but it can call dlopen to load the shared library manully.
3.Generate symbols source file from all exectuable unresolved symbols
Note: For kernel build, symbol table should be empty as note 2.b.
4.Compile and archieve symbols source file into libapps.a
Note: this is required for two reason:
a.binfmt need the symbol table to load the exectutable at runtime
b.symbol table ensure the linker don't remove these unresolved symbols
5.libapps.a link with other libarary to get nuttx.bin or userspace.bin
Note: this just happen for flat and protected build, all exectuable need put into the file system and load through binfmt subsytem in kernel build, no big image blob(nuttx.bin or userspace.bin) exist at all.

yamt: so, flock in $(APPDIR)/Makefile (step 4) should not be necessary because there should not be concurrent ARCHIVE activities (step 1) at that point?

xiaoxiang781216: Maybe, but it isn't harm to do the lock like other place, and make the code more consistent.

xiaoxiang781216:
@yamt do you plan to update this PR per comment?

yamt:
@xiaoxiang781216 
no for now.
i guess i will revisit this after i learn a bit more about KERNEL build.

xiaoxiang781216:
Ok, let wait your result.

v01d:
@yamt I believe I actually addressed this in a separate PR #380 
If you think it does, we can close this.

xiaoxiang781216:
@yamt I will close this PR, let's focus on PR #380.

PR #101, opened 2020-02-26
Make sure examples/elf testbuild pass in parallel build

liuguo09:


examples/elf/tests/errno/Makefile
```diff
@@ -86,6 +86,6 @@ clean:
 	$(call DELFILE, $(BIN))
 	$(call CLEAN)
 
-install:
+install: $(BIN)
```
yamt: if we do this at this level, we should remove "build" step from examples/elf/tests/Makefile.

an alternative: https://github.com/apache/incubator-nuttx-apps/pull/103

examples/elf/Makefile
```diff
@@ -57,6 +57,8 @@ else
   FSIMG_HDR = $(TESTS_DIR)/cromfs.h
 endif
 
+elf_main.c: tests/symtab.c
```
yamt: i guess it's more straightforward to require $(FSIMG_HDR) instead.

examples/elf/tests/errno/Makefile
liuguo09: You are right. I think your approach is simple and more general.

examples/elf/Makefile
xiaoxiang781216: But, who do trigger thetest folder build? we have to duplicated the test/symtab.c action here, but this change will trigger test folder build twice.

yamt: i'm not sure what you mean.
symtab.c is in CSRCS. (L40)
it triggers the test folder build. (with VPATH)
tests/symtab.c and $(FSIMG_HDR) are built with the same rule below, L62-63.

xiaoxiang781216: Yes, if we let elf_main.c depends on tests/symtab.c, then make know that it should build symtab.c then test before elf_main.c, but if we change elf_main.c depends on $(FSIMG_HDR), but without any action, then how make ensure the header generated before it start to compile elf_main.c?

yamt: * symtab.c and $(FSIMG_HDR) are built with the same rule (L62-63)
* elf_main.c (well, strictly speaking elf_main.o) depends on $(FSIMG_HDR) and dirlist.h, not symtab.c
* usually Make.dep will know elf_main.o depends on tests/dirlist.h (it was the reason for 45bcfa8b2eb1b0a84214a2bfc23f555ad8cf135e)

i don't understand why you prefer to make this depend on symtab.c over those headers.

xiaoxiang781216: > * symtab.c and $(FSIMG_HDR) are built with the same rule (L62-63)

Yes.

> * elf_main.c (well, strictly speaking elf_main.o) depends on $(FSIMG_HDR) and dirlist.h, not symtab.c

Yes, that is why is it better to let elf_main.c depends on $(FSIMG_HDR) and dirlist.h.
 
> * usually Make.dep will know elf_main.o depends on tests/dirlist.h (it was the reason for [45bcfa8](https://github.com/apache/incubator-nuttx-apps/commit/45bcfa8b2eb1b0a84214a2bfc23f555ad8cf135e))
> 

Yes, but $(FSIMG_HDR) and dirlist.h is generated in another Makefile, there isn't these files in the first time invoke. If elf_main.c has no relationship with symtab.c, elf_main.c may compile first and than can't find these header file.

> i don't understand why you prefer to make this depend on symtab.c over those headers.

xiaoxiang781216: @yamt we remove the first patch and use your solution, do you have any concern about this patch?

PR #111, opened 2020-02-28
apps: rename Makefile/Make.defs/Make.dep to Nuttx.mk/Nuttx.defs/Nuttx.dep

anchao:
Rename "Makefile" to "Nuttx.mk" to ensure the uniqueness
of "Makefile" in different projects, which is meaningful
to avoid frequent overwriting of "Makefile" if we want
to use the same project to compile in different platforms.
which also makes "Nuttx.mk" easier to join other open source
projects (like Android.mk) to avoid unnecessary porting work.

Rename:

Makefile   ->  Nuttx.mk
Make.defs  ->  Nuttx.defs
Make.dep   ->  Nuttx.dep


patacongo:
See pr399

PR #113, opened 2020-03-05
Initial checkin nuttx-apps github action CI workflow

liuguo09:
Github action CI workflow steps as below:
1. Use docker container with build essential tools preinstalled
2. nxstyle check pull request with checkpatch.sh
3. Call testing cibuild.sh to do jobs matrix check builds

Now split full testlist into sim/mips-riscv-x86/arm01~arm15 totally 17 jobs to do parallel builds.
All the jobs could be finished in half an hour at most. We could reduce the build time later by refine the configs in each job.

Also note that the pull request check is by default whole file check.

For example, refer to the check build in my own fork:
https://github.com/liuguo09/incubator-nuttx-apps/pull/3/checks

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,83 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
LGTM.

xiaoxiang781216:
The similar change for nuttx side:
https://github.com/apache/incubator-nuttx/pull/261

jerpelea:
+1

PR #119, opened 2020-03-12
Update workflow to let build jobs do not depend on check nxstyle job

liuguo09:
As discussion in https://github.com/apache/incubator-nuttx/pull/549 shows,
update github action workflow to let build jobs do not depend on nxstyle
check job in order to make sure each PR (including some nxstyle waiving cases)
be built at least.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,6 @@ jobs:
         ../nuttx/tools/checkpatch.sh -g $commits
 
   build:
-    needs: check
     runs-on: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@yamt another one.

yamt:
lgtm

davids5:
Is the style check still run but the result ignored, or is it not run?

liuguo09:
Style check job still runs and the result isn't ignored. But the build jobs behavior is different. In the former workflow, if style check failed, build jobs wouldn't run. Now the build jobs do not depend on style check job result in the current workflow. Build jobs would run anyway.

PR #118, opened 2020-03-11
Fix parallel buid errors in examples/elf

masayuki2009:
### Summary

- This PR fixes build errors in examples/elf in parallel build.
- I applied the same logic in examples/posix_spawn/Makefile which adds an explicit context:: target.
- Also, I modified tests/Makefile to avoid compiling incomplete symtab.c

### Impact

- This PR affects examples/elf only.

### Testing

- I tested this PR for maix-bit:elf configuration with -j1 to -j24 options.

PR #120, opened 2020-03-13
examples/posix_spawn: Fix parallel build errors at link time

liuguo09:
During generating symtab.c, compile seqeuence would start with
incomplete symtab.c. This change will fix such a situation.

### DIFF
--- a/examples/posix_spawn/filesystem/Makefile
+++ b/examples/posix_spawn/filesystem/Makefile
@@ -78,7 +78,8 @@ $(ROMFS_HDR) : $(ROMFS_IMG)
 # Create the exported symbol table
 
 $(SYMTAB_SRC): $(ROMFS_IMG)
-	$(Q) $(FILESYSTEM_DIR)$(DELIM)mksymtab.sh $(ROMFS_DIR) >$@
+	$(Q) $(FILESYSTEM_DIR)$(DELIM)mksymtab.sh $(ROMFS_DIR) >$@_tmp
+	$(Q) mv $@_tmp $@
 
 # Clean each subdirectory
 


liuguo09:
@masayuki2009 I apply your elf change to posix_spawn since same problem occurred.

masayuki2009:
> @masayuki2009 I apply your elf change to posix_spawn since same problem occurred.

@liuguo09 Thanks, I thought I should apply the same change to posix_spawn after the new nightly build completes to check if my approach was correct.

masayuki2009:
According to comments in #118 from @liuguo09 

>You are right. It seems the mv operation to symtab.c reduces the probability 
>of build break(from my local build, it is easy to hit the link error without mv operation), 
>but it doesn't root cause the link errors in the latest nightly build.

Thanks for the comment, though I think we have another root cause of the link errors in nightly build. This PR might reduce the probability as well. So let's merge this PR.

xiaoxiang781216:
> According to comments in #118 from @liuguo09
> 
> > You are right. It seems the mv operation to symtab.c reduces the probability
> > of build break(from my local build, it is easy to hit the link error without mv operation),
> > but it doesn't root cause the link errors in the latest nightly build.
> 
> Thanks for the comment, though I think we have another root cause of the link errors in nightly build. This PR might reduce the probability as well. So let's merge this PR.

Another try:
https://github.com/apache/incubator-nuttx-apps/pull/123

masayuki2009:
I confirmed that this PR would be effective. 
See https://github.com/apache/incubator-nuttx-apps/pull/118#issuecomment-600904711.

PR #121, opened 2020-03-13
apps/system/cle:  Fix error when compiling with C89 compiler.

patacongo:


system/cle/cle.c
```diff
@@ -798,6 +799,9 @@ static int cle_editloop(FAR struct cle_s *priv)
 
   for (; ; )
     {
+#if  1 /* Perhaps here should be a config switch */
```
Ouss4: Do you want to keep the PR open, or should it be merged?

patacongo:
I believe it should be merged.  It jsut moves a local variable from the middle of a function to the top as required by C89.  Without this, NSH with the CLE fails to build with the ZDS-II toolchain.

system/cle/cle.c
xiaoxiang781216: But @patacongo why we add #if/#endif?

patacongo: It was added.  It was already there.  Just copied.  The original authored intended to add a configuration option but left #if1 in place of the missing option.

PR #127, opened 2020-03-19
Nightly parallel build fix for elf/posix_spawn/module build break

liuguo09:
Makefile: remove context dependency to avoid apps context build twice
   
In nuttx pass1dep and pass2dep builds, context firstly would be built once.
It then call 'make -C apps depend' which would trigger the context built twice.
There is race condition between symtab.c generated by mksymtab.sh in second time
and compiling symtab.c in parallel build. So remove context dependency for apps
depend to make sure context build only one time for apps.
    
Parallel build break logs as below:
/home/jenkins/jenkins-slave/workspace/NuttX-Nightly-Build/apps/examples/elf/elf_main.c:357: undefined reference to `g_elf_nexports'
riscv64-unknown-elf-ld: /home/jenkins/jenkins-slave/workspace/NuttX-Nightly-Build/apps/examples/elf/elf_main.c:357: undefined reference to `g_elf_exports'
make[1]: *** [nuttx] Error 1
make: *** [pass2] Error 2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,7 +167,7 @@ endif
 
 .depdirs: $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_depend)
 
-.depend: context Makefile .depdirs
+.depend: Makefile .depdirs
 	$(Q) touch $@
 
 depend: .depend
--- a/examples/elf/tests/Makefile
+++ b/examples/elf/tests/Makefile
@@ -141,8 +141,7 @@ $(DIRLIST_HDR) : populate
 # Create the exported symbol table
 
 $(SYMTAB_SRC): populate
-	$(Q) $(TESTS_DIR)/mksymtab.sh $(FSIMG_DIR) >$@_tmp
-	$(Q) mv $@_tmp $@
+	$(Q) $(TESTS_DIR)/mksymtab.sh $(FSIMG_DIR) >$@
 
 # Clean each subdirectory
 
--- a/examples/posix_spawn/filesystem/Makefile
+++ b/examples/posix_spawn/filesystem/Makefile
@@ -78,8 +78,7 @@ $(ROMFS_HDR) : $(ROMFS_IMG)
 # Create the exported symbol table
 
 $(SYMTAB_SRC): $(ROMFS_IMG)
-	$(Q) $(FILESYSTEM_DIR)$(DELIM)mksymtab.sh $(ROMFS_DIR) >$@_tmp
-	$(Q) mv $@_tmp $@
+	$(Q) $(FILESYSTEM_DIR)$(DELIM)mksymtab.sh $(ROMFS_DIR) >$@
 
 # Clean each subdirectory
 


Makefile
```diff
@@ -167,7 +167,7 @@ endif
 
 .depdirs: $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_depend)
 
-.depend: context Makefile .depdirs
```
masayuki2009: @liuguo09 
Are there any side effects of removing this dependency?, because this is a top level Makefile.

masayuki2009:
@liuguo09 
Parallel build aging test is working without any problems so far.

Makefile
masayuki2009: @liuguo09 

>So remove context dependency for apps
>depend to make sure context build only one time for apps.

I understood and confirmed that context build for apps is done only once.

liuguo09:
Thanks for merged, let's waiting for the result of Nightly build today. I see it's now building with these changes.

liuguo09:
There is still one maix-bit/module config build break in today's nightly build : (

riscv64-unknown-elf-ld: /home/jenkins/jenkins-slave/workspace/NuttX-Nightly-Build/nuttx/staging/libapps.a(module_main.home.jenkins.jenkins-slave.workspace.NuttX-Nightly-Build.apps.examples.module.o): in function `module_main':
/home/jenkins/jenkins-slave/workspace/NuttX-Nightly-Build/apps/examples/module/module_main.c:164: undefined reference to `g_mod_exports'
riscv64-unknown-elf-ld: /home/jenkins/jenkins-slave/workspace/NuttX-Nightly-Build/apps/examples/module/module_main.c:165: undefined reference to `g_mod_nexports'
make[1]: *** [nuttx] Error 1
make: *** [pass2] Error 2

liuguo09:
@masayuki2009  @xiaoxiang781216 
Besides the context built twice problem, It seems in the time window mksymtab.sh redirect to symtab.c, there is chance symtab.c would also be built meantime. Or should we still consider to not revert your mv operation to reduce the time window?

masayuki2009:
> @masayuki2009 @xiaoxiang781216
> Besides the context built twice problem, It seems in the time window mksymtab.sh redirect to symtab.c, there is chance symtab.c would also be built meantime. Or should we still consider to not revert your mv operation to reduce the time window?

@liuguo09 can you reproduce the same error by adding sleep in mksymtab.sh which you tried before?

liuguo09:
@masayuki2009 I cannot reproduce the same error by adding sleep in mksymtab.sh.
But I found there is some dependency issue in example/module, I make a PR in https://github.com/apache/incubator-nuttx-apps/pull/129. It may make some difference.

masayuki2009:
> @masayuki2009 I cannot reproduce the same error by adding sleep in mksymtab.sh.
> But I found there is some dependency issue in example/module, I make a PR in #129. It may make some difference.

@liuguo09 Thanks for the information. I’ve just merged #129.

masayuki2009:
> @masayuki2009 @xiaoxiang781216
> Besides the context built twice problem, It seems in the time window mksymtab.sh redirect to symtab.c, there is chance symtab.c would also be built meantime. Or should we still consider to not revert your mv operation to reduce the time window?

Hi, @liuguo09 

Unfortunately today's nightly build failed again.
https://builds.apache.org/job/NuttX-Nightly-Build/70/consoleText

I'm still not sure if our previous patches (i.e. mv operation) resolve the issue but we should reconsider to apply them again.

What do you think?

liuguo09:
@masayuki2009 I think we could consider it. I have just made the PR https://github.com/apache/incubator-nuttx-apps/pull/130 to use our former mv operation. Hope it works.

PR #128, opened 2020-03-20
Makefile don't change symtab_apps.c timestamp if no any change

xiaoxiang781216:
Change-Id: I55576aec46bd057bc98cf2b6579d7ac96446ae27


xiaoxiang781216:
> @xiaoxiang781216 Let me confirm. In my understanding, this PR is not to fix a bug but to optimize total build time when a loadable application code is changed but symbols in symtab_apps.c are not changed. Is my understanding correct?

Yes, its' just an optimization.

masayuki2009:
> Yes, its' just an optimization.

@xiaoxiang781216 Thanks for the comment.

PR #130, opened 2020-03-21
examples: Fix elf/module/posix_spawn sometimes parallel build errors …

liuguo09:
…at link time

It seems in the time window mksymtab.sh redirect to symtab.c, there is chance symtab.c
would also be built meantime. Then if built before symtab.c write done, link errors may
occur. So use rename and mv operation to reduce the possibility.

### DIFF
--- a/examples/elf/tests/Makefile
+++ b/examples/elf/tests/Makefile
@@ -141,7 +141,8 @@ $(DIRLIST_HDR) : populate
 # Create the exported symbol table
 
 $(SYMTAB_SRC): populate
-	$(Q) $(TESTS_DIR)/mksymtab.sh $(FSIMG_DIR) >$@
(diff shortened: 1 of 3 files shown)


liuguo09:
@masayuki2009 
the link error seems still happened in today's nightly build : (
It's weird I haven't reproduced in oneshot build recently.
since the full build log is very huge, we may see it in tomorrow's nightly build.

masayuki2009:
> @masayuki2009
> the link error seems still happened in today's nightly build : (
> It's weird I haven't reproduced in oneshot build recently.
> since the full build log is very huge, we may see it in tomorrow's nightly build.

Of course I can't reproduce the link errors on my machine.

And it's very curios because there are 4 defconfigs which have CONFIG_EXAMPLES_POSIXSPAWN=y and 7 defconfigs which has CONFIG_EXAMPLES_ELF=y but almost always lc823450-xgevk:posix_spawn and maix-bit:elf fail.

PR #131, opened 2020-03-21
Remove support for CONFIG_FS_WRITABLE

patacongo:
And also CONFIG_FS_READABLE.

This resolves issue 484 and is a companion to the main PR 602 in incubator_nuttx


patacongo:
@xiaoxiang781216 I am still checking, but I think that the cause of the failure here is that the changes to apps/ of this PR depend on corresponding changes to nuttx/ from PR602.

I don't think it will be able to build the checks.  I will verify that the failed configurations work on my PC, but I do not expected the apps changes to build correctly with the nuttx/ changes or vice versa.

patacongo:
@xiaoxiang781216 Yes, I verified the problem.  These apps/ changes will not build without the corresponding nuttx/ changes in place.  Right now, the PR Check CI / build (arm-14) is failing on the  nucleo-f091rc:nsh.

However, I just built that configuration with no errors or warnings of any kind when both the apps/ changes and the nuttx/ changes are combined.  We need a solution for this.

xiaoxiang781216:
> @xiaoxiang781216 Yes, I verified the problem. These apps/ changes will not build without the corresponding nuttx/ changes in place. Right now, the PR Check CI / build (arm-14) is failing on the nucleo-f091rc:nsh.
> 

I saw some nxstyle errors, do you want to fix them before merge? or let me merge this PR directly.

> However, I just built that configuration with no errors or warnings of any kind when both the apps/ changes and the nuttx/ changes are combined. We need a solution for this.

@liuguo09, could you find some solution to fix this issue?
Before we get a fixed, we have to ignore this false alarm.

btashton:
@liuguo09 one thing I was thinking to help support this flow is to have a keyword in the PR body to refer to a PR from apps.  For example to target apps PR #00 you would do something like this

```
PR description....
blah blah
blah
BuildApps=#00
```

liuguo09:
@xiaoxiang781216 @btashton 
Yes, I think we could parse the PR body to refer to a PR in the other project. Then pull the PR and do the testbuild. I'll do some tests and verify later.

btashton:
@liuguo09 I also have a custom action around somewhere that I use to read PR comments, I suspect it is very similar if you want me to look.

liuguo09:
> @liuguo09 I also have a custom action around somewhere that I use to read PR comments, I suspect it is very similar if you want me to look.

That's good if we could reuse the custom action you used before.

patacongo:
> > @xiaoxiang781216 Yes, I verified the problem. These apps/ changes will not build without the corresponding nuttx/ changes in place. Right now, the PR Check CI / build (arm-14) is failing on the nucleo-f091rc:nsh.
> 
@xiaoxiang781216 
> I saw some nxstyle errors, do you want to fix them before merge? or let me merge this PR directly. Other change looks fine.

Yes, thanks for pointing that out.  I was running a script that had "-m 4" in it is a few long lines were missed.  It should be okay now.

patacongo:
@xiaoxiang781216 Thanks for the merge.  If there any build problems let me know and I will fix them as a priority.

liuguo09:
> @liuguo09 I also have a custom action around somewhere that I use to read PR comments, I suspect it is very similar if you want me to look.

Hi, @btashton have you found your custom action to read PR comments?

PR #132, opened 2020-03-21
Don't generate .depend anymore

xiaoxiang781216:


patacongo:
This is a companion to PR603.  Both should be merged together.

PR #136, opened 2020-03-22
Let install depend on the target binary

xiaoxiang781216:
to ensure the binary get built and install once

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -151,6 +151,7 @@ PROGLIST := $(addprefix $(BINDIR)$(DELIM),$(PROGLIST))
 PROGOBJ := $(MAINOBJ)
 
 $(PROGLIST): $(MAINOBJ)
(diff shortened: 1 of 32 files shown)


Application.mk
```diff
@@ -151,6 +151,7 @@ PROGLIST := $(addprefix $(BINDIR)$(DELIM),$(PROGLIST))
 PROGOBJ := $(MAINOBJ)
 
 $(PROGLIST): $(MAINOBJ)
+	$(Q) mkdir -p $(BINDIR)
```
patacongo: 1) mkdir should not be used explicitly.  This is incompatible with Window native builds.  You should define something.
2) If my recollection is correct, mkdir -p does not work on some Unix platforms, I think macOS.

patacongo: Certainly this is a big risk because I don't see any reference to mkdir in any other Make* or *.mk file.  We need to be careful here.

PR #140, opened 2020-03-26
sotest: Fix a printf format mismatch ("%s" vs int)

yamt:


yamt:
nxstyle complaints seem like problems in nxstyle to me.
https://github.com/apache/incubator-nuttx/issues/627
https://github.com/apache/incubator-nuttx/issues/628

PR #141, opened 2020-03-26
[DON'T MERGE]Fix parallel build

xiaoxiang781216:


patacongo:
Probably needs to be closed without merging.  The companion PR #625 breaks all Cygwin-based builds.  PR #631 reverts it (All Cygwin users are dead in the water until that change is reverted).

xiaoxiang781216:
> Probably needs to be closed without merging. The companion PR #625 breaks all Cygwin-based builds. PR #631 reverts it (All Cygwin users are dead in the water until that change is reverted).

@patacongo but the inclusion sequence issue doesn't exist in apps folder, since INCDIR here is defined after $(TOPDIR)/Make.defs is fully included.

xiaoxiang781216:
Don't merge it, I will update with the better method.

PR #142, opened 2020-03-26
Use MODULECC/MODULELD when building modules

yamt:


yamt:
this depends on https://github.com/apache/incubator-nuttx/pull/629

PR #144, opened 2020-03-27
Clean *.bdat and *.pdat on clean target

yamt:


builtin/registry/Makefile
```diff
@@ -54,8 +54,8 @@ depend:
 
 clean:
 	$(call CLEAN)
-
-distclean: clean
```
xiaoxiang781216: But *.pdat and *.bdat is generated by context target, the stuff generated by context is normally removted in distclean. For example, this come from nuttx/tools/Makefile.unix:
```
clean: subdir_clean
	$(call DELFILE, $(BIN))
	$(call DELFILE, nuttx.*)
	$(call DELFILE, *.map)
	$(call DELFILE, _SAVED_APPS_config)
	$(call DELFILE, nuttx-export*.zip)
	$(call DELDIR, nuttx-export*)
	$(call DELFILE, nuttx_user*)
	$(call DELDIR, staging)
	$(call CLEAN)

subdir_distclean:
	$(Q) for dir in $(CLEANDIRS) ; do \
		if [ -e $$dir/Makefile ]; then \
			$(MAKE) -C $$dir TOPDIR="$(TOPDIR)" distclean ; \
		fi \
	done

distclean: clean subdir_distclean clean_context
ifeq ($(CONFIG_BUILD_2PASS),y)
	$(Q) $(MAKE) -C $(CONFIG_PASS1_BUILDIR) TOPDIR="$(TOPDIR)" distclean
endif
	$(call DELFILE, include/nuttx/config.h)
	$(call DELFILE, include/nuttx/version.h)
	$(call DELFILE, Make.defs)
	$(call DELFILE, defconfig)
	$(call DELFILE, .config)
	$(call DELFILE, .config.old)
	$(call DELFILE, .gdbinit)
	$(Q) $(DIRUNLINK) include/arch/board
	$(Q) $(DIRUNLINK) include/arch/chip
	$(Q) $(DIRUNLINK) include/arch
	$(Q) $(DIRUNLINK) $(ARCH_SRC)/board/board
	$(Q) $(DIRUNLINK) $(ARCH_SRC)/board
	$(Q) $(DIRUNLINK) $(ARCH_SRC)/chip
	$(Q) $(DIRUNLINK) $(TOPDIR)/drivers/platform
```

patacongo: > But *.pdat and *.bdat is generated by context target, the stuff generated by context is normally removed in distclean.

Yes, that is the matching way to under to context target.  The basic idea is that each target in the make should have a matching target that undoes only the actions of that target.  Xiang is correct, make clean should not remove any products created by make context.

There is a make clean_context or make distclean which are the appropriate places.

yamt: ok, i made it clean_context target.
i want a way to clean these files without distclean.

builtin/registry/Makefile
```diff
@@ -55,7 +55,9 @@ depend:
 clean:
 	$(call CLEAN)
 
-distclean: clean
+clean_context:
 	$(call DELFILE, *.bdat)
 	$(call DELFILE, *.pdat)
+
+distclean: clean
```
xiaoxiang781216: need depend on clean_context too?

yamt: i don't think it's necessary here as the top level Makefile have it

PR #145, opened 2020-03-27
Use DEFAULT_TASK_STACKSIZE

yamt:


xiaoxiang781216:
Nice patch! how about we change the default stack size which small than 2048 to DEFAULT_STACK_SIZE, but keep other big stack size as it:
   examples/igmp/Makefile:STACKSIZE = 1024
    examples/powerled/Makefile:STACKSIZE = 1024
    examples/powermonitor/Makefile:STACKSIZE = 768
    examples/relays/Makefile:STACKSIZE = 512
    examples/smps/Makefile:STACKSIZE = 1024
    graphics/screenshot/Makefile:STACKSIZE = 4096
    system/flash_eraseall/Makefile:STACKSIZE = 1024
    testing/cxxtest/Makefile:STACKSIZE = 4096
    testing/smart_test/Makefile:STACKSIZE = 4096
Since the change is safe.

yamt:
using more memory might or might not be safe for small systems.
for examples and testing, it might be fine though.

xiaoxiang781216:
> using more memory might or might not be safe for small systems.
> for examples and testing, it might be fine though.

Yes, but:
1.The memory for each application increase is no more than 1.5K
2.We most likely don't run all this application at the same time
3.These application normally is launched by user manually
So the suggestion in most case:
1.Don't increase the memory consumption after boot finish
2.Just increase 1.5KB when user launch these application
But:
1.Unify the whole code base
2.Make this application can run on sim

yamt:
@xiaoxiang781216 ok. done

xiaoxiang781216:
> @xiaoxiang781216 ok. done

Should we change SYSTEM_RAMTEST_STACKSIZE too?

yamt:
i guess stack size is not too important for what ramtest does. done.

PR #147, opened 2020-03-28
Use new docker image container and add xtensa build support

liuguo09:

### DIFF
--- a/.github/actions/ci-container/action.yaml
+++ b/.github/actions/ci-container/action.yaml
@@ -1,0 +1,20 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


liuguo09:
@btashton apps github action align to nuttx, please help to review.

btashton:
Not sure the failure is related, but any thoughts?

Also I realized that since we share this across the NuttX and Apps we might want to just put it in testing and refer to it there from the workflow. But this can be done later.

liuguo09:
Yes, the failure in arm-13.dat is quite odd. But I think it doesn't matter with this change.
In config stm32f4discovery/kostest: fail log as below:
make[1]: *** No rule to make target '/github/workspace/nuttx/include/nuttx/semapho', needed by 'sched_self.o'.  Stop.
make[1]: *** Waiting for unfinished jobs....
make: *** [tools/LibTargets.mk:74: sched/libsched.a] Error 2
make: *** Waiting for unfinished jobs....
It seems semaphore.h while not semapho. 
I just force pushed again to see if this can be reproduced.

Issue #150, opened 2020-03-30
tcpblaster is using NuttX's CONFIG_ENDIAN_BIG for host OS program


yamt:

Ouss4:
I guess we should remove the HTONS/L macros definition and replace them with htons/l (same as UDP blaster). The host implementation should worry about endianness.

yamt:
@Ouss4 i agree. we need to adjust the usage as htons() is not usable in initializers on linux.

PR #152, opened 2020-03-30
examples/tcpblaster: When a host PC is used, take its implementation of hton/ntoh functions.

Ouss4:
@yamt This should address #150 
Please take a look.

### DIFF
--- a/examples/tcpblaster/tcpblaster.h
+++ b/examples/tcpblaster/tcpblaster.h
@@ -52,33 +52,17 @@
  ****************************************************************************/
 
 #ifdef TCPBLASTER_HOST
-#undef HTONS
-#ifdef CONFIG_ENDIAN_BIG
-#  define HTONS(ns) (ns)
-#else
-#  define HTONS(ns) \
-     (unsigned short) \
-       (((((unsigned short)(ns)) & 0x00ff) << 8) | \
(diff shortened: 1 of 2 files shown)


examples/tcpblaster/tcpblaster.h
```diff
@@ -52,33 +52,17 @@
...
-#  define HTONL(nl) (nl)
-#else
-#  define HTONL(nl) \
-     (unsigned long) \
-       (((((unsigned long)(nl)) & 0x000000ffUL) << 24) | \
-        ((((unsigned long)(nl)) & 0x0000ff00UL) <<  8) | \
-        ((((unsigned long)(nl)) & 0x00ff0000UL) >>  8) | \
-        ((((unsigned long)(nl)) & 0xff000000UL) >> 24))
-#endif
+/* HTONS/L macros are unique to uIP */
```
patacongo: It is kind of weird to be calling these uIP macros.  Yes, a dozen years ago some parts of the network came from uIP.  But I think "HTONS/L macros unique to NuttX" would be more relevant today.

Ouss4: I copied it from the UDP one.
Fixed now.

yamt:
lgtm. thank you.

PR #154, opened 2020-03-31
CI: Tweak workflow to make it simpler to sync with other repos

yamt:


.github/workflows/main.yml
```diff
@@ -16,32 +16,6 @@ on:
   pull_request:
 
 jobs:
-  check:
```
xiaoxiang781216: Should we change line 13 to:
name: PR Check CI (build)
and rename main.yml to build.yml?

liuguo09:
LGTM. Besides xiang's comment, one thing I worry a little bit is: would we keep on matrix build jobs donot depends on check job always as in https://github.com/apache/incubator-nuttx/pull/552 does? If the answer is yes, it's good and reasonable to split into check.yml and build.yml.

yamt:
@liuguo09 we can merge it back if/when we want to put back the dependency. i separated it for now because at this point it seems little point to run it on testing repo.

.github/workflows/main.yml
yamt: replied in https://github.com/apache/incubator-nuttx/pull/666

liuguo09:
@yamt I understand. Thanks.

PR #158, opened 2020-04-03
nsh/command: do not show the module application in built-in list

anchao:
Do not show the module application in built-in list


xiaoxiang781216:
@masayuki2009 please take look which fix nsh help you report several weeks ago.

PR #159, opened 2020-04-03
wireless/wapi: add save_config/reconnect/scan_results support

anchao:


anchao:
so, i think we can ignore the warning here like:
...
wireless/wapi/src/util.c:79:6: error: Mixed case identifier found
wireless/wapi/src/util.c:106:9: error: Mixed case identifier found
wireless/wapi/src/util.c:122:33: error: Mixed case identifier found
wireless/wapi/src/util.c:128:6: error: Mixed case identifier found
...
since the cJson api is mixed case.

netutils/netinit/netinit_associate.c
```diff
@@ -77,7 +77,8 @@ int netinit_associate(FAR const char *ifname)
       conf.phraselen   = strlen(conf.passphrase);
     }
 
-  ret = wpa_driver_wext_associate(&conf);
+  if (conf.ssidlen > 0)
+    ret = wpa_driver_wext_associate(&conf);
```
Ouss4, line 81:
```suggestion
  if (conf.ssidlen > 0)
    {
      ret = wpa_driver_wext_associate(&conf);
    }
```

netutils/netinit/netinit_associate.c
```diff
@@ -60,7 +60,7 @@
 int netinit_associate(FAR const char *ifname)
 {
   struct wpa_wconfig_s conf;
-  int ret;
+  int ret = -EINVAL;
```
Ouss4, line 63: This is causing the build to fail. errno.h is not included.

netutils/netinit/netinit_associate.c
anchao: Oops, Done

netutils/netinit/netinit_associate.c
anchao: Done

PR #161, opened 2020-04-03
netutils/dhcpc: configurable timeout and retry count.

anchao:


netutils/dhcpc/dhcpc.c
```diff
@@ -433,7 +433,7 @@ FAR void *dhcpc_open(FAR const char *interface, FAR const void *macaddr,
 
       /* Configure for read timeouts */
 
-      tv.tv_sec  = 10;
+      tv.tv_sec  = CONFIG_NETUTILS_DHCPC_RECV_TIMEOUT;
```
jerpelea, line 450: is there are reason for changing the 10 second dely with 3?

netutils/dhcpc/Kconfig
```diff
@@ -16,4 +16,17 @@ config NETUTILS_DHCPC_HOST_NAME
 	string "DHCP client host name"
 	default "nuttx"
 
+config NETUTILS_DHCPC_RECV_TIMEOUT
+	int "Number of receive timeout in secound"
```
jerpelea, line 20: typo

netutils/dhcpc/dhcpc.c
anchao: DHCP client uses the UDP protocol to transmit, which is not reliable because sometime the dhcp request packet may not be accurately sent to the server if the signal is very weak, configurable options are also a good choose to change the behavior of  dhcpc.

netutils/dhcpc/Kconfig
anchao: Done

netutils/dhcpc/dhcpc.c
jerpelea: I am asking because the hard coded value was 10 sec while now the default is 3 seconds

xiaoxiang781216: 10sec for the timeout is too large for the most nowaday network, especially for DHCP transaction which always happen between two node directly without any intermediate node.

PR #162, opened 2020-04-03
CI: remove the additional pull docker container step

liuguo09:
There is chance pull docker container failure in build jobs,
remove the additional pull docker container step may make it
more stable. Meanwhile, it also saves some time for build job.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,9 +57,6 @@ jobs:
         username: ${GITHUB_ACTOR}
         password: ${{ secrets.GITHUB_TOKEN }}
 
-    - name: Run Pull Container
-      uses: ./testing/.github/actions/ci-container
-
     - name: Run builds
       uses: ./testing/.github/actions/ci-container
       env:


btashton:
This should have zero performance impact. It was intentionally it's own step so we can see how much time is being spent pulling the image so we can decide if we need to split it up.

xiaoxiang781216:
> This should have zero performance impact. It was intentionally it's own step so we can see how much time is being spent pulling the image so we can decide if we need to split it up.

But it's still consume 50s to finish from check tab.  @liuguo09  intentionally don't change the testing repo, so we can still get the number if we want.

btashton:
Yes it will consume the 50sec, but the next step will take just the amount of time to check that it has what it needs locally which is why it should have no performance difference we can even tag it which would further reduce the cost since it won't even check the remote server.

xiaoxiang781216:
If github can skip the second downloading, my understanding is wrong. Do you want to revert the change?

PR #170, opened 2020-04-11
Remove CONFIG_CAN_PASS_STRUCT

patacongo:
This is the companion to PR #766.  It removes the CONFIG_CAN_PASS_STRUCT option as recommended by Issue #620

NuttX PR #766 depends on PR being in place but not vice versa.  This PR should be merge-able without #766 and then PR #766 should also pass its checks.


patacongo:
Please ignore the nxstyle CI check.  It shows onlyt the expected Mixed case identifier problems noted above.

PR #176, opened 2020-04-15
Fix echo NULL character correctly

xiaoxiang781216:


yamt:
it isn't clear to me why you reverted the change.

xiaoxiang781216:
> it isn't clear to me why you reverted the change.

This patch add -f option to force write back null char, but I think that the correct behaviour should always write back all recevie char, do you think so?

image

yamt:
maybe. i don't know what was the motivation of the -f option.

yamt:
@Ouss4 @lukegluke does this change solve what you wanted to solve with the -f option?

xiaoxiang781216:
> maybe. i don't know what was the motivation of the -f option.

Yes, that's why I remove -f option, writing back what cu receive from serial is always a good action. no any reason don't enable it by default.

Ouss4:
I just created the PR on behalf of Oleg, I had nothing to do with the change itself.
@xiaoxiang781216  the old commit sends back everything without any special treatment, even for commands (with ~), was that not desirable?

lukegluke:
Hi @xiaoxiang781216 

Sorry for late reply, for some reason github do not inform me about discussion!
I totally agree that writing back what cu receive from serial is always a good action, and I was surprised why it is not default.
But -f option is **not only** to force write back null char! @Ouss4 is right with -f option cu doesn't stop on escape characters ~.

xiaoxiang781216:
@lukegluke thanks for explanation, I can understand -f option now. I will close my PR without merge.

lukegluke:
@xiaoxiang781216, but I think https://github.com/apache/incubator-nuttx-apps/pull/176/commits/f00bca3250698da686def67a9a60060e254f2fa6 should be merged to always write back all recevied chars even without -f option

xiaoxiang781216:
@lukegluke Ok, I will bring back this patch.

Ouss4:
@yamt If you are good with the change please merge.
(Please Rebase and merge to not create a merge commit).

yamt:
lgtm

PR #177, opened 2020-04-15
CI: Make job descriptions less redundant

yamt:


yamt:
the macOS/arm-11 failure looks like incomplete Make.dep.
probably https://github.com/apache/incubator-nuttx/pull/707

xiaoxiang781216:
> the macOS/arm-11 failure looks like incomplete Make.dep.
> probably [apache/incubator-nuttx#707](https://github.com/apache/incubator-nuttx/pull/707)

I will look this patch in this week. I think there are many similar issues in Makefile.

PR #182, opened 2020-04-16
Use MODULESTRIP where appropriate

yamt:


yamt:
this depends on https://github.com/apache/incubator-nuttx/pull/806

PR #183, opened 2020-04-16
nshlib/nsh_codeccmd.c: fix potential NULL dereference and check mallo…

acassis:
…c return values


include/netutils/md5.h
```diff
@@ -10,7 +10,6 @@
  *
  *   This code implements the MD5 message-digest algorithm.
  *   The algorithm is due to Ron Rivest.  This code was
- *   written by Colin Plumb in 1993, no copyright is claimed.
```
patacongo, line 13: Did you intend to remove this line?

netutils/codecs/md5.c
```diff
@@ -119,15 +119,15 @@ static void byteReverse(FAR unsigned char *buf, unsigned longs)
...
+ * Name: md5_init
  *
  * Description:
  *   Start MD5 accumulation.  Set bit count to 0 and buffer to mysterious
  *   initialization constants.
  *
  ****************************************************************************/
 
-void MD5Init(struct MD5Context *ctx)
+void md5_init(struct md5_context_s *ctx)
```
patacongo, line 130: This is a global function.  This will need a corresponding change to the prototype in apps/include/netutils/md5.h

include/netutils/md5.h
```diff
@@ -74,24 +73,24 @@ extern "C"
...
-struct MD5Context
+struct md5_context_s
 {
   uint32_t buf[4];
   uint32_t bits[2];
   uint8_t in[64];
 };
 
-typedef struct MD5Context MD5_CTX;
+typedef struct md5_context_s MD5_CTX;
```
patacongo, line 83: The coding standard does not permit upper case letters in type name.  Types must always end in _t.  The correct name would be md5_ctx_t

patacongo: Hmm... there are other examples where upper case types are used for handles (after Microsoft conventions).  I suppose we should end that practice because it is non-compliant with the standard.

include/netutils/md5.h
acassis: No, I didn't want to remove it and also the above line will become incomplete if it was removed.

include/netutils/md5.h
acassis: Ok, I will fix line 13 and keep line 83 since you agree to merge without changing MD5_CTX.

PR #186, opened 2020-04-17
Backport #181  to releases/9.0

Ouss4:
Backport #181


Ouss4:
Included in #188 as there is a circular dependency.

PR #193, opened 2020-04-19
Fix minor issue in Makefile

xiaoxiang781216:


examples/flowc/.gitignore
```diff
@@ -3,6 +3,8 @@
 /.built
 /host
 /config.h
+/receiver
+/sender
```
patacongo, line 7: This and all of the following need to have *.exe to exclude the files in a Windows build.

xiaoxiang781216: *.exe is already added in all .gitignore

PR #210, opened 2020-04-25
Backport #209

btashton:


btashton:
Verified that it is pulling the correct branches
```
Working with ref: refs/heads/releases/9.0
Working with a branch: releases/9.0
Triggered by change in APPS
```

PR #220, opened 2020-04-30
Fix uninitialized variable warning in ostest

btashton:
In lowpri_thread if sem_wait fail nwaiting will not be set even
though the rest of the function uses it.  This is a failure
so really just handling the compiler warning with this.

Signed-off-by: Brennan Ashton <bashton@brennanashton.com>


btashton:
@patacongo  This really papers over the issue, but at this point the ostest has failed so I am not sure how much error handling to add around this vs just being explicit in the behaviour.

btashton:
A lot of unrelated style issues that I am resolving.

PR #226, opened 2020-05-02
Fix nxstyle issue

xiaoxiang781216:

### DIFF
--- a/builtin/exec_builtin.c
+++ b/builtin/exec_builtin.c
@@ -138,9 +138,9 @@ int exec_builtin(FAR const char *appname, FAR char * const *argv,
       goto errout_with_actions;
     }
 
-   /* If robin robin scheduling is enabled, then set the scheduling policy
-    * of the new task to SCHED_RR before it has a chance to run.
-    */
+  /* If robin robin scheduling is enabled, then set the scheduling policy
+   * of the new task to SCHED_RR before it has a chance to run.
+   */
 
 #if CONFIG_RR_INTERVAL > 0
   ret = posix_spawnattr_setschedpolicy(&attr, SCHED_RR);
--- a/fsutils/mkfatfs/mkfatfs.c
+++ b/fsutils/mkfatfs/mkfatfs.c
@@ -104,13 +104,13 @@ static uint32_t fat_systime2fattime(void)
               uint16_t fattime;
               uint16_t fatdate;
 
-              fattime  = (tm.tm_sec       >>  1) & 0x001f; /* Bits 0-4: 2 second count (0-29) */
-              fattime |= (tm.tm_min       <<  5) & 0x07e0; /* Bits 5-10: minutes (0-59) */
-              fattime |= (tm.tm_hour      << 11) & 0xf800; /* Bits 11-15: hours (0-23) */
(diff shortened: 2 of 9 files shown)


patacongo:
Have you tried this in protected and/or kernel build modes. There might be probablys write to the errno in those cases since the errno resides in the TCB which is not accessible.  In those modes set_errno() is a function call that goes through a syscall and error is defined to be get_errno() which will only support read access.

So reading the errno in the modes is okay, but writing must explicitly use set_errno().  That does break the standard but I am open to other technical solutions.

xiaoxiang781216:
> Have you tried this in protected and/or kernel build modes. There might be probablys write to the errno in those cases since the errno resides in the TCB which is not accessible. In those modes set_errno() is a function call that goes through a syscall and error is defined to be get_errno() which will only support read access.
> 

Yes, you are right. userspace can't access TCB directly in protected/kernel build.

> So reading the errno in the modes is okay, but writing must explicitly use set_errno(). That does break the standard but I am open to other technical solutions.

What I can imagine:
1.Move errno to userspace(e.g. at the begining of thread stack)
2.Overwrite assignment operatior(but require C++)

patacongo:
> What I can imagine:
> 1.Move errno to userspace(e.g. at the begining of thread stack)

Except how would you find the beginning of the thread stack in user space?

This is how TLS works.  TLS aligns the stack and then, by masking the current stack pointer, you can get the address of the base of the current stack.  However, this does not work in FLAT or PROTECTED modes.  That is because the size of that mask also effects the maximum size of the stack.  For example, if I use the a 4Kb mask 0x00000fff, then the maximum stack size would be 4Kb which is not big enough.  Large stack alignments effect memory usage on platforms with minimal RAM resources.

So TLS can only really be used in the KERNEL build mode which has highly aligned virtual stacks.

> 2.Overwrite assignment operatior(but require C++)

Or use the non-standard set_errno();

patacongo:
I looked at how GLIBC handles errno.  It uses TLS.  Here is a generl overview:

glibc's errno.h has multiple definitions of errno.h but if nothing is defined, it includes stdlib/errno.h and stdlib/errno.h has this:

```
    /* The error code set by various library functions.  */
    extern int *__errno_location (void) __THROW __attribute_const__;
    # define errno (*__errno_location ())
```
 
There are various implementations of __errno_location().  Most default to this:
 
```
  #include <errno.h>
  #include <tls.h>
 
  int *
  __errno_location (void)
  {
    return &errno;
  }
  libc_hidden_def (__errno_location)
```

Where errno is then defined (errno.h) as a thread-specific integer:

`    extern __thread int errno attribute_tls_model_ie;`

patacongo:
I think that we could implement a different kind of TLS that does not require aligned stacks.  We could so a system call to get the stack bas address.  There is already a pthread_get_stackaddr_np(), we could generalize that more and use it to get the base of the stack and then provide an alternative implementation of TLS that used the non-aligned stack base.

That would not be too difficult.  However, it is still inefficient since we would have to do a system call to get the stack base address.  With traditional TLS, we only have to mask the stack pointer address and that can all be done in user space.

patacongo:
I moved this discussion to Issue #227 so that we will not forget about this.  It is important in the long term to come up with a proper fix for writing to errno from user space.

xiaoxiang781216:
Sure.

PR #229, opened 2020-05-05
Restore Git Credentials on NuttX repo

btashton:
Something change on the GitHub side that has broken credentials
between the the fetch and buld workflow steps.  There is no
way to tell the checkout action to only re-auth so we have it
checkout the repo again, but it will use what it already has.

In the future we can change fetch step to just compute the
target refs and store them in a script that we can call to
set the environment variables for the targets much like
this patch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -162,13 +162,23 @@ jobs:
           retry_wait_seconds: 10
           command: docker pull docker.pkg.github.com/apache/incubator-nuttx-testing/nuttx-ci-linux
 
+      - name: Export NuttX Repo SHA
+        run:  echo "::set-env name=nuttx_sha::`git -C sources/nuttx rev-parse HEAD`"
+      - name: Refresh Git Credentials
+        uses: actions/checkout@v2
+        with:
+          repository: apache/incubator-nuttx
+          ref: ${{ env.nuttx_sha }}
+          path: sources/nuttx
+          fetch-depth: 1
+      - name: Get Tags for NuttX Repo
+        run: git -C sources/nuttx fetch --tags
       - name: Run builds
         uses: ./sources/testing/.github/actions/ci-container
         env:
           BLOBDIR: /tools/blobs
         with:
           run: |
-            git -C sources/nuttx fetch --tags
             cd sources/testing
             ./cibuild.sh -x testlist/${{matrix.boards}}.dat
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -162,13 +162,23 @@ jobs:
           retry_wait_seconds: 10
           command: docker pull docker.pkg.github.com/apache/incubator-nuttx-testing/nuttx-ci-linux
 
+      - name: Export NuttX Repo SHA
+        run:  echo "::set-env name=nuttx_sha::`git -C sources/nuttx rev-parse HEAD`"
+      - name: Refresh Git Credentials
+        uses: actions/checkout@v2
+        with:
+          repository: apache/incubator-nuttx
```
xiaoxiang781216, line 170: Does this mean we fetch the source code again?

.github/workflows/build.yml
```diff
@@ -162,13 +162,23 @@ jobs:
...
+          ref: ${{ env.nuttx_sha }}
+          path: sources/nuttx
+          fetch-depth: 1
+      - name: Get Tags for NuttX Repo
+        run: git -C sources/nuttx fetch --tags
```
xiaoxiang781216, line 175: I may ask the same question before: does it make sense to move the tag fetch to 'Fetch-Source->Checkout nuttx repo'?

.github/workflows/build.yml
btashton: It cleans the repo and resets all the refs (no way to avoid it, I opened a ticket on the checkout action), but we still have all the objects cached in `.git`

.github/workflows/build.yml
btashton: Causes a huge slow down because the cache becomes substantially larger.  I want to implement a better solution later this week that goes back to just doing the checkouts in the build job but computes the refs in the shared job, but I want to unblock the CI for now.

xiaoxiang781216: I am curious why tag can increase the size so much? as I understand that tag is just a simple marker.

.github/workflows/build.yml
xiaoxiang781216: The current method require each job(20~):
1.Clean the repo and reset all refs
2.Fetch the tag from server
Can we save these two step if we move the tag fetch to 'Fetch-Source->Checkout nuttx repo'? I am wondering which method is more fast and stable.

.github/workflows/build.yml
btashton: because we loose the clone of depth=1 when you pull down tags since it has to pull the full history :/

.github/workflows/build.yml
btashton: I am testing a build where we just pass the refs

.github/workflows/build.yml
xiaoxiang781216: Just saw this from https://github.com/actions/checkout:
## Fetch all tags

```yaml
- uses: actions/checkout@v2
- run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
```

## Fetch all branches

```yaml
- uses: actions/checkout@v2
- run: |
    git fetch --no-tags --prune --depth=1 origin +refs/heads/*:refs/remotes/origin/*
```

## Fetch all history for all tags and branches

```yaml
- uses: actions/checkout@v2
- run: |
    git fetch --prune --unshallow
```
Can we use them to fix the problem without dropping the speed.

PR #230, opened 2020-05-05
CI: Simplify flow for checking out repos

btashton:


btashton:
@xiaoxiang781216 How about this?  This seems like a much better way to go. It also addresses both of our concerns with pulling the tags

.github/workflows/build.yml
```diff
@@ -162,17 +109,38 @@ jobs:
...
           command: docker pull docker.pkg.github.com/apache/incubator-nuttx-testing/nuttx-ci-linux
 
-      - name: Export NuttX Repo SHA
-        run:  echo "::set-env name=nuttx_sha::`git -C sources/nuttx rev-parse HEAD`"
-      - name: Refresh Git Credentials
+      - uses: actions/download-artifact@v2
+        with:
+          name: gitrefs
+          path: gitrefs-rt/
+      - name: Export Repo Refs
```
xiaoxiang781216, line 122: move before "uses: actions/download-artifact@v2"

.github/workflows/build.yml
```diff
@@ -162,17 +109,38 @@ jobs:
...
         uses: actions/checkout@v2
         with:
           repository: apache/incubator-nuttx
-          ref: ${{ env.nuttx_sha }}
+          ref: ${{ env.os_ref }}
           path: sources/nuttx
           fetch-depth: 1
-      - name: Get Tags for NuttX Repo
-        run: git -C sources/nuttx fetch --tags
+      - run: git -C sources/nuttx pull --tags
```
xiaoxiang781216, line 132: Should we use "git -C sources/nuttx fetch --prune --unshallow" to get both branch and tag? I am afraid that we may want the branch at the late time.

.github/workflows/build.yml
```diff
@@ -73,52 +73,18 @@ jobs:
...
-        uses: actions/cache@v1
-        with:
-          path: sources
-          key: build-sources-${{ github.run_id }}
+          name: gitrefs
+          path: gitenv.sh
 
   Linux:
-    needs: Fetch-Source
+    needs: Compute-Source-Refs
```
xiaoxiang781216, line 93: But Compute-Source-Refs don't generate and upload the cache anymore, why don't we move os_ref/apps_ref/testing_ref into a common function? So we can invoke it in macOS/Linux directly. The current workflow which upload/download gitenv.sh seems too complex.

.github/workflows/build.yml
```diff
@@ -217,17 +172,38 @@ jobs:
...
           key: ${{ runner.os }}-tools-${{ hashFiles('./sources/testing/cibuild.sh') }}
 
-      - name: Export NuttX Repo SHA
-        run:  echo "::set-env name=nuttx_sha::`git -C sources/nuttx rev-parse HEAD`"
-      - name: Refresh Git Credentials
+      - uses: actions/download-artifact@v2
+        with:
+          name: gitrefs
+          path: gitrefs-rt/
+      - name: Export Repo Refs
```
xiaoxiang781216, line 179: move before "uses: actions/download-artifact@v2"

.github/workflows/build.yml
btashton: This is how github recommends that you pass information between jobs. Once they implement templating like every other yaml based CI system we can remove this.

As for why it is not a function. At this point prior to doing cloning there is not repo to put the script in, only the workflow yaml so it must be in the yaml. Making it a job that others depend on saves us from having to copy it to every job (currently Linux and macOS, but soon windows as well)

As for the bit about the cache, I had the cache before because it was much faster to move the larger git repos. Now we just move 100 bytes of data so there is no reason to do the cache since we had to support the fallback of artifacts anyway.

.github/workflows/build.yml
btashton: We should consider that this might not be a branch and instead the merge ref that is being tested (that's what you get by default with the checkout action if no ref is explicitly set). This matches the existing functionally so I don't see  a reason to change it.

xiaoxiang781216: Ok.

.github/workflows/build.yml
xiaoxiang781216: All steps have a name but "uses: actions/download-artifact@v2" don't have. so I suppose that it's one step from line 118 to line 123, and should order like this:
      - name: Export Repo Refs
      - uses: actions/download-artifact@v2
        with:
          name: gitrefs
          path: gitrefs-rt/
        run: source gitrefs-rt/gitenv.sh

.github/workflows/build.yml
```diff
@@ -162,17 +115,38 @@ jobs:
           retry_wait_seconds: 10
           command: docker pull docker.pkg.github.com/apache/incubator-nuttx-testing/nuttx-ci-linux
 
-      - name: Export NuttX Repo SHA
-        run:  echo "::set-env name=nuttx_sha::`git -C sources/nuttx rev-parse HEAD`"
-      - name: Refresh Git Credentials
+      - uses: actions/download-artifact@v2
+        with:
+          name: gitrefs
+          path: gitrefs-rt/
```
xiaoxiang781216, line 121: should we directly donwload to current the directory?

.github/workflows/build.yml
btashton: Ah I see the confusion. That is actually two steps. Every '-' is a step, you cannot have a 'run' and a 'uses' in the same step.

.github/workflows/build.yml
btashton: Sounds good.

.github/workflows/build.yml
xiaoxiang781216: > This is how github recommends that you pass information between jobs. Once they implement templating like every other yaml based CI system we can remove this.
> 
> As for why it is not a function. At this point prior to doing cloning there is not repo to put the script in, only the workflow yaml so it must be in the yaml. Making it a job that others depend on saves us from having to copy it to every job (currently Linux and macOS, but soon windows as well)
> 

I mean the funciton in action yml not in bash script. Maybe the step is better term than function in this context.

> As for the bit about the cache, I had the cache before because it was much faster to move the larger git repos. Now we just move 100 bytes of data so there is no reason to do the cache since we had to support the fallback of artifacts anyway.

sharing the code by job need save the result to script and then load/execute this script in other jobs. Can we use the common step to share the code like:
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions
```
Example using action in the same repository as the workflow
./path/to/dir

The path to the directory that contains the action in your workflow's repository. You must check out your repository before using the action.

jobs:
  my_first_job:
    steps:
      - name: Check out repository
        uses: actions/checkout@v2
      - name: Use local my-action
        uses: ./.github/actions/my-action
```
Or even put the common step in the same yml file:
https://github.com/actions/checkout/blob/master/action.yml
```
name: 'Checkout'
description: 'Checkout a Git repository at a particular version'
...
runs:
  using: node12
  main: dist/index.js
  post: dist/index.js
```
I guess that Compute-Source-Refs become a common step if we move Compute-Source-Refs out of the job.
From my understand, once the common code become to a step, we can use it inside Linux/macOS job and get the result directly.

btashton: That is a good option in the since that it actually can be stored in the testing repo (see what I did for the Linux build step). But it means that we would need to write it in nodejs since you cannot use the docker action on macOS or Windows. I was trying to avoid writing nodejs from a maintenance standpoint. I can try it if you think it's ok.

xiaoxiang781216: Can we use bash script here?

btashton: Nope only if you make it a Docker action which limits you to Linux. Our only option would be JavaScript.
https://help.github.com/en/actions/building-actions/about-actions#types-of-actions

.github/workflows/build.yml
```diff
@@ -74,51 +79,18 @@ jobs:
...
+          echo "#!/bin/sh" > gitenv.sh
+          echo "echo ::set-env name=os_ref::$OS_REF" >> gitenv.sh
+          echo "echo ::set-env name=apps_ref::$APPS_REF" >> gitenv.sh
+          echo "echo ::set-env name=testing_ref::$TESTING_REF" >> gitenv.sh
+      - name: Upload gitrefs
+        uses: actions/upload-artifact@v2
         with:
-          path: sources
-          key: build-sources-${{ github.run_id }}
+          name: gitrefs
```
xiaoxiang781216, line 89: change gitrefs to gitenv?

.github/workflows/build.yml
```diff
@@ -74,51 +79,18 @@ jobs:
...
-          path: sources.tar.gz
-
-      - name: Cache Source
-        id: cache-source
-        uses: actions/cache@v1
+          echo "#!/bin/sh" > gitenv.sh
+          echo "echo ::set-env name=os_ref::$OS_REF" >> gitenv.sh
+          echo "echo ::set-env name=apps_ref::$APPS_REF" >> gitenv.sh
+          echo "echo ::set-env name=testing_ref::$TESTING_REF" >> gitenv.sh
+      - name: Upload gitrefs
```
xiaoxiang781216, line 86: gitrefs to gitenv?

.github/workflows/build.yml
```diff
@@ -21,17 +21,22 @@ on:
     tags:
 
 jobs:
-  Fetch-Source:
+  Compute-Source-Refs:
```
xiaoxiang781216, line 24: Compute-Source-Refs==>Environment?

.github/workflows/build.yml
```diff
@@ -74,51 +79,18 @@ jobs:
...
+          echo "echo ::set-env name=os_ref::$OS_REF" >> gitenv.sh
+          echo "echo ::set-env name=apps_ref::$APPS_REF" >> gitenv.sh
+          echo "echo ::set-env name=testing_ref::$TESTING_REF" >> gitenv.sh
+      - name: Upload gitrefs
+        uses: actions/upload-artifact@v2
         with:
-          path: sources
-          key: build-sources-${{ github.run_id }}
+          name: gitrefs
+          path: gitenv.sh
```
xiaoxiang781216, line 90: After some thought, env.sh may a better name because we may add new variable not related to git in the furture.

btashton:
@xiaoxiang781216 I am going to wait on this until the weekend and see if I can just write the javascript action to managed the repos logic.  It will make it easy to add the PR linking logic that we have been needing and bit us today.

xiaoxiang781216:
Sure, thanks for your effort to improve the github action so much.

PR #235, opened 2020-05-07
netutils/usrsock_rpmsg:  Fix build failures

patacongo:
## Summary

Not all files were correctly modified when usrsock_rpmsg was moved from apps/system to apps/netutils.  Changes were overlooked in the Makefile and Make.defs file.

## Impact

Should correct the PR check system which is failing on this problem.


xiaoxiang781216:
@patacongo thannks to fix my fault.

PR #238, opened 2020-05-07
Remove use of set_errno() from application code.

patacongo:
## Summary

The normal assignment errno = errcode should work fine now.


patacongo:
This PR will NOT compile on all architectures with the matching nuttx/ PR 997

Ouss4:
> This PR will NOT compile on all architectures without the matching nuttx/ PR 997

The matching PR is in a separate branch.  The check will pull in master.

PR #244, opened 2020-05-09
examples/thttpd/content/tasks/tasks.c:  Update OS internal name

patacongo:
This logic uses the OS internal nxsched_foreach() function.  This is a violation of the POSIX OS interface.  But for now, we just update the name.


patacongo:
In order to complete the renaming of sched functions, two things should be done:

1. All of the functions in sched/sched/sched.h need to modified to follow the same naming conventions used in include/nuttx/sched.h:

    nxsched_<verb>_<object>

2. Of much less importance, all of the numbers of static functions within sched/sched/*.c functions must be changed to follow that same convention.  This is less important because it is not exposed outside of the file.

PR #250, opened 2020-05-14
Update make dependency for elf/module/nxflat/posix_spawn/sotest/thttpd

liuguo09:
## Summary
Update make dependency for elf/module/nxflat/posix_spawn/sotest/thttpd examples which make use of mksymtab.sh. This could avoid them built twice in 'make depend' and 'make all' which would result in file truncated build break. Now only build once by 'make all'.

## Testing
Pass local nightly build

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -219,7 +219,7 @@ else
 context::
 endif
 
-.depend: Makefile $(SRCS)
+.depend: Makefile $(wildcard $(SRCS))
 ifeq ($(filter %$(CXXEXT),$(SRCS)),)
 	$(Q) $(MKDEP) $(ROOTDEPPATH) "$(CC)" -- $(CFLAGS) -- $(filter-out Makefile,$^) >Make.dep
 else
--- a/examples/elf/Makefile
+++ b/examples/elf/Makefile
@@ -37,7 +37,14 @@
 
 # ELF Example
 
-CSRCS = symtab.c
+ifeq ($(CONFIG_EXAMPLES_ELF_ROMFS),y)
+CSRCS = romfs.c
+endif
+ifeq ($(CONFIG_EXAMPLES_ELF_CROMFS),y)
+CSRCS = cromfs.c
+endif
+CSRCS += dirlist.c
+CSRCS += symtab.c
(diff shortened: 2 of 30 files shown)


liuguo09:
_examples/sotest/sotest_main.c:203:25: error: Operator/assignment must be preceded with whitespace
examples/sotest/sotest_main.c:223:25: error: Operator/assignment must be preceded with whitespace
examples/sotest/sotest_main.c:243:25: error: Operator/assignment must be preceded with whitespace_

It seems the above nxstyle print is wrongly reported, the source code as below:

_203   testfunc = (CODE void (*)(FAR const char *))dlsym(handle2, "testfunc1");
223   testfunc = (CODE void (*)(FAR const char *))dlsym(handle2, "testfunc2");
243   testfunc = (CODE void (*)(FAR const char *))dlsym(handle2, "testfunc3");_

patacongo:
> It seems the above nxstyle print is wrongly reported

This is addressed in issue #628

Application.mk
```diff
@@ -219,7 +219,7 @@ else
 context::
 endif
 
-.depend: Makefile $(SRCS)
+.depend: Makefile $(wildcard $(SRCS))
```
yamt, line 222: @liuguo09 @xiaoxiang781216
doesn't this break apps using VPATH/--dep-path?
my app's Make.dep is empty after this change. (as wildcard expands to empty as no source files in the current directory)

xiaoxiang781216: Yes, we have to remove the generated source from Make.dep. The build sequence is:
1.context
2.depend
3.compile
The code generated dynamically need the result from phase 3(compile), but phase 2(depend) need all source code ready. This is a chicken and egg problem. We decide to not generate the dependence for the dynamically code since the dependence relationship is optinal, but the compile is critical.

yamt: my app doesn't involve any generated code at all. everything is static. just in the other directory.

liuguo09: Just update https://github.com/apache/incubator-nuttx-apps/pull/265 to fix it. Please take a look.  Thanks.

PR #255, opened 2020-05-17
apps/nshlib:  Restore correct load order for binaries.

patacongo:
## Summary
An important feature of ELF binaries for NSH is the capability to replace built-in applications with updated ELF applictions on a file system.  This works because of this ording in NSH:

- Load file from file system if possible
- If not, run built-in application if possible
- If not, run NSH command line command if possible
- If not, report that the command was not found

This critical behavior was broken by this commit:

    commit 9a28ccf836c1b9f0eb5e1163964042eddc207697
    Author: chao.an <anchao@xiaomi.com>
    Date:   Fri Feb 21 09:54:47 2020 +0800

        nsh/parse: try the builtin configuration first

        In the case of enable the BUILTIN_APPS/FILE_APPS at the same time,
        try the builtin list first to ensure that the relevant configuration
        (stacksize, priority) can be set normally.

This commit breaks the feature because it changes the order to that the built-in application is tried first.  Hence, the version on the file system will never replace the built-in version.

That commit must be reverted in order to restore the correct functionality.

## Impact

None other than the ruined feature should become functional again.

## Testing

Under test by floweh@gmx.net.  Refer to discussion of mid-May 2020.

### DIFF
--- a/nshlib/nsh_parse.c
+++ b/nshlib/nsh_parse.c
@@ -1,36 +1,20 @@
 /****************************************************************************
  * apps/nshlib/nsh_parse.c
  *
- *   Copyright (C) 2007-2013, 2014, 2017-2018 Gregory Nutt. All rights
- *     reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ * Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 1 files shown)


patacongo:
nxstyle fixes forthcoming.

PR #258, opened 2020-05-18
nsh/parse: Try FILE_APPS first in the case of builtin

anchao:
## Summary
nsh/parse: Try FILE_APPS first in the case of builtin
Bring back the BUILTIN_APPS / FILE_APPS calling sequence


patacongo:
BTW:  The code that Florian was running when he found the problem was the example at https://cwiki.apache.org/confluence/display/NUTTX/Updating+a+Release+System+with+ELF+Programs

He was running the "Fully linked" example.  That one does not use a symbol table.  it is fully linked with libc and gets address information from System.map

PR #263, opened 2020-05-20
Build improvment

xiaoxiang781216:
## Summary
The remain nxstyle error is:
```
 error: Mixed case identifier found
```
depend on https://github.com/apache/incubator-nuttx/pull/1088


examples/audio_rttl/Makefile
```diff
@@ -33,10 +33,8 @@
 #
 ############################################################################
 
--include $(TOPDIR)/.config
--include $(TOPDIR)/Make.defs
+include $(TOPDIR)/Make.defs
 -include $(SDKDIR)/Make.defs
-include $(APPDIR)/Make.defs
```
Ouss4, line 39: This makefile needs DELIM, it gets it from apps/Make.defs.  I'm not sure if there is any indirect definition somehow.

examples/elf/tests/hello/Makefile
```diff
@@ -33,16 +33,7 @@
 #
 ############################################################################
 
--include $(TOPDIR)/Make.defs
-
-OBJEXT ?= .o
-DELIM ?= /
-
-ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-NUTTXLIB = "${shell cygpath -w $(TOPDIR)$(DELIM)staging}"
```
Ouss4, line 42: If NUTTXLIB is removed, where doesn't come from? It's used below.

examples/audio_rttl/Makefile
xiaoxiang781216: $(TOPDIR)/Make.defs->$(TOPDIR)/tools/Config.mk->DELIM define here.

xiaoxiang781216: BTW, there is only one DELIM definition now.

PR #266, opened 2020-05-21
Build simplification

xiaoxiang781216:


examples/tcpblaster/tcpblaster_client.c
```diff
@@ -37,8 +37,6 @@
  * Included Files
  ****************************************************************************/
 
-#include "config.h"
```
Ouss4, line 40: This is not an error.  It's the same config.h file but for the "host".  It's copied to the examples directory during build time.
(https://github.com/apache/incubator-nuttx-apps/blob/master/examples/tcpblaster/Makefile#L112)

xiaoxiang781216: Yes, you are right. Iwill remove this patch from PR.

PR #278, opened 2020-05-25
webclient: Fix http -> https redirection

yamt:
http -> https redirection is rather common. The old code was
just broken in that case.

Also, this commit is a step towards https support.

* Switch to netlib_parseurl
* Fix error propagation in wget_parseheaders
* Bail out on a redirect to a URL with unsupported scheme


yamt:
this fixes https://github.com/apache/incubator-nuttx-apps/issues/273

PR #285, opened 2020-06-05
system: nxplayer: Refactor nxplayer.c

masayuki2009:
## Summary

- This PR removes redundant code regarding audio buffer configuration in nxplayer.c
- In previous implementation, there were many #ifdef CONFIG_AUDIO_DRIVER_SPECIFIC_BUFFERS which resulted in complicated code.
- This PR simplifies nxplayer.c which ** always ** calls ioctl() to the nuttx audio driver to obtain buffer configurations and stores the information to buf_info. If ioctl() failed which means that underlying audio driver does not support 'audio driver specific buffers', then use default values such as CONFIG_AUDIO_BUFFER_NUMBYTES and CONFIG_AUDIO_NUM_BUFFERS

## Impact

- This PR affects nxplayer

## Testing

- I tested this PR with both spresense which has CONFIG_AUDIO_DRIVER_SPECIFIC_BUFFERS=y and lc823450-xgevk which ** does not ** have CONFIG_AUDIO_DRIVER_SPECIFIC_BUFFERS=y

### DIFF
--- a/system/nxplayer/nxplayer.c
+++ b/system/nxplayer/nxplayer.c
@@ -80,14 +80,6 @@
 #define NXPLAYER_STATE_PLAYING   1
 #define NXPLAYER_STATE_PAUSED    2
 
-#ifndef CONFIG_AUDIO_NUM_BUFFERS
-#  define CONFIG_AUDIO_NUM_BUFFERS  2
-#endif
-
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
LGTM, nxrecorder can do the same improvement too.

PR #289, opened 2020-06-10
examples/lvgldemo: Make touch calibration optional

no1wudi:
## Summary
Some touch screen don't need calibration,  so let the touch calibration as optional
## Impact
No impact for current board configs, touch calibration is enabled by default
## Testing
It tested with stm32f746g-disco and my custom board

### DIFF
--- a/examples/lvgldemo/Kconfig
+++ b/examples/lvgldemo/Kconfig
@@ -11,6 +11,13 @@ menuconfig EXAMPLES_LVGLDEMO
 
 if EXAMPLES_LVGLDEMO
 
+config EXAMPLES_LVGLDEMO_CALIBRATE
+	bool "Calibrate touchscreen"
+	default y
+	---help---
+		Calibrate touchscreen before demo start, but some touchscreen
+		don't need it, like capacitive touchscreen.
+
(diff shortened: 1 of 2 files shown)


examples/lvgldemo/Kconfig
```diff
@@ -11,6 +11,13 @@ menuconfig EXAMPLES_LVGLDEMO
 
 if EXAMPLES_LVGLDEMO
 
+config EXAMPLES_LVGLDEMO_CALIBRATE
+	bool "Calibrate touchscreen"
+	default y
```
xiaoxiang781216, line 16: Change to n? more and more touch technology don't need calibrate anymore

no1wudi: But most boards currently supported need it， like stm32f429i-disco, etc
If change it to n, we need to update these boards defconfig

xiaoxiang781216: Ok.

acassis: Yes, some boards that support the lvgl-demo are resistive touchscreen...

PR #291, opened 2020-06-11
system: nxrecorder: Fix compile warnings in nxrecorder.c

masayuki2009:
## Summary

- This PR fixes compile warnings in nxrecorder.c

## Impact

- This PR affects nxrecorder only.

## Testing

- I tested this PR with spresense:wifi (need to add nxrecorder to defconfig)

### DIFF
--- a/system/nxrecorder/nxrecorder.c
+++ b/system/nxrecorder/nxrecorder.c
@@ -216,7 +216,7 @@ static int nxrecorder_enqueuebuffer(FAR struct nxrecorder_s *precorder,
   bufdesc.session   = precorder->session;
 #endif
   bufdesc.numbytes  = apb->nbytes;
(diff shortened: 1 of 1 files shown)


btashton:
@masayuki2009 Can we add this to a defconfig as well that way we can get better coverage on the CI for this?

masayuki2009:
> @masayuki2009 Can we add this to a defconfig as well that way we can get better coverage on the CI for this?

Yes. I will send another PR to the nuttx after this PR is merged.

btashton:
Ok, I'll merge this now since the fix is minor and the CI system wont capture this anyway.

masayuki2009:
> Ok, I'll merge this now since the fix is minor and the CI system wont capture this anyway.

@btashton 
Thanks!

PR #293, opened 2020-06-15
Include malloc.h instead of stdlib.h for mallinfo()

yamt:
Adapt to the change in the main repo.

mallinfo is meant to be API compatible with Linux,
where it's provided by malloc.h.

(I think the API actually originated with System V. I don't
remember how it was there though. Anyway, I guess the
compatibility with Linux is more important than System V
these days.)


yamt:
the associated PR in the main repo is https://github.com/apache/incubator-nuttx/pull/1236

PR #294, opened 2020-06-15
net: SocketCAN support and SocketCAN CAN-utils tools

PetervdPerk-NXP:
## Summary
Adds SocketCAN (AF_CAN) support to netlib require for https://github.com/apache/incubator-nuttx/pull/1238.

And adds SocketCAN cansend & candump utilities ported from https://github.com/linux-can/can-utils

## Testing

Verified on rddrone-uavcans32k146 (S32K146) & rddrone-fmuk66 (Kinetis K66).


patacongo:
Since this change includes third party code, we will need to update the LICENSE file.  Perhaps we can do this after a merge.

PR #303, opened 2020-06-24
Make programs under apps/bin executable to generate symtab file rightly

liuguo09:
## Summary
Make programs under apps/bin executable since tools/mksymtab.sh called with
'find $dir -type f -perm -a=x 2>/dev/null'. So generate symtab file rightly.

## Testing
Verify with sim:loadable config. Or running nuttx with sim loadable config got stuck.


masayuki2009:
@liuguo09 @xiaoxiang781216 

Thanks fot the fix. It works!.

PR #306, opened 2020-06-24
 make/clean: clean the OBJ directly if declare in subdirectory 

anchao:
## Summary
 make/clean: clean the OBJ directly if declare in subdirectory 

## Impact
Depends on https://github.com/apache/incubator-nuttx/pull/1297

## Testing
make clean


anchao:
well, I gave up, conflict with VPATH

PR #316, opened 2020-07-01
apps: Remove all stuff related to CONFIG_xxx_CXXINITIALIZE

xiaoxiang781216:
since it is moved to the central place in nuttx side instead

Change-Id: I544d6110f1ca6460f7c82f970870aa9b1e7ab3dd

## Summary
depends on: https://github.com/apache/incubator-nuttx/pull/1341


xiaoxiang781216:
> This should NOT me merged [apache/incubator-nuttx#1341](https://github.com/apache/incubator-nuttx/pull/1341) is stalled. It is the wrong approach .

Yes, we shouldn't merge this patch before [apache/incubator-nuttx#1341](https://github.com/apache/incubator-nuttx/pull/1341). I provide this patch just show the current siutation as reference. Whether this is wrong approach is still in the discussion, please give you comment in [apache/incubator-nuttx#1341](https://github.com/apache/incubator-nuttx/pull/1341) instead.

patacongo:
@xiaoxiang781216 I will go ahead and merge this now.  it still cannot pass the PR checks  because some defconfig files set appliation C++ settings that no longer exist, for example.  esp32-core/nsh

This will break all all PR checks until all of the relevant defconfig files are updated in the incubator_nuttx repository.

patacongo:
> This will break all all PR checks until all of the relevant defconfig files are updated in the incubator_nuttx repository.

I will submit the the PR that fixes the PR checks in a few minutes.

xiaoxiang781216:
Yes, here is the patch: https://github.com/apache/incubator-nuttx/pull/1347

davids5:
@xiaoxiang781216 would you be kind enough to fill in the #Impact section and give porting instruction?

xiaoxiang781216:
@davids5 the change is compatable as before since all config which utilize the global constructor should already define CONFIG_HAVE_CXXINITIALIZE. The user don't take any special action and the feature still work as it, the only difference is that some config doesn't need anymore(e.g. CONFIG_EXAMPLES_HELLOXX_CXXINITIALIZE or SYSTEM_NSH_CXXINITIALIZE), but there isn't any bad side effect if user forget to remove these config from their defconfig file.

PR #317, opened 2020-07-01
nshlib: Call symlink if user pass -s for ln command

xiaoxiang781216:
## Summary
since NuttX kernel support really support symlink not link.
Note: link equal symlink now because the hard link doesn't support yet

## Impact
No, since link equal symlink in the current implementation


xiaoxiang781216:
The related patch:
https://github.com/apache/incubator-nuttx/pull/1279
https://github.com/apache/incubator-nuttx/pull/1343

PR #318, opened 2020-07-01
zmodem: Remove CONFIG_HAVE_INLINE from compiler.h

xiaoxiang781216:
## Summary
follow up the kernel side change

## Impact
No function chnage since host binary don't use these variables.


xiaoxiang781216:
Follow up the kernel change: https://github.com/apache/incubator-nuttx/pull/1321

PR #334, opened 2020-07-15
Don't copy headers outside of LittlevGL library

w8jcik:
## Summary

LittlevGL is an example of library that copies it's header outside of it's folder.

This PR removes such behavior. Library exposes one of it's folders as include directory instead.

LittlevGL is just an example and this PR is a question if you accept such change.

## Impact

Applications are not separable because they have some files outside of their folders. In particular LittlevGL is copying `lvgl.h` to `include/graphics/lvgl.h` when building. This file organisation doesn't seem to help with anything. If someone finds it useful to have one `include` directory with all possible libraries it also doesn't work because normally `lvgl.h` header is not there.

Some other libraries also have files in both their directories and in `include`.

If this change is accepted I can alter the other libraries with same issue.

Long term purpose of this is to be able to plug-in and out libraries. (package management)

## Testing

```
mkdir test
(
  cd test
  git clone https://github.com/apache/incubator-nuttx.git nuttx
  git clone --branch app-separation https://github.com/w8jcik/incubator-nuttx-apps.git apps
  (
    cd nuttx
    tools/configure.sh stm32f4discovery/nsh
    make menuconfig
    # enable GRAPHICS_LVGL, EXAMPLES_LVGLDEMO and INPUT
    make -j 12
  )
)
```

http://nuttx-config.nxtlabs.pl/#/?filter=LVGL

It is just a build system change and it builds. Also if anything is wrong with `apps/graphics/lvgl`, the `apps/examples/lvgldemo` fails to build which shows that one is using the other. I didn't run it on a board.

## Breaking change

Use `#include <lvgl/lvgl.h>` or `#include <lvgl.h>` in place of `#include <graphics/lvgl.h>`.


graphics/lvgl/Makefile
```diff
@@ -78,20 +78,21 @@ $(LVGL_UNPACKNAME): $(LVGL_TARBALL)
...
-lvgl/lvgl.h: $(LVGL_UNPACKNAME)
+$(LVGL_UNPACKNAME)/lvgl.h: $(LVGL_UNPACKNAME)
 
-$(APPDIR)/include/graphics/lvgl.h: lvgl/lvgl.h
-	@echo "CP: lvgl/lvgl.h"
-	$(Q) cp lvgl/lvgl.h $(APPDIR)/include/graphics/lvgl.h
+exports/graphics/lvgl.h: $(LVGL_UNPACKNAME)/lvgl.h
+	$(Q) mkdir -p exports/graphics
+	@echo "CP: $(LVGL_UNPACKNAME)/lvgl.h"
+	$(Q) cp $(LVGL_UNPACKNAME)/lvgl.h exports/graphics/lvgl.h
```
xiaoxiang781216, line 86: Why copy lvgl.h to exports directory? How about we export?
CFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" $(APPDIR)/graphics/lvgl}
CXXFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" $(APPDIR)/graphics/lvgl}

w8jcik: Sure, but then we need to `#include <lvgl.h>` instead of `#include <graphics/lvgl.h>`. I don't mind but I though that this prefix is intentional to keep things within namespaces like `graphics` here. Do we want keep prefixes or do they go? So far they were used consistently.

Here is additional PR which shows necessary change to introduce this suggestion https://github.com/w8jcik/incubator-nuttx-apps/pull/1/files

w8jcik: Update:

Acutally it will be `#include <lvgl/lvgl.h>` not <lvgl.h>, there was one more include present in `lvgldemo` which I just noticed.

https://github.com/w8jcik/incubator-nuttx-apps/pull/1/commits/7f62736c771922aa8c8e6bd3d4976d083ef66ef8

But this is LittlevGL specific. The question remains, what to do with prefixes.

v01d: Usually libraries assume some sort of layout for the include directory and the user only needs to account for installation prefix. In this case, lvgl is assuming a layout of <prefix>/lvgl/lvgl.h, so any extra component in the path should be part of the -I flag passed to the compiler. Adding extra prefixes could break existing programs following the expected inclusion format.

xiaoxiang781216: We should follow LVGL convention:
https://github.com/lvgl/lv_examples/blob/master/lv_examples.h#L16
either ```#include <lvgl.h>``` or ```#include <lvgl/lvgl.h> is good, but ```#include <graphics/lvgl.h>``` just make the life of programmer is hard.

acassis: I think "#include <lvgl/lvgl.h>" as Maciej modified is fine. Currently only "lvgldemo" is using the LVGL. Maciej, avoid using the name LittlevGL because it is "deprecated", see: https://blog.lvgl.io/2020-06-01/announcement

w8jcik: Hi @v01d the prefix is already present, because currently littlevgl is copying some header to `graphics`. Change suggested by @xiaoxiang781216 removes this prefix. I was asking should we really remove it, because this way we break compatibility. Breaking compatibility is fine for me. But what is the official strategy. Do we want `graphics` in front of all lvgl includes? Right now it seems to be in front of one of the headers and probably not in front of the others. I think it should be still possible to have `#include <graphics/lvgl/...>`. The question is do we want it.

w8jcik:
Ok, it should be fine now. I have three more questions:

* Is what I put into README correct? Regarding framebuffer device.
* Where should I put deprecation note that users should use `lvgl/lvgl.h` instead of `graphics/lvgl.h` now?
* Can I fix other libraries that sit in `apps/include`?

examples/lvgldemo/Makefile
```diff
@@ -86,11 +86,18 @@ endif
...
-LVGL_DIR=$(APPDIR)/graphics/lvgl
+# Upstream `lv_examples/lv_examples.h` includes LVGL
+# in following way `#include "../lvgl/lvgl.h"`, for no clean reason.
+#
+# The way that we export LVGL offers `#include <lvgl/lvgl.h>`, so it won't work.
+# Solution below is not pretty but it fixes `lv_examples` without patching.
+#
+# Remove this fix when you will get better idea.
+
+LVGL_EXAMPLE_INCLUDE_FIX=$(APPDIR)/graphics/lvgl/lvgl
```
xiaoxiang781216, line 97: We can mdoify the inclusion to lvgl.h by LV_LVGL_H_INCLUDE_SIMPLE macro like this:
CFLAGS += ${shell $(DEFINE) "$(CC)" LV_LVGL_H_INCLUDE_SIMPLE}
Here is the related code: https://github.com/lvgl/lv_examples/blob/master/lv_examples.h#L16

graphics/lvgl/Makefile
```diff
@@ -78,20 +76,15 @@ $(LVGL_UNPACKNAME): $(LVGL_TARBALL)
...
 
-lvgl/lvgl.h: $(LVGL_UNPACKNAME)
-
-$(APPDIR)/include/graphics/lvgl.h: lvgl/lvgl.h
-	@echo "CP: lvgl/lvgl.h"
-	$(Q) cp lvgl/lvgl.h $(APPDIR)/include/graphics/lvgl.h
+$(LVGL_UNPACKNAME)/lvgl.h: $(LVGL_UNPACKNAME)
 
-context:: $(LVGL_UNPACKNAME) $(APPDIR)/include/graphics/lvgl.h
+context:: $(LVGL_UNPACKNAME) lvgl/lvgl.h
```
xiaoxiang781216, line 81: Remove lvgl/lvgl.h, don't need anymore.

graphics/lvgl/Makefile
```diff
@@ -78,20 +76,15 @@ $(LVGL_UNPACKNAME): $(LVGL_TARBALL)
 	$(Q) mv	lvgl-$(LVGL_VERSION) $(LVGL_UNPACKNAME)
 	$(Q) touch $(LVGL_UNPACKNAME)
 
-lvgl/lvgl.h: $(LVGL_UNPACKNAME)
-
-$(APPDIR)/include/graphics/lvgl.h: lvgl/lvgl.h
-	@echo "CP: lvgl/lvgl.h"
-	$(Q) cp lvgl/lvgl.h $(APPDIR)/include/graphics/lvgl.h
+$(LVGL_UNPACKNAME)/lvgl.h: $(LVGL_UNPACKNAME)
```
xiaoxiang781216, line 79: Remove too

xiaoxiang781216:
@w8jcik please squash all commit into one by git rebase --interactive HEAD~8. Also we need remove lvgl.h from apps/include/graphics/.gitignore too.

graphics/lvgl/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+Upstream example is present at `examples/lvgldemo`.
+
+LVGL can be used with framebuffer device. To find example boards with this
+preconfigured, search for `CONFIG_GRAPHICS_LVGL=y` in `defconfig` files.
+All of them have also `CONFIG_VIDEO_FB=y` present.
+
+As a second option, LVGL can talk to a display driver and explicitly draw line by line.
+For this case, there is no preconfigured board present. Go to _Porting_ section
+of upstream documentation for more hints.
```
xiaoxiang781216, line 13: Not sure the second approach is possible.

xiaoxiang781216:
> Ok, it should be fine now. I have three more questions:
> 
> * Is what I put into README correct? Regarding framebuffer device and direct drawing?

The description for framebuffer is good, but I am not sure thtat it's possible to do the direct drawing. 

> * Where should I put deprecation note that users should use `lvgl/lvgl.h` instead of `graphics/lvgl.h` now?

Yes, I think we should follow the convention from the upstream project.

> * Can I fix other libraries that sit in `apps/include`?

The modulization is a good practice, but other people may has other idle, you can send a query @dev.

graphics/lvgl/README.md
w8jcik: Yes, I am using it. Although I bypassed NuttX display driver because it was considerably slower at that time (NuttX 7.24). Or I didn't know how to use it better.

NuttX framebuffer as far as I understand needs width x height x depth amount of memory. It means that it requires external RAM. But LVGL don't have such limitation. It can draw almost directly with some few kilobyte buffer.

v01d: In my own use of littevgl in NuttX I actually draw using a custom LCD character driver to avoid the overhead of using the framebuffer->LCD emulation layer of NuttX (you can see the driver here: https://gitlab.com/nuttx_projects/lcd_dev). 
Just FYI.

w8jcik: So there is some emulation layer that can be enabled? Maybe we should add this information and your custom solution to the README to help people who want to use it.

v01d: No, this is not standard, I use it on my own code. I could propose it in a pull-request. I never did since I extended the LCD interface to accept drawing multiple rows at once (also due to overhead of individual putrun()), but I could leave that out. I would not add something to the README if it is not part of NuttX codebase itself.

w8jcik: The topic of drawing multiple-lines comes back from time to time. I would also be in favor of having such interface beacuse it seems to be a basic use case. Somebody was trying to add it few months ago, but I don't know where it ended up.

I understand, these READMEs seems to be more serious. For me README is just a way to communicate. If someone needs to know something and there are some solutions, README would be the place.

Anyway I think it is enough for this PR then.

xiaoxiang781216: @v01d why not enhance the driver layer? @w8jcik also look for a solution which mean this is a common requirement.

examples/lvgldemo/Makefile
w8jcik: For this to work I added one more export in LVGL. Now LVGL can be included both with `lvgl/lvgl.h` and `lvgl.h` paths.

`CXXFLAGS` as well, right?

Done.

graphics/lvgl/Makefile
w8jcik: Done

graphics/lvgl/Makefile
w8jcik: Done

w8jcik:
> The modulization is always a good practice, but other people may has other ideal, you can send a query @dev.

I was particularly interested in LittlevGL right now. I might come back to this later.

w8jcik:
Done

examples/lvgldemo/Make.defs
```diff
@@ -36,4 +36,6 @@
 
 ifneq ($(CONFIG_EXAMPLES_LVGLDEMO),)
 CONFIGURED_APPS += $(APPDIR)/examples/lvgldemo
+CFLAGS += ${shell $(DEFINE) "$(CC)" LV_LVGL_H_INCLUDE_SIMPLE}
```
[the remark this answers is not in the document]
w8jcik: Done

graphics/lvgl/README.md
v01d: @xiaoxiang781216 I can open a PR for my LCD character driver. As per the ability to send multiple rows, I propose that on the mailing list and did not get much response. We can discuss it further again.

xiaoxiang781216: At least, @w8jcik is also interesting your solution. We can talk about the detail in your PR.

PR #336, opened 2020-07-19
Update boards matrix to add AVR builds.

Ouss4:
## Summary
The checks are failing because the change is only in testing/. This updates apps/.


xiaoxiang781216:
The error is due to https://github.com/apache/incubator-nuttx/pull/1423, let's merge this first.

Issue #340, opened 2020-07-24
Is preconfig necessary for application directories?


Ouss4:
This is somewhat related to some of the issues pointed out here: https://github.com/apache/incubator-nuttx/issues/1407

Most of the configuration phase output is coming from running the `preconfig` target inside all the application directories.
(By application directories I'm referring to the lowest level, for instance: apps/examples/adc)
That target is inherited from `Application.mk` and applications and examples could then extend it.
As of right now, none of the applications present in the repo use that target.  I also looked for some examples in other repositories but I couldn't find any.

The addition came as part of https://github.com/apache/incubator-nuttx-apps/commit/a9aa753fa4d282bc0aa48333ece593bdb5dc707b which added the dynamic top-level Kconfig generation.
This prepares the top-level Kconfig files, at the level of application directory there is no more preparation.  So I'm wondering if someone is using this for any other purpose?  If so, shouldn't that be part of `context`?

Removing this part will significantly reduce configuration time and also provides a cleaner output.
I have the following patch:

```diff
diff --git a/Application.mk b/Application.mk
index d23492eb..9549f7ac 100644
--- a/Application.mk
+++ b/Application.mk
@@ -96,7 +96,7 @@ VPATH += :.
 # Targets follow

 all:: .built
-.PHONY: clean preconfig depend distclean
+.PHONY: clean depend distclean
 .PRECIOUS: $(BIN)

 define ELFASSEMBLE
@@ -195,8 +195,6 @@ install::

 endif # BUILD_MODULE

-preconfig::
-
 ifeq ($(CONFIG_NSH_BUILTIN_APPS),y)
 ifneq ($(PROGNAME),)
 ifneq ($(PRIORITY),)
diff --git a/Directory.mk b/Directory.mk
index 66fbbdab..e04b414e 100644
--- a/Directory.mk
+++ b/Directory.mk
@@ -43,7 +43,6 @@ all: nothing

 .PHONY: nothing context depend clean distclean

-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),preconfig)))
 $(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),context)))
 $(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),depend)))
 $(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),clean)))
[...]

PR #341, opened 2020-07-24
Don't preconfig, clean or distclean when it's not necessary

Ouss4:
## Summary
This PR has 2 main commits.  The first one related to #340.  It prevents the `preconfig` target from executing in application directories that don't actually need any pre-configuration.
The second one limits the execution scope of `clean` and `distclean` to the folder that we know we have built, instead of recursively going through every sub-directory.

## Impact
The `preconfig` commit reduces configuration time.
The `clean` and `distclean` commit reduces cleaning time.
Both remove a bunch of unnecessary output.
The whole `./tools/configure.sh make make distclean` procedure shall be faster.
## Testing
A bunch of configurations with different examples, hello, helloxx, ostest, etc.

### DIFF
--- a/Directory.mk
+++ b/Directory.mk
@@ -35,38 +35,38 @@
 
 include $(APPDIR)/Make.defs
 
-# Sub-directories
+# Sub-directories that have been built or configured.
 
-SUBDIRS = $(dir $(wildcard */Makefile))
+SUBDIRS       := $(dir $(wildcard *$(DELIM)Makefile))
+CONFIGSUBDIRS := $(filter-out $(dir $(wildcard *$(DELIM)Kconfig)),$(SUBDIRS))
+CLEANSUBDIRS  := $(dir $(wildcard *$(DELIM).built))
+CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).depend))
+CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).kconfig))
 
 all: nothing
 
-.PHONY: nothing context depend clean distclean
+.PHONY: nothing clean distclean
 
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),preconfig)))
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),context)))
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),depend)))
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),clean)))
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),distclean)))
+$(foreach SDIR, $(CONFIGSUBDIRS), $(eval $(call SDIR_template,$(SDIR),preconfig)))
+$(foreach SDIR, $(CLEANSUBDIRS), $(eval $(call SDIR_template,$(SDIR),clean)))
+$(foreach SDIR, $(CLEANSUBDIRS), $(eval $(call SDIR_template,$(SDIR),distclean)))
 
 nothing:
 
 install:
 
-preconfig: $(foreach SDIR, $(SUBDIRS), $(SDIR)_preconfig)
+preconfig: $(foreach SDIR, $(CONFIGSUBDIRS), $(SDIR)_preconfig)
 ifneq ($(MENUDESC),)
 	$(Q) $(MKKCONFIG) -m $(MENUDESC)
+	$(Q) touch .kconfig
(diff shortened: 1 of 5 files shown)


patacongo:
> The second one limits the execution scope of `clean` and `distclean` to the folder that we know we have built, instead of recursively going through every sub-directory.

This behavior was intentional.  If you do not clean ALL directories, then when you reconfigure and disable an app, the object files and other products will be stranded in the directory with no way to clean them.  You should not make this change; we must continue to clean in all directories.

If this is merged there will be much breakage (it used to work this way years ago, but the behavior was wrong on a reconfiguration).

Ouss4:
> This behavior was intentional. If you do not clean ALL directories, then when you reconfigure and disable an app, the object files and other products will be stranded in the directory with no way to clean them

Yes, I was careful regarding this part.
The directories that are cleaned or distcleaned are the directories that contain those object files.  So even if an app is disabled, its directory will still be cleaned during the next clean/distclean.

patacongo:
> The original bad version just cleaned in CONFIGURED_APPS. That was the version that did not work because it stranded objects when apps were removed from CONFIGURED_APPS. Your logic is smarter; it looks at some landmarks to see if the directory was built.

I believe that the nuttx/ repository still has this problems.  Objects will get left in directories if you configure and CLEANDIRS changes.

Ouss4:
> I believe that the nuttx/ repository still has this problem. Objects will get left in directories if you reconfigure and CLEANDIRS changes.

I gave it a quick try by enabling a driver, building, disabling the driver then clean.  It seems that its object files are removed.  But that wasn't a thorough investigation.

Regarding the config that's failing, it's not related to this change.  It seems that stm32f4discovery:hciuart defconfig has changed.  It builds fine but refresh.sh fails.

patacongo:
> > I believe that the nuttx/ repository still has this problem. Objects will get left in directories if you reconfigure and CLEANDIRS changes.
> 
> I gave it a quick try by enabling a driver, building, disabling the driver then clean. It seems that its object files are removed. But that wasn't a thorough investigation.

I think that the CLEANDIRS only addresses top-level directories.  So you would only see the issue, for example, if you built with networking, then disabled networking and rebuilt.

I have often seen leftover objects under nuttx/ and I am think that I often saw them under boards and arch, but I can't remember.  Anyway, that is a different subject than this PR.

Directory.mk
```diff
@@ -35,36 +35,30 @@
...
-.PHONY: nothing context depend clean distclean
+.PHONY: nothing clean distclean
 
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),preconfig)))
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),context)))
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),depend)))
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),clean)))
-$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),distclean)))
+$(foreach SDIR, $(CLEANSUBDIRS), $(eval $(call SDIR_template,$(SDIR),clean)))
+$(foreach SDIR, $(DISTCLEANSUBDIRS), $(eval $(call SDIR_template,$(SDIR),distclean)))
```
btashton, line 48: Should these be moved to a make target so they can use parallelism of '-j'?

Ouss4:
>  I often saw them under boards and arch

This rings a bell.  I've seen some myself too, I'll take a look after I'm done with this.

@patacongo @btashton I'm seeing some Kconfig leftovers with this PR, let me sleep on it and check tomorrow with a fresher mind.

Directory.mk
xiaoxiang781216: @btashton the current form already run parallelly.

Ouss4:
@patacongo @btashton @xiaoxiang781216 could you please give it another look?  All the checks have passed.
I added a bunch of changes to fix some corner cases.

xiaoxiang781216:
LGTM. @patacongo @btashton could you merge the patch if no more concern.

Directory.mk
btashton: @xiaoxiang781216  I was under the impression that a foreach with eval call would not be run with j parallelism

xiaoxiang781216: foreach SDIR_template with will expand to something like this:
```
xxx_clean:
  make -C xxx clean
```
And the follow line:
```
clean: $(foreach SDIR, $(CLEANSUBDIRS), $(SDIR)_clean)
```
will expand to:
```
clean: xxx_clean ...
```
so make will execute xxx_clean concurrently.

PR #333, opened 2020-07-12
webclient improvements

yamt:
## Summary
Highlights:

* TLS support (a hook to allow users to provide TLS implementation)
* ability to add extra request headers
* ability to use PUT method
* ability to report http status
* error handling improvements

Proposed on the ML while ago:
https://www.mail-archive.com/dev@nuttx.apache.org/msg03803.html

## Impact
The original API is kept for now.
I plan to remove them after adapting the existing users.
(examples in this repo)

## Testing
tested with my local app

PR #346, opened 2020-07-28
Update wget

SPRESENSE:
## Summary

Improve examples/wget

- Enable URL from command line
- It can be used with ifup command
- Add HTTPS protocol

## Impact

only wget

## Testing

Tested by spresense:wifi


yamt:
a competing PR: https://github.com/apache/incubator-nuttx-apps/pull/333

xiaoxiang781216:
@yamt there is nxstyle warning need to fix.

yamt:
@xiaoxiang781216 this is not my PR.

xiaoxiang781216:
@yamt, sorry I make a mistake, #346 and #333 are modifing the same files. @SPRESENSE please add a new patch to address the style issue, thanks.

yamt:
the system-global tls design in this patch doesn't work for my use cases, where i need to use a few types of tls sessions. (different set of credentials etc)

@SPRESENSE otoh, does https://github.com/apache/incubator-nuttx-apps/pull/333 work for your use case?

SPRESENSE:
@yamt 

Sorry too late.
I removed SSL support patch, because you already pushed.
However, I have some questions for your patch, maybe I'll comment on #333 later.

@xiaoxiang781216 

Could you please review again?

netutils/webclient/webclient.c
```diff
@@ -658,12 +658,18 @@ static int wget_base(FAR const char *url, FAR char *buffer, int buflen,
...
+          ret = send(sockfd, buffer + ((dest - buffer) - len), len, 0);
+          if (ret < 0)
+            {
+              nerr("ERROR: send failed: %d\n", errno);
+              goto errout;
+            }
+
+          len -= ret;
         }
+      while (len > 0);
```
yamt, line 672: does this mean short-writes on a blocking socket?

examples/wget/Kconfig
```diff
@@ -12,11 +12,18 @@ config EXAMPLES_WGET
 
 if EXAMPLES_WGET
 
+config EXAMPLES_WGET_STACKSIZE
+	int "wget stack size"
+	default 2048
```
xiaoxiang781216, line 17: should we use DEFAULT_TASK_STACKSIZE here?

examples/wget/Kconfig
```diff
@@ -12,11 +12,18 @@ config EXAMPLES_WGET
...
+	---help---
+		Stack size of "wget" command. If you use SSL connection,
+		you should increase this value around 8096.
+
 config EXAMPLES_WGET_URL
```
xiaoxiang781216, line 22: should we remve this option? since it get from command line now.

examples/wget/Kconfig
```diff
@@ -12,11 +12,18 @@ config EXAMPLES_WGET
...
 	string "File URL"
 	default ""
 	---help---
-	The URL of the file to get
+		The URL of the file to get
 
 config EXAMPLES_WGET_NOMAC
```
xiaoxiang781216, line 28: Should we guard this by NSH_NETINIT?

netutils/webclient/webclient.c
SPRESENSE: Yes.

examples/wget/Kconfig
SPRESENSE: Actually, this patch is for SSL support, but useful in the future.

examples/wget/Kconfig
SPRESENSE: I have no idea.
It looks that not related with NSH_NETINIT, isn't it?

examples/wget/Kconfig
SPRESENSE: Is it possible to use both of command line and option? I found it used in some defconfigs.

examples/wget/Kconfig
xiaoxiang781216: Yes, but it's better to change 2048 to DEFAULT_TASK_STACKSIZE, so we can chagne all program stack size(e.g. sim).

examples/wget/Kconfig
xiaoxiang781216: But your patch don't reference EXAMPLES_WGET_URL anymore and the code return -1 if user don't pass a URL. So please make your code and Kconfig consistent with each other.

examples/wget/Kconfig
xiaoxiang781216: I mean change to:
```
config EXAMPLES_WGET_NOMAC
depends on !NSH_NETINIT
```
It doesn't make sense to let user select mac related option if NSH_NETINIT turn on because your code skip all related code in this case.

examples/wget/Kconfig
SPRESENSE: Sorry, my previous comment was mistake.

I know this patch is wrong.
I think it has 2 ways for fix it. 1) completely remove EXAMPLES_WGET_URL option and 2) to be able to use both of command line and option.
Do you think which better ones?

examples/wget/Kconfig
SPRESENSE: I misunderstood. I'll fix it.

examples/wget/Kconfig
xiaoxiang781216: Both are fine for me.

SPRESENSE: Thanks. I'll fix it to both.

examples/wget/Kconfig
SPRESENSE: I see. I've fixed it and main code.

xiaoxiang781216:
LGTM.

PR #353, opened 2020-08-07
interpreters/duktape: Fix tarball download url

no1wudi:
## Summary
Fix error line wrap for download URL in Makefile.
## Testing
Tested on sim & stm32f746g-disco.


xiaoxiang781216:
@no1wudi could you upstream sim config to enable the check?

PR #356, opened 2020-08-08
interpreters/duktape: Fix build warning for sim on macOS

no1wudi:
## Summary
See #355 
## Impact
sim:duktape on macOS


interpreters/duktape/duk_cmdline.patch
```diff
@@ -21,3 +21,15 @@ diff --color -ur duktape-2.5.0/src-noline/duk_config.h duktape-2.5.0-modifed/src
 -#define DUK_USE_DATE_PRS_STRPTIME
 +#undef DUK_USE_DATE_PRS_STRPTIME
  #define DUK_USE_DATE_FMT_STRFTIME
+
+
+diff --color -ur duktape-2.5.0/src-noline/duk_config.h duktape-2.5.0-modifed/src-noline/duk_config.h
```
xiaoxiang781216, line 26: can we merge the change into one?

no1wudi: The config header detect the sim platform as Linux or macOS even unix, so we need these changes.

Maybe we can add NuttX platform defination into duktape.

PR #357, opened 2020-08-09
netutils/ftpc: Embed wdog_s in ftpc_session_s directly

xiaoxiang781216:
follow up the kernel side change: https://github.com/apache/incubator-nuttx/pull/1545

### DIFF
--- a/netutils/ftpc/ftpc_connect.c
+++ b/netutils/ftpc/ftpc_connect.c
@@ -103,10 +103,6 @@ SESSION ftpc_connect(FAR union ftpc_sockaddr_u *server)
 
   session->homeldir = strdup(ftpc_lpwd());
 
-  /* Create up a timer to prevent hangs */
-
-  session->wdog = wd_create();
-
   /* And (Re-)connect to the server */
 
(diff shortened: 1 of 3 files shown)


netutils/ftpc/ftpc_connect.c
```diff
@@ -152,8 +148,8 @@ int ftpc_reconnect(FAR struct ftpc_session_s *session)
 
   /* Set up a timer to prevent hangs */
 
-  ret = wd_start(session->wdog,
-                 session->conntimeo, ftpc_timeout, 1, session);
+  ret = wd_start(&session->wdog, session->conntimeo,
```
patacongo, line 151: Hmm.. this is another violation of the portable POSIX OS interface.  wd_start(0 is not an application accessible interface.  It is only for use inside the OS.

xiaoxiang781216:
Yes, I will open an issue track this problem.

xiaoxiang781216:
Just see you already opened an issue for this: https://github.com/apache/incubator-nuttx-apps/issues/360

PR #358, opened 2020-08-10
Remove the unnecessary touch and delete from Makefile

xiaoxiang781216:
## Impact
No functionality change.

### DIFF
--- a/graphics/lvgl/Makefile
+++ b/graphics/lvgl/Makefile
@@ -78,9 +78,6 @@ $(LVGL_UNPACKNAME): $(LVGL_TARBALL)
 
 context:: $(LVGL_UNPACKNAME)
 
-clean::
-	$(call DELDIR, build)
-
 distclean::
 	$(call DELDIR, $(LVGL_UNPACKNAME))
 	$(call DELFILE, $(LVGL_TARBALL))
--- a/testing/unity/Makefile
+++ b/testing/unity/Makefile
@@ -68,7 +68,7 @@ $(UNITY_TARBALL):
 $(UNITY_UNPACKNAME): $(UNITY_TARBALL)
 	@echo "Unpacking: $(UNITY_TARBALL) -> $(UNITY_UNPACKNAME)"
 	$(Q) $(UNPACK) $(UNITY_TARBALL)
-	@touch $(UNITY_UNPACKNAME)
+	$(Q) touch $(UNITY_UNPACKNAME)
 
 $(UNITY_SRCDIR)$(DELIM)unity.h: $(UNITY_UNPACKNAME)
 
@@ -82,12 +82,8 @@ $(APPS_INCDIR)$(DELIM)unity_internals.h: $(UNITY_SRCDIR)$(DELIM)unity_internals.
 
 context:: $(APPS_INCDIR)$(DELIM)unity.h $(APPS_INCDIR)$(DELIM)unity_internals.h
 
-clean::
-	$(call DELDIR, build)
-
 distclean::
 	$(call DELDIR, $(UNITY_UNPACKNAME))
-	$(call DELFILE, .downloaded)
 	$(call DELFILE, $(UNITY_TARBALL))
 	$(call DELFILE, $(APPDIR)/include/testing/unity.h)
 	$(call DELFILE, $(APPDIR)/include/testing/unity_internals.h)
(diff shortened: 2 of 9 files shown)


graphics/lvgl/Makefile
```diff
@@ -74,13 +74,9 @@ $(LVGL_UNPACKNAME): $(LVGL_TARBALL)
 	@echo "Unpacking: $(LVGL_TARBALL) -> $(LVGL_UNPACKNAME)"
 	$(Q) $(UNPACK) $(LVGL_TARBALL)
 	$(Q) mv	lvgl-$(LVGL_VERSION) $(LVGL_UNPACKNAME)
-	$(Q) touch $(LVGL_UNPACKNAME)
```
Ouss4, line 77: @xiaoxiang781216 I think touch is still necessary here.
I was puzzled by this one, and thought it was unnecessary, please confirm the following:
1. We extract the folder, say LVGL-v2.2.0 (this folder will keep the timestamps when it was tar'ed)
2. We rename it as LVGL (it also keeps the timestamps of the above)

With that, the tarball is newer than the folder LVGL and the target will execute again with the next `make` and copy an extra LVGL-v2.2.0 inside LVGL.  Now LVGL has changed and no more copies.

With `touch` we update the folder's timestamps and we won't have an extra copy.

Ouss4: BTW, `tar` has the option --touch (or -m) that does make updates the extracted folder's timestaps.  But I believe explicitly calling `touch` is cleaner.

xiaoxiang781216: I tries two package(lvgl and libcanard), both don't hit the double downloading issue. @Ouss4 could you help me confirm?

Ouss4: libcanard's Makefile deletes the folder before unpacking, so we can't see the issue there.
I just tried with lvgl, and I do get a second folder. 
Is that an `mv` implementation difference?  I'm running on Manjaro with Coretutils 8.32, what's your platform?
Could you please try to manually inspect the timestamps with ls -l?

> both don't hit the double downloading issue

It doesn't download the archive again, it just unpacks the existing tar or zip file.

xiaoxiang781216: Here is my test:
```
drwxr-xr-x 9 xiaoxiang xiaoxiang    4096 2020-08-11 17:35:48.659398700 +0800 lvgl
-rw-r--r-- 1 xiaoxiang xiaoxiang 6847211 2020-08-11 17:35:14.369398700 +0800 v7.0.2.zip
```
and mv --version:
```
xiaoxiang@DESKTOP-JH2VEP6:~/nuttx/nuttx$ mv --version
mv (GNU coreutils) 8.28
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Parker, David MacKenzie, and Jim Meyering.
```

Ouss4: So it's your `mv` that's changing the timestamps?
It looks like it's an `mv` version issue.  There are some explanations [here](https://unix.stackexchange.com/questions/503233/when-does-mv-change-the-timestamp-of-a-file-and-while-does-it-not).

I don't think we should rely on the details of `mv` implementation.

xiaoxiang781216: Ok, I keep the touch immediately follow mv, but remove others. Please take a look, @Ouss4.

testing/unity/Makefile
```diff
@@ -68,7 +68,6 @@ $(UNITY_TARBALL):
 $(UNITY_UNPACKNAME): $(UNITY_TARBALL)
 	@echo "Unpacking: $(UNITY_TARBALL) -> $(UNITY_UNPACKNAME)"
 	$(Q) $(UNPACK) $(UNITY_TARBALL)
-	@touch $(UNITY_UNPACKNAME)
```
Ouss4, line 71: This should stay too, unzip or tar will keep the timestamps of what's inside the archive.

xiaoxiang781216: Ok, keep touch for unzip too.

PR #362, opened 2020-08-12
system: add libuv port for NuttX

spiriou:
Summary
From http://docs.libuv.org/en/v1.x/
"libuv is a multi-platform support library with a focus on asynchronous I/O. It was primarily developed for use by Node.js, but it’s also used by Luvit, Julia, pyuv, and others."

Initial port to NuttX from libuv v1.38.1.

Impact
None, disabled by default (CONFIG_LIBUV)

Testing
Select CONFIG_LIBUV and CONFIG_LIBUV_UNIT_TESTS. Then run "libuvtest" in nsh.
All CONFIG_LIBUV_* options must be enabled for all the tests to pass.
Validated on "sim" and "stm32f4discovery" targets.

### DIFF
--- a/system/libuv/Makefile
+++ b/system/libuv/Makefile
@@ -1,0 +1,23 @@
+############################################################################
+# system/libuv/Makefile
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+MENUDESC = "libuv async i/o Library"
+
+include $(APPDIR)/Directory.mk
--- a/system/libuv/libuv/Makefile
+++ b/system/libuv/libuv/Makefile
@@ -1,0 +1,121 @@
+############################################################################
+# system/libuv/Makefile
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+LIBUV_PATCHS ?= $(sort $(wildcard $(LIBUV_DIR)/000*.patch))
+WGET ?= wget
+UNPACK ?= unzip -q
+
+$(LIBUV_DIR)/$(CONFIG_LIBUV_TARBALL):
+	@echo "Downloading: $(CONFIG_LIBUV_TARBALL)"
+	$(Q) $(WGET) $(CONFIG_LIBUV_URL)/$(CONFIG_LIBUV_TARBALL) \
+		-O $(LIBUV_DIR)/$(CONFIG_LIBUV_TARBALL)
+
+$(LIBUV_UNPACKDIR): $(LIBUV_DIR)/$(CONFIG_LIBUV_TARBALL)
+	@echo "Unpacking: $(CONFIG_LIBUV_TARBALL) -> $(LIBUV_UNPACKDIR)"
+	$(call DELDIR, $(LIBUV_UNPACKDIR))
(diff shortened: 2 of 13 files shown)


patacongo:
We need to be careful with this.  We have to handle third party code very differently:  We cannot use the Apache license header and we have to document the use of third party software in the LICENSE file.  See also https://www.apache.org/legal/resolved.html

Most people opt correctly to NOT explicitly include third party software in the repository, but rather to download it on demand and build time and patch it as needed.

xiaoxiang781216:
> We need to be careful with this. We have to handle third party code very differently: We cannot use the Apache license header and we have to document the use of third party software in the LICENSE file. See also https://www.apache.org/legal/resolved.html
> 
> Most people opt correctly to NOT explicitly include third party software in the repository, but rather to download it on demand and build time and patch it as needed.

@papatience libuv is downloaded from the official location, the code in here is the porting layer written by @spiriou , so I think we don't need to mention license specially.

spiriou:
@patacongo @xiaoxiang781216 indeed most of the code+unit-tests is downloaded from https://github.com/libuv/libuv/. I added in the patch the apache license (with the libuv license below) to protect the modifications to support NuttX on those files.

There are still few exceptions. Those files are issued from a copy/paste, and modified so they can work on NuttX. The modifications are too heavy to be managed with #ifdef #endif. What I did is reuse the libuv license, with apache license on top of it. Those files can be moved in the patch instead.
system/libuv/libuv/nuttx_async.c
system/libuv/libuv/nuttx_stream.c
system/libuv/libuv/nuttx_threadpool.c
system/libuv/libuv/nuttx.c
system/libuv/libuv/nuttx_tcp.c

system/libuv/libuv/nuttx_timer.c: that one is done from scratch so no license issue. It just implements the libuv API for that feature.

system/libuv/Makefile
```diff
@@ -0,0 +1,52 @@
...
+	$(Q) touch $@
+
+depend: .depend
+
+distclean:
+	$(call DELFILE, .depend)
+	$(call DELDIR, $(LIBUV_UNPACKDIR))
+	$(call DELFILE, $(CONFIG_LIBUV_TARBALL))
+
+include $(APPDIR)/Directory.mk
```
Ouss4, line 52: Is this really a "Directory" type Makefile?  It looks like an Application type.

spiriou: I'm not sure about this. I want to build the test executable (system/libuv/tests/Makefile) which uses "Application.mk", and also the libuv library (system/libuv/libuv/Makefile) that I declared as another "Application.mk". Is this the right approach ? How can I build both the test tool and the library in the same Makefile ?

Ouss4: You can do that with an Application.mk makefile as well.
The reason I'm saying this is because you are not using any of the prepared targets from Directory.mk and you are recreating targets that already exist in Application.mk.
For instance Application.mk's file already creates the .depend file in its depend target and it cleans it during its distclean.

spiriou: The Makefiles should be fine now, I managed to remove the .depend hack.

Ouss4: So you went with the Directory.mk?  It's all right if it fits your need, however you need to add a .gitignore file and ignore the autogenerated Kconfig.

system/libuv/libuv/Makefile
```diff
@@ -0,0 +1,107 @@
...
+CSRCS += $(LIBUV_UNPACKDIR)/src/strscpy.c
+
+ifeq ($(CONFIG_LIBUV_NET),y)
+CSRCS += $(LIBUV_UNPACKDIR)/src/inet.c
+endif
+
+context:: $(LIBUV_UNPACKDIR)
+
+clean::
+	$(call DELFILE, $(OBJS))
```
Ouss4, line 101: Is this necessary?  Application.mk already deletes object files.  If the object files are generated else where then you need to extend clean::.

spiriou: my understanding is CLEAN only removes "$(Q) rm -f *$(OBJEXT) *$(LIBEXT)", so "*.o" and "*.a".
It can be tricky to use VPATH here due to potential file name collisions between libuv/{libuv,tests} and libuv/libuv-1.38.1/.
So I added the files in CSRCS variable with absolute path. Hence the CLEAN target does not work anymore because the ".o" files are located in another folder, "libuv-1.38.1/src/[unix]".

If it's an issue, I can switch to VPATH and ensure there is no filename collision.

Ouss4: It's not an issue it's exactly how it should work.  But are your $(LIBUV_UNPACKDIR)/src/*.o files being cleaned? As I see it, it cleans only at the libuv/libuv level.

spiriou: Yes, all the objects are cleaned with this trick, even the ones not in libuv/ or tests/ folders like "./libuv-1.38.1/src/unix/nuttx_stream.path.to.nuttx.incubator-nuttx-apps.system.libuv.libuv.o".
OBJS is defined based on $(CSRCS:.c=$(SUFFIX)$(OBJEXT)) so it works fine.
How do you feel about this ?
What would be the correct solution to ensure all the objects are cleaned without extending the clean:: target ?

Ouss4: No. You are right.
The clean target called from Application only removes the files at the current level, it doesn't clean subdirectories, we need to extend it to clean any other file generated. The way you extended it to clean all OBJS does go through all the subdirectories.

patacongo: The CLEAN operation does more than just remove object files.  See definitions in tools/Config.mk

patacongo: The CLEAN operation of tools/Config.mk is also over-ridden by architecture-specific operations that generate other kinds of files.  See for example, tools/zds/Config.mk.  If you simply remove the $(OBJS) in the sub-directory, the clean will be incomplete in some cases.

Ouss4: The clean target as called from the general Application.mk does the regular CLEAN.  The application specific Makefile extends that to clean any other files that can not be reached by CLEAN.

PR #364, opened 2020-08-13
system/cu: refactor, add option -c to disable \n -> \r\n conversion

lukegluke:
option -c will disable \n -> \r\n conversion both for serial device and console if it is a tty

refactor: use only one function to set termios parameters, use termious related code and options only if CONFIG_SERIAL_TERMIOS is defined

This options is handy (and very vital) to interact with serial devices with binary protocol together with option '-f' (endless mode without escape sequence) introduced in df5d4cd033167c775fd69692446a146fe3453a0f.

due to ifdef construction nxstyle complains
`cu_main.c:246:44: error: Unmatched right parentheses`
should I (and how) eliminate this?

I've built and tested this as cherry pick to 8.2 and only with CONFIG_SERIAL_TERMIOS defined. Haven't built it without this.

### DIFF
--- a/system/cu/cu_main.c
+++ b/system/cu/cu_main.c
@@ -81,6 +81,11 @@ enum parity_mode
  ****************************************************************************/
 
 static struct cu_globals_s g_cu;
+#ifdef CONFIG_SERIAL_TERMIOS
+int fd_std_tty;
+static struct termios g_tio_std;
+static struct termios g_tio_dev;
+#endif
 
 /****************************************************************************
  * Public Data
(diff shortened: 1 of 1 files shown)


system/cu/cu_main.c
```diff
@@ -342,8 +366,37 @@ int main(int argc, FAR char *argv[])
...
+    fd_std_tty = fileno(stderr);
+  else if (isatty(fileno(stdout)))
+    fd_std_tty = fileno(stdout);
+  else if (isatty(fileno(stdin)))
+    fd_std_tty = fileno(stdin);
+  else
+    fd_std_tty = -1;
+
+  if (fd_std_tty >= 0)
+    tcgetattr(fd_std_tty, &g_tio_std);
```
Ouss4, line 393: Curly brackets are required even for one-line/empty if/for/while statements.

patacongo: Coding Standard:  https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard#ifthenelse

PR #367, opened 2020-08-17
nsh: Accept the command line arguments like sh

xiaoxiang781216:
## Summary
See the dissussion here: https://github.com/apache/incubator-nuttx/pull/1597

## Testing
Please ignore the false nxstyle alarm:
```
nshlib/nsh_timcmds.c:115:35: error: Multiple data definitions
```

### DIFF
--- a/nshlib/nsh_timcmds.c
+++ b/nshlib/nsh_timcmds.c
@@ -99,6 +99,7 @@ static inline int date_month(FAR const char *abbrev)
           return i;
         }
     }
+
   return ERROR;
 }
 #endif
@@ -108,7 +109,8 @@ static inline int date_month(FAR const char *abbrev)
  ****************************************************************************/
 
 #ifndef CONFIG_NSH_DISABLE_DATE
-static inline int date_showtime(FAR struct nsh_vtbl_s *vtbl, FAR const char *name)
+static inline int date_showtime(FAR struct nsh_vtbl_s *vtbl,
+                                FAR const char *name)
 {
   static const char format[] = "%a, %b %d %H:%M:%S %Y";
   struct timespec ts;
@@ -127,7 +129,7 @@ static inline int date_showtime(FAR struct nsh_vtbl_s *vtbl, FAR const char *nam
 
   /* Break the current time up into the format needed by strftime */
 
-  if (gmtime_r((FAR const time_t*)&ts.tv_sec, &tm) == NULL)
+  if (gmtime_r((FAR const time_t *)&ts.tv_sec, &tm) == NULL)
     {
       nsh_error(vtbl, g_fmtcmdfailed, name, "gmtime_r", NSH_ERRNO);
       return ERROR;
@@ -152,8 +154,8 @@ static inline int date_showtime(FAR struct nsh_vtbl_s *vtbl, FAR const char *nam
  ****************************************************************************/
 
 #ifndef CONFIG_NSH_DISABLE_DATE
-static inline int date_settime(FAR struct nsh_vtbl_s *vtbl, FAR const char *name,
-                               FAR char *newtime)
+static inline int date_settime(FAR struct nsh_vtbl_s *vtbl,
+                               FAR const char *name, FAR char *newtime)
 {
   struct timespec ts;
   struct tm tm;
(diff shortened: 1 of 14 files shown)


xiaoxiang781216:
@patacongo @v01d @davids5 could you review this patch which is paired with https://github.com/apache/incubator-nuttx/pull/1597? so user can let nsh exit after the initialization finish by:
```
CONFIG_INIT_ARGS="\"-c\", \"exit\""
```

nshlib/nsh_timcmds.c
```diff
@@ -386,11 +392,11 @@ int cmd_date(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
         }
       else /* option = '?' */
         {
-          /* We need to parse to the end anyway so that getopt stays healthy */
+          /* We need parse to the end anyway so that getopt stays healthy */
```
v01d, line 395: this comment change does not make sense

xiaoxiang781216: Done.

v01d:
I'm not that familiar with this part of NuttX but I couldn't find anything that caught my attention.

I'm curious, how do you manage nsh to exit? I'm assuming nsh starts some other task via init script, which
you fork, otherwise it this child task would also exit once nsh exits.
In that case, wouldn't be simpler to just add an "exit" at the end? Also, it is typical to end this kind of scripts with an exec()
call so that the calling process (nsh) is replaced by your second task. Not sure if that works in NuttX though.

Anyway, this PR is more general than that, so I agree with the change and aligns with your previous PR.

xiaoxiang781216:
> I'm not that familiar with this part of NuttX but I couldn't find anything that caught my attention.
> 
> I'm curious, how do you manage nsh to exit? I'm assuming nsh starts some other task via init script, which
> you fork, otherwise it this child task would also exit once nsh exits.

Yes, we have many services which is started by init.d/rcS run on the background.

> In that case, wouldn't be simpler to just add an "exit" at the end? Also, it is typical to end this kind of scripts with an exec()
> call so that the calling process (nsh) is replaced by your second task. Not sure if that works in NuttX though.
> 

Yes, adding 'exit' at the end of init.d/rcS should work, but we want to control the behaviour through defconfig since developers are more familiar to modify defconfig than init.d/rcS. And this also make nsh align with sh's behaviour.

> Anyway, this PR is more general than that, so I agree with the change and aligns with your previous PR.

davids5:
@xiaoxiang781216 - I would like to test this and the sister PR. But I need some time. I am fixing other breakage in kinetis at the moment.

xiaoxiang781216:
Here is my test(sim:nsh), the behaviour is good:
```
xiaoxiang@DESKTOP-JH2VEP6:~/nuttx/nuttx$ ./nuttx 
nsh: mkrd: command not found
login: admin
password: 
User Logged-in!

NuttShell (NSH) NuttX-9.1.0
MOTD: username=admin password=Administrator
nsh> poweroff
```
Could you set breakpoint in fopen? so I can see the context when the error happen, or try on sim:nsh.

davids5:
This was a complicated debug, it was the last open on nested scripts.....

image

The info you asked for is above. 

I believe this is wrong https://github.com/apache/incubator-nuttx-apps/pull/367/files#diff-1037d1ad14b4ee447c6ca07d40ccb9f7R131 
image

I have no clue why you do not  see it on the sim.

davids5:
@xiaoxiang781216 this is the problem the PR apache/incubator-nuttx#1597 did not do what @patacongo asked for with CONFIG_INIT_ARGS

from config.h
`#define CONFIG_INIT_ARGS ""`

xiaoxiang781216:
But I am different from yours, here is nsh arguments:
image
and include/nuttx/config.h:
image
Yes, .config contain the empty string:
image
But mkconfig will remove the empty string macro, do you have this line:
https://github.com/apache/incubator-nuttx/blob/master/tools/cfgdefine.c#L69

davids5:
@xiaoxiang781216 - yes I have it, 

Sorry, the problem was old version of mkconfig binary in my tools dir. I need to fix the out of tree build on our side.

PR #374, opened 2020-08-29
Bluetooth: Use the new l2cap socket address type

btashton:
## Summary
Update btsak application to use the new BTPROTO_L2CAP `sockaddr_l2` struct instead of the generic one we had for Bluetooth before.

## Impact
No change in functionality.

## Testing
Tool works the same as before.


patacongo:
I imagine that this will fail the PR checks without the matching change to nuttx/?

btashton:
> I imagine that this will fail the PR checks without the matching change to nuttx/?

Yes

> I imagine that this will fail the PR checks without the matching change to nuttx/?

Yes here https://github.com/apache/incubator-nuttx/pull/1666

I see there are some nxstyle issues with `btsak` so I will fix those up now.

btashton:
Ok style fixes are in.

patacongo:
Okay, fails with:

    btsak.h:74:22: error: field 'ep_sockaddr' has incomplete type
       74 |   struct sockaddr_l2 ep_sockaddr;    /* AF_BLUETOOTH endpoint address */
          | 

I will have to merge this and apache/incubator-nuttx#1666 together and hope there will be no breakage.

PR #378, opened 2020-09-08
interpreters: Initial wasm3 WebAssembly runtime support

no1wudi:
## Summary

Introduce wasm3 WebAssembly runtime support, with baisc test utility.

## Impact
None

## Testing
Tested on stm32

### DIFF
--- a/interpreters/wasm3/.gitignore
+++ b/interpreters/wasm3/.gitignore
@@ -1,0 +1,2 @@
+wasm3*
+*.tar.gz
--- a/interpreters/wasm3/Kconfig
+++ b/interpreters/wasm3/Kconfig
@@ -1,0 +1,42 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config INTERPRETERS_WASM3
+	tristate "WASM3 Webassembly Runtime"
+	default n
+
+if INTERPRETERS_WASM3
+
+config INTERPRETERS_WASM3_PRIORITY
+	int "WASM3 Webassembly priority"
+	default 100
+
+config INTERPRETERS_WASM3_STACKSIZE
+	int "WASM3 Webassembly stack size"
+	default DEFAULT_TASK_STACKSIZE
+
+config INTERPRETERS_WASM3_CODEPAGE
+	int "WASM3 code page align size"
+	default 512
+
+config INTERPRETERS_WASM3_FIXEDHEAP
+	int "WASM3 Fixed heap size"
+	default 0
+	---help---
+	0 : Use malloc/free instead of internal heap
+
+config INTERPRETERS_WASM3_LOG
+	bool "WASM3 log"
+	default n
+
+if INTERPRETERS_WASM3_LOG
+
+config INTERPRETERS_WASM3_LOG_VERBOSE
+	bool "WASM3 log verbose"
+	default n
+
+endif
+
+endif
--- a/interpreters/wasm3/Make.defs
+++ b/interpreters/wasm3/Make.defs
@@ -1,0 +1,29 @@
+############################################################################
+# interpreters/wasm3/Make.defs
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifneq ($(CONFIG_INTERPRETERS_WASM3),)
+CONFIGURED_APPS += $(APPDIR)/interpreters/wasm3
+
+CFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" \
+    $(APPDIR)/interpreters/wasm3/wasm3/source}
+CXXFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" \
+    $(APPDIR)/interpreters/wasm3/wasm3/source}
+
+endif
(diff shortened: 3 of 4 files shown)


interpreters/wasm3/.gitignore
```diff
@@ -0,0 +1,2 @@
+wasm3*
+*.tar.gz
```
xiaoxiang781216, line 2: add a newline

patacongo:
wasm3 has an MIT license.  The MIT license is compatible with Apache, but don't you think we should at least let the user know that the resulting binaries follow MIT rules, not Apache rules?  I am thinking in the README.md or in the top-level LICENSE file.

One significant difference is that the Apache license grants use of patents to end-users; MIT does not.  So conceivable, an end user could inadvertently violate any patent thinking that the code is Apache licensed.

@justinmclean @Apache9 Opinions?

xiaoxiang781216:
How about we mark the license in Kconfig like this:
```
config INTERPRETERS_WASM3
	tristate "WASM3 Webassembly Runtime(MIT)"
	default n
```
So the user know the license impact before enabling the component.

btashton:
> How about we mark the license in Kconfig like this:
> 
> ```
> config INTERPRETERS_WASM3
> 	tristate "WASM3 Webassembly Runtime(MIT)"
> 	default n
> ```
> 
> So the user know the license impact before enabling the component.

Another option in this vain would be to have an allowed licences Kconfig where we would fail the build if a non-allowed license was enabled.  We could also make the Kconfig dependent, the downside being it might be hard to find features behind a license gate.

patacongo:
> Another option in this vain would be to have an allowed licences Kconfig where we would fail the build if a non-allowed license was enabled. We could also make the Kconfig dependent, the downside being it might be hard to find features behind a license gate.

I don't have any concern about 3rd party BSD and MIT code in the system.  That is spelled out in the LICENSE file.  I just don't have a clue about the legal standing of non-Apache licensed code that is downloaded and built into the binaries.  The licenses in general place no limits on the content of the binaries.  I think that the difference in patent grants between the licenses is the only issue that could compromise binaries.

btashton:
I'm just extrapolating from my experience with Yocto and Buildroot for Linux.  With those you can limit the acceptable licenses and also get the License files for the external software that was included in the build.  Up to the user to then do what they feel is needed to abide by the restrictions.

There are a bunch of other Apache projects where you can turn on build flags that would not match the Apache binary requirements, especially software that uses Intel or Nvidia libraries.  As far as AFS is concerned we are not bundling any of these nor are we generating binary releases, so I don't see the problem.

v01d:
Maybe something useful would be to set a configuration value indicating the "clean" licensing status of a build so that on a binary you can check if the build includes projects with other licenses.

patacongo:
> 
> 
> Maybe something useful would be to set a configuration value indicating the "clean" licensing status of a build so that on a binary you can check if the build includes projects with other licenses.

I'm not sure what you mean by "clean" license.  The ASF permits third party software into the repositories PROVIDED that we follow the rules for inclusion of third part software.  See, for example, https://www.apache.org/legal/resolved.html .  I think this means, basically that the code has to have a compatible license and we also have to document that license and copyright in the LICENSE file.  There is also this vague issue that has come up a few times that ways were also supposed to have the blessing of the copyright holder to use the code, but I don't really understand what that means in any legal sense.

There is a lot of BSD third party code in the repositories now and you could not even build NuttX of that code were not included in the build.

My curiosity is for the case of downloading third party code with a compatible license at build time and NOT following the ASF rules for third party code.  Honestly, I don't really know the answer.  But it seems to me that there is an issue with documenting the license, making sure that the end user is aware of the licensing, and protecting ourselves legally from being a party to a possible patent infringement.

I don't know what the answer is and I don't have a strong recommendation.  The safest thing to do would be to follow the ASF rules for third party software for downloaded software even though it does not reside in the repository.

It would be good to have a mentor weigh in on this.  I don't think any of us really have the appropriate knowledge to answer this question.

v01d:
Is the licensing really something to block this PR? We're not including third-party code here. It downloads code as we do with other projects. One example is LVGL, which is in fact MIT licensed as well.

justinmclean:
Hi,

From an ASF policy point of view it’s fine to depend on 3rd party software that is compatible with the Apache license. It doesn’t need to be added to LICENSE but you may want to inform users about it, some projects have a DEPENDANCIES file for this. Using 3rd party MIT and BSD license code like this is fine.

Thanks,
Justin

justinmclean:
Hi,
> I don't have any concern about 3rd party BSD and MIT code in the system. That is spelled out in the LICENSE file. I just don't have a clue about the legal standing of non-Apache licensed code that is downloaded and built into the binaries. The licenses in general place no limits on the content of the binaries.
> 

ASF policy does. Anything in the binary produced by an Apache project (with a few exceptions) needs to be compatible with the Apache license. Basically you can include anything from Category A [1] and Category B [2] but not Category X. [3]

MIT is fine in this case.

Thanks,
Justin

1. https://www.apache.org/legal/resolved.html#category-a
2. https://www.apache.org/legal/resolved.html#category-b
3. https://www.apache.org/legal/resolved.html#category-x

justinmclean:
Hi,
> There are a bunch of other Apache projects where you can turn on build flags that would not match the Apache binary requirements, especially software that uses Intel or Nvidia libraries.
> 

Optional non compatible dependancies are allowed, but they do need to be optional. [1] The user need to be clearly informed that this is the case.

Take care making assumption that is is OK to do this, for instance one incubating project MXNet recently got itself into serious trouble assuming this was OK.

Thanks,
Justin

1. https://www.apache.org/legal/resolved.html#optional <https://www.apache.org/legal/resolved.html#optional>

v01d:
Thanks @justinmclean. That really clears up a lot of questions. I think we're in the green since these kind of dependencies are optional and they are generally in category A even. 

Regarding letting the user now, that is what I was suggesting with my comment above. Specifically, we could use Kconfig to select `INCLUDES_MIT`, `INCLUDES_BSD3`, etc. From that, this can be exposed somewhere in the final binary, in the output of some command or via `procfs`, for example.

v01d:
BTW, this information should go somewhere in our documentation, since it pertains to any further contributions.
I will add a task to the "documentation roadmap" issue.

xiaoxiang781216:
@v01d, it's a good idea to let user enable BSD or MIT before they can enable the 3rd party library. But it's better to create a new PR to convert all 3rd party projects with the new facility. Do you think so?

v01d:
My suggestion was to simply have a "notice" not necessarily a user choice. I understand we don't actually need a per-license choice (that would really be annoying since you would not discover feature hidden behind those choices unless you explore them).
The idea is, for example: have LVGL config `select INCLUDES_MIT`. Then, in some part of the code, generate output like: "this binary is distributed under the license terms of" and the result of the mix of chosen licenses.

But again, don't let this hold the PR. Please merge this if you think it is correct. We can discuss these ideas in a separate issue.

xiaoxiang781216 (approved):
LGTM.

PR #379, opened 2020-09-09
build: Support both .s & .S

no1wudi:
## Summary
See title

## Impact
None
## Testing
Tested on stm32

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -70,7 +70,11 @@ PROGNAME := $(shell echo $(PROGNAME))
 
 # Object files
 
-AOBJS = $(ASRCS:.S=$(SUFFIX)$(OBJEXT))
+RASRCS = $(filter %.s,$(ASRCS))
+CASRCS = $(filter %.S,$(ASRCS))
+
+AOBJS = $(RASRCS:.s=$(SUFFIX)$(OBJEXT))
+AOBJS += $(CASRCS:.S=$(SUFFIX)$(OBJEXT))
 COBJS = $(CSRCS:.c=$(SUFFIX)$(OBJEXT))
 CXXOBJS = $(CXXSRCS:$(CXXEXT)=$(SUFFIX)$(OBJEXT))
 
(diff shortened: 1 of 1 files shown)


patacongo:
By convention, .s and .S files are normally different file types.  .s is raw assembly code; .S files need to be run through the C pre-processor before being assembled.  .s files are assembled with an assembler like as; .S files are assembled with the compile such as gcc.  We need to support both .s and .S files, not one or the other.

no1wudi:
> By convention, .s and .S files are normally different file types. .s is raw assembly code; .S files need to be run through the C pre-processor before being assembled. .s files are assembled with an assembler like as; .S files are assembled with the compile such as gcc. We need to support both .s and .S files, not one or the other.

Both .s & .S supported now, 
I pass the .s file to gcc as .S, is there any problem ?

no1wudi:
The build break is from sim/up_tapdev.c

xiaoxiang781216:
Fix here: https://github.com/apache/incubator-nuttx/pull/1743

xiaoxiang781216:
> By convention, .s and .S files are normally different file types. .s is raw assembly code; .S files need to be run through the C pre-processor before being assembled. .s files are assembled with an assembler like as; .S files are assembled with the compile such as gcc.

Two issues here:
1.Toolchain.defs doesn't define assembler related macro(e.g. AS, ASFLAGS...).
2.There is no huge side effect that let *.s also pass through the preprocess.

> We need to support both .s and .S files, not one or the other.

How about we add a new .s rule but has the same action as .S?

no1wudi:
> > By convention, .s and .S files are normally different file types. .s is raw assembly code; .S files need to be run through the C pre-processor before being assembled. .s files are assembled with an assembler like as; .S files are assembled with the compile such as gcc.
> 
> Two issues here:
> 1.Toolchain.defs doesn't define assembler related macros(e.g. AS, ASFLAGS...).
> 2.There is no huge side effect that let *.s also pass through the preprocess.
> 
> > We need to support both .s and .S files, not one or the other.
> 
> How about we add a new .s rule but has the same action as .S?

In fact, asm in apps side is less common, I think it's ok otherwise the symbol in raw assembly code (.s) is dulpicated with existing pre-processor's definition

xiaoxiang781216:
@patacongo do you have more comment with the new approach?

xiaoxiang781216 (approved):

PR #380, opened 2020-09-12
Fix: ensure archive files do not carry object files from prior builds

v01d:
## Summary

This is the corresponding change to the one on main NuttX repo (see https://github.com/apache/incubator-nuttx/pull/1765).
In this case this involves splitting the build of libapps.a into: a) building
all applications (which is safely parallelizable), b) adding each
application's object files to the archive in turns (serial by nature).

This removes the need for the flock used to protect the parallel build.

NOTE: this will not pass CI since it depends on the other PR

## Impact

Build

## Testing

Builds correctly

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -95,7 +95,7 @@ VPATH += :.
 
 # Targets follow
 
-all:: .built
+all:: $(OBJS)
 .PHONY: clean depend distclean
 .PRECIOUS: $(BIN)
 
@@ -131,13 +131,12 @@ $(CXXOBJS): %$(SUFFIX)$(OBJEXT): %$(CXXEXT)
 	$(if $(and $(CONFIG_BUILD_LOADABLE),$(CXXELFFLAGS)), \
 		$(call ELFCOMPILEXX, $<, $@), $(call COMPILEXX, $<, $@))
 
-.built: $(OBJS)
+archive:
 ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-	$(call ARLOCK, "${shell cygpath -w $(BIN)}", $^)
+	$(call ARCHIVE_ADD, "${shell cygpath -w $(BIN)}", $(OBJS))
 else
-	$(call ARLOCK, $(BIN), $^)
+	$(call ARCHIVE_ADD, $(BIN), $(OBJS))
 endif
-	$(Q) touch $@
 
 ifeq ($(BUILD_MODULE),y)
 
@@ -234,7 +233,6 @@ endif
 depend:: .depend
 
 clean::
-	$(call DELFILE, .built)
 	$(call CLEAN)
 
 distclean:: clean
--- a/Make.defs
+++ b/Make.defs
@@ -108,10 +108,6 @@ define REGISTER
 	$(Q) touch "$(BUILTIN_REGISTRY)$(DELIM).updated"
 endef
 
-define ARLOCK
-	$(Q) flock $1.lock $(call ARCHIVE, $1, $(2))
-endef
-
 # Standard include path
 
 CFLAGS   += ${shell $(INCDIR) "$(CC)" "$(APPDIR)$(DELIM)include"}
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,13 @@ SYMTABOBJ = $(SYMTABSRC:.c=$(OBJEXT))
 
 # Build targets
 
-all: $(BIN)
+# We first remove libapps.a before letting the other rules add objects to it
+# so that we ensure libapps.a does not contain objects from prior build
+
+all:
+	$(RM) $(BIN)
+	$(MAKE) $(BIN)
+  
 .PHONY: import install dirlinks context context_serialize clean_context context_rest export .depdirs preconfig depend clean distclean
 .PRECIOUS: $(BIN)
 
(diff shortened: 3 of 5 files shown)


Application.mk
```diff
@@ -131,13 +131,12 @@ $(CXXOBJS): %$(SUFFIX)$(OBJEXT): %$(CXXEXT)
...
-.built: $(OBJS)
+link:
 ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-	$(call ARLOCK, "${shell cygpath -w $(BIN)}", $^)
+	$(call ARCHIVE_ADD, "${shell cygpath -w $(BIN)}", $(OBJS))
 else
-	$(call ARLOCK, $(BIN), $^)
+	$(call ARCHIVE_ADD, $(BIN), $(OBJS))
 endif
-	$(Q) touch $@
```
Ouss4, line 140: Directories that are going to be cleaned are filtered-out using this marker ".built".  Removing it will keep object files (and other autogenerated files) after a clean/distclean.

Make.defs
```diff
@@ -108,10 +108,6 @@ define REGISTER
 	$(Q) touch "$(BUILTIN_REGISTRY)$(DELIM).updated"
 endef
 
-define ARLOCK
-	$(Q) flock $1.lock $(call ARCHIVE, $1, $(2))
```
Ouss4, line 112: Hmmmm flock was necessary to get the nightly build in a stable situation.  @xiaoxiang781216 could you please comment here.

Application.mk
Ouss4: Hmm.. is ".depend" covering everything?

Makefile
```diff
@@ -87,6 +87,10 @@ else
 ifeq ($(CONFIG_BUILD_LOADABLE),)
 
 $(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
+	$(RM) $(BIN)
```
Ouss4, line 90: Is this to remove old members that are not going to get replaced with the 'r' of `ar crs`?

Makefile
```diff
@@ -100,9 +104,9 @@ $(SYMTABOBJ): %$(OBJEXT): %.c
 
 $(BIN): $(SYMTABOBJ)
 ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-	$(call ARLOCK, "${shell cygpath -w $(BIN)}", $^)
+	$(call ARCHIVE, "${shell cygpath -w $(BIN)}", $^)
```
xiaoxiang781216, line 107: Should we use ARCHIVE_ADD here?

Makefile
v01d: Yes

Makefile
v01d: I understand this is the only call to the archive being created (it is either the previous recursive call into subdirectories or this one, according to the `if`). So in this case we need to create the library from scratch. But I'm not really familiar with this SYMTABOBJ rule.

Application.mk
v01d: I will take a look at the clean.

v01d: Yes, it appears .depend covers it, don't really understand the purpose of the extra search for .built.

Ouss4: I tried again to see if there is any leftover if we omit the search for ".built".  I'm not sure if there is a place where only .built is generated.  We need to let the checks of this PR run to know (after we merge the dependency).
If ".built" is not needed, please remove the search from Direcotry.mk too.

BTW, can you build stm32f4discovery:wifi?  I pulled your changes (both os&apps) I can build other cofings like esp32-core:nsh but the stm32f4discovery:wifi is giving me the libapp.a file not found error.

v01d: > I tried again to see if there is any leftover if we omit the search for ".built", but things seem clean. I'm not sure if there is a place where only .built is generated. We need to let the checks of this PR run to know (after we merge the dependency).
> If ".built" is not needed, please remove the search from Direcotry.mk too.

Yes, I did the same check. I have already modified my local commit to include it but did not want to push to step on your review.

> 
> BTW, can you build stm32f4discovery:wifi? I pulled your changes (both os&apps) I can build other cofings like esp32-core:nsh but the stm32f4discovery:wifi is giving me the libapp.a file not found error.

Thanks for testing, yes it fails. It seems this is the LOADABLE case. I will look into how to fix this.

v01d: I managed to solve this, I force pushed the change.

Makefile
v01d: You were right, this needed ARCHIVE_ADD and remove the library earlier. I just pushed a change which consideres both the CONFIG_BUILD_LOADABLE and !CONFIG_BUILD_LOADABLE case.

Application.mk
```diff
@@ -131,13 +131,12 @@ $(CXXOBJS): %$(SUFFIX)$(OBJEXT): %$(CXXEXT)
 	$(if $(and $(CONFIG_BUILD_LOADABLE),$(CXXELFFLAGS)), \
 		$(call ELFCOMPILEXX, $<, $@), $(call COMPILEXX, $<, $@))
 
-.built: $(OBJS)
+link:
```
xiaoxiang781216, line 134: should we change to archive to reflect what this tep really do?

Make.defs
xiaoxiang781216, line 112: Please don't forget to remove the flock from CI script and the document.

Application.mk
v01d: yes, sounds better

xiaoxiang781216:
LGTM, @Ouss4 please merge it if you are fine too.

xiaoxiang781216:
Sorry, just remeber that we can remove .built and .lock from .gitignore? @v01d

Make.defs
v01d: Right. The documentation is on nuttx repo so I will add the change there.
As per removing flock from CI this is on the testing repo. I can do that once these two PRs are merged.

v01d:
> Sorry, just remeber that we can remove .built and .lock from .gitignore? @v01d

Done

Ouss4:
> LGTM, @Ouss4 please merge it if you are fine too.

@xiaoxiang781216 did you had the chance to take a look at https://github.com/apache/incubator-nuttx/pull/1765?  If all is good I'd like that we merge that one first and restart the checks here.
I just want to make sure we won't have any leftovers.  Some configs are tricky.

PR #381, opened 2020-09-12
CI: don't cache sources, use artifacts only

v01d:
## Summary

Same change as https://github.com/apache/incubator-nuttx/pull/1762 applied here

## Impact

CI

## Testing

CI itself


v01d:
This is failing for unrelated reason, maybe we should merge this anyway so that we can actually have CI for -apps working

btashton:
Yes I agree.

PR #384, opened 2020-09-15
Fix multi MAINSRC match suffix in single Makefile

Donny9:
## Summary
Fix multi MAINSRC match suffix in single Makefile 

## Impact
We can compile multi program in single makefile with different suffix file in MAINSRC.
eg: MAINSRC = xxx1.c
      PROGNAME = xxx1
      MAINSRC += xxx2.cpp
      RROGNAME += xxx2

## Testing
Built locally

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -74,17 +74,16 @@ AOBJS = $(ASRCS:.S=$(SUFFIX)$(OBJEXT))
 COBJS = $(CSRCS:.c=$(SUFFIX)$(OBJEXT))
 CXXOBJS = $(CXXSRCS:$(CXXEXT)=$(SUFFIX)$(OBJEXT))
 
-ifeq ($(suffix $(MAINSRC)),$(CXXEXT))
-  MAINOBJ = $(MAINSRC:$(CXXEXT)=$(SUFFIX)$(OBJEXT))
-else
-  MAINOBJ = $(MAINSRC:.c=$(SUFFIX)$(OBJEXT))
-endif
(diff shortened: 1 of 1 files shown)


v01d:
This feels conceptually wrong, why would need two main files? There should be only a single entry function, regardless of C vs C++ (which is currently supported). If there are two "main" this are two separate applications. An app with two mains does not exist in other OSs.

Donny9:
> This feels conceptually wrong, why would need two main files? There should be only a single entry function, regardless of C vs C++ (which is currently supported). If there are two "main" this are two separate applications. An app with two mains does not exist in other OSs.

Hello @v01d ,you can see following makefile about zmode, it use double main file.
https://github.com/apache/incubator-nuttx-apps/blob/master/system/zmodem/Makefile

v01d:
I'm not sure I follow how that works or why is designed that way.

v01d:
Ah, I see there are two PROGNAME. Well, I can't say I like the idea. In any other OS you would have to different applications built separately sharing common code via a library. This is actually possible in NuttX, it only takes more work. Anyway, I understand you are simple allowing different languages for each main.
I can approve the change, even if I don't really like the idea (which is independent of this PR)

v01d (approved):

xiaoxiang781216:
@v01d should we merge this PR? We can create a new issue to track the problem of application structure.

v01d:
Yes, I approved, you can merge if you agree as well.

xiaoxiang781216 (approved):

xiaoxiang781216:
Ok, let's go ahead.

PR #385, opened 2020-09-15
apps/MAKE: Add register staging to context

GUIDINGLI:
## Summary

apps/MAKE: Add register staging to context

## Impact

When compiling an external library, and the library need "configure" firstly.
Then there are 2 ways to doing the "configure" :
1. preconfig stage:
   If put "configure" in this stage, the global config.h (external library also needed) hasn't ready.
2. context stage:
   If put "configure" in this stage, the action of registering will be failed, because PROGNAME in library
   has dependency to action "configure".

In another word, compiler should follow this step for this library:
preconfig(config.h) ->  configure to library -> PROGNAME -> action of registering to built-in list

So, to resolve this.

preconfig(config.h) ->  configure to library -> PROGNAME -> action of registering to built-in list
preconfig stage &emsp; &emsp; &emsp;   context stage   &emsp;  &emsp; &emsp; &emsp; &emsp;&emsp; &emsp; register stage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,12 +52,14 @@ all:
 	$(RM) $(BIN)
 	$(MAKE) $(BIN)
   
-.PHONY: import install dirlinks context context_serialize clean_context context_rest export .depdirs preconfig depend clean distclean
+.PHONY: import install dirlinks export .depdirs preconfig depend clean distclean
+.PHONY: context clean_context context_all register register_all
 .PRECIOUS: $(BIN)
 
 $(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),all)))
 $(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),install)))
 $(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),context)))
+$(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),register)))
 $(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),depend)))
 $(foreach SDIR, $(CLEANDIRS), $(eval $(call SDIR_template,$(SDIR),clean)))
 $(foreach SDIR, $(CLEANDIRS), $(eval $(call SDIR_template,$(SDIR),distclean)))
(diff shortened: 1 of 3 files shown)


v01d:
What is the purpose of this change? Please be more descriptive in the PR.

GUIDINGLI:
@v01d fill the description

Makefile
```diff
@@ -152,13 +156,12 @@ endif # CONFIG_BUILD_KERNEL
 dirlinks:
 	$(Q) $(MAKE) -C platform dirlinks TOPDIR="$(TOPDIR)" APPDIR="$(APPDIR)"
 
-context_rest: $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_context)
+context_all: $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_context)
+register_all: $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_register)
 
-context_serialize:
-	$(Q) $(MAKE) -C builtin context TOPDIR="$(TOPDIR)" APPDIR="$(APPDIR)"
```
v01d, line 158: aren't you removing the special case of running context on builtin first? I don't see where builtin is called after your change

v01d:
Thanks, I now understand why the change. I think the idea is reasonable.
Maybe someone else wants to take a look and see if there isn't anything critical not considered?

BTW, if you don't mind I'd rather we wait for this to be merged https://github.com/apache/incubator-nuttx/pull/1785 
since I have that PR ready and the corresponding one for apps. Adapting this PR to that is really simple, but this way
we make it in one go.

Makefile
GUIDINGLI: the special case of running context on builtin first, is useless, so I removed.

You can see there is no "context:: " in builtin/Makefile, then it will use common context:: in Application.mk.
That is do the action of register, but builtin folder has no PROGNAME itself.

so we can remove this:
$(Q) $(MAKE) -C builtin context TOPDIR="$(TOPDIR)" APPDIR="$(APPDIR)"

GUIDINGLI:
Sure, wait for your patch merged.

v01d:
FYI, this is the PR. Once it is merged, please rebase here. You will have to remove the passing of TOPDIR and APPDIR in Make calls (but only if they are passed like TOPDIR=$(TOPDIR), if the value is different, keep it).

Makefile
v01d: I'm unsure we should remove the call to a target such as context for a subdirectory because it is unused. context might someday be required in builtin and it may not be trivial to understand it is not being invoked. These targets are invoked for all subdirectories in NuttX, so let's keep it even while right now it is not used.
From what I understand $(CONFIGURED_APPS) does not contain "builtin", so unless you explicitly call it, it will not be called at all. Just add "builtin" instead the foreach of the context target. That will be enough I think.

xiaoxiang781216: @v01d $(CONFIGURED_APPS) actually contain "builtin", here is the code from https://github.com/apache/incubator-nuttx-apps/blob/master/builtin/Make.defs#L37:
```
ifeq ($(CONFIG_BUILTIN),y)
CONFIGURED_APPS += $(APPDIR)/builtin
endif
```

the old Makefile build the context target of builtin twice, this patch fix this issue by removing the explicit invocation.

v01d: Ok then!

PR #387, opened 2020-09-16
interpreters: Inroduce support for WAMR

no1wudi:
## Summary

Support interpreter/aot mode, and this patch had submitted to WAMR (Apache License). See 

https://github.com/bytecodealliance/wasm-micro-runtime/pull/377

The patch file will be removed on WAMR's next release.

## Impact
None
## Testing
Tested on stm32.


Ouss4 (approved):
LGTM

PR #392, opened 2020-09-17
wireless/wapi: ignore the fail case in show command

anchao:
## Summary

wireless/wapi: ignore the fail case

do not break the show process if some cases failure

Change-Id: I5d5768ba116a9e8a8ea62ae9455d773838328bd5


xiaoxiang781216 (approved):
LGTM.

PR #395, opened 2020-09-17
make/clean: clean the OBJ directly if declare in subdirectory

anchao:
## Summary
make/clean: clean the OBJ directly if declare in subdirectory

Change-Id: Iabfb3c5ed4d37a5534e3bab63a6b3cffb4816230

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -238,6 +238,7 @@ depend:: .depend
 
 clean::
 	$(call CLEAN)
+	$(call DELFILE, $(wildcard $(foreach obj, $(OBJS), $(addsuffix /$(obj), $(subst :, ,$(VPATH))))))
 
 distclean:: clean
 	$(call DELFILE, Make.dep)


Application.mk
```diff
@@ -238,6 +238,7 @@ depend:: .depend
 
 clean::
 	$(call CLEAN)
+	$(Q) rm -f $(wildcard $(foreach obj, $(OBJS), $(addsuffix /$(obj), $(subst :, ,$(VPATH)))))
```
Ouss4, line 241: Use DELFILE instead of rm?

anchao: Done

v01d:
This feel a bit hackish. Shouldn't CLEAN take care of this?
I think I know why this is needed: if you declare SRCS inside subdirectories from a Makefile.defs in another location, since CLEAN only considers objects in current directory, this indeed would clean them. But I think that we should fix CLEAN then instead, shouldn't we?

Also, note that I think this is thought to be used to that every subdirectory with sources has a Makefile.defs, but if you want to build some third-party app which you can't modify (by adding the Make.defs everywhere), this isn't feasible.

So, I think it is better to modify CLEAN IMHO.

xiaoxiang781216:
Yes, I think too.

v01d:
It is possible that once you do this change, some leftover .o will remain after a clean (since right now the CLEAN target 
simply removes all .o files instead of looking for $(OBJS) (see below). If you do find leftover .o's somewhere we need to check if something else needs fixing.

```
ifeq ($(CONFIG_WINDOWS_NATIVE),y)
define CLEAN
        $(Q) if exist *$(OBJEXT) (del /f /q *$(OBJEXT))
        $(Q) if exist *$(LIBEXT) (del /f /q *$(LIBEXT))
        $(Q) if exist *~ (del /f /q *~)
        $(Q) if exist (del /f /q  .*.swp)
endef
else
define CLEAN
        $(Q) rm -f *$(OBJEXT) *$(LIBEXT) *~ .*.swp
endef
endif
```

anchao:
Hi @v01d,

Yes, thanks for review, totally agree with you!

Let me first introduce the cause of this issue first,
In some open source projects, we often encounter files with the same source file name but in different directories, 
for example:

```
chao@archer:~/code/apps/examples/hello$ tree
.
├── hello_main.c
├── Kconfig
├── Make.defs
├── Makefile
├── test
│   └── test.c
└── test2
    └── test.c
```


If we use VPATH to build this project, we will find that only one file is compiled because VPATH mistakenly believes that another file is already dependent:

Makefile:

```
include $(APPDIR)/Make.defs
...
VPATH := test test2
CSRCS += test.c test.c
....
include $(APPDIR)/Application.mk
```

Compile object:

```
chao@archer:~/code/apps/examples/hello$ tree
.
├── hello_main.c
├── hello_main.home.chao.code.apps.examples.hello.o
├── Kconfig
├── Make.defs
├── Make.dep
├── Makefile
├── test
│   └── test.c
├── test2
│   └── test.c
└── test.home.chao.code.apps.examples.hello.o
```

So VPATH is not a good choice in this case, we must use relative paths to ensure that the files are compiled:

Makefile:
```
include $(APPDIR)/Make.defs
...
CSRCS += test/test.c test2/test.c
...
include $(APPDIR)/Application.mk
```

compile object:

```
chao@archer:~/code/apps/examples/hello$ tree
.
├── hello_main.c
├── hello_main.home.chao.code.apps.examples.hello.o
├── Kconfig
├── Make.defs
├── Make.dep
├── Makefile
├── test
│   ├── test.c
│   └── test.home.chao.code.apps.examples.hello.o
└── test2
    ├── test.c
    └── test.home.chao.code.apps.examples.hello.o
```

As you can see, the source code was compiled correctly, then we encountered a new problem, "How to clear it?"

I have traversed a lot of source code in nuttx , the build system of nuttx is basically built with VPATH, so there is no case where the source code file names are the same.

Therefore, this clean process will be a special and partially effective process. In order to reduce the damage, I only take effect when building external applications, I think this is a comprehensive consideration for a lot of 3rd open source project.

v01d:
I understand completely and your analysis is correct. I agree that VPATH is problematic and that whole "use the path to build object name" only complicates it. We should eventually improve that, but that is a whole problem in itself.
On the other hand, I see no problem in modifying CLEAN to work as I suggested. We only need to ensure it works OK for all cases. I think that extending CLEAN macro this way is a good step towards a direction without depending on VPATH (in that scenario I would guess that each source file has a full path, so doing `rm` on $(OBJS) should be the best way).

Can you try the suggested approach and test it? This would of course be a change in nuttx repo.

anchao:
Emm.... Seriously, I don't really want to put this hack on the main line. This kind of processing will make the NUTTX repo seems weird.... I think if nuttx mainline keeps VPATH style, there will be no such problem.. Of course, I will try your suggestions further

v01d:
What I'm suggesting is not a hack, it is actually the safest approach (remove $(OBJS), not just any .o in current directory). Please try it.

anchao:
Hi @v01d ,

I uploaded a pull request on https://github.com/apache/incubator-nuttx/pull/1843 
please review again

xiaoxiang781216:
Let's close this PR, since the better approach exist.

PR #402, opened 2020-09-23
Network Monitor: Runtime selection of DHCP/Static IP with fall back set by the board, and a polled mode for HW lacking a PHY interrupt 

davids5:
## Summary

Sister to https://github.com/apache/incubator-nuttx/pull/1886 - merge this APP pr second.

 ### netinit:Network Monitor add a polled option
 
  Not all boards have in interrupt line from the phy to the Soc. This allows the phy to be polled for link status.
   This may not work on all MAC/PHY combination that have mutually exclusive link management and operating
   modes. The STM32F7 and LAN8742AI do not have such a limitation.

   To use this option the arch must supply CONFIG_ARCH_PHY_POLLED  and not CONFIG_ARCH_PHY_INTERRUPT.

 If the network is not connected the phy init may time out and netinit_net_bringup will return with out completing DHCP or initializing the ntp client.

This change uses the lack of the DHCP lease, to contiune the netinit_net_bringup, once the interface does come up. It also will renew the lease over time, and will re run the dhcp request after a network down/up transition.

###  netinit Net monitior support getting address from board

   Add the use of CONFIG_BOARDCTL_NETCONF to retrieve the address configurations from the board. Allowing setting
   static or dhcp addressing at run time.

 ###  netinit:Net monitior support fallback to static IP

   After CONFIG_NETINIT_FALLBACK dhcp attempts, fall back to  the static IP provided by the board via CONFIG_BOARDCTL_NETCONF.

## Impact

None All options have Kconfig knobs to turn them off by default

## Testing

FMUV5X with a test vector of all the new Kconfig options.

### DIFF
--- a/netutils/netinit/netinit.c
+++ b/netutils/netinit/netinit.c
@@ -221,13 +221,22 @@
 #  define AF_INETX AF_INET6
 #endif
 
-/* While the network is up, the network monitor really does nothing.  It
- * will wait for a very long time while waiting, it can be awakened by a
- * signal indicating a change in network status.
+/* If using signal notification, raise by a the PHY, then while the network
+ * is up, the network monitor really does nothing.  It will wait for a very
+ * long time while waiting, it can be awakened by a signal indicating a
+ * change in network status.
+ *
+ * If the network monitor is used in polled mode these values are set by the
+ * CONFIG_NETINIT_LOSS_POLL_RATE and CONFIG_NETINIT_ESTABLISH_POLL_RATE
  */
 
-#define LONG_TIME_SEC    (60*60) /* One hour in seconds */
-#define SHORT_TIME_SEC   (2)     /* 2 seconds */
+#if !(defined(CONFIG_ARCH_PHY_POLLED))
+#  define LONG_TIME_SEC    (60*60) /* One hour in seconds */
+#  define SHORT_TIME_SEC   (2)     /* 2 seconds */
+#else
+#  define LONG_TIME_SEC    (CONFIG_NETINIT_LOSS_POLL_RATE)
+#  define SHORT_TIME_SEC   (CONFIG_NETINIT_ESTABLISH_POLL_RATE)
+#endif
 
 /****************************************************************************
  * Private Data
@@ -537,7 +546,7 @@ static void netinit_configure(void)
  *
  ****************************************************************************/
 
-#ifdef CONFIG_NETINIT_MONITOR
+#if defined(CONFIG_NETINIT_MONITOR) && !defined(CONFIG_ARCH_PHY_POLLED)
 static void netinit_signal(int signo, FAR siginfo_t *siginfo,
                                FAR void * context)
 {
@@ -573,8 +582,10 @@ static int netinit_monitor(void)
   struct timespec abstime;
   struct timespec reltime;
   struct ifreq ifr;
+#if !defined(CONFIG_ARCH_PHY_POLLED)
   struct sigaction act;
   struct sigaction oact;
+#endif
   bool devup;
   int ret;
   int sd;
@@ -599,6 +610,7 @@ static int netinit_monitor(void)
       goto errout;
     }
 
+#if !defined(CONFIG_ARCH_PHY_POLLED)
   /* Attach a signal handler so that we do not lose PHY events */
 
   act.sa_sigaction = netinit_signal;
@@ -614,6 +626,7 @@ static int netinit_monitor(void)
       goto errout_with_socket;
     }
 
+#endif
   /* Now loop, waiting for changes in link status */
 
   for (; ; )
@@ -626,6 +639,7 @@ static int netinit_monitor(void)
       ifr.ifr_mii_notify_event.sigev_notify = SIGEV_SIGNAL;
       ifr.ifr_mii_notify_event.sigev_signo  = CONFIG_NETINIT_SIGNO;
 
+#if !defined(CONFIG_ARCH_PHY_POLLED)
       ret = ioctl(sd, SIOCMIINOTIFY, (unsigned long)&ifr);
       if (ret < 0)
         {
@@ -636,6 +650,7 @@ static int netinit_monitor(void)
           goto errout_with_sigaction;
         }
 
+#endif
       /* Does the driver think that the link is up or down? */
 
       ret = ioctl(sd, SIOCGIFFLAGS, (unsigned long)&ifr);
@@ -776,11 +791,13 @@ static int netinit_monitor(void)
   /* TODO: Stop the PHY notifications and remove the signal handler. */
 
 errout_with_notification:
+#if !defined(CONFIG_ARCH_PHY_POLLED)
   ifr.ifr_mii_notify_event.sigev_notify = SIGEV_NONE;
   ioctl(sd, SIOCMIINOTIFY, (unsigned long)&ifr);
 errout_with_sigaction:
   sigaction(CONFIG_NETINIT_SIGNO, &oact, NULL);
 errout_with_socket:
+#endif
   close(sd);
 errout:
   nerr("ERROR: Aborting\n");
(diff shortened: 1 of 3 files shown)


netutils/netinit/netinit.c
```diff
@@ -533,6 +554,10 @@ static void netinit_configure(void)
 #ifndef CONFIG_NETINIT_NETLOCAL
   /* Bring the network up. */
 
+#ifdef CONFIG_NETINIT_DHCPC
+    memset(&g_ds, 0, sizeof(g_ds));
```
xiaoxiang781216, line 658: don't need, global variable is already zerod

netutils/netinit/netinit.c
```diff
@@ -627,6 +655,18 @@ static int netinit_monitor(void)
...
 
 #endif
+
+#ifdef CONFIG_NETINIT_DHCPC
+  /* We assume that the netinit_thread was able to get a lease
+   * If it did we set up the renew point at lease_time / 2
+   * if it did not the renew point is now.
+   */
+
+  DEBUGVERIFY(clock_gettime(CLOCK_REALTIME, &dhcprenew));
```
xiaoxiang781216, line 768: should we use time API to simplify the timing calculator?

netutils/netinit/netinit.c
```diff
@@ -357,6 +365,85 @@ static void netinit_set_macaddr(void)
...
+static int netinit_get_ipaddrs(void)
+{
+#ifdef CONFIG_NETINIT_DHCPC
+  bool use_dhcp   = false;
+#endif
+  bool use_static = false;
+#ifdef CONFIG_NET_IPv4
+  struct in_addr addr;
+
+  int ret = boardctl(BOARDIOC_NETCONF, (uintptr_t) &g_netconf);
```
xiaoxiang781216, line 389: why not support ipv6

xiaoxiang781216 (changes requested):

netutils/netinit/netinit.c
patacongo: As I disucss in apache/incubator-nuttx#1886, I do not think it is appropriate to use boardctl() for this purpose when the established, mostly standard ways of getting this information.  NuttX is ALL about conformance to standards and we must be faithful to that.

boardctl() is unfortunately necessary, but we cannot let it be an expedient, low-effort solution for system interface issues that already have standard solutions.  We must to the work as necessary and not cut corners.

netutils/netinit/netinit.c
```diff
@@ -343,6 +365,85 @@ static void netinit_set_macaddr(void)
 #  define netinit_set_macaddr()
 #endif
 
+#ifdef CONFIG_BOARDCTL_NETCONF
+/****************************************************************************
```
patacongo, line 369: This should depend on the existence of etc/sysconfig/network-scripts/ifcfg. boardctl() should not be used as discussed in apache/incubator-nuttx#1886

netutils/netinit/netinit.c
```diff
@@ -411,45 +519,26 @@ static void netinit_set_ipaddrs(void)
...
-  if (netlib_ifup(NET_DEVNAME) < 0)
-    {
-      return;
-    }
-
-#ifdef CONFIG_WIRELESS_WAPI
-  /* Associate the wlan with an access point. */
-
-  if (netinit_associate(NET_DEVNAME) < 0)
+  int ret = -1;
```
patacongo, line 534: What does the return value -1 mean?  -EPERM?

netutils/netinit/netinit.c
```diff
@@ -411,45 +519,26 @@ static void netinit_set_ipaddrs(void)
...
+#ifdef CONFIG_BOARDCTL_NETCONF
+  if ((g_netconf.flags & BOARDIOC_NETCONF_DHCP) == 0)
     {
-      return;
+      return 1;
```
patacongo, line 538: What does the return value of 1 mean?

netutils/netinit/netinit.c
antmerlino: Sorry for the long, somewhat tangent reply.

I was curious about this comment. While I agree it is not necessary for the normal call vector, who says that `netinit_configure()` cannot be called a second time. In this case, I'd think ensuring a known state is a good idea.

I was curious if it was called anywhere twice, so I searched where `netinit_configure()` is called from. The only place it is _currently_ called is `netinit_bringup()` or `netinit_thread()` (which `netinit_bringup()` launches) 

But where is `netinit_bringup()` called from? 
`nsh_consolemain()`
Looking at the documentation I see:

``` 
 * Name: nsh_consolemain (Normal character device version)
 *
 * Description:
 *   This interfaces may be to called or started with task_start to start a
 *   single an NSH instance that operates on stdin and stdout.  This
 *   function does not normally return (see below).```

I think this raises a concern. If I wanted multiple NSH consoles, does this mean that my network gets re-configured each time any of them come up?

antmerlino (commented):
LGTM.

I've looked at this twice now and nothing stands out. I'll let the others comments get responded to before approving.

netutils/netinit/netinit.c
patacongo: There are three nsh_consolemain() functions, each of them calls netinit_bringup() unconditionally.  There is only one system console, but potentially many telnet NSH sessions.  There are at least two ways to get multiple session:

- Telnet.  But in this case, nsh_consolemain() will not be called.  A special nsh_telnetmain() is called instead and it does not re-initialize the network.  
- You can also create a new NSH session from the `nsh>` command line.  This, I suspect will be a problem because I think that nsh_consolemain() (and, hence, netinit_bringup()) WILL be called again.

I don't see any protections in netinit_bringup() or netinit_configure() to make them idempotent.  What does happen in this case?  My guess is that you are correct to be concerned.

patacongo:
> I've looked at this twice now and nothing stands out. I'll let the others comments get responded to before approving.

The use of boardctl() in this way is unacceptable.  I discuss in the matching PR apache/incubator-nuttx#1886

antmerlino:
> 
> 
> > I've looked at this twice now and nothing stands out. I'll let the others comments get responded to before approving.
> 
> The use of boardctl() in this way is unacceptable. I discuss in the matching PR [apache/incubator-nuttx#1886](https://github.com/apache/incubator-nuttx/pull/1886)

Not sure how I didn't see the nuttx changes PR or realize there would have needed to be one. That ones a little more active, I'll spend some time reviewing it since this change is meaningless until the other one is resolved.

patacongo:
> Not sure how I didn't see the nuttx changes PR or realize there would have to be one. That one's a little more active, I'll spend some time reviewing it since this change is meaningless until the other one is resolved.

I want that change to be made in a more standard and compatible way, using existing mechanisms, proper separation of application and OS functionality (DHCP is application functionality) and without using the KLUDGE boardctl() interface.

My concern is with the integrity and roadmap of the OS.  I don't want quick-n-dirty changes when a proper system solution is required.  That degrades the OS.

netutils/netinit/netinit.c
davids5: @antmerlino  Yes that make sense, I was only looking at the init thread case and overlooked it.

davids5:
> > Not sure how I didn't see the nuttx changes PR or realize there would have to be one. That one's a little more active, I'll spend some time reviewing it since this change is meaningless until the other one is resolved.
> 
> I want that change to be made in a more standard and compatible way, using existing mechanisms, proper separation of application and OS functionality (DHCP is application functionality) and without using the KLUDGE boardctl() interface.
> 
> My concern is with the integrity and roadmap of the OS. I don't want quick-n-dirty changes when a proper system solution is required. That degrades the OS.

Thank you all for all the feedback. 

While this solve the problem on a small system using netinit thread, it is far from what is needed as a general solution So I am closing this.

xiaoxiang781216:
@davids5 I think the first two patches are reasonable, could you remove the last one?

PR #410, opened 2020-09-29
Add LTP support

ttnie:
## Summary
Add LTP support

## Testing
test on local environment

### DIFF
--- a/testing/ltp/Kconfig
+++ b/testing/ltp/Kconfig
@@ -1,0 +1,10 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config TESTING_LTP
+	tristate "Linux Test Project"
+	default n
+	---help---
+        Enable support for the Linux Test Project(LTP) framework
--- a/testing/ltp/Make.defs
+++ b/testing/ltp/Make.defs
@@ -1,0 +1,24 @@
+############################################################################
+# testing/ltp/Make.defs
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+#
+############################################################################
+
+ifneq ($(CONFIG_TESTING_LTP),)
+CONFIGURED_APPS += $(APPDIR)/testing/ltp
+endif
--- a/testing/ltp/Makefile
+++ b/testing/ltp/Makefile
@@ -1,0 +1,115 @@
+
+# apps/testing/ltp/ltp/Makefile
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include $(APPDIR)/Make.defs
+
+LTP_UNPACK   = ltp
+LTP_URL      = https://github.com/linux-test-project/ltp.git
+
+TESTDIR      = $(LTP_UNPACK)/testcases/open_posix_testsuite
+
+ifeq ($(CONFIG_FS_AIO),)
+BLACKWORDS  += "aio.h"
+BLACKWORDS  += "SIGPOLL"
+endif
+BLACKWORDS  += "affinity.h"
+BLACKWORDS  += "langinfo.h"
+BLACKWORDS  += "ucontext.h"
+BLACKWORDS  += "noatime.h"
+BLACKWORDS  += "CLOCK_PROCESS_CPUTIME_ID"
+BLACKWORDS  += "CLOCK_THREAD_CPUTIME_ID"
+BLACKWORDS  += "fork()"
+BLACKWORDS  += "pthread_attr_setscope"
+BLACKWORDS  += "RLIMIT_MEMLOCK"
+BLACKWORDS  += "PTHREAD_SCOPE_PROCESS"
+BLACKWORDS  += "SIGABRT"
+BLACKWORDS  += "SIGBUS"
+BLACKWORDS  += "SIGFPE"
+BLACKWORDS  += "SIGHUP"
+BLACKWORDS  += "SIGILL"
+BLACKWORDS  += "SIGPROF"
+BLACKWORDS  += "SIGSEGV"
+BLACKWORDS  += "SIGSYS"
+BLACKWORDS  += "SIGTRAP"
+BLACKWORDS  += "SIGTSTP"
+BLACKWORDS  += "SIGTTIN"
+BLACKWORDS  += "SIGTTOU"
+BLACKWORDS  += "SIGURG"
+BLACKWORDS  += "SIGVTALRM"
+BLACKWORDS  += "SIGXCPU"
+BLACKWORDS  += "SIGXFSZ"
+BLACKWORDS  += "SIGSTKSZ"
+BLACKWORDS  += "stack_t"
+BLACKWORDS  += "siginterrupt"
+BLACKWORDS  += "threads_scenarii.c"
+
+BLACKWORDS  += "ILL_[A-Z]"
+BLACKWORDS  += "FPE_[A-Z]"
+BLACKWORDS  += "BUS_[A-Z]"
+BLACKWORDS  += "TRAP_[A-Z]"
+BLACKWORDS  += "SEGV_[A-Z]"
+BLACKWORDS  += "POLL_[A-Z]"
+BLACKWORDS  += "SA_ONSTACK"
+BLACKWORDS  += "MINSIGSTKSZ"
+BLACKWORDS  += "FPE_FLTINV"
+
+BLACKSRCS += testfrmw.c
+BLACKSRCS += 7-3-buildonly.c
+BLACKSRCS += 7-4-buildonly.c
+BLACKSRCS += 7-5-buildonly.c
+BLACKSRCS += 11-1-buildonly.c
+BLACKSRCS += 17-1-buildonly.c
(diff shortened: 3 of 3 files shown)


testing/ltp/Make.defs
```diff
@@ -0,0 +1,39 @@
+############################################################################
+# apps/testing/Make.defs
```
xiaoxiang781216, line 2: replace the copyright

testing/ltp/Kconfig
```diff
@@ -0,0 +1,10 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config TESTING_CXXTEST
```
xiaoxiang781216, line 6: change the name of option

xiaoxiang781216 (changes requested):

testing/ltp/Kconfig
ttnie: @xiaoxiang781216  update done.

testing/ltp/Make.defs
ttnie: @xiaoxiang781216 update done

acassis: Strange, it still displaying BSD License instead of Apache, is it some issue on github?

acassis (changes requested):

testing/ltp/Make.defs
ttnie: @acassis Thanks for your reminder, update done.

acassis:
@ttnie is it ready to merge?

btashton:
I would want to know more about the 20k line patch file before merging this in. It's GPL code in the patch.

acassis:
Since it will be used only for testing and will not be included in the final product's firmware, maybe it is not a big concern. What you think?

xiaoxiang781216:
> It's 20k lines of GPL we cannot have that in the Apache release bundle. Also it's not clear to me how we would maintain it.

The change isn't specific to NuttX, @ttnie and @PeterBee97 is preparing to upstream the patch to mainline.

xiaoxiang781216:
> @ttnie is it ready to merge?

There are few compiling warnings need to resolve before merging. We ares in China Nation Holiday, the progress is a little bit slow.

btashton:
> > It's 20k lines of GPL we cannot have that in the Apache release bundle. Also it's not clear to me how we would maintain it.
> 
> @btashton all changes aren't specific to NuttX, @ttnie and @PeterBee97 is preparing to upstream the patch to mainline.

Awesome that would be the best. If we hit a roadblock maybe we can consider handling it like we do with the other restricted tools. This is certainly something we want to be able to support

acassis:
> > @ttnie is it ready to merge?
> 
> @acassis there are few compiling warnings need to resolve before merging. We ares in China Nation Holiday, the progress is a little bit slow.

Hi @xiaoxiang781216 I'm sorry, I forgot about it. When people will return? By Thursday?

xiaoxiang781216:
Friday, but people may take one ro two days vacation.

justinmclean:
Just trying to understand the GPL comment above. The GPL license is not compatible with the Apache one and we can't have GPL code in our repo. ASF projects also can't have a dependancy on GPL code either. Is this the case with this PR?

btashton:
The this PR would enable users to optionally include the LTP as an application. To do so requires a very large patch applied (included in the PR) to what the user would choose to download.

That is why I had raised the issue about accepting this PR. The contents of the patch are being upstreamed to the LTP project. Once that is done we should be able to accept this as the GPL code will only be included if the user chooses to download it in addition to our Apache release.

patacongo:
@justinmclean Can you please clarify this.  Particularly the phrase "ASF projects also can't have a dependancy on GPL code either."
> 
> Just trying to understand the GPL comment above. The GPL license is not compatible with the Apache one and we can't have GPL code in our repo. ASF projects also can't have a dependancy on GPL code either. Is this the case with this PR?

This could would download and build the GPL Linux Test Program at build time.  So, even if they work around the issues of the GPL patches, the binaries built with option would include GPL code.  Is this a "dependency on GPL".  It seems so to me.

There are several other places where this download trick is used to download third part code into NuttX binaries at build time.  This is done to avoid ASF third party rules, to avoid conformance with the NuttX coding standard, and to deal with third party code that is undergoing changes.  It would be good to understand if the ASF has any restrictions or thoughts on this practice -- with both compatible and incompatible licenses.

btashton:
There are lot of other Apache projects that support being built with non allowed code such as some intel and Nvidia libraries. It is my understanding that the Apache project just still needs to be functional without it which certainly is the case here.

patacongo:
> 
> 
> There are lot of other Apache projects that support being built with non allowed code such as some intel and Nvidia libraries. It is my understanding that the Apache project just still needs to be functional without it which certainly is the case here.

That seems like a reasonable interpretation to me.  But it would still be good to have an official ASF position.

justinmclean:
Proprietary licenses like some Intel and Nvidia ones are different are in general not compatible with the Apache license as they add restrictions above and beyond what the Apache license allows. If a license add any condition (not just copyleft) then it cannot be included in a release or depended on.If you want an example of where a project can get into trouble look at the issues with MXNet project and it use of Nvidia licensed code. You can find this in the last several Incubator reports https://whimsy.apache.org/board/minutes/Incubator.html. The likely outcome of this is that either Nvidia will change it license or MXNet will no longer be an Apache project.
Category  X license dependancies are allowed if the feature is truely optional and the user need to install that themselves. See https://www.apache.org/legal/resolved.html#optional

justinmclean:
Now in general using non compatible licenses for testing is fine as no incapable nothing ends up in the final source or binary release. e.g JUnit But it seem like this issue may be more similar with one of the MXNet issue it that GPL licensed code ends up in the binary. Is this correct? If so there is a dependency on GPL and would not be allowed unless it's an optional feature that is not built by default and the user is made aware of the licensing implication if they do enable it.

justinmclean:
Here's the main information on GPL (and other Category X) licenses. https://www.apache.org/legal/resolved.html#prohibited

patacongo:
> Category X license dependancies are allowed if the feature is truely optional and the user need to install that themselves. See https://www.apache.org/legal/resolved.html#optional

Certainly, this code is _optional_ by that definition.  In this case, the NuttX project code would provide the logic to automatically download the code without requiring that the user perform the download.  So the use of GPL is completely transparent.

btashton:
Thanks @justinmclean yeah I had been watching the MXNet conversation.

Here I think we meet the Optional definition in that link. In this case a user would enable the LTP as an application specifically to test NuttX compliance with the POSIX interfaces, it would never be expected to be included in an end product using NuttX and no LTP GPL code would be included in the release.

We include some helpers in the build scripts to download the source but only if a user opts in to that feature.

The only users of this feature are likely NuttX contributors and maintainers.

patacongo:
> 
> 
> Now in general using non compatible licenses for testing is fine as no incapable nothing ends up in the final source or binary release. e.g JUnit But it seem like this issue may be more similar with one of the MXNet issue it that GPL licensed code ends up in the binary. Is this correct? If so there is a dependency on GPL and would not be allowed unless it's an optional feature that is not built by default and the user is made aware of the licensing implication if they do enable it.

We don't release binaries and the GPL code would not be present in the source releases either. In the build process, the GPL source would be automatically downloaded on demand and built into the binaries by the user.  This code would not be downloaded or built in the _normal_ OS build; it would only be included the if the user were making test-related build and if the user explicitly selected inclusion of the GPL code from a configuration menu.

justinmclean:
That all sounds fine to me, thanks for the explanation.

ttnie:
> @ttnie is it ready to merge?

@acassis back from holiday.

btashton:
@ttnie why do we need to make all of the functions and variables static here?  Seems like we should be able to work around this in the build without having to patch that in?

ttnie:
> @ttnie why do we need to make all of the functions and variables static here? Seems like we should be able to work around this in the build without having to patch that in?

please ignore this PR, we'll apply LTP modifications to GPL not nuttx.

btashton:
Ok, just wanted to see if there was a way to work around it if we face resistance upstream :)

ttnie:
> @ttnie why do we need to make all of the functions and variables static here? Seems like we should be able to work around this in the build without having to patch that in?



> Ok, just wanted to see if there was a way to work around it if we face resistance upstream :)

I make sense, the modifications of LTP is plan to added to GPL by our team member. And no need patch at all.

xiaoxiang781216:
> @ttnie why do we need to make all of the functions and variables static here? Seems like we should be able to work around this in the build without having to patch that in?

In Linux, each test case build into a separated ELF binary, but NuttX build all soruces into one big image(execpt we enable ELF binfmt). So the linker complain the duplicated symbols on NuttX. But anyway, it's always a good egnineering practice to limit the symbol scope as small as possible even it isn't so critical on Linux.

btashton:
> In Linux, each test case build into a separated ELF binary, but NuttX build all soruces into one big image(execpt we enable ELF binfmt). So the linker complain the duplicated symbols on NuttX. But anyway, it's always a good egnineering practice to limit the symbol scope as small as possible even it isn't so critical on Linux.

Yeah I was wondering if we could just create standalone ELF for each, but I agree lets see if we can just improve the LTP and use that as a fallback otherwise.

testing/ltp/Makefile
```diff
@@ -0,0 +1,113 @@
...
+CFLAGS       += -I$(LTP_UNPACK)
+CFLAGS       += -I$(TESTDIR)/include
+
+$(LTP_UNPACK):
+	$(Q) echo "git clone $(LTP_URL)"
+	$(Q) git clone $(LTP_URL)
+
+context:: $(LTP_UNPACK)
+
+
```
xiaoxiang781216, line 112: need add distclean to keep git status clear.

ttnie: > need add distclean to keep git status clear.

add clean part to makefile, pls review

testing/ltp/Makefile
```diff
@@ -0,0 +1,117 @@
...
+CFLAGS       += -I$(TESTDIR)/include
+
+$(LTP_UNPACK):
+	$(Q) echo "git clone $(LTP_URL)"
+	$(Q) git clone $(LTP_URL)
+
+context:: $(LTP_UNPACK)
+
+clean::
+	$(call DELFILE, .built)
```
xiaoxiang781216, line 113: remove since .built isn't generated anymore

testing/ltp/Makefile
```diff
@@ -0,0 +1,117 @@
...
+CFLAGS       += -I$(LTP_UNPACK)
+CFLAGS       += -I$(TESTDIR)/include
+
+$(LTP_UNPACK):
+	$(Q) echo "git clone $(LTP_URL)"
+	$(Q) git clone $(LTP_URL)
+
+context:: $(LTP_UNPACK)
+
+clean::
```
xiaoxiang781216, line 112: change to distclean

testing/ltp/Makefile
```diff
@@ -0,0 +1,117 @@
...
+	$(call DELFILE, .built)
+	$(call CLEAN)
```
xiaoxiang781216, line 114: remove,

ttnie: > remove,

update to distclean and remove clean

xiaoxiang781216:
LGTM, @btashton could you review again?

jerpelea:
LGTM

PR #414, opened 2020-09-30
fsutils/ipcfg fixes

davids5:
## Summary

1. fsutils:ipcfg Add extern 'C' under cpp
2. fsutils:ipcfg Fix range checking for PRTOCOL in bin mode
3. fsutils:ipcfg Fix fd propagation as return value in bin mode 

## Impact
Fixes use with C++
Fixes use with CONFIG_IPCFG_BINARY set

## Testing
px4_fmuv5x

### DIFF
--- a/fsutils/ipcfg/ipcfg.c
+++ b/fsutils/ipcfg/ipcfg.c
@@ -201,8 +201,6 @@ static int ipcfg_open(FAR const char *netdev, int oflags, mode_t mode)
   /* Now open the file */
 
   fd  = open(path, oflags, mode);
-  ret = OK;
-
   if (fd < 0)
     {
       ret = -errno;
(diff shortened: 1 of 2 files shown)


fsutils/ipcfg/ipcfg.c
```diff
@@ -220,11 +218,14 @@ static int ipcfg_open(FAR const char *netdev, int oflags, mode_t mode)
...
       ret = -errno;
+      close(fd);
       fprintf(stderr, "ERROR: Failed to seek to $ld: %d\n",
               (long)CONFIG_IPCFG_OFFSET, ret);
-      close(fd);
+      goto errout_with_path;
     }
+
 #endif
+  ret = fd;
```
patacongo, line 229: I don't see any logic difference with this change.  What is the problem that you are solving?  Or is this just NIH?

patacongo: Never mind.  It is clearer looking at the whole file.

davids5: No. It's a bug.
Valid fd is 3 but code returns 0 (OK)
image

Ouss4 (approved):
LGTM

PR #415, opened 2020-09-30
Network Monitor: Runtime selection of DHCP/Static IP with fall back set by fsutil/ipcfg and a polled mode for HW lacking a PHY interrupt 

davids5:
## Summary

Sister to https://github.com/apache/incubator-nuttx/pull/1924 - **merge this APP pr second**. 

**Please merge** https://github.com/apache/incubator-nuttx/pull/1924 and https://github.com/apache/incubator-nuttx-apps/pull/414 **first**.

 ### netinit:Network Monitor add a polled option
 
  Not all boards have in interrupt line from the phy to the Soc. This allows the phy to be polled for link status.
   This may not work on all MAC/PHY combination that have mutually exclusive link management and operating
   modes. The STM32F7 and LAN8742AI do not have such a limitation.

   To use this option the arch must supply CONFIG_ARCH_PHY_POLLED  and not CONFIG_ARCH_PHY_INTERRUPT.

 If the network is not connected the phy init may time out and netinit_net_bringup will return with out completing DHCP or initializing the ntp client.

This change uses the lack of the DHCP lease, to contiune the netinit_net_bringup, once the interface does come up. It also will renew the lease over time, and will re run the dhcp request after a network down/up transition.

###  netinit Net monitior support getting address from fsutils/ipcfg

Uses new fsutils/ipcfg to read the networks settings. 

 ###  netinit:Net monitior support fallback to static IP

   After CONFIG_NETINIT_FALLBACK dhcp attempts, fall back to  the static IP provided by the board via fsutils/ipcfg

## Impact

None All options have Kconfig knobs to turn them off by default

## Testing

FMUV5X with a test vector of all the new Kconfig options

### DIFF
--- a/netutils/netinit/Kconfig
+++ b/netutils/netinit/Kconfig
@@ -73,7 +73,7 @@ if NETINIT_THREAD
 config NETINIT_MONITOR
 	bool "Monitor link state"
 	default n
-	depends on ARCH_PHY_INTERRUPT && NETDEV_PHY_IOCTL && NET_UDP
+	depends on (ARCH_PHY_INTERRUPT || ARCH_PHY_POLLED) && NETDEV_PHY_IOCTL && NET_UDP
 	---help---
 		By default the net initialization thread will bring-up the network
 		then exit, freeing all of the resources that it required.  This is a
@@ -87,10 +87,26 @@ config NETINIT_MONITOR
 		required for network initialization are never released.
 
 if NETINIT_MONITOR
+config NETINIT_ESTABLISH_POLL_RATE
+	int "The poll rate in seconds, to check for link establishment."
+	default 2
+	depends on !ARCH_PHY_INTERRUPT && ARCH_PHY_POLLED
+	---help---
+		The network monitor will check the PHY link state every
+		NETINIT_ESTABLISH_POLL_RATE seconds for link establishment.
+
+config NETINIT_LOSS_POLL_RATE
+	int "The poll rate in seconds, to check for link loss."
+	default 3
+	depends on !ARCH_PHY_INTERRUPT && ARCH_PHY_POLLED
+	---help---
+		The network monitor will check the PHY link state every
+		NETINIT_LOSS_POLL_RATE seconds for link loss.
 
 config NETINIT_SIGNO
 	int "Notification signal number"
 	default 18
+	depends on ARCH_PHY_INTERRUPT && !ARCH_PHY_POLLED
 	---help---
 		The network monitor logic will receive signals when there is any
 		change in the link status.  This setting may be used to customize
--- a/netutils/netinit/Kconfig
+++ b/netutils/netinit/Kconfig
@@ -103,6 +103,18 @@ config NETINIT_LOSS_POLL_RATE
 		The network monitor will check the PHY link state every
 		NETINIT_LOSS_POLL_RATE seconds for link loss.
 
+config NETINIT_DHCP_FALLBACK
+	int "The number of failed DHCP attempts to fall back to using static settings"
+	default 5
+	depends on NETINIT_DHCPC && !ARCH_PHY_INTERRUPT && ARCH_PHY_POLLED && FSUTILS_IPCFG
+	---help---
+		When the network monitor is used with FSUTILS_IPCFG if, the
+		the BOOTPROTO_FALLBACK is the active protocol. NETINIT_DHCP_FALLBACK will
+		be used as the number of times, DHCP will be be attempted before
+		falling back to using the static settings. The time will be roughly
+		NETINIT_LOSS_POLL_RATE * NETINIT_DHCP_FALLBACK seconds.
+		Setting this value to 0, will disable this feature.
+
 config NETINIT_SIGNO
 	int "Notification signal number"
 	default 18
--- a/netutils/netinit/netinit.c
+++ b/netutils/netinit/netinit.c
@@ -221,13 +221,22 @@
 #  define AF_INETX AF_INET6
 #endif
 
-/* While the network is up, the network monitor really does nothing.  It
- * will wait for a very long time while waiting, it can be awakened by a
- * signal indicating a change in network status.
+/* If using signal notification, raise by a the PHY, then while the network
+ * is up, the network monitor really does nothing.  It will wait for a very
+ * long time while waiting, it can be awakened by a signal indicating a
+ * change in network status.
+ *
+ * If the network monitor is used in polled mode these values are set by the
+ * CONFIG_NETINIT_LOSS_POLL_RATE and CONFIG_NETINIT_ESTABLISH_POLL_RATE
  */
 
-#define LONG_TIME_SEC    (60*60) /* One hour in seconds */
-#define SHORT_TIME_SEC   (2)     /* 2 seconds */
+#if !(defined(CONFIG_ARCH_PHY_POLLED))
+#  define LONG_TIME_SEC    (60*60) /* One hour in seconds */
+#  define SHORT_TIME_SEC   (2)     /* 2 seconds */
+#else
+#  define LONG_TIME_SEC    (CONFIG_NETINIT_LOSS_POLL_RATE)
+#  define SHORT_TIME_SEC   (CONFIG_NETINIT_ESTABLISH_POLL_RATE)
+#endif
 
 /****************************************************************************
  * Private Data
@@ -537,7 +546,7 @@ static void netinit_configure(void)
  *
  ****************************************************************************/
 
-#ifdef CONFIG_NETINIT_MONITOR
+#if defined(CONFIG_NETINIT_MONITOR) && !defined(CONFIG_ARCH_PHY_POLLED)
 static void netinit_signal(int signo, FAR siginfo_t *siginfo,
                                FAR void * context)
 {
@@ -573,8 +582,10 @@ static int netinit_monitor(void)
   struct timespec abstime;
   struct timespec reltime;
   struct ifreq ifr;
+#if !defined(CONFIG_ARCH_PHY_POLLED)
   struct sigaction act;
   struct sigaction oact;
+#endif
   bool devup;
   int ret;
   int sd;
@@ -599,6 +610,7 @@ static int netinit_monitor(void)
       goto errout;
     }
 
+#if !defined(CONFIG_ARCH_PHY_POLLED)
   /* Attach a signal handler so that we do not lose PHY events */
 
   act.sa_sigaction = netinit_signal;
@@ -614,6 +626,7 @@ static int netinit_monitor(void)
       goto errout_with_socket;
     }
 
+#endif
   /* Now loop, waiting for changes in link status */
 
   for (; ; )
@@ -626,6 +639,7 @@ static int netinit_monitor(void)
       ifr.ifr_mii_notify_event.sigev_notify = SIGEV_SIGNAL;
       ifr.ifr_mii_notify_event.sigev_signo  = CONFIG_NETINIT_SIGNO;
 
+#if !defined(CONFIG_ARCH_PHY_POLLED)
       ret = ioctl(sd, SIOCMIINOTIFY, (unsigned long)&ifr);
       if (ret < 0)
         {
@@ -636,6 +650,7 @@ static int netinit_monitor(void)
           goto errout_with_sigaction;
         }
 
+#endif
       /* Does the driver think that the link is up or down? */
 
       ret = ioctl(sd, SIOCGIFFLAGS, (unsigned long)&ifr);
@@ -776,11 +791,13 @@ static int netinit_monitor(void)
   /* TODO: Stop the PHY notifications and remove the signal handler. */
 
 errout_with_notification:
+#if !defined(CONFIG_ARCH_PHY_POLLED)
   ifr.ifr_mii_notify_event.sigev_notify = SIGEV_NONE;
   ioctl(sd, SIOCMIINOTIFY, (unsigned long)&ifr);
 errout_with_sigaction:
   sigaction(CONFIG_NETINIT_SIGNO, &oact, NULL);
 errout_with_socket:
+#endif
   close(sd);
 errout:
   nerr("ERROR: Aborting\n");
--- a/netutils/netinit/netinit.c
+++ b/netutils/netinit/netinit.c
@@ -1,9 +1,10 @@
 /****************************************************************************
  * apps/netutils/netinit/netinit.c
  *
- *   Copyright (C) 2010-2012, 2014-2016, 2019 Gregory Nutt. All rights
+ *   Copyright (C) 2010-2012, 2014-2016, 2019-2020 Gregory Nutt. All rights
  *     reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ *   AuthorS: Gregory Nutt <gnutt@nuttx.org>
+ *            David Sidrane <david.sidrane@nscdg.com>
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -242,6 +243,10 @@
  * Private Data
  ****************************************************************************/
 
+#ifdef CONFIG_NETINIT_DHCPC
+static struct dhcpc_state g_ds;
+#endif
+
 #ifdef CONFIG_NETINIT_MONITOR
 static sem_t g_notify_sem;
 #endif
@@ -420,6 +425,65 @@ static void netinit_set_ipaddrs(void)
 #endif
 
 /****************************************************************************
+ * Name: netinit_dhcp()
+ *
+ * Description:
+ *   Renew or obtain a IP Address over DHCP
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_NETINIT_DHCPC
+static int netinit_dhcp(struct dhcpc_state *ds)
+{
+  uint8_t mac[IFHWADDRLEN];
+  FAR void *handle;
+  int ret = -1;
+
+  /* Get the MAC address of the NIC */
+
+  netlib_getmacaddr(NET_DEVNAME, mac);
+
+  /* Set up the DHCPC modules */
+
+  handle = dhcpc_open(NET_DEVNAME, &mac, IFHWADDRLEN);
+  if (handle == NULL)
+    {
+      return ret;
+    }
+
+  /* Get an IP address.  Note that there is no logic for renewing the
+   * IP address in this example. The address should be renewed in
+   * (ds.lease_time / 2) seconds.
+   */
+
+  ret = dhcpc_request(handle, ds) ;
+
+  if (ret == OK)
+    {
+      netlib_set_ipv4addr(NET_DEVNAME, &ds->ipaddr);
+
+      if (ds->netmask.s_addr != 0)
+        {
+          netlib_set_ipv4netmask(NET_DEVNAME, &ds->netmask);
+        }
+
+      if (ds->default_router.s_addr != 0)
+        {
+          netlib_set_dripv4addr(NET_DEVNAME, &ds->default_router);
+        }
+
+      if (ds->dnsaddr.s_addr != 0)
+        {
+          netlib_set_ipv4dnsaddr(&ds->dnsaddr);
+        }
+    }
+
+  dhcpc_close(handle);
+  return ret;
+}
+#endif
+
+/****************************************************************************
  * Name: netinit_net_bringup()
  *
  * Description:
@@ -430,12 +494,6 @@ static void netinit_set_ipaddrs(void)
 #if defined(NETINIT_HAVE_NETDEV) && !defined(CONFIG_NETINIT_NETLOCAL)
 static void netinit_net_bringup(void)
 {
-#ifdef CONFIG_NETINIT_DHCPC
-  uint8_t mac[IFHWADDRLEN];
-  struct dhcpc_state ds;
-  FAR void *handle;
-#endif
-
   /* Bring the network up. */
 
   if (netlib_ifup(NET_DEVNAME) < 0)
@@ -459,44 +517,7 @@ static void netinit_net_bringup(void)
 #endif
 
 #ifdef CONFIG_NETINIT_DHCPC
-  /* Get the MAC address of the NIC */
-
-  netlib_getmacaddr(NET_DEVNAME, mac);
-
-  /* Set up the DHCPC modules */
-
-  handle = dhcpc_open(NET_DEVNAME, &mac, IFHWADDRLEN);
-  if (handle == NULL)
-    {
-      return;
-    }
-
-  /* Get an IP address.  Note that there is no logic for renewing the
-   * IP address in this example. The address should be renewed in
-   * (ds.lease_time / 2) seconds.
-   */
-
-  if (dhcpc_request(handle, &ds) == OK)
-    {
-      netlib_set_ipv4addr(NET_DEVNAME, &ds.ipaddr);
-
-      if (ds.netmask.s_addr != 0)
-        {
-          netlib_set_ipv4netmask(NET_DEVNAME, &ds.netmask);
-        }
-
-      if (ds.default_router.s_addr != 0)
-        {
-          netlib_set_dripv4addr(NET_DEVNAME, &ds.default_router);
-        }
-
-      if (ds.dnsaddr.s_addr != 0)
-        {
-          netlib_set_ipv4dnsaddr(&ds.dnsaddr);
-        }
-    }
-
-  dhcpc_close(handle);
+  netinit_dhcp(&g_ds);
 #endif
 
 #ifdef CONFIG_NETUTILS_NTPCLIENT
@@ -532,6 +553,10 @@ static void netinit_configure(void)
 
 #ifndef CONFIG_NETINIT_NETLOCAL
   /* Bring the network up. */
+
+#ifdef CONFIG_NETINIT_DHCPC
+    memset(&g_ds, 0, sizeof(g_ds));
+#endif
 
   netinit_net_bringup();
 #endif
@@ -586,6 +611,9 @@ static int netinit_monitor(void)
   struct sigaction act;
   struct sigaction oact;
 #endif
+#ifdef CONFIG_NETINIT_DHCPC
+  struct timespec dhcprenew;
+#endif
   bool devup;
   int ret;
   int sd;
@@ -627,6 +655,18 @@ static int netinit_monitor(void)
     }
 
 #endif
+
+#ifdef CONFIG_NETINIT_DHCPC
+  /* We assume that the netinit_thread was able to get a lease
+   * If it did we set up the renew point at lease_time / 2
+   * if it did not the renew point is now.
+   */
+
+  DEBUGVERIFY(clock_gettime(CLOCK_REALTIME, &dhcprenew));
+  dhcprenew.tv_sec  += g_ds.lease_time / 2;
+  dhcprenew.tv_nsec = 0;
+#endif
+
   /* Now loop, waiting for changes in link status */
 
   for (; ; )
@@ -735,6 +775,43 @@ static int netinit_monitor(void)
             {
               /* The link is still up.  Take a long, well-deserved rest */
 
+#ifdef CONFIG_NETINIT_DHCPC
+              /* Get Current time */
+
+              DEBUGVERIFY(clock_gettime(CLOCK_REALTIME, &abstime));
+
+              /* Did we ever get a lease */
+
+              if (g_ds.lease_time == 0)
+                {
+                  /* No so, let's try to bring up the network */
+
+                  netinit_net_bringup();
+
+                  /* Did we get a lease */
+
+                  if (g_ds.lease_time != 0)
+                    {
+                      /* Yes, set next renewal time */
+
+                      DEBUGVERIFY(clock_gettime(CLOCK_REALTIME, &dhcprenew));
+                      dhcprenew.tv_sec  += g_ds.lease_time / 2;
+                    }
+                }
+              else if (abstime.tv_sec > dhcprenew.tv_sec)
+                {
+                  /* Do the renew */
+
+                  if (netinit_dhcp(&g_ds) == OK)
+                    {
+                      /* Yes, set next renewal time */
+
+                      DEBUGVERIFY(clock_gettime(CLOCK_REALTIME, &dhcprenew));
+                      dhcprenew.tv_sec  += g_ds.lease_time / 2;
+                    }
+                }
+
+#endif
               reltime.tv_sec  = LONG_TIME_SEC;
               reltime.tv_nsec = 0;
             }
@@ -763,6 +840,13 @@ static int netinit_monitor(void)
                 }
             }
 
+#ifdef CONFIG_NETINIT_DHCPC
+          /* Link went down, next up could be a different network,
+           * so force a renew, on the next up.
+           */
+
+          dhcprenew.tv_sec = 0;
+#endif
           /* In either case, wait for the short, configurable delay */
 
           reltime.tv_sec  = CONFIG_NETINIT_RETRYMSEC / 1000;
--- a/netutils/netinit/netinit.c
+++ b/netutils/netinit/netinit.c
@@ -1,35 +1,21 @@
 /****************************************************************************
  * apps/netutils/netinit/netinit.c
  *
- *   Copyright (C) 2010-2012, 2014-2016, 2019-2020 Gregory Nutt. All rights
- *     reserved.
- *   AuthorS: Gregory Nutt <gnutt@nuttx.org>
- *            David Sidrane <david.sidrane@nscdg.com>
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
  *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
+ *   http://www.apache.org/licenses/LICENSE-2.0
  *
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in the
- *    documentation and/or other materials provided with the distribution.
- * 3. Neither the name of the Institute nor the names of its contributors
- *    may be used to endorse or promote products derived from this software
- *    without specific prior written permission.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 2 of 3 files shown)


davids5:
> @davids5 should we keep the first two patches which isn't related to boardctrl?

@xiaoxiang781216 Done and using ipcfg.

netutils/netinit/netinit.c
```diff
@@ -1,9 +1,10 @@
...
  *
- *   Copyright (C) 2010-2012, 2014-2016, 2019 Gregory Nutt. All rights
+ *   Copyright (C) 2010-2012, 2014-2016, 2019-2020 Gregory Nutt. All rights
  *     reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ *   AuthorS: Gregory Nutt <gnutt@nuttx.org>
+ *            David Sidrane <david.sidrane@nscdg.com>
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
```
patacongo, line 10: Why not an Apache 2.0 header?

patacongo:
@antmerlino Anthony, your reviewed and approved the previous iteration this PR, can you extend that to this updated version?

netutils/netinit/Kconfig
```diff
@@ -103,6 +103,18 @@ config NETINIT_LOSS_POLL_RATE
 		The network monitor will check the PHY link state every
 		NETINIT_LOSS_POLL_RATE seconds for link loss.
 
+config NETINIT_FALLBACK
+	int "The number of failed DHCP attempts to fall back to using static settings"
+	default 5
+	depends on !ARCH_PHY_INTERRUPT && ARCH_PHY_POLLED && BOARDCTL_NETCONF
```
xiaoxiang781216, line 109: remove BOARDCTL_NETCONF, why the fallback can't work with ARCH_PHY_INTERRUPT?

netutils/netinit/netinit.c
```diff
@@ -357,6 +367,86 @@ static void netinit_set_macaddr(void)
...
+{
+#ifdef CONFIG_NETINIT_DHCPC
+  bool use_dhcp   = false;
+#endif
+  bool use_static = false;
+#ifdef CONFIG_NET_IPv4
+  struct in_addr addr;
+  int ret;
+
+  ret = ipcfg_read(NET_DEVNAME, &g_netconf);
```
xiaoxiang781216, line 392: why not call ipcfg_read for IPv6?

netutils/netinit/Kconfig
davids5: @xiaoxiang781216 - Thank you I missed the one. I think it need to be FSUTIL_IPCFG to work. This is what provides the static IP. 

As far as ARCH_PHY_POLLED.  I do not have HW with the IRQ connected. The timing on the fall back is condition by the ARCH_PHY_POLLED, so I am not sure how it will work.

Do you have HW with the IRQ connected? Can you test it?

netutils/netinit/netinit.c
davids5: I do not have a way to test it.  Can you test it?, it it works we can move the call up out of the #ifdef CONFIG_NET_IPv4

netutils/netinit/netinit.c
davids5: I will take that as you grant, I will update it.

netutils/netinit/netinit.c
patacongo: There is no support for IPv6 in ipcfg_read/write at present.  That would have to be extended first.  There are some semi-standard IPv6 definitions (although many of them do not apply):

https://www.thegeekdiary.com/understanding-the-network-interface-configuration-file-etc-sysconfig-network-scripts-ifcfg-eth/
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/s1-networkscripts-interfaces

But it is not clear how to integrate IPv6 support into the binary interface without useless explosive growth since the binary form contains all possible settings. IPv6 support would require some additional design work. Certainly, if IPv6 support is added, that will delay this so that it misses the Window for the 10.0 release.  So, the options are IPv4 only for now (10.0) or with IPv6 support in 10.1.

davids5: I totally agree. I will push the minor changes in a bit....

patacongo: > 
> 
> I totally agree. I will push the minor changes in a bit....

Don't get me wrong.  I think IPv6 support is critical, at least for static IP assignment (IPv6 does not use DHCP but has a separate auto-configuration feature).  The only issue if we merge this now and incorporate another PR later, or if we essentially close this and delay merging to add the IPv6 support.

Delaying the merge provides better assurance that the IPv6 support will not be swept under the rug but pushes the feature to 10.1.  Really your choice.  I have no strong opinion.  I am just trying to point features.

netutils/netinit/netinit.c
davids5: Done in 22493084fcb81fdd055eb1075f6636a0475721dd

netutils/netinit/Kconfig
davids5: done in 664208ece0b24ac2b219694111fc1847dcadc444 fixup

netutils/netinit/netinit.c
davids5: I think this should come in as is and be in the 10.0 release. I will commit to help implement IPv6 in the 10.1 release.   I am not set up currently to test IPv6.

netutils/netinit/netinit.c
```diff
@@ -720,6 +867,80 @@ static int netinit_monitor(void)
...
+                    }
+#if defined(USE_FALLBACK)
+                  else
+                    {
+                      if ((g_netconf.proto & BOOTPROTO_FALLBACK) ==
+                          BOOTPROTO_FALLBACK &&
+                          ++dhcp_fail_count > CONFIG_NETINIT_FALLBACK)
+                        {
+                          struct in_addr addr;
+                          g_netconf.proto &= ~BOOTPROTO_DHCP;
```
antmerlino, line 903: @davids5 Will this create the desired behavior? I would think that if the cable gets unplugged, and then replugged in where DHCP may not fail, we'd want to use DHCP again. Right?  Perhaps I am seeing it wrong, but with the current logic, once you fallback to static once, that's it.  Is that the desired behavior?

antmerlino (commented):
Not able to test it right now, but this looks good to me, besides for my one comment.

netutils/netinit/netinit.c
davids5: Yes. It is.  The idea being that it stays at a know address once the timeout occurs.

netutils/netinit/netinit.c
```diff
@@ -357,6 +367,86 @@ static void netinit_set_macaddr(void)
...
+#endif
+  bool use_static = false;
+#ifdef CONFIG_NET_IPv4
+  struct in_addr addr;
+  int ret;
+
+  ret = ipcfg_read(NET_DEVNAME, &g_netconf);
+  if (ret < 0)
+    {
+      return ret;
```
xiaoxiang781216, line 381: can we initialize g_netconf here with the default value? so we don't need duplicate netlib_set_xxx.

netutils/netinit/Kconfig
```diff
@@ -87,10 +87,38 @@ config NETINIT_MONITOR
...
+	default 3
+	depends on !ARCH_PHY_INTERRUPT && ARCH_PHY_POLLED
+	---help---
+		The network monitor will check the PHY link state every
+		NETINIT_LOSS_POLL_RATE seconds for link loss.
+
+config NETINIT_FALLBACK
+	int "The number of failed DHCP attempts to fall back to using static settings"
+	default 5
+	depends on !ARCH_PHY_INTERRUPT && ARCH_PHY_POLLED && FSUTILS_IPCFG
```
xiaoxiang781216, line 109: Since DHCP isn't relate to the PHY operation, I don't think there are any reason to disable FALLBACK in this case. Let's desgin the thing right at the first place. BTW, the option should depend on DHCPC instead, add rename to NETINIT_DHCPC_FALLBACK.

xiaoxiang781216:
Let's do the thing right and correctly, I don't think this feature is so critical to rush into 10.0 release timeframe.

xiaoxiang781216 (changes requested):

antmerlino:
> 
> 
> Let's do the thing right and correctly, I don't think this feature is so critical to rush into 10.0 release timeframe. It's better to get IPv6 support and then release in 10.1.

@xiaoxiang781216 I disagree. As long as there isn't anything wrong with these IPv4 changes, and it doesn't break or change anything that worked before for IPv6, then we should merge it. 

Does calling it for IPv6 even get us anything right now anyway? Currently, the ipcfg files ONLY have a setting that is specific to DHCP use-cases. We don't even have a DHCPv6 implementation for this to be relevant for.

patacongo:
> Does calling it for IPv6 even get us anything right now anyway? Currently, the ipcfg files ONLY have a setting that is specific to DHCP use-cases. We don't even have a DHCPv6 implementation for this to be relevant for.

We do not have DHCPv6 but we do use ICMPv6 auto-configuration which is the more common IPv6 mechanism for IP address assignment.  That is the analogous step for IPv6.  So both auto-configuration and static IPv6 addresses would be supported, just as DHCP and static IPv4 addresses are be supported.

netutils/netinit/Kconfig
davids5: @xiaoxiang781216 without indications of state change either via interrupts of polling fallback will not work correctly.

davids5: > BTW, the option should depend on DHCPC instead, add rename to NETINIT_DHCPC_FALLBACK.

Rename done. dependency on NETINIT_DHCPC added in 078d39e5e and force pushed.

netutils/netinit/netinit.c
davids5: No. That can be done when we add IPv6.

davids5:
> > Let's do the thing right and correctly, I don't think this feature is so critical to rush into 10.0 release timeframe. It's better to get IPv6 support and then release in 10.1.
> 
> @xiaoxiang781216 I disagree. As long as there isn't anything wrong with these IPv4 changes, and it doesn't break or change anything that worked before for IPv6, then we should merge it.

This PR is to the benefit IPv4.  There is nothing wrong or incorrect other than it does not solve your desire for IPv6.  In fact, it can be totally turned off.

I suggest you merge this PR as it stands now. You can then take over the IPv6 work. I do not have a use case (in ASF terms. it is not my ITCH) for it nor am I set up to test IPv6. 

All I ask is you do not break the IPv4 functionality as it solves a problem with multi system disparate boot times.

xiaoxiang781216:
> > > Let's do the thing right and correctly, I don't think this feature is so critical to rush into 10.0 release timeframe. It's better to get IPv6 support and then release in 10.1.
> > 
> > 
> > @xiaoxiang781216 I disagree. As long as there isn't anything wrong with these IPv4 changes, and it doesn't break or change anything that worked before for IPv6, then we should merge it.
> 
> This PR is to the benefit IPv4. There is nothing wrong or incorrect other than it does not solve your desire for IPv6. In fact, it can be totally turned off.
> 
> I suggest you merge this PR as it stands now. You can then take over the IPv6 work. I do not have a use case (in ASF terms. it is not my ITCH) for it nor am I set up to test IPv6.
> 
> All I ask is you do not break the IPv4 functionality as it solves a problem with multi system disparate boot times.

No, as @patacongo mention that the addition of saving IPv6 in binary format will definitely break the binary compability on EEPROM/file. Before we get the solution on this, this patch shouldn't merge. Nobody want that 10.1 release can't load the config saved by 10.0.

netutils/netinit/netinit.c
xiaoxiang781216: It's not related to IPv6 even IPv4 we have the duplicaed code from line 393-403 and line 468-478.

davids5:
What binary compatibility? In Binary mode, it supports IPv4 now on one interface, nothing more. 

As Greg said: 

> But it is not clear how to integrate IPv6 support into the binary interface without *********useless******  ********explosive****** growth since the binary form contains all possible settings.

We can version the structure if you think it is necessary.  Or the user can use the Non binary version until you make it perfect. 

Now we can not be changed setting at run time and that is a BIG problem.

patacongo:
> We can version the structure if you think it is necessary. Or the user can use the Non binary version until you make it perfect.

We are not designing code here.  But my thought was to put two different binary structures back-to-back, one for IPv4 and one for IPv6.  Either or both could be present.  Each would add only:

- An byte size offset to the next structure (zero meaning the end of the list)
- An byte-size identifier to identify the type of structure IPv4 or IPv6

This would eliminate the "explosive growth".  If we were to add those two bytes to the beginning of the structure now (both set to zero and type == 0 meaning IPv4), then we would have binary backward compatibility in the future.  No versioning is really needed.

The ipcfg_read() interface would change in that you would requires an address family type as an additional argument.

We freeze 10.0 tomorrow.  There is no time to road test any significant changes now.  None at all.

xiaoxiang781216:
> What binary compatibility? In Binary mode, it supports IPv4 now on one interface, nothing more.

I mean that the new code still can read the old config after we add IPv6 support.

> 
> As Greg said:
> 
> > But it is not clear how to integrate IPv6 support into the binary interface without *********useless******  ********explosive****** growth since the binary form contains all possible settings.
> 
> We can version the structure if you think it is necessary. Or the user can use the Non binary version until you make it perfect.
> 

Yes, the version is the mininal requirement. Once we announce this feature, the user may select to use the binary format.

> Now we can not change the setting at run time and that is a BIG problem.

It is very easy to cherry-pick the patch once this PR get merged. I don't think it's so critical to rush it into release 10.0. Especially, this PR contain the code to save/load the user data. It's always the highest issue to ensure the use never lose their data in any situation.

netutils/netinit/netinit.c
davids5: It is not the same 

image

image

the compile time #ifdefs will make it not readable just setting the protocol.

netutils/netinit/netinit.c
```diff
@@ -221,18 +212,37 @@
...
  * Private Data
  ****************************************************************************/
 
+#ifdef CONFIG_NETINIT_DHCPC
+static struct dhcpc_state g_ds;
+#endif
+#if defined(CONFIG_FSUTILS_IPCFG)
+static struct ipcfg_s g_netconf;
+#  if defined(CONFIG_NETINIT_DHCP_FALLBACK) && CONFIG_NETINIT_DHCP_FALLBACK > 0
+#    define USE_FALLBACK 1
```
xiaoxiang781216, line 242: USE_FALLBACK to USE_DHCP_FALLBACK?

xiaoxiang781216 (changes requested):

netutils/netinit/Kconfig
xiaoxiang781216: > @xiaoxiang781216 without indications of state change either via interrupts of polling fallback will not work correctly.

Could you point out where the dhcp fallback code depend on the poll? I can't find there is the coupling between the fallback and polling. I expect that the dependence is:
```
config NETINIT_FALLBACK
	int "The number of failed DHCP attempts to fall back to using static settings"
	default 5
	depends on (ARCH_PHY_INTERRUPT || ARCH_PHY_POLLED) && FSUTILS_IPCFG
```

netutils/netinit/netinit.c
xiaoxiang781216: Do you think the follow code is reasonable and more simple?
```
ret = ipcfg_read(NET_DEVNAME, &g_netconf);
if (ret < 0)
   {
     /* Set the default value from Kconfig here */
     g_netconf.proto = BOOTPROTO_STATIC;
     g_netconf.s_addr = CONFIG_NETINIT_IPADDR;
     ...
   }

/* Share the same code regardless where we get the config from EEPROM or Kconfig */
...
```

xiaoxiang781216:
@davids5 all patches except patch 3 is in the good shape, you could split patch 3 from this PR and then merge the rest first if you want. BTW, I am wondering why patch 3 is part of this PR, after all patch 3 is unrelated to other patches in this PR.

netutils/netinit/netinit.c
davids5: Done in 9f21ad3e4 and FP

davids5:
> An byte size offset to the next structure (zero meaning the end of the list)
> An byte-size identifier to identify the type of structure IPv4 or IPv6

@patacongo - This makes good sense.  Will you do the PR? I  can then merge it and fix up this one.

davids5:
> BTW, I am wondering why patch 3 is part of this PR, after all patch 3 is unrelated to other patches in this PR.

What is the shal of the commit you are questioning?

netutils/netinit/netinit.c
```diff
@@ -221,13 +221,22 @@
...
+ * change in network status.
+ *
+ * If the network monitor is used in polled mode these values are set by the
+ * CONFIG_NETINIT_LOSS_POLL_RATE and CONFIG_NETINIT_ESTABLISH_POLL_RATE
  */
 
-#define LONG_TIME_SEC    (60*60) /* One hour in seconds */
-#define SHORT_TIME_SEC   (2)     /* 2 seconds */
+#if !(defined(CONFIG_ARCH_PHY_POLLED))
+#  define LONG_TIME_SEC    (60*60) /* One hour in seconds */
```
xiaoxiang781216, line 225: Why not make NETINIT_ESTABLISH_POLL_RATE/NETINIT_LOSS_POLL_RATE available when ARCH_PHY_INTERRUPT enable by removing the dependence(depends on !ARCH_PHY_INTERRUPT && ARCH_PHY_POLLED)?

patacongo:
> 
> 
> > An byte size offset to the next structure (zero meaning the end of the list)
> > An byte-size identifier to identify the type of structure IPv4 or IPv6
> 
> @patacongo - This makes good sense. Will you do the PR? I can then merge it and fix up this one.

Yes, it is just a matter of adding two fields to struct ipcfg_s.  In anticipation, it should be renamed ipv4cfg_s.

netutils/netinit/netinit.c
davids5: If you look ad the code this is enabled from the bottom up.  Formally by ARCH_PHY_INTERRUPT and now also by ARCH_PHY_INTERRUPT.

It is this way because I am respecting the previous functionality 100%. I find that to be respectful of all users. I explained to you before I do not have HW with a PHY that has an IRQ, I test my changes.  This approach ensures I did not break other peoples code.  If you can test it with **all** the options submit a PR and I will be happy to retest what I need working.

netutils/netinit/netinit.c
davids5: Will it compile if DCP is selected. CONFIG_NETINIT_IPADDR is not defined

netutils/netinit/Kconfig
davids5: It is the times,

netutils/netinit/netinit.c
xiaoxiang781216: I don't think the below code change will break the previous functionality:
```
config NETINIT_ESTABLISH_POLL_RATE
	int "The poll rate in seconds, to check for link establishment."
	default 2
	---help---
		The network monitor will check the PHY link state every
		NETINIT_ESTABLISH_POLL_RATE seconds for link establishment.

config NETINIT_LOSS_POLL_RATE
	int "The poll rate in seconds, to check for link loss."
	default 3600
	---help---
		The network monitor will check the PHY link state every
		NETINIT_LOSS_POLL_RATE seconds for link loss.

#  define LONG_TIME_SEC   CONFIG_NETINIT_LOSS_POLL_RATE
#  define SHORT_TIME_SEC CONFIG_NETINIT_ESTABLISH_POLL_RATE
```
It's the responsibility of maintainer to keep the code quality and avoid the code become "spanish noodles" as possible as we can.

patacongo:
> Yes, it is just a matter of adding two fields to struct ipcfg_s. In anticipation, it should be renamed ipv4cfg_s.

See PR #416

netutils/netinit/netinit.c
xiaoxiang781216: The code just give my rough idea to you, please adjust as need. The keypoint is that we can initialzie g_netconf with the default value if  ipcfg_read return fail, then we don't duplicate netlib_set_xxx twice.

netutils/netinit/Kconfig
xiaoxiang781216: Please either fix this issue or give a technical reason to not fix it. The time isn't a good reason to ignore the right concern.

davids5: LOL By time I meat the time the task waits. 

Have you read the code?  Interrupt mode is just that: Edges are created on interrupts. In Interrupt mode the task waits "Forever".  In polled mode the task waits at one of the 2 POLL rates. This is dependent  on the state. Since Polling is a waste of CPU, it is only used it this mode. But without the IRQ line it is the only way to do this.  The time the task waits is the N * the POLL rate for the fall back. It will not work, of should I say it will be WAY to long in IRQ mode.

netutils/netinit/netinit.c
davids5: I understand the original code to do the right thing in using interrupts.  I will not break that, nor should the user be exposed to it.

xiaoxiang781216: I read the code long time ago, but if my memory is correct, the polling logic is already supported to handle the hardware which can't generate the interrupt with the correct timing. It's very easy to support the PHY which can't generate the interrupt at all by making LONG_TIME_SEC/SHORT_TIME_SEC configurable.

netutils/netinit/Kconfig
xiaoxiang781216: The polling logic exist for a long time before ARCH_PHY_POLLED is added by you recently. The only real difference is that ARCH_PHY_POLLED need a short poll interval but ARCH_PHY_INTERRUPT need a long poll interval. So the simple change is let user config the interval through Kconfig.
But DHCP should happen only after the network establish the physical link. Why you think the network status polling ncopule with DHCP?

netutils/netinit/netinit.c
davids5: > I read the code long time ago, but if my memory is correct,

your recollection is not correct.

I ask the question on dev first - see ponymail for Greg's response and his concern about management while in link active. 

It does not support polling.

All of this has been discussed. 

See the title block:
```
Not all boards have in interrupt line from the phy to the Soc. This allows the phy to be polled for link status.
This may not work on all MAC/PHY combination that have mutually exclusive link management and operating
modes. ***********************The STM32F7 and LAN8742AI do not have such a limitation.********

To use this option the arch must supply CONFIG_ARCH_PHY_POLLED and not CONFIG_ARCH_PHY_INTERRUPT.
```
ARCH_PHY_POLLED - is the enabler set by the HW knowing it can deal with polling. 

Without being intimately familiar with the code and all the issues that have been addressed you are not adding value here.

davids5:
The last commit requires PR #416 to come in first.

davids5:
Rebased and force pushed

netutils/netinit/Kconfig
davids5: Just like the HW had to have the IRQ line the HW has to also support link management while active. See the mailing list for Greg concern regarding this.   

!ARCH_PHY_INTERRUPT && ARCH_PHY_POLLED is that enable. 

I t would not be good to allow the user to set poll rates on HW that does not tolerate it,  This is all documented in the HW Kconfig.

netutils/netinit/netinit.c
davids5: I will not be changing this. 1) It is stable, 2) easy to read, 3) compiles with all the permutations of the CONFIG settings. 

I suspect this code will change when the V6 part is done. Look to this is a major code saving you can add with all the IPv6.

netutils/netinit/Kconfig
xiaoxiang781216: @davids5 I am talking about DHCP fallback here, it isn't relate to the polling, rihgt? It is very bad to combine the unrelated patch into one PR, so please move DHCP fallback to the separated PR. So we can disscuss the patch individually.

netutils/netinit/netinit.c
xiaoxiang781216: > I will not be changing this. 1) It is stable, 2) easy to read, 3) compiles with all the permutations of the CONFIG settings.
> 
> I suspect this code will change when the V6 part is done. Look to this is a major code saving you can add with all the IPv6.

I don't agree, let's make the code in a good shape in the first place. If you code get merged as the current state, it will encourage people who add IPv6 support duplicate the similar code again.

netutils/netinit/netinit.c
xiaoxiang781216: Why do you think that I am not familiar with the net monitor code? I fixed a bug in it several months ago:
https://github.com/apache/incubator-nuttx-apps/commit/d96bc5a4fbb90754f5afececbdc97f3d5395076e
And create an issue suggest to decouple the monitor from ethernet hardware:
https://github.com/apache/incubator-nuttx-apps/commit/d96bc5a4fbb90754f5afececbdc97f3d5395076e

Let's come back to your change, here is the original code snippet:
···
static int netinit_monitor(void)
{
  /* Now loop, waiting for changes in link status */

  for (; ; )
    {
      /* Check for link up or down */

      if ((ifr.ifr_mii_val_out & MII_MSR_LINKSTATUS) != 0)
        {
          /* Link up... does the drive think that the link is up? */

          if (!devup)
            {
              /* No... We just transitioned from link down to link up.
               * Bring the link up.
               */

              /* And wait for a short delay.  We will want to recheck the
               * link status again soon.
               */

              reltime.tv_sec  = SHORT_TIME_SEC;
              reltime.tv_nsec = 0;
            }
          else
            {
              /* The link is still up.  Take a long, well-deserved rest */

              reltime.tv_sec  = LONG_TIME_SEC;
              reltime.tv_nsec = 0;
            }
        }
      else
        {
          /* Link down... Was the driver link state already down? */

          if (devup)
            {
              /* No... we just transitioned from link up to link down.  Take
               * the link down.
               */
                }
            }

          /* In either case, wait for the short, configurable delay */

          reltime.tv_sec  = CONFIG_NETINIT_RETRYMSEC / 1000;
          reltime.tv_nsec = (CONFIG_NETINIT_RETRYMSEC % 1000) * 1000000;
        }

      /* Now wait for either the semaphore to be posted for a timed-out to
       * occur.
       */

      sched_lock();
      DEBUGVERIFY(clock_gettime(CLOCK_REALTIME, &abstime));

      abstime.tv_sec  += reltime.tv_sec;
      abstime.tv_nsec += reltime.tv_nsec;
      if (abstime.tv_nsec >= 1000000000)
        {
          abstime.tv_sec++;
          abstime.tv_nsec -= 1000000000;
        }

      sem_timedwait(&g_notify_sem, &abstime);
      sched_unlock();
    }
}
···
The keypoint for polling is sem_timedwait which already exist for a long time. What your change is:
1. Skip the signal related setting if ARCH_PHY_POLLED
2. Make SHORT_TIME_SEC/LONG_TIME_SEC configurable
All change looks good, what I suggest is making Kconfig available in both(interrupt/polled) cases.

davids5: AGAIN **_I do not have HW to test your use case_**. If you test it. If it woks, do a PR again this branch.

netutils/netinit/Kconfig
davids5: No the polling adds the temporal component, 

There is a **great** possibility we are miss-communicating here (you and I have been is this sort of cycle before)

@xiaoxiang781216 At this point rather continue to say the same things over and over and over again. I invite you to do a PR against this branch with any changes you want. Verify it builds 100% with all the CONFIG* permutations. If it 1) works the same for me with, my setting AND 2) It does not expose the user to be able to configure the system in a way that will not work with their SoC's MAC and PHY, I will be happy to test it with my settings and merge your work into this PR.

Does that work for you?

netutils/netinit/netinit.c
xiaoxiang781216: Here is my suggestion:
image
image
There isn't any logic change and the timeout also same as before. If you can't guarantee the correctness, which mean you don't have confidence with your own change.

netutils/netinit/Kconfig
xiaoxiang781216: > No the polling adds the temporal component,
> 
> There is a **great** possibility we are miss-communicating here (you and I have been is this sort of cycle before)
> 

That is because you reject other people's comments, but don't give the technical reason.

> @xiaoxiang781216 At this point rather continue to say the same things over and over and over again. I invite you to do a PR against this branch with any changes you want. Verify it builds 100% with all the CONFIG* permutations. If it 1) works the same for me with, my setting AND 2) It does not expose the user to be able to configure the system in a way that will not work with their SoC's MAC and PHY, I will be happy to test it with my settings and merge your work into this PR.
> 
> Does that work for you?

Could you follow my suggestion to split the PR first?

davids5: I am afraid I have run out of time. I need all the features herein to fix all the issues I have documented.

netutils/netinit/netinit.c
```diff
@@ -62,6 +49,10 @@
 
 #include <nuttx/net/mii.h>
 
+#if defined(CONFIG_FSUTILS_IPCFG)
+#  include <fsutils/ipcfg.h>
+#endif
```
patacongo, line 54: Technically, should be enclosed in quotes because this is not a system include.

netutils/netinit/netinit.c
```diff
@@ -358,6 +448,13 @@ static void netinit_set_macaddr(void)
     defined(CONFIG_NET_IEEE802154)
 static void netinit_set_ipaddrs(void)
 {
+#ifdef CONFIG_FSUTILS_IPCFG
+  if (netinit_get_ipaddrs() == OK)
+    {
+      return;
+    }
+
+#endif
```
patacongo, line 457: It is a C90 violation to place executable logic before data definitions.

netutils/netinit/netinit.c
```diff
@@ -411,45 +508,26 @@ static void netinit_set_ipaddrs(void)
...
-      return;
-    }
-
-#ifdef CONFIG_WIRELESS_WAPI
-  /* Associate the wlan with an access point. */
-
-  if (netinit_associate(NET_DEVNAME) < 0)
+  int ret = -EAGAIN;
+#ifdef CONFIG_FSUTILS_IPCFG
+  if ((g_netconf.proto & IPv4PROTO_DHCP) == 0)
```
patacongo, line 525: Hmmm... & makes not sense here.  The prototype is not bit encoded, it is an enumeration.

netutils/netinit/netinit.c
```diff
@@ -343,6 +353,86 @@ static void netinit_set_macaddr(void)
...
+      return ret;
+    }
+  else
+    {
+      use_static = g_netconf.proto & BOOTPROTO_STATIC;
+
+#ifdef CONFIG_NETINIT_DHCPC
+      use_dhcp = g_netconf.proto & IPv4PROTO_DHCP;
+      addr.s_addr = use_dhcp ? 0 : HTONL(g_netconf.ipaddr);
+#else
```
patacongo, line 390: g_netconf.ipaddr is already in network order.  HTONL should not be used.

netutils/netinit/netinit.c
```diff
@@ -343,6 +353,86 @@ static void netinit_set_macaddr(void)
...
+      addr.s_addr = HTONL(g_netconf.ipaddr);
+#endif
```
patacongo, line 392: Ditto.  And for all over places where HTONL is used on addresses returned by ipcfg.

netutils/netinit/netinit.c
patacongo: Because of the particular definitions, that does test work.  But that is merely a quirk of fate, not by design.  By designed it was just an enumerated value.

netutils/netinit/netinit.c
davids5: @patacongo - good feedback. Thank you.

netutils/netinit/netinit.c
davids5: That should be easy to fix, in the re-write.

netutils/netinit/netinit.c
davids5: That was by design. I the original design, if you recall , it was called it flags. and I purposely spaced the code points. Were I obviously failed was documenting the bit of cleverness.  What was merely a quirk of fate, is that in the transitions from the boardctl to the ipcfg that  code point spacing was preserved.  did check it.

netutils/netinit/netinit.c
davids5: Since the original code has the #defined values in host order.

https://github.com/nuttx-to-asf/incubator-nuttx-apps/blob/5a6246c3f2adfd0a4cc10639eb9923682e8ce1f4/netutils/netinit/netinit.c#L472-L477

 I choose that convention.  Values in code and memory on the host are in host order. My application does a hton on load (after the read) and ntol on save (before the save). but avoid all that would be good.  

For the PX4 use case this was necessary for future exposing the configuration data in QCG.

netutils/netinit/netinit.c
davids5: "

netutils/netinit/netinit.c
patacongo: > 
> 
> Since the original code has the #defined values in host order.
> 
> https://github.com/nuttx-to-asf/incubator-nuttx-apps/blob/5a6246c3f2adfd0a4cc10639eb9923682e8ce1f4/netutils/netinit/netinit.c#L472-L477
> 
> I choose that convention. Values in code and memory on the host are in host order. My application does a hton on load (after the read) and ntoh on save (before the save). but avoid all that would be good.
> 
> For the PX4 use case this was necessary for future exposing the configuration data in QCG.

Then you should use NTOHL not HTONL.  It would be a better design to use all network order internally.  Of course the user provided addresses from the Kconfigs will be in host order but applying HTONL in that case has zero overhead; the endian-ness is resolved at compile time.  Converting variable data to host order only so that it can be converted back to network order is very inefficient.

It is also expensive because NTOHL and HTON is relatively big when they operate on variable data (as well as non-standard.)  The better (and standard) interfaces might be ntohl() and htonl() in these cases.  Best of all:  Keep everything in network order.

patacongo: Well, looking at the code again, it is wrong.  The s_addr must always be in network order.  You save it here in host order which is inconsistent with all other uses and would fail.  All other settings in the original:

    netinit.c:  addr.s_addr = HTONL(CONFIG_NETINIT_IPADDR);
    netinit.c:  addr.s_addr = 0;
    netinit.c:  addr.s_addr = HTONL(CONFIG_NETINIT_DRIPADDR);
    netinit.c:  addr.s_addr = HTONL(CONFIG_NETINIT_NETMASK);
    netinit.c:  addr.s_addr = HTONL(CONFIG_NETINIT_DNSIPADDR);

The configuration settings are ALL in host order and the HTONL converts them to network order which is correct.  But this is incorrect:

    addr.s_addr = use_dhcp ? 0 : HTONL(g_netconf.ipaddr);

Because g_netconfig.ipaddr is in network order.  The HTONL is incorrect, it should be NTOHL.  But that is still wong because s_addre must be in network order.  This fix is to remove the HTONL altogether.

PR #416, opened 2020-10-02
ipcfg:  Add hooks for future IPv6 support.

patacongo:
## Summary

This commit adds structures, modifies function prototypes, and renames data to support the future addition of IPv6 support.  This commit does _NOT_ add that IPv6, only the hooks for backward compatible future support.

## Impact

At present this has no impact because there is no existing use of the ipcfg access helpers in the repository.  However, it does affect the logic of the pending change of PR #415 

## Testing

Verified with examples/ipcfg in binary mode using stm32f4discovery:netnsh

### DIFF
--- a/fsutils/ipcfg/ipcfg.c
+++ b/fsutils/ipcfg/ipcfg.c
@@ -40,19 +40,33 @@
  ****************************************************************************/
 
 #define MAX_LINESIZE  80
-#define MAX_BOOTPROTO BOOTPROTO_FALLBACK
+#define MAX_IPv4PROTO IPv4PROTO_FALLBACK
+
+/* Values for the record type field */
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+/* IP Configuration record header. */
+
+struct ipcfg_header_s
+{
+  uint8_t next;         /* Offset to the next IP configuration record */
+  sa_family_t type;     /* Must be AF_INET */
+};
 
 /****************************************************************************
  * Private Data
  ****************************************************************************/
 
 #if defined(CONFIG_IPCFG_WRITABLE) && !defined(CONFIG_IPCFG_BINARY)
-static const char *g_proto_name[] =
+static const char *g_ipv4proto_name[] =
 {
-  "none",      /* BOOTPROTO_NONE */
-  "static",    /* BOOTPROTO_STATIC */
-  "dhcp",      /* BOOTPROTO_DHCP */
-  "fallback"   /* BOOTPROTO_FALLBACK */
+  "none",      /* IPv4PROTO_NONE */
+  "static",    /* IPv4PROTO_STATIC */
+  "dhcp",      /* IPv4PROTO_DHCP */
+  "fallback"   /* IPv4PROTO_FALLBACK */
 };
 #endif
 
@@ -321,7 +335,10 @@ static void ipcfg_putaddr(FAR FILE *stream, FAR const char *variable,
  * Input Parameters:
  *   netdev - The network device.  For examplel "eth0"
  *   ipcfg  - Pointer to a user provided location to receive the IP
- *            configuration.
+ *            configuration.  Refers to either struct ipv4cfg_s or
+ *            ipv6cfg_s, depending on the value of af.
+ *   af     - Identifies the address family whose IP configuration is
+ *            requested.  May be either AF_INET or AF_INET6.
  *
  * Returned Value:
  *   Zero is returned on success; a negated errno value is returned on any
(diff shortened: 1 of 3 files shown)


include/fsutils/ipcfg.h
```diff
@@ -31,23 +31,35 @@
...
+  IPTYPE_IPv6        = 1  /* IPv6 configuration */
+};
+
+/* Generic IP Configuration record header.  Must be cast compatible with
+ * both struct ipv4cfg_s and struct ipv6cfg_s
+ */
+
+struct ipcfg_s
+{
+  uint8_t next;                 /* Offset to the next IP configuration record */
```
xiaoxiang781216, line 48: Who is responsible to initialize next field in the read or write case? I don't see there is the code to reference this filed in ipcfg_read/ipcfg_write.

include/fsutils/ipcfg.h
```diff
@@ -137,7 +227,8 @@ int ipcfg_read(FAR const char *netdev, FAR struct ipcfg_s *ipcfg);
  ****************************************************************************/
 
 #ifdef CONFIG_IPCFG_WRITABLE
-int ipcfg_write(FAR const char *netdev, FAR const struct ipcfg_s *ipcfg);
+int ipcfg_write(FAR const char *netdev, FAR const struct ipcfg_s *ipcfg,
+                sa_family_t af);
```
xiaoxiang781216, line 231: If ipcfg_s already contain type field, why the call stll have to pass af argument?

include/fsutils/ipcfg.h
```diff
@@ -31,23 +31,35 @@
...
+};
+
+/* Generic IP Configuration record header.  Must be cast compatible with
+ * both struct ipv4cfg_s and struct ipv6cfg_s
+ */
+
+struct ipcfg_s
+{
+  uint8_t next;                 /* Offset to the next IP configuration record */
+  enum ipcfg_iptype_e type;     /* Must be IPTYPE_IPv4 */
```
davids5, line 49: @patacongo did you consider using the AF_INET/AF_INET6 here or did you want the extra abstraction>?

fsutils/ipcfg/ipcfg.c
```diff
@@ -498,14 +504,16 @@ int ipcfg_read(FAR const char *netdev, FAR struct ipcfg_s *ipcfg)
  ****************************************************************************/
 
 #ifdef CONFIG_IPCFG_WRITABLE
-int ipcfg_write(FAR const char *netdev, FAR const struct ipcfg_s *ipcfg)
```
davids5, line 501: After Line 498 needs the af param doc added

davids5 (changes requested):

include/fsutils/ipcfg.h
patacongo: Fixed.  the header is not managed internally by ipcfg_write() and ipcfg_read();

include/fsutils/ipcfg.h
patacongo: Done.  sa_family_t is a little larger, but more straightforward.

fsutils/ipcfg/ipcfg.c
patacongo: Fixed

include/fsutils/ipcfg.h
patacongo: Fixed.  the header is not managed internally by ipcfg_write() and ipcfg_read();

patacongo:
I will flesh out the IPv6 support after 10.0 is branched.

include/fsutils/ipcfg.h
davids5: is this the classic not->now?

davids5 (approved):
LGTM, We could debate the removing the AF from the calls, and setting the AF type in the struct prior to the call for read and write but that can be changed in the future, with almost 0 impact,

So let's get his in and the PR that uses it.

patacongo:
> LGTM, We could debate the removing the AF from the calls, and setting the AF type in the struct prior to the call for read and write but that can be changed in the future, with almost 0 impact,

That is a possibility on ipcfg_write(), but makes not sense for ipcfg_read().  It is better being symmetric the way that it is.

davids5:
> That is a possibility on ipcfg_write(), but makes not sense for ipcfg_read(). It is better being symmetric the way that it is.

good point, shall we merge this now?

patacongo:
> good point, shall we merge this now?

I know of no reason why we should not.  I will finish off the IPv6 logic in a subsequent PR after the merge.

PR #407, opened 2020-09-28
Add IPv4 Configuration File Access Helper

patacongo:
## Summary

This commit adds support to access an IPv4 Configuration file similar to the Linux dhpc.client ipfcg file.  This version, tailored for deeply embedded systems supports several options to tailor the file and file access to different environments.  It supports:

- Writable as well as read-only configuration files.
- ASCII human readable files as well as smaller binary files.
- It supports using character driver access to constrained media (such as EEPROM).

## Impact

Since nothing uses the IPv4 Configuration file at present, there should be no impact from this addition.

## Testing

To be provided after initial review.

PR #417, opened 2020-10-03
apps/fsutils/ipcfg:  Add support for IPv6

patacongo:
## Summary

The IP Configuration File access library at apps/fsutils/ipcfg was originally merged with IPv4 support only.  Only hooks were provided for IPv6.  This PR adds that missing IPv6 support.

Related PRs:  #407 #414 #415 #416 

## Impact

This could potentially effect apps/netutils/netinit which is the only client of apps/fsutils/ipcfg.

## Testing

The changes were verified using apps/examples/ipcfg with the base configuration stm32f4discovery:netnsh

Testing was fairly thorough.  I verified:

- IPv4
- IPv6
- IPv4 + IPv6
- All possible binary or human-readable text formats
- Assertions on and off

### DIFF
--- a/fsutils/ipcfg/ipcfg.h
+++ b/fsutils/ipcfg/ipcfg.h
@@ -1,0 +1,238 @@
+/****************************************************************************
+ * apps/fsutils/ipcfg/ipcfg.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __APPS_FSUTILS_IPCFG_IPCFG_H
+#define __APPS_FSUTILS_IPCFG_IPCFG_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdint.h>
+
+#include <netinet/in.h>
+
+#include "fsutils/ipcfg.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define MAX_LINESIZE  80
+#define MAX_IPv4PROTO IPv4PROTO_FALLBACK
+#define MAX_IPv6PROTO IPv6PROTO_FALLBACK
+
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+/* IP Configuration record header. */
+
+struct ipcfg_header_s
+{
+  uint8_t next;         /* Offset to the next IP configuration record */
+  sa_family_t type;     /* Either AF_INET or AF_INET6 */
+};
+
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: ipcfg_read_binary_ipv4
+ *
+ * Description:
+ *   Read IPv4 configuration from a binary IP Configuration file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   ipv4cfg - Location to read IPv4 configration to
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_IPCFG_BINARY) && defined(CONFIG_NET_IPv4)
+int ipcfg_read_binary_ipv4(FAR const char *path,
+                           FAR struct ipv4cfg_s *ipv4cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_read_binary_ipv6
+ *
+ * Description:
+ *   Read IPv6 configuration from a binary IP Configuration file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   ipv6cfg - Location to read IPv6 configration to
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_IPCFG_BINARY) && defined(CONFIG_NET_IPv6)
+int ipcfg_read_binary_ipv6(FAR const char *path,
+                           FAR struct ipv6cfg_s *ipv6cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_write_binary_ipv4
+ *
+ * Description:
+ *   Write the IPv4 configuration to a binary IP Configuration file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   ipv4cfg - The IPv4 configration to write
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_IPCFG_WRITABLE) && defined(CONFIG_IPCFG_BINARY) && \
+    defined(CONFIG_NET_IPv4)
+int ipcfg_write_binary_ipv4(FAR const char *path,
+                            FAR const struct ipv4cfg_s *ipv4cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_write_binary_ipv6
+ *
+ * Description:
+ *   Write the IPv6 configuration to a binary IP Configuration file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   ipv6cfg - The IPv6 configration to write
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_IPCFG_WRITABLE) && defined(CONFIG_IPCFG_BINARY) && \
+    defined(CONFIG_NET_IPv6)
+int ipcfg_write_binary_ipv6(FAR const char *path,
+                            FAR const struct ipv6cfg_s *ipv6cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_read_text_ipv4
+ *
+ * Description:
+ *   Read IPv4 configuration from a human-readable, text IP Configuration
+ *   file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   netdev  - Network device name string
+ *   ipv4cfg - Location to read IPv4 configration to
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_NET_IPv4
+int ipcfg_read_text_ipv4(FAR const char *path, FAR const char *netdev,
+                         FAR struct ipv4cfg_s *ipv4cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_read_text_ipv6
+ *
+ * Description:
+ *   Read IPv6 configuration from a human-readable, text IP Configuration
+ *   file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   netdev  - Network device name string
+ *   ipv6cfg - Location to read IPv6 configration to
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_NET_IPv6
+int ipcfg_read_text_ipv6(FAR const char *path, FAR const char *netdev,
+                         FAR struct ipv6cfg_s *ipv6cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_write_text_ipv4
+ *
+ * Description:
+ *   Write the IPv4 configuration to a human-readable, text IP Configuration
+ *   file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   netdev  - Network device name string
+ *   ipv4cfg - The IPv4 configration to write
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_IPCFG_WRITABLE) && defined(CONFIG_NET_IPv4)
+int ipcfg_write_text_ipv4(FAR const char *path, FAR const char *netdev,
+                          FAR const struct ipv4cfg_s *ipv4cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_write_text_ipv6
+ *
+ * Description:
+ *   Write the IPv6 configuration to a human-readable, text IP Configuration
+ *   file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   netdev  - Network device name string
+ *   ipv6cfg - The IPv6 configration to write
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_IPCFG_WRITABLE) && defined(CONFIG_NET_IPv6)
+int ipcfg_write_text_ipv6(FAR const char *path, FAR const char *netdev,
+                          FAR const struct ipv6cfg_s *ipv6cfg);
+#endif
+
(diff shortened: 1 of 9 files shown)


patacongo:
@xiaoxiang781216 @davids5 There is no impact to the release from this PR (because there is still no user of ipcfg in master) so this could be merged whenever you are comfortable.  I have tested pretty thoroughly and I am confident in the state of the changes.  I think David has become exasperated.  After this is merged, I will submit a PR to get the IPv4 and IPv6 configuration from the configuration file.  I will not implement David's full polling logic, only the logic that gets the IP addresses.  It would be nice if you two could resolve your differences and we could make forward progress on that.

fsutils/ipcfg/ipcfg_text.c
```diff
@@ -0,0 +1,810 @@
...
+  fclose(stream);
+  return ret;
+}
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_write_text_ipv6
+ *
+ * Description:
+ *   Write the IPv6 configuration to a binary IP Configuration file.
```
xiaoxiang781216, line 726: binary->text

fsutils/ipcfg/ipcfg_text.c
```diff
@@ -0,0 +1,810 @@
...
+  "fallback"   /* IPv4PROTO_FALLBACK */
+};
+#endif
+
+#if defined(CONFIG_IPCFG_WRITABLE) && defined(CONFIG_NET_IPv6)
+static const char *g_ipv6proto_name[] =
+{
+  "none",      /* IPv6PROTO_NONE */
+  "static",    /* IPv6PROTO_STATIC */
+  "dhcp",      /* IPv6PROTO_AUTOCONF */
```
xiaoxiang781216, line 58: dhcp->autoconf to match the read path

fsutils/ipcfg/ipcfg_binary.c
```diff
@@ -0,0 +1,567 @@
...
+  close(fd);
+  return ret;
+}
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_read_binary_ipv6
+ *
+ * Description:
+ *   Read IPv4 configuration from a binary IP Configuration file.
```
xiaoxiang781216, line 310: IPv4->IPv6

fsutils/ipcfg/ipcfg_binary.c
```diff
@@ -0,0 +1,567 @@
...
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   ipv4cfg - Location to read IPv4 configration to
```
xiaoxiang781216, line 314: ipv6cfg - Location to read IPv6 configration to

fsutils/ipcfg/ipcfg.h
```diff
@@ -0,0 +1,237 @@
...
+#if defined(CONFIG_IPCFG_WRITABLE) && defined(CONFIG_NET_IPv4)
+int ipcfg_write_text_ipv4(FAR const char *path, FAR const char *netdev,
+                          FAR const struct ipv4cfg_s *ipv4cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_write_text_ipv6
+ *
+ * Description:
+ *   Write the IPv6 configuration to a binary IP Configuration file.
```
xiaoxiang781216, line 219: binary->text

xiaoxiang781216:
> @xiaoxiang781216 @davids5 There is no impact to the release from this PR (because there is still no user of ipcfg in master) so this could be merged whenever you are comfortable. I have tested pretty thoroughly and I am confident in the state of the changes. I think David has become exasperated since he has closed PR #415 . After this is merged, I will submit a PR to get the IPv4 and IPv6 configuration from the configuration file. I will not implement David's full polling logic from PR #415 , only the logic that gets the IP addresses which I think is non-controversial. It would be nice if you two could resolve your differences and we could make forward progress on that.

Sure, let's go ahead. Actually, all my suggestions are very simple:
1. Make NETINIT_ESTABLISH_POLL_RATE/NETINIT_LOSS_POLL_RATE available in both interrupt/poll case
2. Make NETINIT_DHCP_FALLBACK available in both interrupt/poll case
3. Initialize g_netconf with the default if ipcfg_read return fail to avoid the code duplication

The first two only involve the minor change in Kconfig, the 3rd one just need reorder the initialition code.

patacongo:
> Actually, all my suggestions are very simple:
> 
>     1. Make NETINIT_ESTABLISH_POLL_RATE/NETINIT_LOSS_POLL_RATE available in both interrupt/poll case
> 
>     2. Make NETINIT_DHCP_FALLBACK available in both interrupt/poll case
> 
>     3. Initialize g_netconf with the default if ipcfg_read return fail to avoid the code duplication
> 
> 
> The first two only involve the minor change in Kconfig, the 3rd one just need reorder the initialition code.

Unfortunately, this will not happen.  David and closed #415 and you can read his comments here:  PX4/NuttX-apps#7

fsutils/ipcfg/ipcfg.h
```diff
@@ -0,0 +1,237 @@
...
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+/* IP Configuration record header. */
+
+struct ipcfg_header_s
+{
+  uint8_t next;         /* Offset to the next IP configuration record */
+  sa_family_t type;     /* Must be AF_INET */
```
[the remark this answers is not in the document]
patacongo: fixed

fsutils/ipcfg/ipcfg.h
```diff
@@ -0,0 +1,237 @@
...
+
+/****************************************************************************
+ * Name: ipcfg_read_binary_ipv6
+ *
+ * Description:
+ *   Read IPv4 configuration from a binary IP Configuration file.
+ *
+ * Input Parameters:
+ *   path    - The full path to the IP configuration file
+ *   ipv4cfg - Location to read IPv4 configration to
```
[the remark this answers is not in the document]
patacongo: fixed

fsutils/ipcfg/ipcfg.h
```diff
@@ -0,0 +1,237 @@
...
+#if defined(CONFIG_IPCFG_BINARY) && defined(CONFIG_NET_IPv4)
+int ipcfg_read_binary_ipv4(FAR const char *path,
+                           FAR struct ipv4cfg_s *ipv4cfg);
+#endif
+
+/****************************************************************************
+ * Name: ipcfg_read_binary_ipv6
+ *
+ * Description:
+ *   Read IPv4 configuration from a binary IP Configuration file.
```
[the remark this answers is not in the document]
patacongo: fixed

fsutils/ipcfg/ipcfg_text.c
patacongo: fixed

fsutils/ipcfg/ipcfg_binary.c
patacongo: fixed

fsutils/ipcfg/ipcfg_binary.c
patacongo: fixed

fsutils/ipcfg/ipcfg.h
patacongo: fixed

fsutils/ipcfg/ipcfg_text.c
patacongo: fixed

patacongo:
@xiaoxiang781216 I believe that all of you comments have been addressed.

xiaoxiang781216:
One more comment: can we replace all printf in fsutils/ipcfg with ferr? It will save a lot of code size.

patacongo:
> 
> 
> One more comment: can we replace all printf in fsutils/ipcfg with ferr? It will save a lot of code size.

Done

xiaoxiang781216:
@patacongo there are some nxstyle warnning need to resolve:
/home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/fsutils/ipcfg/ipcfg_text.c:393:78: error: Long line found
/home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/fsutils/ipcfg/ipcfg_text.c:416:82: error: Long line found
/home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/fsutils/ipcfg/ipcfg_text.c:543:78: error: Long line found
/home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/fsutils/ipcfg/ipcfg_text.c:566:82: error: Long line found
All looks fine, I will merge the change once it pass the precheck.

patacongo:
> 
> 
> @patacongo there are some nxstyle warnning need to resolve:
> /home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/fsutils/ipcfg/ipcfg_text.c:393:78: error: Long line found
> /home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/fsutils/ipcfg/ipcfg_text.c:416:82: error: Long line found
> /home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/fsutils/ipcfg/ipcfg_text.c:543:78: error: Long line found
> /home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/fsutils/ipcfg/ipcfg_text.c:566:82: error: Long line found
> All looks fine, I will merge the change once it pass the precheck.

Fixed

xiaoxiang781216 (approved):

patacongo:
@xiaoxiang781216 Thanks for the merge.

I have some concerns about byte ordering.

- apps/fsutils/ipcfg does not care about network vs. host byte ordering.  It saves and returns addresses in the same order that they were provided.  Perhaps this should be specified as network order?
- But it does call inet_addr() which expects the string representation in host order, but returns the binary in network order.
- It also calls inet_pton() and inet_ntop().  inet_pton() expects the string representation in host order.  The comments say that it returns the binary in network order, but I don't see that.  It looks like it returns the data in host order (but I might be wrong).
- inet_ntop() comments say that the input in network byte ordering.  That is true, but it seems to be hard coded to work only on a little endian machine of the IPv4 conversion.  The IPv6 conversion calls ntohs() for each 16-bit value.  I think that is okay.
- apps/examples/provides addresses in host order.  But in testing, the final addresses are printed in the correct host byte order.  Now I am thinking that is suspicious.
- David's netinit change assumes that the IPv4 values returned by ipcfg are in host order and uses HTONL() to convert the addresses to network order.

This is confusing.  I think the byte ordering needs to be clearly stated (network ordering), inet_pton() et al should be verified, and netinit.c should not call HTONL.  I will look at this more when I incorporate the ipcfg changes in netinit.  Perhaps there is no problem that a few comments cannot fix.

patacongo:
>     * apps/fsutils/ipcfg does not care about network vs. host byte ordering.  It saves and returns addresses in the same order that they were obtained.  Perhaps this should be specified as network order?

It already is specified as little endian.  In apps/fsutils/ipcfg.h:

    All of the following addresses are in network order.  The special value
    zero is used to indicate that the address is not available:

>     * inet_ntop() comments say that the input in network byte ordering.  That is true, but it seems to be hard coded to work only on a little endian machine for the IPv4 conversion.  The IPv6 conversion calls ntohs() for each 16-bit value.  I think that is okay.

No, the inet_ntop() IPv4 logic is correct, but very confusing.  I added a PR that adds comments (only) to clarify the endian-ness issues.  See PR apache/incubator-nuttx#1935

>     * David's netinit change assumes that the IPv4 values returned by ipcfg are in host order and uses HTONL() to convert the addresses to network order.

That would then be wrong.  It does not matter since that PR was not merged.

> This is confusing. I think the byte ordering needs to be clearly stated (network ordering), inet_pton() et al should be verified, and netinit.c should not call HTONL.

I am thinking everything is OK, but very confusing.  Endian-ness issues are always confusing.

xiaoxiang781216:
> @xiaoxiang781216 Thanks for the merge.
> 
> I have some concerns about byte ordering.
> 
> * apps/fsutils/ipcfg does not care about network vs. host byte ordering.  It saves and returns addresses in the same order that they were obtained.  Perhaps this should be specified as network order?

Could be, but it isn't so critifal except the user extract config file to other device. Otherwise, the real requirement is that the load/save use the same order(either little, big or native).

> * But it does call inet_addr() which expects the string representation in host order, but returns the binary in network order.

string don't have the order concept, only the binary bigger than one byte has.

> * It also calls inet_pton() and inet_ntop().  inet_pton() expects the string representation in host order.  The comments say that it returns the binary in network order, but I don't see that.  It looks like it returns the data in host order (but I might be wrong; the code all works okay as it).

The addres return in network order:
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/net/lib_inetpton.c#L145
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/net/lib_inetpton.c#L270

> * inet_ntop() comments say that the input in network byte ordering.  That is true, but it seems to be hard coded to work only on a little endian machine for the IPv4 conversion.  The IPv6 conversion calls ntohs() for each 16-bit value.  I think that is okay.

IPv4 work for big endian machine too:
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/net/lib_inetntop.c#L112
because the code access the address through char * from low to high.

> * apps/examples/provides addresses in network order.  In testing, the final addresses are printed in the correct host byte order.  So that seems to be fine.
> * David's netinit change assumes that the IPv4 values returned by ipcfg are in host order and uses HTONL() to convert the addresses to network order.
> 
> This is confusing. I think the byte ordering needs to be clearly stated (network ordering), inet_pton() et al should be verified, and netinit.c should not call HTONL. I will look at this more when I incorporate the ipcfg changes in netinit. Perhaps there is no problem that a few comments cannot fix.

Since these routines don't validate the range of input and the conversion happen inside the implementation, the result of load is always good regardless what's order the caller use. But it's always good to appoint the order(network), so the user don't surprise when he/she cat the saved file and it also make the config file truely independent from device.

davids5:
> > @xiaoxiang781216 @davids5 There is no impact to the release from this PR (because there is still no user of ipcfg in master) so this could be merged whenever you are comfortable. I have tested pretty thoroughly and I am confident in the state of the changes. I think David has become exasperated since he has closed PR #415 . After this is merged, I will submit a PR to get the IPv4 and IPv6 configuration from the configuration file. I will not implement David's full polling logic from PR #415 , only the logic that gets the IP addresses which I think is non-controversial. It would be nice if you two could resolve your differences and we could make forward progress on that.

@xiaoxiang781216 Per @patacongo request:

 I do not "reject other people's comments."  I do look for them to acknowledged the points and concern I have and respond to it.  This does not happen with you and I. (I truly believe it is miscommunication)  I take more responsibility in that failing as I can not speak Chinese.  I wish I had the capability to speak Chinese as well as you speak English so that I could ensure you are acknowledging my points and concerns and have a counter argument that nullifies them.  Our common languages is "C" this is why I ask you to do a PR.
 
> 
> Sure, let's go ahead. Actually, all my suggestions are very simple:
> 
> 1. Make NETINIT_ESTABLISH_POLL_RATE/NETINIT_LOSS_POLL_RATE available in both interrupt/poll case


Prior to my PR HW that could support network monitoring enabled the whole network monitor via ARCH_PHY_INTERRUPT. 
This was safe. The state (edges) of the network were provided  to the monitor ONLY on transitions.
On the UP edge a **DELAY** was used to the bring up networking. On a DOWN edge no delay was used to bring down the network it was the  checked at **CONFIG_NETINIT_RETRYMSEC** .  Once the network was up the PHY was ONLY queried ASYNCHRONOUSLY every 3600. 

When I asked if it supported polled mode. Greg responded with no. AND a concern that HW may not support link management while active:

image

My first test was to do poling in the F7 driver using a LAN8742AI-CZ-TR and simulate the interrupts. 

1) I PROVED that the specific combination of the STM32F7' MAC and LAN8742AI PHY do not disabled the MAC during management.

2) The polling logic in the driver would be a waste of resources to simulate the interrupt as it would be a duplicate of what the the client task could be modified to do polling with **only a semantic** change.. 

Since other SOC / PHY may have the limitation.  I did not want to cause some poor user to have that heartache. It would waste their time and reflect badly on NuttX,

My reason for not accepting your change is the your change can case **user specified** _periodic_ networking issues with HW that disable the MAC in management mode.  I can not imagine how hard that will be to trouble shoot.  I feel strongly about this, I would not expose a user to that. I would not want it done to me. 

Further more the word polled and interrupt to the 180% opposite. The time components in the monitor are DELAYS from edges, not POLLING rates. I introduced that notion by setting  a set of DELAYS  = to a set of POLLING RATES based on  CONFIG_ARCH_PHY_POLLED.

> 2. Make NETINIT_DHCP_FALLBACK available in both interrupt/poll case

This may work. But is will be only based on the CONFIG_NETINIT_RETRYMSEC, not the polling rates.

It needs to be tested on HW that can support it.

> 3. Initialize g_netconf with the default if ipcfg_read return fail to avoid the code duplication

I like the idea of saving space and reducing complexity, My attempt at it gave me concerns that the combination of the CONFIG_* permutations with the run time choices made it very hard to follow and test. You  may have better results with it. This is why I asked for a PR from you if you wanted the change in.

> The first two only involve the minor change in Kconfig, the 3rd one just need reorder the initialition code.

I did run out of time on this. I have products that needed all the functionality that I added. The concerns of  binary compatibility will be an issue based on the host/network ordering.

davids5:
@patacongo based on the last comment above, can you please use the extra 16 bit in the header as a version number?

patacongo:
> 
> 
> @patacongo based on the last comment above, can you please use the extra 16 bit in the header as a version number?

No, but you are welcome to submit a PR.

I will be integrating IPv6 support into netinit today.  This may conflict with your plans, but I hope not.

davids5:
> > @patacongo based on the last comment above, can you please use the extra 16 bit in the header as a version number?
> 
> No, but you are welcome to submit a PR.
> 
> I will be integrating IPv6 support into netinit today. This may conflict with your plans, but I hope not.

PR is here. https://github.com/apache/incubator-nuttx-apps/pull/419

PR #418, opened 2020-10-05
Incorporate IP configuration files into netinit.

patacongo:
## Summary

Uses logic from apps/fsutils/ipcfg to obtain the IP address configuration from a file, if available.  This only effects the behavior of netinit if CONFIG_FSUTILS_IPCFG is selected and if an IP configuration file is available on a file system. Otherwise, it reverts to the preceding behavior.

Also adds definitions to apps/include/fsutils/ipcfg.h to formalize a bit encoding of the prototype which was already in use, but not documented.

This PR also corrects a few problems with fsutils/ipcfg:

- Corrects some bad assertions (duplicated in PR #420, we can close that PR if this one is merged first.
- Corrects some inconsitent variable naming between apps/includes/fsutils/ipcfg.c and apps/fsutils/ipcfg/ipcfg_text.c

## Impact

Obviously, the apps/netutils/netinit logic is affected.  The use of the ipcfg IP configuration file is conditional so, if all was done right, there should be no impact.

## Testing

Everything tests out.  Due to complexities of different test setups, only a few configurations could be verified:

**Test strategy**

- Create an ASCII IP configuration file on an SD card using an editor.  The name of the IP configuration file must be ipcfg-eth0.
- Use stm32f4discovery:netnsh configuration.  This configuration already supports IPv4 networking, the netinit thread, and SD card.  It has this default IP configuration:

    CONFIG_NETINIT_IPADDR=0x0a000002
    CONFIG_NETINIT_DRIPADDR=0x0a000001
    CONFIG_NETINIT_NETMASK=0xffffff00
    
    CONFIG_NETINIT_IPv6ADDR_1=0xfc00
    CONFIG_NETINIT_IPv6ADDR_2=0x0000
    CONFIG_NETINIT_IPv6ADDR_3=0x0000
    CONFIG_NETINIT_IPv6ADDR_4=0x0000
    CONFIG_NETINIT_IPv6ADDR_5=0x0000
    CONFIG_NETINIT_IPv6ADDR_6=0x0000
    CONFIG_NETINIT_IPv6ADDR_7=0x0000
    CONFIG_NETINIT_IPv6ADDR_8=0x0002
    CONFIG_NETINIT_DRIPv6ADDR_1=0xfc00
    CONFIG_NETINIT_DRIPv6ADDR_2=0x0000
    CONFIG_NETINIT_DRIPv6ADDR_3=0x0000
    CONFIG_NETINIT_DRIPv6ADDR_4=0x0000
    CONFIG_NETINIT_DRIPv6ADDR_5=0x0000
    CONFIG_NETINIT_DRIPv6ADDR_6=0x0000
    CONFIG_NETINIT_DRIPv6ADDR_7=0x0000
    CONFIG_NETINIT_DRIPv6ADDR_8=0x0001
    CONFIG_NETINIT_IPv6NETMASK_1=0xffff
    CONFIG_NETINIT_IPv6NETMASK_2=0xffff
    CONFIG_NETINIT_IPv6NETMASK_3=0xffff
    CONFIG_NETINIT_IPv6NETMASK_4=0xffff
    CONFIG_NETINIT_IPv6NETMASK_5=0xffff
    CONFIG_NETINIT_IPv6NETMASK_6=0xffff
    CONFIG_NETINIT_IPv6NETMASK_7=0xffff
    CONFIG_NETINIT_IPv6NETMASK_8=0x0000

- Enable IPv6 and ICMPv6
- Enable IPCFG with Text file support

    CONFIG_IPCFG_WRITABLE=y
    \# CONFIG_IPCFG_BINARY is not set
    CONFIG_IPCFG_PATH="/ipcfg"

- Build and execute with SD card in place
- Mount the SD card prior to running NSH and netinit (requires a small code change to the board bring-up logic).
- Use the NSH ifconfig command to see if the network is properly configured.

```
diff --git a/boards/arm/stm32/stm32f4discovery/src/stm32_bringup.c b/boards/arm/stm32/stm32f4discovery/src/stm32_bringup.c
index fedd8ecec0..80cd76fa94 100644
--- a/boards/arm/stm32/stm32f4discovery/src/stm32_bringup.c
+++ b/boards/arm/stm32/stm32f4discovery/src/stm32_bringup.c
@@ -526,5 +526,12 @@ int stm32_bringup(void)
     }
 #endif /* CONFIG_LPWAN_SX127X */

+  ret = mount("/dev/mmcsd0", CONFIG_IPCFG_PATH, "vfat", 0, NULL);
+  if (ret < 0)
+    { 
+      int errcode = errno; 
+      _err("ERROR: Failed to mount " CONFIG_IPCFG_PATH "\n: %d", errcode);
+    } 
+
   return ret;
 }
```

Testing only with a text file is not limitation since the unit under test is netinit, not ipcfg.  Various text files can be created with an editor and many possible IP configurations can be verified.  For example,

    IPv4PROTO=static
    IPv4IPADDR=10.0.0.42
    IPv4NETMASK=255.255.0.0
    IPv4ROUTER=10.0.0.77
    IPv4DNS=10.0.0.32
    IPv6PROTO=static
    IPv6IPADDR=fc00::42
    IPv6NETMASK=ffff:ffff::
    IPv6ROUTER=fc00::77

**Test Result**

_Without_ SD card inserted, we get the configured default settings:

    eth0    Link encap:Ethernet HWaddr 00:e0:de:ad:be:ef at UP
            inet addr:10.0.0.2 DRaddr:10.0.0.1 Mask:255.255.255.0
            inet6 addr: fc00::2/112
            inet6 DRaddr: fc00::1/112

_With_ SD card inserted, we get the settings from the ipcfg-eth0 file:

    eth0    Link encap:Ethernet HWaddr 00:e0:de:ad:be:ef at UP
            inet addr:10.0.0.42 DRaddr:10.0.0.77 Mask:255.255.0.0
            inet6 addr: fc00::42/32
            inet6 DRaddr: fc00::77/32

### DIFF
--- a/netutils/netinit/netinit.c
+++ b/netutils/netinit/netinit.c
@@ -1,34 +1,21 @@
 /****************************************************************************
  * apps/netutils/netinit/netinit.c
  *
- *   Copyright (C) 2010-2012, 2014-2016, 2019 Gregory Nutt. All rights
- *     reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
  *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
+ *   http://www.apache.org/licenses/LICENSE-2.0
  *
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in the
- *    documentation and/or other materials provided with the distribution.
- * 3. Neither the name of the Institute nor the names of its contributors
- *    may be used to endorse or promote products derived from this software
- *    without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
- * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
- * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
- * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
- * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
- * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
- * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
- * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
- * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
- * SUCH DAMAGE.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+
  ****************************************************************************/
 
 /****************************************************************************
@@ -63,7 +50,7 @@
 #include <nuttx/net/mii.h>
 
 #include "netutils/netlib.h"
-#if defined(CONFIG_NETINIT_DHCPC) || defined(CONFIG_NETINIT_DNS)
+#if defined(CONFIG_NETUTILS_DHCPC) || defined(CONFIG_NETINIT_DNS)
 #  include "netutils/dhcpc.h"
 #endif
 
@@ -77,6 +64,10 @@
 
 #ifdef CONFIG_NETUTILS_NTPCLIENT
 #  include "netutils/ntpclient.h"
+#endif
+
+#if defined(CONFIG_FSUTILS_IPCFG)
+#  include "fsutils/ipcfg.h"
 #endif
 
 #include "netutils/netinit.h"
@@ -237,6 +228,10 @@
 static sem_t g_notify_sem;
 #endif
 
+#ifdef CONFIG_NETUTILS_DHCPC
+bool g_use_dhcpc;
+#endif
+
 #if defined(CONFIG_NET_IPv6) && !defined(CONFIG_NET_ICMPv6_AUTOCONF) && \
    !defined(CONFIG_NET_6LOWPAN)
 /* Host IPv6 address */
@@ -295,7 +290,8 @@ static const uint16_t g_ipv6_netmask[8] =
  *
  ****************************************************************************/
 
-#if defined(NETINIT_HAVE_NETDEV) && defined(CONFIG_NETINIT_NOMAC) && defined(HAVE_MAC)
+#if defined(NETINIT_HAVE_NETDEV) && defined(CONFIG_NETINIT_NOMAC) && \
+    defined(HAVE_MAC)
 static void netinit_set_macaddr(void)
 {
 #if defined(CONFIG_NET_ETHERNET)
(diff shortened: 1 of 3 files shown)


davids5:
@patacongo As soon as I can I will take a look.

netutils/netinit/netinit.c
```diff
@@ -344,66 +340,239 @@ static void netinit_set_macaddr(void)
...
+        {
+          /* We are not using DHCPC.  We need an IP address */
 
-  addr.s_addr = HTONL(CONFIG_NETINIT_DRIPADDR);
-  netlib_set_dripv4addr(NET_DEVNAME, &addr);
+#ifdef CONFIG_NETINIT_IPADDR
+          /* Check if we have a static IP address in the configuration file */
 
-  /* Setup the subnet mask */
+          if (ipv4cfg.proto != 0)
```
davids5, line 388: in the DHCP case the `else` limits the scope but would it it make sense to check the static bit here?

netutils/netinit/netinit.c
```diff
@@ -344,66 +340,239 @@ static void netinit_set_macaddr(void)
...
+          addr.s_addr = 0;
+          netlib_set_dripv4addr(NET_DEVNAME, &addr);
+          netlib_set_ipv4netmask(NET_DEVNAME, &addr);
+        }
+#endif
+      else
+        {
+          /* Otherwise, set up the configured default router address */
+
+          addr.s_addr = HTONL(CONFIG_NETINIT_DRIPADDR);
```
davids5, line 438: This case seems redundant to the one on line 411. But I am having a super hard time with all the #ifdef and 'else' and maintaining visual alignment. - that might just be my deficit. Which is fine. But I cannot help I wonder if it would be better from a readability stand point to assemble a state variable and have switch statement ?

davids5 (commented):
Over all it looks good. Per my inline comment, I am afraid I do not add much value here,

netutils/netinit/netinit.c
```diff
@@ -344,66 +340,239 @@ static void netinit_set_macaddr(void)
...
 {
-#ifdef CONFIG_NET_IPv4
   struct in_addr addr;
+#ifdef CONFIG_FSUTILS_IPCFG
+  struct ipv4cfg_s ipv4cfg;
+  int ret;
+#endif
 
-  /* Set up our host address */
+#ifdef CONFIG_FSUTILS_IPCFG
```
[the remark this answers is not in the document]
patacongo: Done

netutils/netinit/netinit.c
```diff
@@ -344,66 +340,239 @@ static void netinit_set_macaddr(void)
...
+  /* Attempt to obtain IPv4 address configuration from the IP configuration
+   * file.
+   */
 
-#ifndef CONFIG_NETINIT_DHCPC
-  addr.s_addr = HTONL(CONFIG_NETINIT_IPADDR);
+  ret = ipcfg_read(NET_DEVNAME, (FAR struct ipcfg_s *)&ipv4cfg, AF_INET);
+#ifdef CONFIG_NETUTILS_DHCPC
+  if (ret >= 0 && ipv4cfg.proto != IPv4PROTO_NONE)
```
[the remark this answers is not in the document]
patacongo: No, that is impossible in a modular, layered architecture.  User-space applications must be completely ignorant of the underlying media and hence must never make any assumptions about media properties, whether or not it is eraseable or what the erased value it.  Reading from an erased FLASH would fail for the base the binary file:  The address family of 0xffff would be invalid; it would not match AF_INET or AF_INET6  The hdr.next field would be 0xff so that might go off into never-never-land.  I suppose there should be a sanity check on the hdr.next field too.

I think your PR #419 addresses the versioning issues.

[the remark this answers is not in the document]
patacongo: Resolving this conversation because the question, while important, does not apply to this PR.

netutils/netinit/netinit.c
patacongo: The test` if (ipv4cfg.proto != 0)` is equivalent but a little stronger.  If the STATIC bit is not set, the addresses should all be zero, but not necessarily.

But I think you are right, better to test the static bit because if the user wants to force a zero address I suppose that should be permitted.

netutils/netinit/netinit.c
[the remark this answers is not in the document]
davids5: > he hdr.next field would be 0xff so that might go off

yes - the version, can catch this. May be it should  be checked first.

netutils/netinit/netinit.c
patacongo: Changes per your suggestion.

netutils/netinit/netinit.c
patacongo: It is, but sometimes a small code duplication can save a lot of complexity.  I don't know how to do the switch statement you are referring to.

The complexity is due to:

- CONFIG_NETUTILS_DHCPC meas that the DHCPC code has been built into the system and is available.

DHCPC may used under the following conditions if CONFIG_NETUTILS_DHCPC=y:

- CONFIG_FSUTILS_IPCFG is selected, a configuration file is read, and the boot protocol if DHCPC
- CONFIG_FSUTILS_IPCFG is NOT selected or a configuration file is no available and CONFIG_NETINIT_DHCPC is selected.

There are lots of ways that there can be mismatches and the complex logic tries to resolve the complexities.  For example, if the configuration file asks for DHCPC but CONFIG_NETUTILS_DHCPC is not selected.

A little redundancy is a good thing if it gets us out of some of the complexity.

You comment is noted, but I don't see anything actionable here.

patacongo:
I have finished verification of this change.  Looks good.

davids5:
@patacongo - thank you for this!

patacongo:
> @patacongo - thank you for this!

You are welcome.  I am of the belief that if you are going to bitch and moan about someone else's solution, then it is your responsibility to provide a better one.  My complaints about boardctl() cost me several days of effort!

PR #419, opened 2020-10-06
ipcfg:Version the binary structure

davids5:
## Summary

 This adds a version for the binary file structure. 

## Impact

Should be none as this is not used yet.

## Testing

inspection.

### DIFF
--- a/fsutils/ipcfg/ipcfg.h
+++ b/fsutils/ipcfg/ipcfg.h
@@ -37,9 +37,10 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-#define MAX_LINESIZE  80
-#define MAX_IPv4PROTO IPv4PROTO_FALLBACK
-#define MAX_IPv6PROTO IPv6PROTO_FALLBACK
+#define IPCFG_VERSION   1  /* Increment this if the data/structure changes */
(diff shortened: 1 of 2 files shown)


fsutils/ipcfg/ipcfg.h
```diff
@@ -49,7 +50,8 @@
 
 struct ipcfg_header_s
 {
-  uint8_t next;         /* Offset to the next IP configuration record */
+  uint8_t     next;     /* Offset to the next IP configuration record */
+  uint8_t     version;  /* For binary compatibility */
   sa_family_t type;     /* Either AF_INET or AF_INET6 */
 };
```
patacongo, line 56: This is really insufficient.  If the version number changes, then the size will also change.  This won't work in most cases.

patacongo: Nevermind.  I see that you ignore the data if the version number is bad.  That is okay if you don't try to read it.

patacongo (approved):
LGTM (ignore my other comments)

fsutils/ipcfg/ipcfg.h
davids5: Yes. The ides is that returning to factory default are better than using corrupted data.

PR #422, opened 2020-10-09
wireless: gs2200m: Fix recvfrom_reques() in gs2200m_main.c

masayuki2009:
## Summary

- During Wi-Fi stress test, I noticed ASSERT happens in gs2200m_main.c
- Actually, req->max_buflen was 0 but the case was not handled correctly
- This commit fixes this issue

## Impact

- Affects read/recvfrom with len=0 cases

## Testing

- Tested with spresense:wifi_smp
- Run nxplayer for audio streaming
- Run ps/free/ifconfig commands via telnet
- Run smp command via telnet


masayuki2009:
@jerpelea

PR #425, opened 2020-10-13
Remove all fclose with stdin, stdout and stderr

xiaoxiang781216:
## Summary
since it is wrong to close the builtin stream and specially note
https://pubs.opengroup.org/onlinepubs/9699919799/functions/fclose.html:

Since after the call to fclose() any use of stream results in
undefined behavior, fclose() should not be used on stdin, stdout,
or stderr except immediately before process termination (see XBD
Process Termination), so as to avoid triggering undefined behavior
in other standard interfaces that rely on these streams. If there
are any atexit() handlers registered by the application, such a
call to fclose() should not occur until the last handler is
finishing. Once fclose() has been used to close stdin, stdout, or
stderr, there is no standard way to reopen any of these streams.

and it is also unnecessary because the stream always get flushed.

## Impact
Remove the undefined behaviour

## Testing
reported here:
https://github.com/apache/incubator-nuttx/commit/b0797263ca2817cf536be8b843ac5b5aac5d4d9c
https://github.com/apache/incubator-nuttx/pull/1611

### DIFF
--- a/graphics/nxwm/src/cnxterm.cxx
+++ b/graphics/nxwm/src/cnxterm.cxx
@@ -504,22 +504,10 @@ int CNxTerm::nxterm(int argc, char *argv[])
   std::fflush(stderr);
 
 #ifdef CONFIG_NXTERM_NXKBDIN
-  std::fclose(stdin);
-#endif
-  std::fclose(stdout);
-  std::fclose(stderr);
-
-#ifdef CONFIG_NXTERM_NXKBDIN
   std::dup2(fd, 0);
 #endif
   std::dup2(fd, 1);
   std::dup2(fd, 2);
-
-#ifdef CONFIG_NXTERM_NXKBDIN
-  std::fdopen(0, "r");
-#endif
-  std::fdopen(1, "w");
-  std::fdopen(2, "w");
 
   // And we can close our original driver file descriptor
 
--- a/examples/nxterm/nxterm_main.c
+++ b/examples/nxterm/nxterm_main.c
@@ -391,9 +391,6 @@ int main(int argc, FAR char *argv[])
   fflush(stdout);
   fflush(stderr);
 
-  fclose(stdout);
-  fclose(stderr);
-
   dup2(fd, 1);
   dup2(fd, 2);
 
--- a/graphics/twm4nx/apps/cnxterm.cxx
+++ b/graphics/twm4nx/apps/cnxterm.cxx
@@ -418,22 +418,10 @@ int CNxTerm::nxterm(int argc, char *argv[])
   std::fflush(stderr);
 
 #ifdef CONFIG_NXTERM_NXKBDIN
-  std::fclose(stdin);
-#endif
-  std::fclose(stdout);
-  std::fclose(stderr);
-
-#ifdef CONFIG_NXTERM_NXKBDIN
   std::dup2(fd, 0);
 #endif
   std::dup2(fd, 1);
   std::dup2(fd, 2);
-
-#ifdef CONFIG_NXTERM_NXKBDIN
-  std::fdopen(0, "r");
-#endif
-  std::fdopen(1, "w");
-  std::fdopen(2, "w");
 
   // And we can close our original driver file descriptor
 
--- a/nshlib/nsh_altconsole.c
+++ b/nshlib/nsh_altconsole.c
@@ -78,9 +78,9 @@ static int nsh_clone_console(FAR struct console_stdio_s *pstate)
       return -ENODEV;
     }
 
-  /* Close stderr: we only close stderr if we opened the alternative one */
+  /* Flush stderr: we only flush stderr if we opened the alternative one */
 
-  fclose(stderr);
+  fflush(stderr);
 
   /* Associate the new opened file descriptor to stderr */
 
@@ -101,9 +101,9 @@ static int nsh_clone_console(FAR struct console_stdio_s *pstate)
       return -ENODEV;
     }
 
-  /* Close stdout: we only close stdout if we opened the alternative one */
+  /* Flush stdout: we only flush stdout if we opened the alternative one */
 
-  fclose(stdout);
+  fflush(stdout);
 
   /* Associate the new opened file descriptor to stdout */
 
@@ -122,7 +122,6 @@ static int nsh_clone_console(FAR struct console_stdio_s *pstate)
   pstate->cn_errstream = fdopen(pstate->cn_errfd, "a");
   if (!pstate->cn_errstream)
     {
-      close(pstate->cn_errfd);
       free(pstate);
       return -EIO;
     }
@@ -133,7 +132,6 @@ static int nsh_clone_console(FAR struct console_stdio_s *pstate)
   pstate->cn_outstream = fdopen(pstate->cn_outfd, "a");
   if (!pstate->cn_outstream)
     {
-      close(pstate->cn_outfd);
       free(pstate);
       return -EIO;
     }
@@ -202,10 +200,6 @@ static int nsh_wait_inputdev(FAR struct console_stdio_s *pstate,
     }
   while (fd < 0);
 
-  /* Close stdin: we only closed stdin if we opened the alternative one */
-
-  fclose(stdin);
-
   /* Okay.. we have successfully opened the input device.  Did
    * we just re-open fd 0?
    */
@@ -225,7 +219,6 @@ static int nsh_wait_inputdev(FAR struct console_stdio_s *pstate,
       pstate->cn_constream = fdopen(pstate->cn_confd, "r+");
       if (!pstate->cn_constream)
         {
-          close(pstate->cn_confd);
           free(pstate);
           return -EIO;
         }
--- a/nshlib/nsh_usbconsole.c
+++ b/nshlib/nsh_usbconsole.c
@@ -77,31 +77,16 @@
 
 static void nsh_configstdio(int fd)
 {
-  /* Make sure the stdin, stdout, and stderr are closed */
+  /* Make sure the stdout, and stderr are flushed */
 
-  fclose(stdin);
-  fclose(stdout);
-  fclose(stderr);
+  fflush(stdout);
+  fflush(stderr);
 
   /* Dup the fd to create standard fd 0-2 */
 
   dup2(fd, 0);
   dup2(fd, 1);
   dup2(fd, 2);
-
-  /* fdopen to get the stdin, stdout and stderr streams. The following logic
-   * depends on the fact that the library layer will allocate FILEs in order.
-   * And since we closed stdin, stdout, and stderr above, that is what we
-   * should get.
-   *
-   * fd = 0 is stdin  (read-only)
-   * fd = 1 is stdout (write-only, append)
-   * fd = 2 is stderr (write-only, append)
-   */
-
-  fdopen(0, "r");
-  fdopen(1, "a");
-  fdopen(2, "a");
 }
 
 /****************************************************************************
--- a/nshlib/nsh_usbconsole.c
+++ b/nshlib/nsh_usbconsole.c
@@ -75,7 +75,7 @@
  *
  ****************************************************************************/
 
-static void nsh_configstdio(int fd)
+static void nsh_configstdio(int fd, FAR struct console_stdio_s *pstate)
 {
   /* Make sure the stdout, and stderr are flushed */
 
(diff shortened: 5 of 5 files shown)


xiaoxiang781216:
@juniskane could you try whether this patch can fix your issue?

jerpelea:
LGTM

juniskane:
This fixes the USB console for me, but note that this is still a workaround. The following strictly conforming C program causes a crash on NuttX (tested in Flat mode build):
```
#include <stdio.h>

int main(void)
{
    /* Note: order significant if want to see a crash on NuttX. */
    fclose(stdin);
    fclose(stdout);
    fclose(stderr);
    /* ... open some file here, read/write to it, close it. ... just don't call printf() */
    return 0;
}
```

This kind of usage of fclose() is unlikely to appear in real programs, though.

Also note that fflush(stdin) is undefined behavior in C, but we can allow it since we are the implementor.

xiaoxiang781216:
> This fixes the USB console for me, but note that this is still a workaround.

No, it isn't a workaround, the old way is an undefined behaviour from https://pubs.opengroup.org/onlinepubs/9699919799/functions/fclose.html and we should avoid. 

> The following strictly conforming C program causes a crash on NuttX (tested in Flat mode build):
> 
> ```
> #include <stdio.h>
> 
> int main(void)
> {
>     /* Note: order significant if want to see a crash on NuttX. */
>     fclose(stdin);
>     fclose(stdout);
>     fclose(stderr);
>     /* ... can open some file here, read/write to it, close it. ... just don't call printf() */
>     return 0;
> }
> ```
> 
> This kind of usage of fclose() is unlikely to appear in real programs, though.

To improve the compability, I will provide a patch to ignore fclose to stdin/stdout/stderr anyway.

> 
> Also note that fflush(stdin) is undefined behavior in C, but we can allow it since we are the implementor.

Why is fflush(stdin) undefined? can you point me the related document? It's reasonable to expect that fflush(stdin) flush out the unread data in the buffer.

juniskane:
Calling fclose() on any of the standard text streams is not undefined behavior, just any use of the stream afterwards (including second fclose() and things like perror() inside library functions, but not the implicit closing of files in process exit). Wording in ISO/IEC 9899:1999 7.19.3.4 hints that standard streams can be closed like any other: _"The value of a pointer to a FILE object is indeterminate after the associated file is closed (including the standard text streams)."_

OpenGroup/POSIX bug tracker has interesting discussion about this:
https://www.austingroupbugs.net/view.php?id=555
That shows how the recently added wording you quoted from OpenGroup spec came into being.

The code that was in USB console and daemon() is typical for many UNIX daemons so existing applications might contain a similar fclose/dup2/fdopen sequence.

Gnulib has close_stdout() explicitly for closing stdout and stderr streams, I guess someone finds that feature useful:
https://git.savannah.gnu.org/cgit/gnulib.git/tree/lib/closeout.c

Just making fclose() a no-op if applied to any of the three standard streams feels like a suitable fix for this. There might be difficulties with setbuf() and setvbuf(), however, if application sets a buffer in stack and relied on fclose() disassociating it from the stream before exiting code block were buffer was defined. This is probably very rare scenario.

juniskane:
fflush(stdin) is in comp.lang.c FAQ:
http://c-faq.com/stdio/stdinflush.html

Chapter and verse is ISO/IEC 9899:1999 section 7.19.5.2 The fflush function, clause 2:
_"If stream points to an output stream or an update stream in which the most recent
operation was not input, the fflush function causes any unwritten data for that stream
to be delivered to the host environment to be written to the file; otherwise, the behavior is
undefined."_

Some systems indeed define fflush(stdin) to flush out the unread data in the buffer, it is just not portable. We can keep doing that in NuttX, if we wish.

juniskane:
> There might be difficulties with setbuf() and setvbuf(), however, if application sets a buffer in stack and relied on fclose() disassociating it from the stream before exiting code block were buffer was defined. This is probably very rare scenario.

Actually that case is likely fine, as any operation on stream after fclose() is UB anyway so it does not matter if the IO buffer vanishes silently. Nothing in nuttx libc touches the buffer without user calling a function first. Same applies to re-use of the buffer for some new stream by libc user. But automatically allocated buffer might leak memory (until task exit) if fclose on stdin/stdout/stderr does not free it.

xiaoxiang781216:
stdin/stdout/stderr are still freed when the task exit, only fclose skip these special handlers.

graphics/nxwm/src/cnxterm.cxx
```diff
@@ -500,27 +500,18 @@ int CNxTerm::nxterm(int argc, char *argv[])
   // console).  (2) Don't bother trying to put debug instrumentation in the
   // following because it will end up in the NxTerm window.
 
-  std::fflush(stdout);
-  std::fflush(stderr);
-
 #ifdef CONFIG_NXTERM_NXKBDIN
-  std::fclose(stdin);
+  std::fflush(stdin);
```
patacongo, line 504: Using fflush with stdin is not appropriate.  fflush has no effect on input streams.  See https://pubs.opengroup.org/onlinepubs/007904975/functions/fflush.html

It would be better to remove this and all other places where fflush is used with stdin.

xiaoxiang781216: Done.

masayuki2009:
>This fixes the USB console for me, but note that this is still a workaround. 

@juniskane @xiaoxiang781216 
Hmm, as long as I tried this PR after rebasing to master, 
it still does not work with stm32f4discovery:usbnsh.

xiaoxiang781216:
@masayuki2009 I don't have the device with USB support:(, @juniskane could you help retest the change?

juniskane:
nshlib/nsh_altconsole.c calls non-existent flush(), typo for fflush().

Other than this, change works for me. Tested with proprietary board. I'm using slightly older apps and backported the change though.

xiaoxiang781216:
> nshlib/nsh_altconsole.c calls non-existent flush(), typo for fflush().
> 
> Other than this, change works for me. Tested with proprietary board. I'm using slightly older apps and backported the change though.

My fault, @masayuki2009 could you test my new change? I am wondering why the PR pass the build without error or warning.

masayuki2009:
> > nshlib/nsh_altconsole.c calls non-existent flush(), typo for fflush().
> > Other than this, change works for me. Tested with proprietary board. I'm using slightly older apps and backported the change though.
> 
> My fault, @masayuki2009 could you test my new change? I am wondering why the PR pass the build without error or warning.

@xiaoxiang781216 

At least, stm32f4discovery:usbnsh does not use nsh_altconsole.c.
Anyway, could you fix the file and rebase the fclose branch to the latest master and push the commit again?

Also, I think we need to apply https://github.com/apache/incubator-nuttx/pull/1997 as well.
If so, please rebase the  xiaoxiang781216:fclose3 to the latest master.

xiaoxiang781216:
Done, both rebase to the latest master.

masayuki2009:
> Done, both rebase to the latest master.

@xiaoxiang781216 
I think you pushed the commit to https://github.com/xiaoxiang781216/incubator-nuttx-apps/tree/fclose
However, this PR is from https://github.com/apache/incubator-nuttx-apps/tree/fclose

masayuki2009:
> > Done, both rebase to the latest master.
> 
> @xiaoxiang781216
> I think you pushed the commit to https://github.com/xiaoxiang781216/incubator-nuttx-apps/tree/fclose
> However, this PR is from https://github.com/apache/incubator-nuttx-apps/tree/fclose

@xiaoxiang781216 

I tried 
https://github.com/xiaoxiang781216/incubator-nuttx-apps/tree/fclose
and
https://github.com/xiaoxiang781216/incubator-nuttx/tree/fclose3
with stm32f4discovery:usbnsh

Then start terminal program on Ubuntu 18.04
$ picocom -b 115200 /dev/ttyACM0

Howerver, lib_fwrite() always failes.

```
gdb) where                                                                                                                                                                                     
#0  lib_fwrite (ptr=ptr@entry=0x8011590 <g_nshgreeting>, count=29, stream=stream@entry=0x200012fc) at stdio/lib_libfwrite.c:81                                                                  
#1  0x0800f12a in fputs (s=0x8011590 <g_nshgreeting> "\nNuttShell (NSH) NuttX-8.2.0\n", stream=0x200012fc) at stdio/lib_fputs.c:171                                                             
#2  0x08005b40 in nsh_session (pstate=0x10000bd0, login=<optimized out>, argc=1, argv=0x10000b88) at nsh_session.c:106                                                                          
#3  0x080058ce in nsh_consolemain (argc=argc@entry=1, argv=argv@entry=0x10000b88) at nsh_usbconsole.c:334                                                                                       
#4  0x080057b8 in nsh_main (argc=1, argv=0x10000b88) at nsh_main.c:168                                                                                                                          
#5  0x08003252 in nxtask_startup (entrypt=0x800578d <nsh_main>, entrypt@entry=0x20000188 <inited>, argc=1, argv=0x10000b88) at sched/task_startup.c:165                                         
#6  0x08001cdc in nxtask_start () at task/task_start.c:144                                                                                                                                      
#7  0x00000000 in ?? () 


ssize_t lib_fwrite(FAR const void *ptr, size_t count, FAR FILE *stream)                                                                                                                         
#ifndef CONFIG_STDIO_DISABLE_BUFFERING                                                                                                                                                          
{                                                                                                                                                                                               
  FAR const unsigned char *start = ptr;                                                                                                                                                         
  FAR const unsigned char *src   = ptr;                                                                                                                                                         
  ssize_t ret = ERROR;                                                                                                                                                                          
  unsigned char *dest;                                                                                                                                                                          
                                                                                                                                                                                                
  /* Make sure that writing to this stream is allowed */                                                                                                                                        
                                                                                                                                                                                                
  if (stream == NULL)                                                                                                                                                                           
    {                                                                                                                                                                                           
      set_errno(EBADF);                                                                                                                                                                         
      return ret;                                                                                                                                                                               
    }                                                                                                                                                                                           
                                                                                                                                                                                                
  /* Check if write access is permitted */                                                                                                                                                      
                                                                                                                                                                                                
  if ((stream->fs_oflags & O_WROK) == 0)                                                                                                                                                        
    {                                                                                                                                                                                           
=>    set_errno(EBADF);                                                                                                                                                                         
      goto errout;                                                                                                                                                                              
    }                         
```

xiaoxiang781216:
Very strange, what's value in other FILE's field?

davids5:
@xiaoxiang781216 would you please set this a `draft` until it is sorted out?. You can test on the board I left with you, it has a usb.

masayuki2009:
> Very strange, what's value in other FILE's field?

@xiaoxiang781216 

(gdb) p *stream                                                                                                                                                                             
$1 = {fs_next = 0x7500706d, fs_fd = 8, fs_sem = {semcount = 0}, fs_holder = -32768, fs_counts = 32, fs_bufstart = 0x80000008 "", fs_bufend = 0x0, fs_bufpos = 0x10000010 "\240\v", fs_bufread = 0x0, fs_oflags = 0, fs_flags = 0 '\000', fs_nungotten = 0 '\000', fs_ungotten = "\000"}

xiaoxiang781216:
Campare the data, it look like that stream is freed.

masayuki2009:
> Campare the data, it look like that stream was freed which is what this patch want to fix.

@xiaoxiang781216 
Hmmm, I added breakpoints to fclose() and close() and ran the NuttX on stm32d4discovery again.
However, close(fd=3)  was only hit which is not stdout.

masayuki2009:
@xiaoxiang781216 
Finally, I found bugs in nsh_usbconsole.c and fixed locally.

I think we need to set pstate->cn_outstream and pstete->cnerrstream with fdopen()
which is actually done in nsh_atlconsle.c
I'm still not sure why stm32f4discovery:usbnsh works with nuttx-9.1.0-RC2.

May I add the coomit to this PR?

xiaoxiang781216:
Sure, no problem. Thanks for help me finding the touch bug.

masayuki2009:
>I'm still not sure why stm32f4discovery:usbnsh works with nuttx-9.1.0-RC2.

@xiaoxiang781216 
I think the previous code in nsh_usbconsole.c worked luckly, because file_struct was statically allocated.

xiaoxiang781216:
@masayuki2009 could you merge this PR and https://github.com/apache/incubator-nuttx/pull/1997?

masayuki2009:
> @masayuki2009 could you merge this PR and [apache/incubator-nuttx#1997](https://github.com/apache/incubator-nuttx/pull/1997)?

Both PRs are merged.

xiaoxiang781216:
@btashton this patch need backport to 10.0.0 too.

PR #429, opened 2020-10-16
Fix note_syscall_leave_s to avoid unaligned access

YuuichiNakamura:
## Summary
App code support for https://github.com/apache/incubator-nuttx/pull/2005
"note" and "trace" commands are fixed.

## Impact
See https://github.com/apache/incubator-nuttx/pull/2005

## Testing
See https://github.com/apache/incubator-nuttx/pull/2005

### DIFF
--- a/system/sched_note/note_main.c
+++ b/system/sched_note/note_main.c
@@ -457,6 +457,7 @@ static void dump_notes(size_t nread)
             {
               FAR struct note_spinlock_s *note_spinlock =
                 (FAR struct note_spinlock_s *)note;
+              FAR void *spinlock;
 
               if (note->nc_length != sizeof(struct note_spinlock_s))
                 {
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):
LGTM. let's ignore the false 'Mixed case identifier' alarm

system/sched_note/note_main.c
```diff
@@ -457,6 +457,7 @@ static void dump_notes(size_t nread)
             {
               FAR struct note_spinlock_s *note_spinlock =
                 (FAR struct note_spinlock_s *)note;
+              FAR void *spinlock;
```
xiaoxiang781216, line 460: BTW, @YuuichiNakamura could you update note_main.c to support syscall/irq you add recently?

YuuichiNakamura: @xiaoxiang781216 These are already supported by note_main.c line 592-671.

PR #434, opened 2020-10-20
testing/sensortest: add sensor driver test

Donny9:
## Summary

This testing for sensor driver is so simple, it use ioctl to control sensor and get some event to verify driver is OK.

It and [https://github.com/apache/incubator-nuttx/pull/2039](url) are closely linked.

## Impact

We can use tool to find issue about sensor driver.

## Testing

daily test.

### DIFF
--- a/testing/sensortest/Kconfig
+++ b/testing/sensortest/Kconfig
@@ -1,0 +1,29 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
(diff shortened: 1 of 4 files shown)


xiaoxiang781216 (approved):

xiaoxiang781216:
Let's merge this patch since the kernel side is already merged.

btashton:
Thanks guys. I know I asked a lot of questions, but only because I think this is the way forward.

PR #445, opened 2020-10-28
Android Debug Bridge daemon

spiriou:
## Summary

add Android Debug Bridge daemon for NuttX.
https://developer.android.com/studio/command-line/adb

## Impact

None. Disabled by default.

## Testing

Validated on "sim" board (TCP) and stm32 (both TCP and USB).

There are still some issues that will be address soon:
* Authentication check disabled due to missing RSA crypto lib in NuttX
* Some instabilities and memory leak using "adb shell"
* adb forward/reverse (socket) disabled for now as it needs a refactoring

defconfig for test purposes can be found at https://github.com/spiriou/incubator-nuttx/commits/adb_examples

### DIFF
--- a/system/adb/Kconfig
+++ b/system/adb/Kconfig
@@ -1,0 +1,158 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+menuconfig SYSTEM_ADBD
+	tristate "ADB daemon application"
+	default n
+	---help---
+		Enable support for adb daemon.
+
+if SYSTEM_ADBD
+
+config ADBD_PROGNAME
+	string "Program name"
+	default "adbd"
+	---help---
+		This is the name of the program that will be used.
+
+config ADBD_STACKSIZE
+	int "Stack size"
+	default DEFAULT_TASK_STACKSIZE
+	---help---
+		The size of stack allocated for the adb daemon task.
+
+config ADBD_PRIORITY
+	int "Task priority"
+	default 100
+	---help---
+		The priority of the adb daemon task.
+
+config ADBD_AUTHENTICATION
+	bool "Authentication support"
+	default n
+	---help---
+		Enable authentication for adb daemon.
+
+if ADBD_AUTHENTICATION
+
+config ADBD_AUTH_PUBKEY
+	bool "Public key authentication"
+	default n
+	---help---
+		Enable hook to accept new public keys.
+
+config ADBD_TOKEN_SIZE
+	int "Authentication token size"
+	default 20
+
+endif # ADBD_AUTHENTICATION
+
+if ! BOARDCTL_UNIQUEID
+config ADBD_DEVICE_ID
+	string "Default adb device id"
+	default ""
+endif # BOARDCTL_UNIQUEID
+
+config ADBD_PRODUCT_NAME
+	string "Default adb product name"
+	default "adb dev"
+
+config ADBD_PRODUCT_MODEL
+	string "Default adb product model"
+	default "adb board"
+
+config ADBD_PRODUCT_DEVICE
+	string "Default adb product device"
+	default "NuttX device"
+
+config ADBD_FEATURES
+	string "Default adb server features list"
+	default "cmd"
+
+config ADBD_PAYLOAD_SIZE
+	int "Normal ADB frame size"
+	default 64
+	---help---
+		Normal frame size in bytes.
+
+config ADBD_CNXN_PAYLOAD_SIZE
+	int "Connection frame size"
+	default 1024
+	---help---
+		Connection frame is bigger than others.
+		Can be between 128 to 256 bytes in most of the cases.
+		If authentication is enabled, frame size must bigger
+		to receive public key from host (around 1024 bytes).
+
+config ADBD_FRAME_MAX
+	int "Frame pool size"
+	default 1
+	---help---
+		ADB frame pool size.
+
+config ADBD_TCP_SERVER
+	bool "Network socket transport support"
+	depends on NET_TCP
+	default n
+	---help---
+		Run adb daemon on network socket.
+
+config ADBD_TCP_SERVER_PORT
+	int "Network socket transport port"
+	depends on ADBD_TCP_SERVER
+	default 5555
+	---help---
+		Port used by adb daemon socket server
+
+config ADBD_USB_SERVER
+	bool "USB transport support"
+	depends on USBADB
+	default n
+	---help---
+		Run adb daemon on USB bus
+
+config ADBD_LOGCAT_SERVICE
+	bool "ADB logcat support"
+	depends on RAMLOG_SYSLOG
+	default n
+	---help---
+		Enable "adb logcat" feature.
+
+config ADBD_FILE_SERVICE
+	bool "ADB file sync support"
+	default n
+	---help---
+		Enable "adb ls/push/pull" feature.
+
+ config ADBD_SHELL_SERVICE
+	bool "ADB shell support"
+	depends on NSH_CONSOLE
+	default n
+	---help---
+		Enable "adb shell" feature.
+
+config ADBD_FILE_SYMLINK
+	bool "File service symlink support"
+	default n
+	depends on PSEUDOFS_SOFTLINKS
+	---help---
+		Enable fs symlink support.
+
+config ADBD_BOARD_INIT
+	bool "Board initialization"
+	depends on LIB_BOARDCTL
+	default n
+	---help---
+		Setup board before running adb daemon.
+
+config ADBD_NET_INIT
+    bool "Network initialization"
+    default n
+    depends on NET
+    select NETUTILS_NETINIT
+    ---help---
+        This option enables/disables all network initialization in ADB server.
+
+endif # SYSTEM_ADBD
--- a/system/adb/Makefile
+++ b/system/adb/Makefile
@@ -1,0 +1,92 @@
+############################################################################
+# system/adb/Makefile
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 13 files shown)


system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+menuconfig ADB_SERVER
+	bool "adb server command"
+	default n
```
xiaoxiang781216, line 8: need depends or select kernel adb driver, or adbd can work without usb but tcp?

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+menuconfig ADB_SERVER
+	bool "adb server command"
```
xiaoxiang781216, line 7: change to tristate, so we can move adbd to file system

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
...
+if ADB_SERVER
+
+config ADBD_STACKSIZE
+	int "adb stack size"
+	default 3072
+	---help---
+		The size of stack allocated for the adb task.
+
+config ADBD_PRIORITY
+	int "adb task priority"
```
xiaoxiang781216, line 21: change all adb to adbd?

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+menuconfig ADB_SERVER
```
xiaoxiang781216, line 6: change all ADB_ to ADBD_

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
...
+
+config ADBD_FILE_SERVICE
+	bool "adb file sync support"
+	default n
+	---help---
+		Enable "adb ls/push/pull" feature.
+
+ config ADBD_SHELL_SERVICE
+	bool "adb shell support"
+	default n
```
xiaoxiang781216, line 121: should we depend on NSH

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
...
+config ADBD_TCP_SERVER_PORT
+	int "Network socket transport port"
+	depends on ADBD_TCP_SERVER
+	default 5555
+	---help---
+		Port used by adb socket server
+
+config ADBD_USB_SERVER
+	bool "USB transport support"
+	default n
```
xiaoxiang781216, line 102: depend on adb usb driver

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
...
+
+config ADBD_FRAME_MAX
+	int "Frame pool size"
+	default 1
+	---help---
+		ADB frame pool size.
+
+config ADBD_TCP_SERVER
+	bool "Network socket transport support"
+	default n
```
xiaoxiang781216, line 89: depend on TCP

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
...
+
+config ADBD_PRODUCT_MODEL
+	string "default adb product model"
+	default "adb board"
+
+config ADBD_PRODUCT_DEVICE
+	string "default adb product device"
+	default "NuttX device"
+
+config ADBD_FEATURES
```
xiaoxiang781216, line 62: should we auto generate the feature string by check wether user select  ADBD_FILE_SERVICE, ADBD_SHELL_SERVICE...?

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
...
+	---help---
+		Enable hook to accept new public keys.
+
+config ADBD_TOKEN_SIZE
+	int "authentication token size"
+	default 20
+
+endif # ADBD_AUTHENTICATION
+
+config ADBD_DEVICE_ID
```
xiaoxiang781216, line 46: how about we get device id from BOARDIOC_UNIQUEID

system/adb/Makefile
```diff
@@ -0,0 +1,98 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+ADB_DIR := $(APPDIR)/system/adb
+CONFIG_ADBD_URL ?= "https://github.com/spiriou/microADB.git"
```
xiaoxiang781216, line 24: can we directly import microADB into apps git?

system/adb/Makefile
```diff
@@ -0,0 +1,98 @@
...
+	git remote add origin "$(CONFIG_ADBD_URL)" && \
+	git fetch origin $(CONFIG_ADBD_VERSION) --depth=1 && \
+	git reset --hard FETCH_HEAD
+
+# adb server app
+
+CONFIG_ADBD_PRIORITY ?= SCHED_PRIORITY_DEFAULT
+CONFIG_ADBD_STACKSIZE ?= 2048
+CONFIG_ADBD_PROGNAME ?= adbd$(EXEEXT)
+PROGNAME = $(CONFIG_ADBD_PROGNAME)
```
xiaoxiang781216, line 44: dup with line 46

system/adb/Makefile
```diff
@@ -0,0 +1,98 @@
...
+	$(Q) cd "$@" && git init && \
+	git remote add origin "$(CONFIG_ADBD_URL)" && \
+	git fetch origin $(CONFIG_ADBD_VERSION) --depth=1 && \
+	git reset --hard FETCH_HEAD
+
+# adb server app
+
+CONFIG_ADBD_PRIORITY ?= SCHED_PRIORITY_DEFAULT
+CONFIG_ADBD_STACKSIZE ?= 2048
+CONFIG_ADBD_PROGNAME ?= adbd$(EXEEXT)
```
xiaoxiang781216, line 43: line 41 to line 43 isn't needed

xiaoxiang781216: should we add CONFIG_ADBD_PROGNAME to Kconfig like other apps

system/adb/shell_exec_builtin.c
```diff
@@ -0,0 +1,200 @@
...
+  posix_spawnattr_t attr;
+  posix_spawn_file_actions_t file_actions;
+  struct sched_param param;
+  pid_t pid;
+  int index;
+  int ret;
+
+  /* Verify that an application with this name exists */
+
+  index = builtin_isavail(appname);
```
xiaoxiang781216, line 50: should we support to launch program on file system?

system/adb/shell_exec_builtin.c
```diff
@@ -0,0 +1,200 @@
...
+
+#include "adb.h"
+#include "builtin/builtin.h"
+#include "shell_pipe.h"
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+int shell_exec_builtin(FAR const char *appname, FAR char * const *argv,
```
xiaoxiang781216, line 37: why not call system or nsh api instead?

spiriou: This is a first iteration regarding NSH integration in "adb shell" command. The technical solution is to leverage pipe for both stdout and stdin streams. The issue is I need a custom function to configure fd 0 and fd 1 for NSH process. That's what shell_exec_builtin() does (based on ./builtin/exec_builtin.c but with more posix_spawn_file_actions_addopen()).

Next step could be to try integrating nshlib directly in adbd thread. It may simplify this logic and save lot of RAM.

system/adb/shell_exec_builtin.c
spiriou: Indeed this can be simplified. This function is the same as the one in ./builtin/exec_builtin.c but with more "posix_spawn_file_actions_addopen()" to setup correctly stdin and stdout streams.

system/adb/Makefile
spiriou: Initially this project was not meant for NuttX so I would prefer to keep the platform independent code it in a separate repository for now. Is this an issue ?

system/adb/Kconfig
spiriou: The idea is once the server is enabled we can select at least one of USB or TCP transports. Or both if needed.
ADBD_USB must indeed depend on kernel adb driver.

system/adb/Kconfig
```diff
@@ -0,0 +1,147 @@
...
+	bool "adb server command"
+	default n
+	---help---
+		Enable support for adb server.
+
+if ADB_SERVER
+
+config ADBD_STACKSIZE
+	int "adb stack size"
+	default 3072
```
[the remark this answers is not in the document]
spiriou: Sure. For now the stack requirement is way to high (3k). In the initial adbd version the stack usage was around 1500 bytes but I'll optimize it later.

system/adb/Makefile
xiaoxiang781216: No problem.

system/adb/shell_exec_builtin.c
xiaoxiang781216: Yes, you can reference how telnetd integrate with nshlib here:
https://github.com/apache/incubator-nuttx-apps/blob/master/nshlib/nsh_telnetd.c
https://github.com/apache/incubator-nuttx-apps/blob/master/netutils/telnetd/telnetd_daemon.c

xiaoxiang781216: One more simple example to integrate nsh:
https://github.com/apache/incubator-nuttx-apps/tree/master/examples/pty_test

spiriou: @xiaoxiang781216 thank you for your help. The "adb shell" implementation makes more sense now with dup2().

acassis: LGTM? @xiaoxiang781216 @spiriou should I?

spiriou: The shell implementation could be based on ptmx like in https://github.com/apache/incubator-nuttx-apps/tree/master/examples/pty_test instead of pipes but this can be improved in another patch

xiaoxiang781216: @spiriou is this patch ready for merge?

spiriou: @xiaoxiang781216 it's not perfect but this can be merged as it's working fine. I'll improve it later.

PR #450, opened 2020-10-29
LVGL without framebuffer

w8jcik:
## Summary

LVGL example is using frambuffer device `/dev/fb0`.

Such framebuffer takes width x height x bpp of memory which is a lot.

But it is possible to use LVGL with just a procedure that draws.

This PR changes LVGL example to work without `/dev/fb0`, with low memory footprint and it works for SSD1289, but it is not ready for merge. Not handling other color depths for example, I can add it.

I am opening this draft PR, because I have a question. I consider this change an obvious thing to do, so I would expect somebody doing it before. Am I missing something? Is it possible to configure `/dev/fb0` interface to not use that much memory? I know that `/dev/fb0` can be configured to talk to displays which offer only simple drawing routines, but allocation of width x height x bpp cannot be avoided.

Should I finish adding this here, or should such functionality be added to `/dev/fb0`. Not sure if it would make sense. On one hand it would be a framebuffer without any buffer. On the other hand, most display chips have framebuffer of it's own. Sending rows of pixels is just filling their internal framebuffer. So `/dev/fb0` could represent that framebuffer and offer uniform interface.

On the picture `hymini-stm32v:nsh2` with `64 kB` of RAM, running LVGL calibration example.

IMG_20201029_231750

## Impact

Low memory requirement to use LVGL.

## Testing

This is draft pull request.

### DIFF
--- a/examples/lvgldemo/lvgldemo.c
+++ b/examples/lvgldemo/lvgldemo.c
@@ -45,6 +45,9 @@
 #include <stdio.h>
 #include <pthread.h>
 #include <time.h>
+#ifndef CONFIG_VIDEO_FB
+#include <nuttx/lcd/lcd.h>  // lcd_dev_s
+#endif
 
 #include <lvgl/lvgl.h>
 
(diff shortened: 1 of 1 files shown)


v01d:
Wow, incredible coincidence that I was just adding a PR for lvgl.

Regarding this change: this is not the way to do it since you are calling OS interfaces directly. In fact, I was also about to make a PR for a LCD character driver I wrote sometime ago and never upstreamed. It is exactly for this reason: it offers userspace interface to deal with LCDs. It also avoids the need for the framebuffer -> LCD driver which is quite inefficient in memory and communication.

v01d:
I have opened the relevant PR on main nuttx repo: https://github.com/apache/incubator-nuttx/pull/2166
I'm closing this as it is not the proper way to deal with this. You can try extending the lvgldemo app with
this driver (once it is merged), which would be more appropriate for LCD devices.

PR #451, opened 2020-10-29
lvgl: use malloc/free for memory handling and "tickless" tick interface

v01d:
## Summary

This makes LVGL use malloc/free to handle its memory (instead of
their own memory handling) and supports a tick interface by which
LVGL can ask NuttX the elapsed time, instead of having to periodically
"tick" LVGL internal time in a thread.

## Impact

Improves integration with LVGL

## Testing

Sim with X11 framebuffer


btashton (approved):
LGTM.   I'll test the new config is OS after this is merged.

PR #461, opened 2020-11-04
nxplayer & nxrecorder: make main thread configurable, set to default

GUIDINGLI:
## Summary

nxplayer & nxrecorder: make main thread configurable, set to default

## Impact

This should be commit with:
https://github.com/apache/incubator-nuttx/pull/2124


xiaoxiang781216 (approved):
LGTM.

PR #466, opened 2020-11-07
examples/watchdog/watchdog_main.c: Elapsed var  is now being updated …

saramonteiro:
## Summary

This PR aims to update the elapsed time variable using the current time instead of simply incrementing it with the delay.  It also discounts the "System timer tick period" to not cause delays besides the `pingdelay`

## Impact

It allows a more realistic test. 

## Testing

The example was run successfully.


xiaoxiang781216 (approved):
LGTM.

PR #467, opened 2020-11-08
libuv/adb: add gitignore files

spiriou:
## Summary

Add missing gitignore files for libuv and adb builds


btashton:
I am going to just merge this, the CI is not going to catch anything and it will be further verified in the OS ci with the new configuration for adb.

PR #468, opened 2020-11-08
netutils/webserver: fix build error

spiriou:


btashton:
Thanks for fixing this and all style issues along with it.

PR #472, opened 2020-11-09
Fix random printf format warnings and type mismatches

yamt:
## Summary
see https://github.com/apache/incubator-nuttx/pull/2222

### DIFF
--- a/canutils/candump/candump.c
+++ b/canutils/candump/candump.c
@@ -45,6 +45,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdint.h>
+#include <inttypes.h>
 #include <unistd.h>
 #include <string.h>
 #include <signal.h>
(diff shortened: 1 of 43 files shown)


include/graphics/nxwidgets/itextbox.hxx
```diff
@@ -208,7 +208,7 @@ namespace NXWidgets
      * @return position The cursor position.
      */
 
-    virtual const int getCursorPosition(void) const = 0;
+    virtual const int32_t getCursorPosition(void) const = 0;
```
xiaoxiang781216, line 211: should we change https://github.com/apache/incubator-nuttx-apps/blob/master/include/graphics/nxwidgets/cscrollingtextbox.hxx#L329 to int instead?
moveCursorToPosition at https://github.com/apache/incubator-nuttx-apps/blob/master/include/graphics/nxwidgets/cscrollingtextbox.hxx#L320 also need to change.

yamt: i dunno. i'm not familiar with this piece of code.
is there any reason to prefer int over int32_t here?

xiaoxiang781216: two reason:

1. Other function(e.g. moveCursorToPosition) use int
2. Only CScrollingTextBox use int32_t, all other place use int

yamt: ok, i will fix

yamt: done

yamt:
the following nxstyle errors are expected.
```
    examples/nrf24l01_term/nrf24l01_term.c:125:33: error: Mixed case identifier found
    examples/nrf24l01_term/nrf24l01_term.c:129:15: error: Mixed case identifier found
```

also, canutils has a lot of nxstyle errors because it doesn't seem to follow nxstyle at all.

xiaoxiang781216:
Ok, let's skip canutils nxstyle warning now.

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM, @yamt I will merge this PR now. If there are more nxstyle warnings, please create a new PR to address them.

PR #483, opened 2020-11-17
Fix test failed when some apis are NULL

Donny9:
## Summary
Fix test failed when some apis are NULL.
## Impact
Fix bug for some special driver.
## Testing
daily test


xiaoxiang781216 (approved):
LGTM.

PR #489, opened 2020-11-20
webclient: Distinguish request size and buffer size in body callback

yamt:
## Summary
This fixes the case when webclient_set_static_body is used
for data larger than webclient_context::buflen.

## Impact
Note: as of writing this, webclient_set_static_body is the
only user of body_callback in NuttX apps tree.

## Testing
tested with my local app.


xiaoxiang781216 (approved):
LGTM.

PR #507, opened 2020-11-27
irtest: fix  __attributes__((section))  issue on linux and mac

Donny9:
## Summary
Using global arrays instead of __attributes__((section)) organization because  __attribute__((section)) syntax is different on linux and mac

## Impact
Fix compile break on mac
## Testing
Daily test


xiaoxiang781216 (approved):
LGTM.

PR #511, opened 2020-11-29
examples/pwm: fix help message

diegoherranz:
## Summary
- Fix bug for multiple PWM outputs (CONFIG_PWM_MULTICHAN). Wrong define
was in use and only the information about the first channel was shown.
- Fix formatting (a few line breaks and spaces missing).

## Impact
Only the help message modified. It was very confusing before because only one channel was shown, but it shouldn't impact anything else.

## Testing
Built and tested.

Thanks!


Ouss4 (approved):
LGTM.

PR #514, opened 2020-12-03
examples/touchscreen/tc_main.c: Fix a syslog format

yamt:
## Summary
See https://github.com/apache/incubator-nuttx/pull/2347


Ouss4:
@yamt there is also line 144.  Would you like to fix it too?

yamt:
@Ouss4 
yes, sooner or later.
i'm visiting all configs tested on the ci. i guess i will fix it when i visit olimex-lpc1766stk:hidmouse.

Ouss4:
All right, thanks. So let's get this out of the way.

Ouss4 (approved):

PR #519, opened 2020-12-12
Remove unneeded semicolons and parentheses from Python files

jbampton:


Ouss4 (approved):
LGTM.

C programmers writing Python, happens to me all the time. :-)

PR #521, opened 2020-12-14
nshlib: Fix nsh_usbconsole.c

masayuki2009:
## Summary

- stdin/stdout/stderr are now preallocated in libc and fs_fd
  in file_struct are also initialized to -1
- So we need to call fdopen() for stdin again as we did before.
- Also, cn_outstream and cn_errstream are not needed to be set.
- See apps/nshlib/nsh_console.h as well

## Impact

- nsh_usbconsole.c only

## Testing

- Tested with stm32f4discovery:usbnsh

### DIFF
--- a/nshlib/nsh_usbconsole.c
+++ b/nshlib/nsh_usbconsole.c
@@ -75,7 +75,7 @@
  *
  ****************************************************************************/
 
-static void nsh_configstdio(int fd, FAR struct console_stdio_s *pstate)
+static void nsh_configstdio(int fd)
 {
   /* Make sure the stdout, and stderr are flushed */
 
@@ -88,15 +88,16 @@ static void nsh_configstdio(int fd, FAR struct console_stdio_s *pstate)
   dup2(fd, 1);
   dup2(fd, 2);
 
-  /* Setup the stdout */
+  /* fdopen to get the stdin, stdout and stderr streams.
+   *
+   * fd = 0 is stdin  (read-only)
+   * fd = 1 is stdout (write-only, append)
+   * fd = 2 is stderr (write-only, append)
+   */
 
-  pstate->cn_outfd     = 1;
-  pstate->cn_outstream = fdopen(1, "a");
-
-  /* Setup the stderr */
-
-  pstate->cn_errfd     = 2;
-  pstate->cn_errstream = fdopen(2, "a");
+  fdopen(0, "r");
+  fdopen(1, "a");
+  fdopen(2, "a");
 }
 
 /****************************************************************************
@@ -107,7 +108,7 @@ static void nsh_configstdio(int fd, FAR struct console_stdio_s *pstate)
  *
  ****************************************************************************/
 
-static int nsh_nullstdio(FAR struct console_stdio_s *pstate)
+static int nsh_nullstdio(void)
 {
   int fd;
 
@@ -118,7 +119,7 @@ static int nsh_nullstdio(FAR struct console_stdio_s *pstate)
     {
       /* Configure standard I/O to use /dev/null */
 
-      nsh_configstdio(fd, pstate);
+      nsh_configstdio(fd);
 
       /* We can close the original file descriptor now (unless it was one of
        * 0-2)
@@ -224,7 +225,7 @@ restart:
 
   /* Configure standard I/O */
 
-  nsh_configstdio(fd, pstate);
+  nsh_configstdio(fd);
 
   /* We can close the original file descriptor (unless it was one of 0-2) */
 
@@ -306,7 +307,7 @@ int nsh_consolemain(int argc, FAR char *argv[])
   /* Configure to use /dev/null if we do not have a valid console. */
 
 #ifndef CONFIG_DEV_CONSOLE
-  nsh_nullstdio(pstate);
+  nsh_nullstdio();
 #endif
 
   /* Execute the one-time start-up script (output may go to /dev/null) */
@@ -347,7 +348,7 @@ int nsh_consolemain(int argc, FAR char *argv[])
        * valid console device.
        */
 
-      nsh_nullstdio(pstate);
+      nsh_nullstdio();
     }
 }
 


masayuki2009:
See https://github.com/apache/incubator-nuttx/pull/2526 as well

btashton:
@masayuki2009  This lines up with the change I was making, but I am unclear now why we needed to add  `FAR struct console_stdio_s *pstate` to the function definition in the first place.

masayuki2009:
> @masayuki2009 This lines up with the change I was making, but I am unclear now why we needed to add `FAR struct console_stdio_s *pstate` to the function definition in the first place.

OK, I will remove it later.

xiaoxiang781216:
Final patch to fix: https://github.com/apache/incubator-nuttx/issues/2203

btashton:
I'm confused because this seems to back out the change that came in from @masayuki2009 comment here https://github.com/apache/incubator-nuttx-apps/pull/425#issuecomment-718398878

I just want to make sure we are not missing something.

xiaoxiang781216 (approved):

jerpelea (approved):

xiaoxiang781216:
@michallenc could you try this patch with your new config(https://github.com/apache/incubator-nuttx/pull/2526)?

michallenc:
> @michallenc could you try this patch with your new config([apache/incubator-nuttx#2526](https://github.com/apache/incubator-nuttx/pull/2526))?

@xiaoxiang781216 Sure. This seems to be working fine.

```
NuttShell (NSH) NuttX-9.1.1
nsh> 
nsh> 
nsh> 
nsh> ls dev
/dev:
 console
 null
 ttyACM0
 ttyS0
nsh> ?
help usage:  help [-v] [<cmd>]

  .         cd        echo      hexdump   mh        rm        time      xd        
  [         cp        exec      kill      mount     rmdir     true      
  ?         cmp       exit      ls        mv        set       uname     
  basename  dirname   false     mb        mw        sleep     umount    
  break     dd        free      mkdir     ps        source    unset     
  cat       df        help      mkrd      pwd       test      usleep    

Builtin Apps:
  sh   nsh  
nsh> 
```

xiaoxiang781216:
Thanks @michallenc I will merge both patch now.

xiaoxiang781216 (approved):

nshlib/nsh_usbconsole.c
```diff
@@ -88,15 +88,16 @@ static void nsh_configstdio(int fd, FAR struct console_stdio_s *pstate)
...
+   * fd = 0 is stdin  (read-only)
+   * fd = 1 is stdout (write-only, append)
+   * fd = 2 is stderr (write-only, append)
+   */
 
-  pstate->cn_errfd     = 2;
-  pstate->cn_errstream = fdopen(2, "a");
+  fdopen(0, "r");
+  fdopen(1, "a");
+  fdopen(2, "a");
```
patacongo, line 100: This is generally not needed.  The fact that fdopen() corrects the problem is an artifact of the fact that CONFIG_DEV_CONSOLE is not set in the configuration.

See sched/group/goupr_setupidlefiles.c:

     98 #ifdef CONFIG_DEV_CONSOLE
     99   fd = nx_open("/dev/console", O_RDWR);
    100   if (fd == 0)
    101     {
    102       /* Successfully opened /dev/console as stdin (fd == 0) */
    103
    104       fs_dupfd2(0, 1);
    105       fs_dupfd2(0, 2);
    106     }
    125 #endif

I believe that it is the logic in sched/group/goupr_setupidlefiles.c.  Since stdin, stdout, and stderr must always be available, it is wrong to permit CONFIG_DEV_CONSOLE to not be set.

A better solution would be to use /dev/null for tstdin, stdout, and stderr if CONFIG_DEV_CONSOLE  is not set.

xiaoxiang781216: but dev/null or dev/zero may disable too.

patacongo:
What about Telnet.  If networking were enabled, would Telnet not also fail in exactly the same way?

This solution cannot be implemented in a single application like NSH.  It _must_ be implemented as a system-wide solution that fixes all possible applications that can be started with no fixed console device.

nshlib/nsh_usbconsole.c
patacongo: > but dev/null or dev/zero may disable too.

Yes, it would have to e changed so that it cannot be disabled.  Not a big deal.

xiaoxiang781216: But, I am curious that line 87-89 already duplicate fd to 0-2, why we still need call fdopen at line 98-100 again?

patacongo: > But, I am curious that line 87-89 already duplicate fd to 0-2, why we still need call fdopen at line 98-100 again?

The file descriptors are duplicated, but the the streams are still not opened due to the failure of fs_fdopen() when the IDLE thread was created.  Duplicating the file descriptor does not create the stream.

xiaoxiang781216: > > But, I am curious that line 87-89 already duplicate fd to 0-2, why we still need call fdopen at line 98-100 again?
> 
> The file descriptors are duplicated, but the the streams are still not opened due to the failure of fs_fdopen() when the IDLE thread was created. Duplicating the file descriptor does not create the stream.
> 

Ok, I understand the situation now.

> This should not be done in the the OS (not the application) OR it could be done in _all_ applications and the non-standard requirement should be well documented.

Maybe the right thing we should do:

1. Always enable DEV_ZERO or DEV_NULL
2. Or skip check 0-2 valid in fs_fdopen

> The latter is basically the way things worked before your change that broke this.

But with the patch: https://github.com/apache/incubator-nuttx/pull/2263
I think the new behaviour is same as before now. If CONFIG_DEV_CONSOLE isn't enabled, fdopen is always required regardless whether my patch apply.

masayuki2009:
FYI.

https://github.com/apache/incubator-nuttx-apps/pull/523

PR #525, opened 2020-12-15
examples/ftpc: proper dual-stack support, also fix getopt re-entrancy issues

juniskane:
## Summary
Allow -4 and -6 command line options to select IPv4 or IPv6 on dual-stack devices, instead of always defaulting to IPv6
Uses `getaddrinfo()` so can do host name lookups, option -n to disable (also works without CONFIG_LIBC_NETDB for IP addr only, like before)
`getopt()` always called in a loop so no need for non-portable `optind = -1` construction
Added missing `htons()` for port number

## Impact
Command line syntax slightly changed, syntax `host:port` changed to have `port` as separate argument (easier for numeric IPv6)

### DIFF
--- a/examples/ftpc/ftpc_cmds.c
+++ b/examples/ftpc/ftpc_cmds.c
@@ -46,18 +46,6 @@
 #include "ftpc.h"
 
 /****************************************************************************
- * Public Types
- ****************************************************************************/
-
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):
LGTM.

PR #531, opened 2020-12-18
nimble: fix wrong git ref name to use

v01d:
## Summary

Last PR #528 missed a change, where the branch name to use is no longer "nuttx". For now "master" it is used until next stable release of nimBLE is available, where a tag name will be used instead.

## Impact

Allow building nimBLE

## Testing

sim:nimble (to be provided)


xiaoxiang781216 (approved):
LGTM. let's merge it now, it's impossible to break since no any config enable nimBLE yet.

PR #532, opened 2020-12-21
lvgl porting and lvgl demo decoupling

FASTSHIFT:
## Summary
lvgl porting and lvgl demo decoupling.
## Impact
Use public lvgl porting.
## Testing
daily test

### DIFF
--- a/examples/lvgldemo/Makefile
+++ b/examples/lvgldemo/Makefile
@@ -53,8 +53,6 @@ MODULE = $(CONFIG_EXAMPLES_LVGLDEMO)
 
 # LittleVGL demo Example
 
-CSRCS += fbdev.c tp.c tp_cal.c
-
 # static common assets used in mutiple example
 
 LV_EXAMPLE_ASSETS = $(wildcard ./lv_examples/assets/*.c)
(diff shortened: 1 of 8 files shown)


xiaoxiang781216:
@FASTSHIFT maybe it's better to split the PR into two.

FASTSHIFT:
> @FASTSHIFT maybe it's better to split the PR into two.

okay

v01d:
I'm not entirely convinced of this change. I see some that some of this code is part of `lv_drivers` repo, although not sure if there are no NuttX specific changes there (that would be the reason to ship them ourselves). On the other hand, while I can see benefit of having some generic code to access framebuffer, each project may require different way to initialize and access framebuffer so this generic function would not be appropriate. 
Moreover, it exposes global functions with no LVGL specific prefix so I don't think it is appropriate to use them outside of LVGL demo.

Can you explain what is the expected benefit of this change? Do you intend to reuse this on your own application?

v01d:
In other words, these functions are mostly designed to be used from the demos so I don't think it is entirely right to just expose them at the library level (when they are not really LVGL functions with an lv_ prefix).

FASTSHIFT:
@v01d I understand what you mean. I'll close it.

PR #533, opened 2020-12-21
interpreters/wamr: Support latest snapshot for main branch

no1wudi:
## Summary
Now supports specifying the version number of wamr as main.
## Impact
None
## Testing
Tested on stm32f429i-disco

### DIFF
--- a/interpreters/wamr/Makefile
+++ b/interpreters/wamr/Makefile
@@ -22,7 +22,17 @@ include $(APPDIR)/Make.defs
 
 WAMR_VERSION  = $(patsubst "%",%,$(strip $(CONFIG_INTERPRETERS_WAMR_VERSION)))
 WAMR_UNPACK   = wamr
-WAMR_TARBALL  = WAMR-$(WAMR_VERSION).tar.gz
+
+# Tarball of main branch and tagged release have little difference
+# in name after unpack and URL
+
+ifeq ($(WAMR_VERSION), main)
+WAMR_VERSION_STUB  = 
+else
+WAMR_VERSION_STUB  = WAMR-
+endif
+
+WAMR_TARBALL  = $(WAMR_VERSION_STUB)$(WAMR_VERSION).zip
 WAMR_URL_BASE = https://github.com/bytecodealliance/wasm-micro-runtime/archive/
 WAMR_URL      = $(WAMR_URL_BASE)/$(WAMR_TARBALL)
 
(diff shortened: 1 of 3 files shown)


interpreters/wamr/Makefile
```diff
@@ -32,22 +42,25 @@ VPATH += $(WAMR_UNPACK)/product-mini/platforms/nuttx
...
-	$(Q) tar xzvf $(WAMR_TARBALL)
-	$(Q) mv wasm-micro-runtime-WAMR-$(WAMR_VERSION) $(WAMR_UNPACK)
+	$(Q) unzip $(WAMR_TARBALL)
+	$(Q) mv wasm-micro-runtime-$(WAMR_VERSION_STUB)$(WAMR_VERSION) $(WAMR_UNPACK)
 
 context:: $(WAMR_UNPACK)
 
+clean::
+	$(call DELDIR, $(WAMR_UNPACK))
+
```
Ouss4, line 63: What's made during `context` should be removed during `clean_context` or `distclean`.  
If we remove it during `clean` we would always have to re-download it if we want a clean build.

interpreters/wamr/Makefile
```diff
@@ -32,22 +42,25 @@ VPATH += $(WAMR_UNPACK)/product-mini/platforms/nuttx
...
-PROGNAME  = iwamr
+PROGNAME  = iwasm
 PRIORITY  = $(CONFIG_INTERPRETERS_WAMR_PRIORITY)
 STACKSIZE = $(CONFIG_INTERPRETERS_WAMR_STACKSIZE)
 MODULE    = $(CONFIG_INTERPRETERS_WAMR)
 
 $(WAMR_TARBALL):
 	$(Q) echo "Downloading $(WAMR_TARBALL)"
-	$(Q) wget $(WAMR_URL)
+	$(Q) curl -O -L $(WAMR_URL)
```
xiaoxiang781216, line 52: why switch to curl?

interpreters/wamr/Makefile
xiaoxiang781216: Yes, clean should be removed.

no1wudi: **WAMR_UNPACK** is the directory of unpacked context, but source tarball still here.
If WAMR_VERSION changed in menuconfig, the correct source can be unpacked after `make clean`.

interpreters/wamr/Makefile
no1wudi: No functional difference, just because curl does not need to be installed on some systems (like Mac)

interpreters/wamr/Makefile
no1wudi: OK, that requires `make distclean` after Kconfig option changed

interpreters/wamr/Makefile
xiaoxiang781216: But other location use wget too. it's better to make a new patch to replace all wget to curl if the change can simplify the prerequest.

no1wudi: Done, I'll send another patch to do that

xiaoxiang781216 (approved):
LGTM.

Ouss4 (approved):

xiaoxiang781216 (approved):

PR #534, opened 2020-12-21
Replace all wget with curl

no1wudi:
## Summary
wget is missing from some system (like macOS and Windows native), it's better to use curl to simplify build environment.
## Impact
None
## Testing
Test build with stm32f429i-disco on Mac


canutils/libcanard/Makefile
```diff
@@ -55,7 +55,7 @@ CSRCS = $(LIBCANARD_SRCDIR)$(DELIM)canard.c $(LIBCANARD_DRVDIR)$(DELIM)canard_nu
 
 $(LIBCANARD_PACKNAME):
 	@echo "Downloading: $@"
-	$(Q) $(WGET) -O $@ $(LIBCANARD_URL)$(DELIM)$(LIBCANARD_VERSION)$(PACKEXT)
```
xiaoxiang781216, line 58: let's remove all WGET variable too(e.g. line 38).

xiaoxiang781216:
@no1wudi please do the same replacement in nuttx too.

xiaoxiang781216 (commented):
LGTM.

jerpelea:
please add commit description (the same as the PR message should be ok )

jerpelea (approved):

v01d:
Any other uses of wget anywhere else on NuttX? If not, you could look into removing any reference to wget from documentation.

patacongo:
curl should be added to the required tools in the top-level README.md file.  That provides the "getting started" info that people need to know.

I am not sure if there are other tools lists under Documentation/ because I am no long familiar with that.  It is worth checking.

patacongo:
How about in the windows native environment?  Would the user need to install something like:

    https://curl.se/windows/
    https://develop.zendesk.com/hc/en-us/articles/360001068567-Installing-and-using-cURL

The Windows native build has other problems more significant that this.  It is probably sufficient to know that some solution exists.  Cygwin or MSYS are more typical environments and curl is available under Cygwin (don't know about MSYS).

no1wudi:
> How about in the windows native environment? Would the user need to install something like:
> 
> ```
> https://curl.se/windows/
> https://develop.zendesk.com/hc/en-us/articles/360001068567-Installing-and-using-cURL
> ```
> 
> The Windows native build has other problems more significant that this. It is probably sufficient to know that some solution exists. Cygwin or MSYS are more typical environments and curl is available under Cygwin (don't know about MSYS).

Now curl is installed on Windows native by default (tested on my Windows 20H2).

no1wudi:
> please add commit description (the same as the PR message should be ok )

commit message updated

xiaoxiang781216:
curl is much better than wget from compatibility perspective:
image
let's merge it.

PR #536, opened 2020-12-22
netinit: Make the default SSID and passphrase empty

yamt:
## Summary
* It doesn't make much sense to have the default values for
  these highly environment-dependent settings.

* netutils/netinit/netinit_associate.c calls wpa_driver_wext_associate
  if the SSID is not empty. Depending on the environment, it can take
  long to fail. It slows down the boot of some configurations
  considerably. eg. esp32-devkitc:wapi

## Testing
esp32-devkitc:wapi (with some local patches)


xiaoxiang781216 (approved):
LGTM.

PR #538, opened 2020-12-22
system: nxlooper: add nxlooper for loopback test

Dan-guanghua:
which record and play raw data using audio subsystem
to verify audio record and playback function

Signed-off-by: danguanghua <danguanghua@xiaomi.com>

Change-Id: I2c768b8e8234a31f9622cb9955f785329cf402a5
Signed-off-by: danguanghua <danguanghua@xiaomi.com>

### DIFF
--- a/include/system/nxlooper.h
+++ b/include/system/nxlooper.h
@@ -1,0 +1,254 @@
+/****************************************************************************
+ * apps/include/system/nxlooper.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __APPS_INCLUDE_SYSTEM_NXLOOPER_H
+#define __APPS_INCLUDE_SYSTEM_NXLOOPER_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Type Declarations
+ ****************************************************************************/
+
+/* This structure describes the internal state of the NxLooper */
+
+struct nxlooper_s
+{
+  int         loopstate;                   /* Current looper test state */
+  int         recorddev_fd;                /* File descriptor of active
+                                            * record device */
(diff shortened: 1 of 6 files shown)


patacongo:
Should this go under apps/testing instead of apps/system?

xiaoxiang781216:
> Should this go under apps/testing instead of apps/system?

Other two audio tools put to apps/system:
https://github.com/apache/incubator-nuttx-apps/tree/master/system/nxplayer
https://github.com/apache/incubator-nuttx-apps/tree/master/system/nxrecorder
The new place make nxlooper hard to find by people who is already familiar with nxplayer/nxrecorder.
@patacongo do you think so?

patacongo:
> 
> 
> > Should this go under apps/testing instead of apps/system?
> 
> Other two audio tools put to apps/system:
> https://github.com/apache/incubator-nuttx-apps/tree/master/system/nxplayer
> https://github.com/apache/incubator-nuttx-apps/tree/master/system/nxrecorder
> The new place make nxlooper hard to find by people who is already familiar with nxplayer/nxrecorder.
> @patacongo do you think so?

I don't have any strong opinion.  nxplayer and nxrecorder are not tests, they are applications.  nxlooper is a test so it is functionally misplaced.  It is a test that is place in the same directory as its related applications.  I understand the logic for either location.

How would you want this to work in the the future?  Should tests be scattered all over in different directories are consolidated under testing?

There is a similar kind of issue with examples/ and testing/.  A test often makes a good example and most examples were developed as superficial tests.  But the idea is that an example is a HowTo for use of a feature that probably does not have a lot of depth.  A test should have better coverage and may not necessarily be a good HowTo application.

Ideally, it should be possible to automate tests and they should conform to some pass/fail reporting rules.   Examples, just report results in user friendly text.  But none of that has been thought through much less implemented.

xiaoxiang781216:
> > > Should this go under apps/testing instead of apps/system?
> > 
> > 
> > Other two audio tools put to apps/system:
> > https://github.com/apache/incubator-nuttx-apps/tree/master/system/nxplayer
> > https://github.com/apache/incubator-nuttx-apps/tree/master/system/nxrecorder
> > The new place make nxlooper hard to find by people who is already familiar with nxplayer/nxrecorder.
> > @patacongo do you think so?
> 
> I don't have any strong opinion. nxplayer and nxrecorder are not tests, they are applications. nxlooper is a test so it is functionally misplaced. It is a test that is place in the same directory as its related applications. I understand the logic for either location.
> 

The design of nxlooper is very closed to nxplayer, e.g. there is a simple interpreter to accept user command after launching nxlooper. And nxlooper more like a debugging tool to diagnose the audio driver functionality and then require the developer verify the result by ear manually.
 
> How would you want this to work in the the future? Should tests be scattered all over in different directories are consolidated under testing?
> 

Yes, it's better to move the application which design for testing(which mean run and report automatically) to apps/testing.

> There is a similar kind of issue with examples/ and testing/. A test often makes a good example and most examples were developed as superficial tests. But the idea is that an example is a HowTo for use of a feature that probably does not have a lot of depth. A test should have better coverage and may not necessarily be a good HowTo application.
> 
> Ideally, it should be possible to automate tests and they should conform to some pass/fail reporting rules. Examples, just report results in user friendly text. But none of that has been thought through much less implemented.

Yes, this is what I am thinking: let's program under apps/testing report pass/fail with zero/nozero exit code, then we can develop a test driver nsh script to manage all test automatically.

From this perspective, apps/system is more suitable place for nxlooper, because:

1. It's hard to very the result of nxlooper
2. nxlooper is more useful for driver bringup

xiaoxiang781216:
@patacongo do you think this patch is ready for merge after the dissussion?

xiaoxiang781216 (approved):
LGTM.

PR #541, opened 2020-12-29
added: Using nRF24L01+ as A Bluetooth Low Energy Broadcaster/Beacon

yjdwbj:
## Summary
Using nRF24L01+ as A Bluetooth Low Energy Broadcaster/Beacon. This app is inspired by Dmitry Grinberg’s idea [Faking Bluetooth LE](https://dmitry.gr/index.php?r=05.Projects&proj=11.%20Bluetooth%20LE%20fakery). 
## Testing
I have already tested on stm32f103-minimum:nrf24

### DIFF
--- a/examples/nrf24l01_btle/Makefile
+++ b/examples/nrf24l01_btle/Makefile
@@ -1,0 +1,34 @@
+############################################################################
+# apps/examples/nrf24l01_btle/Makefile
+#
+#  Licensed to the Apache Software Foundation (ASF) under one or more
+#  contributor license agreements.  See the NOTICE file distributed with
+#  this work for additional information regarding copyright ownership.  The
+#  ASF licenses this file to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance with the
+#  License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+#  License for the specific language governing permissions and limitations
+#  under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+# Basic nRF24L01+ terminal demonstration
+
+MAINSRC = nrf24l01_btle.c
+
+# built-in application info
+
+PROGNAME  = $(CONFIG_EXAMPLES_NRF24L01_BTLE_PROGNAME)
+PRIORITY  = $(CONFIG_EXAMPLES_NRF24L01_BTLE_PRIORITY)
+STACKSIZE = $(CONFIG_EXAMPLES_NRF24L01_BTLE_STACKSIZE)
+MODULE    = $(CONFIG_EXAMPLES_NRF24L01_BTLE)
+
+include $(APPDIR)/Application.mk
--- a/examples/nrf24l01_btle/nrf24l01_btle.c
+++ b/examples/nrf24l01_btle/nrf24l01_btle.c
@@ -1,0 +1,607 @@
+/****************************************************************************
+ * examples/nrf24l01_btle/nrf24l01_btle.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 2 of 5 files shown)


examples/nrf24l01_btle/Makefile
```diff
@@ -0,0 +1,49 @@
+############################################################################
```
btashton, line 1: This can be converted to use the Apache header.

examples/nrf24l01_btle/nrf24l01_btle.c
```diff
@@ -0,0 +1,635 @@
...
+  while (len--)
+    {
+      uint8_t a = *buf;
+      uint8_t v = 0;
+      *(buf++) = swapbits(a);
+    }
+}
+
+#ifdef CONFIG_DEBUG_WIRELESS
+void nrf24_dumpbuffer(FAR const uint8_t *outbuf)
```
btashton, line 224: Should be able to just use `lib_dumpbuffer`

examples/nrf24l01_btle/nrf24l01_btle.c
```diff
@@ -0,0 +1,635 @@
...
+  if (ret < 0)
+    {
+      wlerr("Error sending packet\n");
+      return ret;
+    }
+
+  return OK;
+}
+
+static inline void random_mac(void)
```
btashton, line 359: why not just call libc `arc4random_buf` to generate the random bytes.

examples/nrf24l01_btle/nrf24l01_btle.c
```diff
@@ -0,0 +1,635 @@
...
+  pls += 3 + 2;
+#else
+  chunk(buffer, pls)->size = 4 + 1;
+  chunk(buffer, pls)->type = 0xff; /* custom data */
+  chunk(buffer, pls)->data[0] = 't';
+  chunk(buffer, pls)->data[1] = 'e';
+  chunk(buffer, pls)->data[2] = 's';
+  chunk(buffer, pls)->data[3] = 't';
+  pls += 4 + 2;
+  nxsig_sleep(1);
```
btashton, line 460: just `sleep(1)` since this is userspace

examples/nrf24l01_btle/nrf24l01_btle.h
```diff
@@ -0,0 +1,87 @@
...
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __EXAMPLES_NRF24L01_BTLE_H
+#define __EXAMPLES_NRF24L01_BTLE_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
```
btashton, line 26: Can you add the correct section headers  here.  pre-processor defines, public types, public functions, etc...

examples/nrf24l01_btle/nrf24l01_btle.c
```diff
@@ -0,0 +1,635 @@
...
+  buffer.mac[2] = ((__TIME__[0] - 0x30) << 4) | (__TIME__[1] - 0x30);
+  buffer.mac[3] = ((__DATE__[4] - 0x30) << 4) | (__DATE__[5] - 0x30);
+  buffer.mac[4] = month(__DATE__);
+
+  /* static random address should have two topmost bits set */
+
+  buffer.mac[5] = ((__DATE__[9] - 0x30) << 4) | (__DATE__[10] - 0x30) | 0xc0;
+}
+
+/**
```
btashton, line 374: block comment style issue

examples/nrf24l01_btle/nrf24l01_btle.c
```diff
@@ -0,0 +1,635 @@
...
+  if (pls > 21)
+    {
+     wlerr("Total payload size must be 21 bytes or less.\n");
+    }
+
+  buffer.payload[pls] = 0x55;
+  buffer.payload[pls + 1] = 0x55;
+  buffer.payload[pls + 2] = 0x55;
+
+  /**
```
btashton, line 472: The `/**` and `**/` will fail style checks (ci is just broken for the moment).

btashton (changes requested):
This is very cool.  Just a few fixups.  The change to get CI working again is in review which I suspect would catch a few of the things I commented on.  Once this is in, would you mind adding this to an existing or new config in the OS repo so we can make sure this code has build test coverage.

examples/nrf24l01_btle/nrf24l01_btle.c
yjdwbj: Hi, @btashton . I have did test of your suggestion, If using libc `arc4random_buf ` , it dependency  enable `CONFIG_CRYPTO_RANDOM_POOL`.  if enable ` CONFIG_CRYPTO_RANDOM_POOL` and `CONFIG_DEBUG_FEATURES` I got report that `arm-none-eabi-ld: region `flash' overflowed by 1104 bytes` on stm32f103-minimum. Also, I  call libc `arc4random_buf` to generate random MAC, this app not working, which means it not discovered  by BLE scanner.

btashton: Yeah I guess that is kind of heavy.  You could also iterate over calls `random()` which should be very cheap and will give you a pseudo random number.

yjdwbj: OK, I added a new option by generate random, give user a choice, maybe the random mac  not suitable for some few people.

btashton (approved):
LGTM.  Thanks for making the changes.

PR #545, opened 2020-12-31
ostest: Don't assign NULL to mq_t varaibles

xiaoxiang781216:
## Summary
let's change to 0 instead since mq_t mayn't be a pointer. Found by PR: https://github.com/apache/incubator-nuttx/pull/2631

## Impact
No real change since NULL always equals 0

## Testing
Pass the build


Ouss4 (approved):
LGTM

PR #551, opened 2021-01-04
make/archive: replace the dependency to preprequisite list

anchao:
## Summary

make/archive: replace the dependency to preprequisite list

Change-Id: I3c3f9354cedb9f4f69c66698cd7556f1de95c660

## Impact

N/A

## Testing

N/A


Application.mk
```diff
@@ -138,11 +138,11 @@ $(CXXOBJS): %$(SUFFIX)$(OBJEXT): %$(CXXEXT)
 	$(if $(and $(CONFIG_BUILD_LOADABLE),$(CXXELFFLAGS)), \
 		$(call ELFCOMPILEXX, $<, $@), $(call COMPILEXX, $<, $@))
 
-archive:
+archive: $(OBJS)
```
xiaoxiang781216, line 141: should we chnage archive to a real target? then we can avoid the achive every time even there is nothing to change.

PR #558, opened 2021-01-13
Ftpd fixes

juniskane:
## Summary
- Fixes crash when ftpd_close() called before account setup
- Fixes compilation on dual-stack (both IPv4 and IPv6) systems
- Rename command RNTO return -EINVAL if no from path (This is just a minor thing, to correct original code that had superfluous return statement)
- nxstyle fixes


xiaoxiang781216 (approved):
LGTM.

PR #561, opened 2021-01-14
webclient: Add AF_LOCAL support

yamt:
## Summary
webclient: Add AF_LOCAL support
## Testing
AF_LOCAL part:
Tested against Docker Engine socket on macOS and Linux. (with a ported version of webclient)
Only build-tested for NuttX.

"webclient: Don't call the sink callback if no data is available" part:
Tested with a local app (only with TCP on NuttX)

### DIFF
--- a/include/netutils/webclient.h
+++ b/include/netutils/webclient.h
@@ -186,18 +186,22 @@ struct webclient_context
 {
   /* request parameters
    *
-   *   method       - HTTP method like "GET", "POST".
-   *                  The default value is "GET".
-   *   url          - A pointer to a string containing the full URL.
-   *                  (e.g., http://www.nutt.org/index.html, or
-   *                   http://192.168.23.1:80/index.html)
-   *   headers      - An array of pointers to the extra headers.
-   *   nheaders     - The number of elements in the "headers" array.
-   *   bodylen      - The size of the request body.
+   *   method           - HTTP method like "GET", "POST".
+   *                      The default value is "GET".
+   *   url              - A pointer to a string containing the full URL.
+   *                      (e.g., http://www.nutt.org/index.html, or
+   *                       http://192.168.23.1:80/index.html)
+   *   unix_socket_path - If not NULL, the path to an AF_LOCAL socket.
+   *   headers          - An array of pointers to the extra headers.
+   *   nheaders         - The number of elements in the "headers" array.
+   *   bodylen          - The size of the request body.
    */
 
   FAR const char *method;
   FAR const char *url;
+#if defined(CONFIG_WEBCLIENT_NET_LOCAL)
+  FAR const char *unix_socket_path;
+#endif
   FAR const char * FAR const *headers;
   unsigned int nheaders;
   size_t bodylen;
--- a/netutils/webclient/webclient.c
+++ b/netutils/webclient/webclient.c
@@ -67,6 +67,9 @@
 
 #include <arpa/inet.h>
 #include <netinet/in.h>
+#if defined(CONFIG_WEBCLIENT_NET_LOCAL)
+#include <sys/un.h>
+#endif
 
 #include <nuttx/version.h>
 
@@ -649,7 +652,6 @@ static int wget_gethostip(FAR char *hostname, FAR struct in_addr *dest)
 
 int webclient_perform(FAR struct webclient_context *ctx)
 {
-  struct sockaddr_in server;
   struct wget_s *ws;
   struct timeval tv;
   bool redirected;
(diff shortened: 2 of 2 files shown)


netutils/webclient/webclient.c
```diff
@@ -724,6 +724,15 @@ int webclient_perform(FAR struct webclient_context *ctx)
         {
           char port_str[sizeof("65535")];
 
+#if defined(CONFIG_WEBCLIENT_NET_LOCAL)
+          if (ctx->unix_socket != NULL)
```
xiaoxiang781216, line 728: If unix_socket doesn't implement TLS, why isn't conn.tls zero in the first place and then the whole block skip at line 723?

include/netutils/webclient.h
```diff
@@ -191,13 +191,17 @@ struct webclient_context
...
+   *   unix_socket  - If not NULL, the path to an AF_LOCAL socket.
    *   headers      - An array of pointers to the extra headers.
    *   nheaders     - The number of elements in the "headers" array.
    *   bodylen      - The size of the request body.
    */
 
   FAR const char *method;
   FAR const char *url;
+#if defined(CONFIG_WEBCLIENT_NET_LOCAL)
+  FAR const char *unix_socket;
```
xiaoxiang781216, line 203: who will set unix_socket? I don't find the related code.

include/netutils/webclient.h
```diff
@@ -191,13 +191,17 @@ struct webclient_context
...
    *                  (e.g., http://www.nutt.org/index.html, or
    *                   http://192.168.23.1:80/index.html)
+   *   unix_socket  - If not NULL, the path to an AF_LOCAL socket.
    *   headers      - An array of pointers to the extra headers.
    *   nheaders     - The number of elements in the "headers" array.
    *   bodylen      - The size of the request body.
    */
 
   FAR const char *method;
   FAR const char *url;
```
xiaoxiang781216, line 201: why not reuse url for unix_socket?

netutils/webclient/webclient.c
```diff
@@ -740,6 +760,40 @@ int webclient_perform(FAR struct webclient_context *ctx)
               return -errno;
             }
 
+#if defined(CONFIG_WEBCLIENT_NET_LOCAL)
```
xiaoxiang781216, line 772: merge line 742-752

netutils/webclient/webclient.c
yamt: to indicate the error to the user.

include/netutils/webclient.h
yamt: the calling app.

include/netutils/webclient.h
yamt: because url is still used for the http request.

netutils/webclient/webclient.c
yamt: what do you mean by "merge"?

include/netutils/webclient.h
xiaoxiang781216: so how do you handle both url and unix_socket not NULL?

netutils/webclient/webclient.c
xiaoxiang781216: like this:
```
#if defined(CONFIG_WEBCLIENT_NET_LOCAL)
          struct sockaddr_un server_un;
#endif
          struct sockaddr_in server_in;
          int domain;
          const struct sockaddr *server_address;
          socklen_t server_address_len;

          /* Create a socket */
#if defined(CONFIG_WEBCLIENT_NET_LOCAL)
          if (ctx->unix_socket != NULL)
            {
              domain = PF_LOCAL;
              memset(&server_un, 0, sizeof(server_un));
              server_un.sun_family = AF_LOCAL;
              strncpy(server_un.sun_path, ctx->unix_socket,
                      sizeof(server_un.sun_path));
#if !defined(__NuttX__) && !defined(__linux__)
              server_un.sun_len = SUN_LEN(&server_un);
#endif
              server_address = (const struct sockaddr *)&server_un;
              server_address_len = sizeof(server_un);
            }
          else
#endif
            {
              domain = PF_INET;
              /* Get the server address from the host name */

              server_in.sin_family = AF_INET;
              server_in.sin_port   = htons(ws->port);
              ret = wget_gethostip(ws->hostname, &server_in.sin_addr);
              if (ret < 0)
                {
                  /* Could not resolve host (or malformed IP address) */

                  nwarn("WARNING: Failed to resolve hostname\n");
                  ret = -EHOSTUNREACH;
                  goto errout_with_errno;
                }

              server_address = (const struct sockaddr *)&server_in;
              server_address_len = sizeof(struct sockaddr_in);
            }
```

netutils/webclient/webclient.c
```diff
@@ -740,6 +769,40 @@ int webclient_perform(FAR struct webclient_context *ctx)
...
+          if (ctx->unix_socket != NULL)
+            {
+              memset(&server_un, 0, sizeof(server_un));
+              server_un.sun_family = AF_LOCAL;
+              strncpy(server_un.sun_path, ctx->unix_socket,
+                      sizeof(server_un.sun_path));
+#if !defined(__NuttX__) && !defined(__linux__)
+              server_un.sun_len = SUN_LEN(&server_un);
```
xiaoxiang781216, line 780: BTW, do we really need set sun_len on macOS?

include/netutils/webclient.h
yamt: the url contains hostname, path, query string, etc.
they will be sent over the unix domain socket.
(while the hostname is less meaningful for unix domain, we need to send something anyway.)

netutils/webclient/webclient.c
yamt: yes

netutils/webclient/webclient.c
yamt: do you mean to move the "prepare server_address" part before socket() call? why?
the number of #ifdef is the same.

xiaoxiang781216: No, we can save one #ifdef/#endif pair and unix_socket check.

include/netutils/webclient.h
xiaoxiang781216, line 203: change to unpath? socket normally reserve to handle.

yamt: i named it after the curl option.
```
--unix-socket PATH    Connect through this Unix domain socket
```

how about `unix_socket_path` then?
`unpath` sounds a bit cryptic to me.

xiaoxiang781216: Ok.

netutils/webclient/webclient.c
yamt: is this what you meant? https://github.com/apache/incubator-nuttx-apps/pull/561/commits/d61d23dd44a4aab892d22ecd0bc736eb631d31c1
(not tested well)

xiaoxiang781216: Yes.

xiaoxiang781216:
BTW, please merge all patches(except webclient: Don't call the sink callback if no data is available) into one.

yamt:
> BTW, please merge all patches(except webclient: Don't call the sink callback if no data is available) into one.

i will squash them after testing with my local app.

yamt:
> > BTW, please merge all patches(except webclient: Don't call the sink callback if no data is available) into one.
> 
> i will squash them after testing with my local app.

done

xiaoxiang781216 (approved):

PR #572, opened 2021-01-25
Add support in printf command to send 8-32 bits value

acassis:
## Summary
A NuttX user reported that the command to turn ON/OFF the userleds from nsh wasn't working:

nsh> printf \x01 > /dev/userleds

It is described in this video tutorial: https://www.youtube.com/watch?v=GzX9okqfxX4&t=150s

After some investigation I discovered that the issue was caused by this modification:
https://github.com/apache/incubator-nuttx/commit/7a18ebe45980a80370b7afaa4b4d0be703402318

## Impact
Now users will be able to use the printf command again to control the LEDs, but they will need to send 32-bit instead of 8-bit:

nsh> printf \x00000001 > /dev/userleds

## Testing
stm32f103-minimum board

### DIFF
--- a/nshlib/nsh_printf.c
+++ b/nshlib/nsh_printf.c
@@ -63,6 +63,8 @@ int cmd_printf(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
 {
   FAR char *fmt;
   char ch;
+  uint32_t value;
+  int len;
   int i;
 
   /* parse each argument, detecting the right action to take
@@ -84,7 +86,25 @@ int cmd_printf(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
             {
               case 'x':
                 fmt++;
-                nsh_output(vtbl, "%c", strtol(fmt, NULL, 16));
+                value = strtoul(fmt, NULL, 16);
+                len = strnlen(fmt, 10);
+
+                if (len >= 7)
+                  {
+                    nsh_output(vtbl, "%c%c%c%c", value & 0xff,
+                                                 (value & 0xff00) >> 8,
+                                                 (value & 0xff0000) >> 16,
+                                                 (value & 0xff000000) >> 24);
+                  }
+                else if (len >= 3)
+                    {
+                      nsh_output(vtbl, "%c%c", value & 0xff,
+                                               (value & 0xff00) >> 8);
+                    }
+                  else if (len >= 1)
+                      {
+                        nsh_output(vtbl, "%c", value & 0xff);
+                      }
                 break;
 
               case 'n':


nshlib/nsh_printf.c
```diff
@@ -84,7 +87,34 @@ int cmd_printf(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
             {
               case 'x':
```
xiaoxiang781216, line 88: should we use something like this for multiple char:
printf \x12 \x34 \x56 \x78
from https://wiki.bash-hackers.org/commands/builtin/printf:
one /x should generate one byte output.

acassis: Hi @xiaoxiang781216, using "printf \x01 \x00 \x00 \x00 > /dev/userleds" works, but the user will need to take care about endianess. Using \x00000001 is more intuitive, because we know the LSB is 1

nshlib/nsh_printf.c
```diff
@@ -84,7 +87,34 @@ int cmd_printf(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
...
                 fmt++;
-                nsh_output(vtbl, "%c", strtol(fmt, NULL, 16));
+                value = strtol(fmt, NULL, 16);
+                len = strnlen(fmt, 10);
+
+                /* Is it a Big Endian MCU ? */
+
+                if (BYTE_ORDER == BIG_ENDIAN)
+                  {
+                    value = __swap_uint32(value);
```
xiaoxiang781216, line 97: Don't manually swap the value.
(0x12345678 & 0xff00) >> 8 always get 0x56.

nshlib/nsh_printf.c
```diff
@@ -84,7 +87,34 @@ int cmd_printf(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
             {
               case 'x':
                 fmt++;
-                nsh_output(vtbl, "%c", strtol(fmt, NULL, 16));
+                value = strtol(fmt, NULL, 16);
```
xiaoxiang781216, line 90: change strtol to strtoul

nshlib/nsh_printf.c
```diff
@@ -84,7 +87,34 @@ int cmd_printf(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
...
+                len = strnlen(fmt, 10);
+
+                /* Is it a Big Endian MCU ? */
+
+                if (BYTE_ORDER == BIG_ENDIAN)
+                  {
+                    value = __swap_uint32(value);
+                  }
+
+                if (len == 8)
```
xiaoxiang781216, line 100: change to len >= 7(len >= 3, len >= 1)

nshlib/nsh_printf.c
```diff
@@ -84,7 +87,34 @@ int cmd_printf(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
...
+                  {
+                    nsh_output(vtbl, "%c%c%c%c", value & 0xff,
+                                                 (value & 0xff00) >> 8,
+                                                 (value & 0xff0000) >> 16,
+                                                 (value & 0xff000000) >> 24);
+                  }
+                else
+                  if (len == 4)
```
xiaoxiang781216, line 108: Why split "else if" into two lines?

nshlib/nsh_printf.c
```diff
@@ -84,7 +87,34 @@ int cmd_printf(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
...
+                    {
+                      nsh_output(vtbl, "%c%c", value & 0xff,
+                                               (value & 0xff00) >> 8);
+                    }
+                  else
+                    if (len == 2)
```
xiaoxiang781216, line 114: same question

nshlib/nsh_printf.c
acassis: Ok! I will remove this check. I need to find a Big Endian MCU supported by NuttX. Doing "git grep "CONFIG_ENDIAN_BIG=y" " inside boards/ reveals some. I will try to get a powered by at32uc3.

nshlib/nsh_printf.c
acassis: ok

nshlib/nsh_printf.c
acassis: I use to do this way, but I could keep it in a single line.

nshlib/nsh_printf.c
acassis: ditto

xiaoxiang781216 (approved):

PR #580, opened 2021-01-29
lvgl: add button, keypad, lcd interface.

FASTSHIFT:
## Summary
lvgl: add button, keypad, lcd interface.
## Impact
let lvgl access button and lcd devices.
## Testing
daily test

### DIFF
--- a/graphics/lvgl/Kconfig
+++ b/graphics/lvgl/Kconfig
@@ -66,6 +66,22 @@ config LV_INV_FIFO_SIZE
 		If too much area is invalidated (greater then this number) then
 		the whole screen will be refreshed
 
+config LV_USE_LCD_INTERFACE
+	bool "LCD interface"
+	default n
+
+if LV_USE_LCD_INTERFACE
+
+config LV_LCD_INTERFACE_DEVICEPATH
+	string "LCD device path"
+	default "/dev/lcd0"
+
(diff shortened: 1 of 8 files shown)


v01d:
I think that this code should be part of lvgldemo and no lvgl, since this part is usually very dependant on how the application is structured (specific LCD, how is input going to be handled, either event-based or polling, etc). On lvgldemo there's an fb interface, so the LCD interface could be made into a choice (either FB or LCD).

xiaoxiang781216:
> I think that this code should be part of lvgldemo and no lvgl, since this part is usually very dependant on how the application is structured (specific LCD, how is input going to be handled, either event-based or polling, etc). On lvgldemo there's an fb interface, so the LCD interface could be made into a choice (either FB or LCD).

But we can achieve the customization through Kconfig, there isn't technical reason to block us creating the generic and flexible lvgl adapter for NuttX driver.

v01d:
I'm not sure how easy would be to make this generic, since the display driver will usually require supplying the "set pixel" and "rounder" callbacks depending on the particular display to use. As long as it can be disabled I think it is OK.
In that case I think we should move the framebuffer interface (and also made optional) from lvgldemo to here.

xiaoxiang781216:
Let's close it now. the new drivers for v8 will provide later.

Issue #584, opened 2021-02-11
Build error in telnetstart if no standard console


lukegluke:
If CONFIG_NSH_CONSOLE=n can't build telnetstart because `nsh_initscript` is called with undeclared `vtbl`.
https://github.com/apache/incubator-nuttx-apps/blob/ca6907ccbe6ee732ba1c179d789fbfb2f8e5e6ae/nshlib/nsh_telnetd.c#L260-L267

PR #586, opened 2021-02-12
nshlib: Fix the build break in nsh_telnetd.c when CONFIG_NSH_CONSOLE=n

xiaoxiang781216:
## Summary

by moving the intializition code to nsh_telnetmain.
Note: the init script postpone the execution when the first user login too.

## Impact

See the issue description:
https://github.com/apache/incubator-nuttx-apps/issues/584

### DIFF
--- a/nshlib/nsh_telnetd.c
+++ b/nshlib/nsh_telnetd.c
@@ -90,6 +90,32 @@ static int nsh_telnetmain(int argc, char *argv[])
   DEBUGASSERT(pstate != NULL);
   vtbl = &pstate->cn_vtbl;
 
+  /* Initialize any USB tracing options that were requested.  If
+   * standard console is also defined, then we will defer this step to
+   * the standard console.
+   */
+
+#if defined(CONFIG_NSH_USBDEV_TRACE) && !defined(CONFIG_NSH_CONSOLE)
+  usbtrace_enable(TRACE_BITSET);
+#endif
+
+  /* Execute the startup script.  If standard console is also defined,
+   * then we will not bother with the initscript here (although it is
+   * safe to call nsh_initscript multiple times).
+   */
+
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@lukegluke please try this path.

lukegluke:
Hi @xiaoxiang781216,
Patch fixed the build, but network not works if CONFIG_NSH_CONSOLE=n.
I suppose it is because netinit_bringup() is called only from nsh_consolemain (usbconsole/altconsole):
https://github.com/apache/incubator-nuttx-apps/blob/39a1950098b331c5a2c203431ffc26f276274863/nshlib/nsh_consolemain.c#L101-L105
that in turn is not called if no console:
https://github.com/apache/incubator-nuttx-apps/blob/f613da4f92d5ae1f719cda3f3f2ae28db3b7559b/system/nsh/nsh_main.c#L165-L168

acassis:
@lukegluke about #584 do you know which PR broke it? Maybe we could revert the offending PR until we have telnet working with it again.

lukegluke:
@acassis I had found it by git blame for that time, but I can't say it to you now.
Why not just moved netinit_bringup() code to main(int argc, FAR char *argv[]) in nsh_main.c (see code parts above)?

acassis:
> @acassis I had found it by git blame for that time, but I can't say it to you now.
> Why not just moved netinit_bringup() code to main(int argc, FAR char *argv[]) in nsh_main.c (see code parts above)?

Because netinit_bringup() doesn't depend on nsh! You can create a standalone (a modified hello_world for example) that needs network initialization.

lukegluke:
Yes you're right, I see. But as it seemed to me that previously netinit_bringup was also in nsh, so it is another problem. I didn't get deep into history of this issue and related net bringup problems, just noticed the issue.

xiaoxiang781216:
See the better fix here: https://github.com/apache/incubator-nuttx-apps/pull/703

PR #593, opened 2021-02-16
Pin Nimble release to working commit until stable tag is ready

btashton:
## Summary
Pin Nimble release to working commit until stable tag is ready
See issue noted here: https://github.com/apache/incubator-nuttx-apps/issues/592

## Impact
CI should function again

## Testing
CI passes this was a previously tested sha.

### DIFF
--- a/wireless/bluetooth/nimble/Kconfig
+++ b/wireless/bluetooth/nimble/Kconfig
@@ -9,7 +9,7 @@ config NIMBLE
 if NIMBLE
   config NIMBLE_REF
   string "Version"
-  default "master"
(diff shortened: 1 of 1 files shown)


btashton:
@xiaoxiang781216 @v01d  FYI.  I think this will unbreak the CI, so we should merge soon.

xiaoxiang781216 (approved):
Sure.

Ouss4:
@xiaoxiang781216 let's not wait for the macOS builds please.  This will take hours and we need to fix the CI.

PR #594, opened 2021-02-16
ntpclient.c: Replace home-grown ntp_addr_u with sockaddr_storage

yamt:
## Summary
ntpclient.c: Replace home-grown ntp_addr_u with sockaddr_storage

## Testing
It tested IPv4 with lm3s6965-ek:qemu-protected on qemu.
IPv6 is not tested.

### DIFF
--- a/netutils/ntpclient/ntpclient.c
+++ b/netutils/ntpclient/ntpclient.c
@@ -154,12 +154,7 @@ struct ntpc_daemon_s
 union ntp_addr_u
 {
   struct sockaddr sa;
-#ifdef CONFIG_NET_IPv4
-  struct sockaddr_in in4;
-#endif
-#ifdef CONFIG_NET_IPv6
-  struct sockaddr_in6 in6;
-#endif
+  struct sockaddr_storage ss;
 };
 
 /* NTP offset. */
@@ -175,7 +170,7 @@ struct ntp_sample_s
 
 struct ntp_servers_s
 {
-  union ntp_addr_u list[CONFIG_NETUTILS_NTPCLIENT_NUM_SAMPLES];
+  struct sockaddr_storage list[CONFIG_NETUTILS_NTPCLIENT_NUM_SAMPLES];
   size_t num;
   size_t pos;
   FAR char *hostlist_str;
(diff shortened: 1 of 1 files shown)


netutils/ntpclient/ntpclient.c
```diff
@@ -625,14 +699,15 @@ static void ntpc_settime(int64_t offset, FAR struct timespec *start_realtime,
  *
  ****************************************************************************/
 
-static bool ntp_address_in_kod_list(FAR const union ntp_addr_u *server_addr)
+static bool ntp_address_in_kod_list(FAR const struct sockaddr *server_addr)
```
xiaoxiang781216, line 702: change all sockaddr to sockaddr_storage to avoid the unnecessary cast

yamt: i'm not sure if i want to do it.

if you feel strong about casts, i can do
```
union ntp_addr_u
  {
    struct sockaddr sa;
    struct sockaddr_storage ss;
  }
```
does it sound better to you?

xiaoxiang781216: > i'm not sure if i want to do it.
> 
> if you feel strong about casts, i can do
> 

Yes, the unnecessary cast mean the design of function prototype is poor and we lose the compiler time check too.

> ```
> union ntp_addr_u
>   {
>     struct sockaddr sa;
>     struct sockaddr_storage ss;
>   }
> ```
> 
> does it sound better to you?

Why not let function accept sockaddr_storage parameter directly?

yamt: > Why not let function accept sockaddr_storage parameter directly?

because it's just awkward. it's what struct sockaddr is for.

xiaoxiang781216: Why sockaddr_storage awkward here?
NTP client support both ipv4 and ipv6, so sockaddr_storage is the better representation to this fact. sockaddr can only express ipv4 address which isn't good enough to express this case.

yamt: it isn't sockaddr_storage itself i feel awkward.

what's awkward is, using sockaddr_storage as a parameter of functions, where sockaddr is more appropriate. ie. for functions where sockaddr_storage is not assumed.

xiaoxiang781216: Yes, we have to cast when interact with socket API by sockaddr_storage. Either approach can't avoid the cast totaly, why not select a more clean interface for our internal function? So ntp_addr_u look like a reasonable compromise to overcome the legacy socket API desgin

yamt: i updated the patch. partially revived the union to reduce casts.

xiaoxiang781216: After review the change again, I don't see too much benefit to switch ipv4/ipv6 to ss, but lose the static type check. On the other hand, sockaddr_cmp and sockaddr_len change look good.

yamt: do you mean, you prefer to use ntp_addr_u all over places, including sockaddr_cmp and sockaddr_len?

xiaoxiang781216: socket API stick to sockaddr, because they need to keep the compatibility, but we don't have this restriction for internal NTP client function. So if the caller declares variables as ntp_addr_u, why we change other function prototype from ntp_addr_u to sockaddr and spread the cast around the code base?

yamt: because it's more convenient to use the standard way (vs app specific types) to implement things like sockaddr_cmp.

xiaoxiang781216: Why? here is sockaddr_cmp with ntp_addr_u:
```
static int sockaddr_cmp(FAR const union ntp_addr_u*sa1,
                        FAR const union ntp_addr_u*sa2)
{
  if (sa1->sa.sa_family != sa2->sa.sa_family)
    {
      return 1;
    }

  switch (sa1->sa.sa_family)
    {
#ifdef CONFIG_NET_IPv4
    case AF_INET:
      {
        FAR const struct sockaddr_in *sin1 = &sa1->in4;
        FAR const struct sockaddr_in *sin2 = &sa2->in4;

        if (sin1->sin_addr.s_addr != sin2->sin_addr.s_addr)
          {
            return 1;
          }

        if (sin1->sin_port != sin2->sin_port)
          {
            return 1;
          }

        break;
      }
#endif

#ifdef CONFIG_NET_IPv6
    case AF_INET6:
      {
        FAR const struct sockaddr_in6 *sin1 = &sa1->in6;
        FAR const struct sockaddr_in6 *sin2 = &sa2->in6;

        if (!memcmp(sin1->sin6_addr.s6_addr,
                    sin2->sin6_addr.s6_addr, 16))
          {
            return 1;
          }

        if (sin1->sin6_port != sin2->sin6_port)
          {
            return 1;
          }

        break;
      }
#endif

    default:
      DEBUGASSERT(0);
      break;
    }

  return 0
}
```
No much different compared with sockaddr version(actually, both have the same number of line).

yamt: ntp_addr_u is not standard.

xiaoxiang781216: sockaddr_cmp is not standard too. If we follow your direction, ntpc_daemon_s, ntp_addr_u, ntp_sample_s, ntp_servers_s and ntp_kod_exclude_s should be removed, all function only accept the primitve type or the standard defined struct.

yamt: as i said, "more convenient to use the standard way (vs app specific types)". that's all.
i'm not arguing that it's impossible to use ntp_addr_u. or force everything to use it. i'm not that extreme. :-)
i can understand your preference. probably i will withdraw this pr.

yamt:
i withdraw this

PR #595, opened 2021-02-16
canutils: changed print_usage function to static

michallenc:
## Summary
Global print_usage function in canutils/candump/ was causing double definition error while compiling NuttX with some extern apps, like pysimCoder for example. Therefore the following change to static function. To be on safe side, the same change was also done in cansend.

## Testing

Tested on Teensy 4.1 with pysimCoder.

### DIFF
--- a/canutils/candump/candump.c
+++ b/canutils/candump/candump.c
@@ -115,7 +115,7 @@ extern int optind, opterr, optopt;
 
 static volatile int running = 1;
 
-void print_usage(char *prg)
+static void print_usage(char *prg)
 {
 	fprintf(stderr, "%s - dump CAN bus traffic.\n", prg);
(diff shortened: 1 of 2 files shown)


michallenc:
Oh, wasn´t expecting that much nx style errors. Gonna have to get some sleep before fixing this, I´ll push nx style changes tomorrow.

btashton:
> Oh, wasn´t expecting that much nx style errors. Gonna have to get some sleep before fixing this, I´ll push nx style changes tomorrow.

I think for canutils we have in the past just left it as is to closely match the upstream and make it easier to backport.

michallenc:
@btashton Well I definitely wouldn´t mind if we just merge it as it is.

xiaoxiang781216 (approved):

xiaoxiang781216:
So, let's ignore the style issue and merge the correct fix.

PR #604, opened 2021-03-03
apps/md5: fix md5 result error

GUIDINGLI:
## Summary

apps/md5: fix md5 result error


nshlib/nsh_codeccmd.c
```diff
@@ -443,8 +443,8 @@ static int cmd_codecs_proc(FAR struct nsh_vtbl_s *vtbl, int argc,
           int i;
 
           md5_final(mac, &ctx);
-          src = (FAR char *)&mac;
-          dest = destbuf;
+          src  = mac;
+          dest = (FAR unsigned char *)destbuf;
```
xiaoxiang781216, line 447: we can keep dest as `char`, only src need change to `unsigned char` to avoid the array index become negative in `hexchars[(*src) >> 4]`.

xiaoxiang781216 (approved):
LGTM.

PR #611, opened 2021-03-16
iperf: plug memory leaks

yamt:
## Summary
iperf: plug memory leaks
## Testing
tested locally on an esp32 board


xiaoxiang781216 (approved):
LGTM.

PR #619, opened 2021-03-18
dhcpd/telnet: don't need to dupcalite listen socket when task_create

Donny9:
## Summary
dhcpd/telnet: don't need to dupcalite listen socket when task_create. this patch associates with https://github.com/apache/incubator-nuttx/pull/3045
## Impact
normal using telnetd/dhcpd.
## Testing
daily test

### DIFF
--- a/netutils/dhcpd/dhcpd.c
+++ b/netutils/dhcpd/dhcpd.c
@@ -1,5 +1,5 @@
 /****************************************************************************
- * netutils/dhcpd/dhcpd.c
+ * apps/netutils/dhcpd/dhcpd.c
  *
  *   Copyright (C) 2007-2009, 2011-2014, 2017, 2020 Gregory Nutt. All rights
  *     reserved.
@@ -846,7 +846,7 @@ static inline int dhcpd_socket(void)
 
   /* Create a socket to listen for requests from DHCP clients */
 
-  sockfd = socket(PF_INET, SOCK_DGRAM, 0);
+  sockfd = socket(PF_INET, SOCK_DGRAM | SOCK_CLOEXEC, 0);
   if (sockfd < 0)
     {
       nerr("ERROR: socket failed: %d\n", errno);
--- a/netutils/telnetd/telnetd_daemon.c
+++ b/netutils/telnetd/telnetd_daemon.c
@@ -1,5 +1,5 @@
 /****************************************************************************
- * netutils/telnetd/telnetd_daemon.c
+ * apps/netutils/telnetd/telnetd_daemon.c
  *
  *   Copyright (C) 2012, 2017 Gregory Nutt. All rights reserved.
  *   Author: Gregory Nutt <gnutt@nuttx.org>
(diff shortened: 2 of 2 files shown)


masayuki2009:
@Donny9 

Thanks, telnetd works now.

By the way, I noticed that webserver and ftpd work without modification.
Could you tell me why only telnetd and dhcpd need to be modified?

Donny9:
> @Donny9
> 
> Thanks, telnetd works now.
> 
> By the way, I noticed that webserver and ftpd work without modification.
> Could you tell me why only telnetd and dhcpd need to be modified?

Okay，i need time to test them and modify them in other patch, thank you～

Donny9:
> @Donny9
> 
> Thanks, telnetd works now.
> 
> By the way, I noticed that webserver and ftpd work without modification.
> Could you tell me why only telnetd and dhcpd need to be modified?

@masayuki2009  I just checked ftpd daemon, it doesn't need to modify because ftpd_startworker will create a work thread for  new connection  by pthread_create API, the filelist duplicate don't occur when pthread_create. And the thttpd webserver used poll(fdwatch) to monitor the fd when a new connections occurs, the new thread or task doesn't be created, so it doesn't need  to modify. 
The root cause is if parent task has socket(listened,not connect), child task will fail when it is created because the dup operate(psock_dup2) will fail.

xiaoxiang781216:
> > @Donny9
> > Thanks, telnetd works now.
> > By the way, I noticed that webserver and ftpd work without modification.
> > Could you tell me why only telnetd and dhcpd need to be modified?
> 
> @masayuki2009 I just checked ftpd daemon, it doesn't need to modify because ftpd_startworker will create a work thread for new connection by pthread_create API, the filelist duplicate don't occur when pthread_create. And the thttpd webserver used poll(fdwatch) to monitor the fd when a new connections occurs, the new thread or task doesn't be created, so it doesn't need to modify.
> The root cause is if parent task has socket(listened,not connect), child task will fail when it is created because the dup operate(psock_dup2) will fail.

https://github.com/apache/incubator-nuttx/pull/3045 check the file duplication result(ignore before this PR):
https://github.com/apache/incubator-nuttx/commit/e96c8b9283ddbdc329e96d262aefc9bb199d5578#diff-3e4edca12a292d53823387fc8d11439c7f1efe7702f80962193bdff8f3c27045L154-R83
and expose an old bug happen in tcp socket dup here:
https://github.com/apache/incubator-nuttx/blob/master/net/socket/net_dup2.c#L102
https://github.com/apache/incubator-nuttx/blob/37d5c1b0d93290212b7a7a246bd5845f8f0dd3cd/net/tcp/tcp_monitor.c#L257-L270
Since the listen socket isn't in TCP_ESTABLISHED or TCP_SYN_RCVD state, tcp_start_monitor return fail and then fail to create the new task.

anchao:
Hi @masayuki2009 san,

Please help to review the below PR, which fixed the TCP socket fd dup(2) issue.

https://github.com/apache/incubator-nuttx/pull/3099

masayuki2009:
@xiaoxiang781216 

>So the fix has two part:
>
>    Add SOCK_CLOEXEC to server who spawn the child task, but the sock isn't really used by child
>    Fix the error in psock_dup2 which will provide by @anchao soon.

Thanks, I understand.

PR #620, opened 2021-03-18
Extend Testing of Sporadic Scheduler in OS test

patacongo:
## Summary

This commit adds the test developed by Jan Staschulat with Issue #apache/incubator-nuttx/2935

## Impact

It is expected to cause the OS test to fail for the time being since that issue reports a bug (that has been corrected; no bug is reported by OS test).

The test is rather long (about 10 minutes), however, that is only if the sporadic scheduler is enabled.  I don't think it ever is.

## Testing

Tested using the (new) stm32f4discovery:sporadic configuration.

### DIFF
--- a/testing/ostest/Makefile
+++ b/testing/ostest/Makefile
@@ -1,35 +1,20 @@
 ############################################################################
 # apps/testing/ostest/Makefile
 #
-#   Copyright (C) 2007-2012, 2014, 2017 Gregory Nutt. All rights reserved.
-#   Author: Gregory Nutt <gnutt@nuttx.org>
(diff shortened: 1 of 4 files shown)


patacongo:
Seems to be a good change.  See sample output in Issue #apache/incubator-nuttx/issues/2935

My only hesitancy is that the test is rather long.  However, I do no believe is a problem because I do not think that the SPORADIC scheduler is ever enabled in any OS tests(?)

patacongo:
I have some problems and lost some changes.  I have temporarily marked this as a draft again to see if I can recover.

xiaoxiang781216 (approved):

PR #626, opened 2021-03-19
build: Remove the null assignment

xiaoxiang781216:
## Summary
clean up the Makefile a little bit

## Impact
Should not

## Testing
Pass CI test


examples/udpblaster/udpblaster.h
```diff
@@ -40,8 +40,6 @@
  * Included Files
  ****************************************************************************/
 
-#include "config.h"
```
Ouss4, line 43: @xiaoxiang781216 I'm not sure if this is not needed.  My recollection is we need this when running from the host.  The Makefile generates this "config.h".

xiaoxiang781216: Yes, you are right. I revert the change.

Ouss4 (approved):

PR #624, opened 2021-03-19
fsutils/mksmartfs: Fix regression caused by BCH separation

gustavonihei:
## Summary
This PR intends to fix a regression caused by PR [3061](https://github.com/apache/incubator-nuttx/pull/3061), which separated the BCH logic from CONFIG_DISABLE_MOUNTPOINT.
Since the Block-to-character driver is not built by default, `mksmartfs` fails to open the file descriptor of the MTD block device.

## Impact
Fix `mksmartfs` NSH command on targets that support SMARTFS file system.

## Testing
`esp32-devkic:spiflash` and also on custom configuration for `sim`
`fstest` also executed to completion.

Previous behavior:
```shell
NuttShell (NSH) NuttX-10.0.1
nsh> mksmartfs /dev/smart0
nsh: mksmartfs: mksmartfs failed: 2
nsh> 
```
With proposed change:
```shell
NuttShell (NSH) NuttX-10.0.1
nsh> mksmartfs /dev/smart0
nsh> 
```

PR #627, opened 2021-03-21
fsutils/mkfatfs/Kconfig: Select BCH

Ouss4:
## Summary
Similar situation as #624 
## Impact
mkfatfs doesn't report issues anymore.
## Testing
sim:nsh

### DIFF
--- a/fsutils/mkfatfs/Kconfig
+++ b/fsutils/mkfatfs/Kconfig
@@ -7,5 +7,6 @@ config FSUTILS_MKFATFS
 	bool "mkfatfs utility"
 	default y
 	depends on FS_FAT && !DISABLE_PSEUDOFS_OPERATIONS
+	select BCH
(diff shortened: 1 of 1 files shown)


Ouss4:
Looks like there are a few defconfigs that need to be refreshed from the kernel side.

xiaoxiang781216 (approved):

xiaoxiang781216:
> Looks like there are a few defconfigs that need to be refreshed from the kernel side.

Before merging this PR, we have to update the kernel first.

Ouss4:
> Before merging this PR, we have to update the kernel first.

I'll do that this evening.  But wouldn't it have the same effect?  Unless both are merged together.

xiaoxiang781216:
Yes, once you prepare the patch on kernel side, I will commit both at the same time.

PR #628, opened 2021-03-24
examples/oneshot: fixed wrong constant and refactored

saramonteiro:
## Summary
This little PR is intended only to:
* Fix a wrong constant that was being used to convert time unities and 
* Refactor the code using constant definitions instead of raw numbers. 

## Impact
N/A

## Testing
After the changes the maximum delay became correct.

### DIFF
--- a/examples/oneshot/oneshot_main.c
+++ b/examples/oneshot/oneshot_main.c
@@ -1,5 +1,5 @@
 /****************************************************************************
- * examples/oneshot/oneshot_main.c
+ * apps/examples/oneshot/oneshot_main.c
  *
  *   Copyright (C) 2015 Gregory Nutt. All rights reserved.
  *   Author: Gregory Nutt <gnutt@nuttx.org>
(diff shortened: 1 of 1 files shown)


davids5 (approved):
LGTM All except but the name of NSEC_PER_USEC seams backwards, doesn't it?

Number for Seconds per micros Second should be NUSEC_PER_SEC, Number for micros Seconds per Second  but that looks like anther PR....

v01d:
> LGTM All except but the name of NSEC_PER_USEC seams backwards, doesn't it?
> 
> Number for Seconds per micros Second should be NUSEC_PER_SEC, Number for micros Seconds per Second but that looks like anther PR....

It refers to nanoseconds per microsecond.

saramonteiro:
> LGTM All except but the name of NSEC_PER_USEC seams backwards, doesn't it?
> 
> Number for Seconds per micros Second should be NUSEC_PER_SEC, Number for micros Seconds per Second but that looks like anther PR....

Hey David, 
these definitions are not mine. They're from `include/nuttx/clock.h` and I don't think it's a good idea to change it.
It will break several files.

patacongo:
> 
> 
> > LGTM All except but the name of NSEC_PER_USEC seams backwards, doesn't it?
> > Number for Seconds per micros Second should be NUSEC_PER_SEC, Number for micros Seconds per Second but that looks like anther PR....
> 
> It refers to nanoseconds per microsecond.

You are interpreting the name incorrectly.  NSEC_PER_USEC does not stand for number seconds per microsecond, it stands for the number nanoseconds per microsecond.  NSEC always refers to nanonseconds in clock.h.

PR #631, opened 2021-03-25
candump.c:  Must not extern optind, et al

patacongo:
## Summary

optind, et al, are not longer simple globals variables (after incubator-nuttx PR 3170).  Redundantly externing them in application code now results in compilation errors.

## Impact

Need for compile with incubator_nuttx PR 3170.  No other impact expected.

## Testing

CI only

### DIFF
--- a/canutils/candump/candump.c
+++ b/canutils/candump/candump.c
@@ -111,8 +111,6 @@ const int canfd_on = 1;
 const char anichar[MAXANI] = {'|', '/', '-', '\\'};
 const char extra_m_info[4][4] = {"- -", "B -", "- E", "B E"};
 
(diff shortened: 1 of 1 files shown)


patacongo:
How did this get into the repository?  It is not even close to the NuttX coding standard!

xiaoxiang781216 (approved):

xiaoxiang781216:
> How did this get into the repository? It is not even close to the NuttX coding standard!

From my remember, there is dissusion to keep the original style to simplify the upgrade since canutlis is still in the active maintain. But, anyware let's merge this PR first which isn't relate to the coding standard.

PR #634, opened 2021-03-28
build: Remove the redundant check from Applicaiton.mk

xiaoxiang781216:
## Summary
It's better to report error if user forget set PRIORITY or STACKSIZE

## Impact
Build sytem

## Testing
Pass CI

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -199,10 +199,7 @@ endif # BUILD_MODULE
 
 context::
 
-ifeq ($(CONFIG_NSH_BUILTIN_APPS),y)
 ifneq ($(PROGNAME),)
-ifneq ($(PRIORITY),)
-ifneq ($(STACKSIZE),)
 
 REGLIST := $(addprefix $(BUILTIN_REGISTRY)$(DELIM),$(addsuffix .bdat,$(PROGNAME)))
 APPLIST := $(PROGNAME)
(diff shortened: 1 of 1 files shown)


v01d:
I imagine not settings this variables probably returns in some weird error right? Have you tried?
Maybe an error message could be printed in that case.

xiaoxiang781216:
Here is the error reported by gcc:
```
builtin_list.h:2:12: error: expected expression before ‘,’ token
    2 | { "hello", , , hello_main },
```
If user open builtin_list.h, he will find the reason easily:
```
{ "sh", 100, 65536, sh_main },
{ "hello", , , hello_main },
{ "nsh", 100, 65536, nsh_main },
```

xiaoxiang781216:
@v01d do you think the error message acceptable?

v01d:
Sorry I missed this one. 
I think it would be good to explicitly print a message stating the missing definitions. But I guess it is not critical as it would be uncommon to have these variables missing.

Ouss4:
What was the purpose of all those `register::` anyway?  I think only one, unconditional, `register::` would have worked the same way.

Ouss4:
> What was the purpose of all those `register::` anyway? I think only one, unconditional, `register::` would have worked the same way.

I mean + one, besides the first `register::` that does all the work...

xiaoxiang781216:
We need one empty register to handle no program(library only) case.

Ouss4:
> We need one empty register to handle no program(library only) case.

Yes, this is what I meant.  Only one empty register would've worked the same way.

Ouss4 (approved):

xiaoxiang781216:
@Ouss4 @v01d could you merge this patch?

Ouss4:
I'm not sure if @v01d have any concerns.   I'll merge this one, please ping if there is anything to look at.

PR #636, opened 2021-03-28
Support for MQTT-C

Ouss4:
## Summary
Add glue code for MQTT-C and an example to use the library.
## Impact
No impact on existing code.  Library disabled by default.
## Testing
Tested with ESP32 DevkitC and Olimex STM32 P407. defconfigs attached, IP addresses need to be adapted.
[esp32_mqttc.txt](https://github.com/apache/incubator-nuttx-apps/files/6220719/esp32_mqttc.txt)
[olimex_mqttc.txt](https://github.com/apache/incubator-nuttx-apps/files/6220720/olimex_mqttc.txt)


mqtt/mqttc/Make.defs
```diff
@@ -0,0 +1,32 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifneq ($(CONFIG_MQTT_MQTTC),)
+CONFIGURED_APPS += $(APPDIR)/mqtt/mqttc
+
+CFLAGS   += -D__unix__
+CXXFLAGS += -D__unix__
```
Ouss4, line 25: Should we contribute `__NuttX__` to the upstream library instead?  We've got similar situations with other 3rd party software as well. [mbedTLS](https://github.com/apache/incubator-nuttx-apps/pull/566/files#diff-6043c39303f7b14be00e75ffcac170a85c3ee0b7369e7db85f4131ef85280c81R39) is one of them.

Ouss4:
Mixed Case style error comes from MQTT-C APIs.

mqtt/mqttc/Make.defs
v01d: Yeah, I think it is a simple change to request. In the meantime we can leave this definition.

v01d:
Thanks for submitting. I will test it on sim if I can (otherwise on ESP32).
BTW, we discussed putting the library under netutils on the list, could you move it there so that we not crowd to toplevel directory?
I think this will make the include path change to `netutils/mqtt.h`, unless we leave it in top-level include for compatibility.

mqtt/mqttc/Make.defs
xiaoxiang781216: ```__unix__``` is only used in implementation, it's better to move CLFAGS/CXXFLAGS into Makefile

mqtt/mqttc/Kconfig
```diff
@@ -0,0 +1,18 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config MQTT_MQTTC
```
xiaoxiang781216, line 6: should we add NETUTILS_ prefix?

xiaoxiang781216: change MQTT_MQTTC to MQTTC?

mqtt/mqttc/Make.defs
```diff
@@ -0,0 +1,32 @@
+############################################################################
+# apps/mqtt/mqttc/Make.defs
```
xiaoxiang781216, line 2: change path to netutils/mqttc?

examples/mqttc/Kconfig
```diff
@@ -0,0 +1,31 @@
...
+
+if EXAMPLES_MQTTC
+
+config EXAMPLES_MQTTC_PROGNAME
+	string "Program name"
+	default "mqttc_pub"
+
+config EXAMPLES_MQTTC_STACKSIZE
+	int "Task's stack size"
+	default 8192
```
xiaoxiang781216, line 21: Can we use the default stack size(2KB) here?

xiaoxiang781216 (approved):

mqtt/mqttc/Make.defs
xiaoxiang781216: @Ouss4 how about this comment?

examples/mqttc/Kconfig
Ouss4: I'm not sure if that's enough.  I remember that I had to adjust the stack size a while back when I had this example.  I'll do some tests later today and see if 2KB is enough.

mqtt/mqttc/Make.defs
Ouss4: @xiaoxiang781216 tried to move them (as is done in other places) but they don't seem to take any effect in Makefile.  Any thing to look for here?

xiaoxiang781216: I just create a patch to mainline: https://github.com/LiamBindle/MQTT-C/pull/131
We can remove these lines once the patch get merged, let's ignore it now.

Ouss4: But we'll still need to wait for the next release.  Or use master?

Ouss4:
I updated the defconfigs in the description to account for the new location.

xiaoxiang781216 (approved):

PR #637, opened 2021-03-29
testing/fstest: Add hash checking of file's name to avoid creating same-name file

donghengqaz:
## Summary

Add hash checking of file's name to avoid creating same-name file.

### DIFF
--- a/testing/fstest/fstest_main.c
+++ b/testing/fstest/fstest_main.c
@@ -91,6 +91,7 @@ struct fstest_filedesc_s
   bool failed;
   size_t len;
   uint32_t crc;
+  uint32_t hash;
 };
 
 /****************************************************************************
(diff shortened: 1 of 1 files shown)


testing/fstest/fstest_main.c
```diff
@@ -188,6 +189,29 @@ static inline char fstest_randchar(void)
...
+static bool fstest_checkexit(FAR struct fstest_filedesc_s *file)
+{
+  int i;
+  bool ret = false;
+
+  for (i = 0; i < CONFIG_TESTING_FSTEST_MAXOPEN; i++)
+    {
+      if (!g_files[i].deleted &&
+          &g_files[i] != file &&
+          g_files[i].hash == file->hash)
```
xiaoxiang781216, line 205: why not directly compare filename to avoid the potential collision?

donghengqaz: Using hash just want to speed up process when files and names are a little of more.

xiaoxiang781216 (approved):
OK, let's same what happen in the new code

PR #639, opened 2021-03-29
remove examples/dsptest

raiden00pl:
## Summary
Remove examples/dsptest.
This example doesn't make sense, tests cases are incomplete, and it's nightmare to maintain with even the slightest changes in libdsp.

Soon I'll provide another example using libdsp that implements a field-oriented control (FOC) for brushless motor.


raiden00pl:
https://github.com/apache/incubator-nuttx/pull/3224 must be merged first

Ouss4 (approved):

PR #640, opened 2021-03-30
examples/mqttc/README.md: Correct a sentence.

Ouss4:
## Summary
examples/mqttc/README.md: Correct a sentence.  "not" was missing.
## Impact
N/A
## Testing
N/A

PR #641, opened 2021-03-30
mqttc example: provide sane hostname default to avoid crash

v01d:
## Summary

Provide a sane hostname default (I used a public test broker for convenience). Otherwise it results in a crash since the
hostname variable was uninitialized. I also made slight improvements to output.

## Impact

Fix potential crash in example app.

## Testing

esp32-devkitc:wapi

### DIFF
--- a/examples/mqttc/README.md
+++ b/examples/mqttc/README.md
@@ -12,7 +12,7 @@ mosquitto is used:
 mosquitto&
 mosquitto_sub -t test
 ```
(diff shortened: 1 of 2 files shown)


v01d:
I've included the change in #640 . If it gets merged before this, it should be able to be rebased easily.

davids5:
> I've included the change in #640 . If it gets merged before this, it should be able to be rebased easily.

I do not see it in #640 - did you force push it yet?

PR #644, opened 2021-03-31
CI: cancel redundant workflows

v01d:
## Summary

Companion change to the one on main nuttx repo.

## Impact

CI

## Testing

CI

### DIFF
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -1,0 +1,17 @@
+name: Cancelling Duplicates
+on:
+  workflow_run:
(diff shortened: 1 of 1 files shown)


Ouss4 (approved):

Ouss4:
We should also get the same to https://github.com/apache/incubator-nuttx-testing.
With all that, we should close https://github.com/apache/incubator-nuttx/issues/2623?

v01d:
Ah yes, it is included in that PR.
Yes, I think we can close the issue when that is merged. It appears to be working OK.

xiaoxiang781216 (approved):

PR #647, opened 2021-04-03
nshlib: Disable mb, mh, and mw by default

swkim101:
## Summary

- Fix https://github.com/apache/incubator-nuttx/issues/3011

## Testing

image

### DIFF
--- a/nshlib/Kconfig
+++ b/nshlib/Kconfig
@@ -365,7 +365,7 @@ config NSH_DISABLE_LS
 
 config NSH_DISABLE_MB
 	bool "Disable mb"
-	default n
+	default y
 
 config NSH_DISABLE_MD5
 	bool "Disable md5"
@@ -402,7 +402,7 @@ config NSH_DISABLE_MKSMARTFS
 
 config NSH_DISABLE_MH
 	bool "Disable mh"
-	default n
+	default y
 
 config NSH_DISABLE_MODCMDS
 	bool "Disable modules commands (insmod, rmmod, lsmod)"
@@ -419,7 +419,7 @@ config NSH_DISABLE_MV
 
 config NSH_DISABLE_MW
 	bool "Disable mw"
-	default n
+	default y
 
 config NSH_DISABLE_NSFMOUNT
 	bool "Disable nfsmount"


xiaoxiang781216:
@swkim101 you need remove all CONFIG_NSH_DISABLE_M[B|H|W]=y from nuttx/boards/ to fix the error reported here:
https://github.com/apache/incubator-nuttx-apps/pull/647/checks?check_run_id=2258212861

Ouss4:
> @swkim101 you need remove all CONFIG_NSH_DISABLE_M[B|H|W]=y from nuttx/boards/ to fix the error reported here:
> https://github.com/apache/incubator-nuttx-apps/pull/647/checks?check_run_id=2258212861

There is this PR https://github.com/apache/incubator-nuttx/pull/3240, but it looks like it addresses only NSH configs.

btashton:
That PR does the opposite it need to remove all the explicit disables since this PR makes it the default

Ouss4:
> That PR does the opposite it need to remove all the explicit disables since this PR makes it the default

Ah you're right, I spoke too fast.

xiaoxiang781216:
The change is minor, and the dependence was merged. Let's merge it directly to save the build resource.

v01d:
@xiaoxiang781216 as @btashton mentioned, the Kconfig change is backwards to the intent of the PR. I don't think this should have been merged.

Also, I see quite a few issues such as:
```
< CONFIG_NSH_DISABLE_MB=y
< CONFIG_NSH_DISABLE_MH=y
63d60
< CONFIG_NSH_DISABLE_MW=y
```
in CI.

Let's try not to make a habit of skipping CI (it is now quite faster and less wasteful when pushing again) as we will loose unexpected errors.

xiaoxiang781216:
> @xiaoxiang781216 as @btashton mentioned, the Kconfig change is backwards to the intent of the PR. I don't think this should have been merged.
> 

This patch is needed to make the default option more safer.

> Also, I see quite a few issues such as:
> 
> ```
> < CONFIG_NSH_DISABLE_MB=y
> < CONFIG_NSH_DISABLE_MH=y
> 63d60
> < CONFIG_NSH_DISABLE_MW=y
> ```
> 

It's strange that I can't find these words in the latest nuttx when I want to fix the issue.

> in CI.
> 
> Let's try not to make a habit of skipping CI (it is now quite faster and less wasteful when pushing again) as we will loose unexpected errors.

Ok.

btashton:
The os repo PR was updated to just remove these entries instead of add them which is what we needed, any CI issue should have just been a matter of the order they were merged. 

But yeah generally I agree we should wait on CI. Sometimes I merge if it's obvious not going to have an issue and I'm trying to unblock things.

PR #649, opened 2021-04-04
OS Test:  Minor improvement to the getopt() test

patacongo:
## Summary

Add logic to assure that the getopt() functions parse exactly the correct number of parameters.  Previously if the test terminated early, the error would be undetected.  Also, prevents indexing past the end of the results array.

Fixes a coding error that was not causing a test failure.

Add some minimal level of testing for invalid options (this found a good bug in getopt_long()).

## Impact

Affects only the getopt() test cases of the OS test.

## Testing

Verified using an NSH simulator configuration set up to add apps/testing/ostest as a built-in command.

### DIFF
--- a/testing/ostest/getopt.c
+++ b/testing/ostest/getopt.c
@@ -68,6 +68,13 @@
     results[n].arg  = "Arg2"; \
  }
 
+#define SHORT_RESULT_X(n) \
+ { \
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):
LGTM.

PR #666, opened 2021-04-05
examples/lvgldemo: Make input options depend on CONFIG_INPUT

gustavonihei:
## Summary
This PR intends to make LVGL demo input-related options depend on CONFIG_INPUT.
The motivation is for easing the usage of the `lvgldemo` with displays that don't support touch input.

## Impact
Now `CONFIG_EXAMPLES_LVGLDEMO_CALIBRATE` will only be selected by default if `CONFIG_INPUT` is also selected. So, in practice, there will be no real impact, since this is the only useful combination.

## Testing
`sim:lvgl` successful build with `CONFIG_INPUT=n` and proper execution with disabled input, as expected.

### DIFF
--- a/examples/lvgldemo/Kconfig
+++ b/examples/lvgldemo/Kconfig
@@ -11,13 +11,6 @@ menuconfig EXAMPLES_LVGLDEMO
 
 if EXAMPLES_LVGLDEMO
 
-config EXAMPLES_LVGLDEMO_CALIBRATE
-	bool "Calibrate touchscreen"
-	default y
-	---help---
-		Calibrate touchscreen before demo start, but some touchscreen
-		don't need it, like capacitive touchscreen.
-
 config EXAMPLES_LVGLDEMO_BUFF_SIZE
 	int "Display buffer size (in line)"
 	default 20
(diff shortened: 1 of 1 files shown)


patacongo:
CONFIG_INPUT enables selection of any driver under drivers/input.  That is not just touchscreen, but all human input devices:  Mice, keyboards, joysticks, nintendo controllers, etc.

gustavonihei:
> CONFIG_INPUT enables selection of any driver under drivers/input. That is not just touchscreen, but all human input devices: Mice, keyboards, joysticks, game controllers, etc.
> 
> (There aren't actually any mice or keyboard driver there, but if we implemented PS/2 mouse or keyboard drivers, drivers/input is where they should go).
> 

You're right, depending on `INPUT` seems to generic and could be more specific.

> I think you could add something like:
> 
> ```
> config TOUCHSCREEN
>   default n
> ```
> 
> and then `select TOUCHSCREEN` if any touchscreen input device is selected. You might look also at how this is down for NX windows under graphics/Kconfig.

Looking at the current state of the `input/Kconfig`, it seems more suitable if `config TOUCHSCREEN` were an explicit `bool` type config. As a bonus it would also cope with the input drivers categorization into sub groups.
So the user would first enable the Touchscreen support and after that select the specific driver.
I've opened a PR with a proposal for this change:
https://github.com/apache/incubator-nuttx/pull/3443

gustavonihei:
I converted this PR to Draft.
Once https://github.com/apache/incubator-nuttx/pull/3443 is closed, I'll extended the changes to this PR (or drop if not applicable).

gustavonihei:
@patacongo I've submitted https://github.com/apache/incubator-nuttx/pull/3475 for implementing the suggested approach and also updated this PR. Could you please take another look?

gustavonihei:
@patacongo @btashton I'd like to have this simple PR merged. The main change has already been merged in https://github.com/apache/incubator-nuttx/pull/3482 and https://github.com/apache/incubator-nuttx/pull/3475. Could you please give your feedback?

btashton:
Apologies @gustavonihei I lost this in the shuffle. Looks good to me.

PR #668, opened 2021-04-06
testing/fstest: Add O_TRUNC flag in fstest_wrfile

xiaoxiang781216:
## Summary
to avoid the interference from the lasting run

## Impact
fstest

## Testing
Pass fstest


davids5:
@xiaoxiang781216 There is an unlink as part of the test, that should remove the file. From a testing perspective: Would it be better to report the existence of the file as an error or warning then truncate it?

xiaoxiang781216:
The problem is that the test forget to clean in the fail path.

PR #672, opened 2021-04-07
testing/ltp: split the ltp source include to 2 part

ttnie:
Signed-off-by: nietingting <nietingting@xiaomi.com>

## Summary
testing/ltp: split the ltp source include to 2 part

## Impact
no

## Testing
test on local

### DIFF
--- a/testing/ltp/Makefile
+++ b/testing/ltp/Makefile
@@ -83,22 +83,34 @@ BLACKSRCS += 35-2-buildonly.c
 
 MAINWORDS    += "main("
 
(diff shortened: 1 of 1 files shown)


btashton:
Why not just make the underlying build infrastructure support longer lists of files which is what I did.

v01d:
I agree, it is better to fix the underlying problem. This can occur again with other projects.

btashton:
I will submit a PR that uses a max args function split the calls for clean and archive. That way they are chunked on  20 files or something reasonable.

v01d:
Great, I was going to suggest that. That way we can still have some efficiency when this problem does not occur.

btashton:
This should be resolved with this https://github.com/apache/incubator-nuttx/pull/3510

xiaoxiang781216:
Ok, let's close this PR.

PR #677, opened 2021-04-09
Add FOC library

raiden00pl:
## Summary
Add FOC library.

Supported features:
- float and fixed16 operations
- helper interface to interact with the kernel-side FOC device
- general interface for FOC controller
- PI current/voltage controller
- 3-phase space vector modulation
- general interface for FOC model
- PMSM model
- general interface for phase angle data
- open-loop handler
- general interface for phase velocity data
- general interface for ramp (eg. velocity ramp, torque ramp)
- (WIP) CORDIC interface for hardware acceleration


xiaoxiang781216 (approved):
LGTM.

PR #684, opened 2021-04-13
interpreters/ficl: Fix compilation issue

acassis:
## Summary
A NuttX user told that ficl wasn't working as suggested in the
README.txt, then I found that the root cause was the order of the
Make.srcs inclusion inside the Makefile. I think we can do better:
we could automate the process to download and configure ficl.

## Impact
Users will get ficl support working on NuttX

## Testing
Compiled correctly

### DIFF
--- a/interpreters/ficl/Makefile
+++ b/interpreters/ficl/Makefile
@@ -39,6 +39,10 @@ include $(APPDIR)/Make.defs
 
 # Tools
 
+# Include the generated Make.srcs
+
+-include Make.srcs
+
 # Include paths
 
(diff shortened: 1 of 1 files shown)


acassis:
@xiaoxiang781216 @Ouss4 please review/merge

xiaoxiang781216 (approved):

xiaoxiang781216:
Maybe we can add a new config to enable ficl to avoid it break in the future.

acassis:
Thank you @xiaoxiang781216 ! Yes, the Interpreters are not correct, it is enabling everything! I will submit a fix

PR #691, opened 2021-04-22
addroute: inet_pton cannot accept slash notation. 

antmerlino:
## Summary
addroute: inet_pton cannot accept slash notation.  We must fixup the argument before calling.

## Impact
Fix addroute for cases using slash notation.

## Testing
Built and ran configuration.

### DIFF
--- a/nshlib/nsh_routecmds.c
+++ b/nshlib/nsh_routecmds.c
@@ -151,6 +151,7 @@ int cmd_addroute(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
   int shift;
   int sockfd;
   int ret;
+  FAR char *sptr;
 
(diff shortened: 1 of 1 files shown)


nshlib/nsh_routecmds.c
```diff
@@ -236,6 +236,17 @@ int cmd_addroute(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
       goto errout;
     }
 
+  /* We need to remove the slash notation before passing it to inet_pton */
+
+  if (shift > 0)
+    {
+      sptr = strchr(argv[1], '/');
```
xiaoxiang781216, line 243: @antmerlino sptr isn't defined in this function.

xiaoxiang781216 (approved):

PR #695, opened 2021-04-28
webclient: Implement non-blocking I/O

yamt:
## Summary
* webclient_perform
    * Add a new flag to use non-blocking mode (WEBCLIENT_FLAG_NON_BLOCKING)
    * Implement restarting

* Add a few associated API functions
    * webclient_get_poll_info: get the descriptor info for poll/select
    * webclient_abort: abort the operation (instead of restarting)

## Testing
tested with my local app.

### DIFF
--- a/include/netutils/webclient.h
+++ b/include/netutils/webclient.h
@@ -384,6 +384,17 @@ struct webclient_context
   size_t http_reason_len;
 
   struct wget_s *ws;
+
+#ifdef CONFIG_DEBUG_ASSERTIONS
+  enum webclient_context_state_e
+  {
+    WEBCLIENT_CONTEXT_STATE_UNINITIALIZED,
+    WEBCLIENT_CONTEXT_STATE_INITIALIZED,
+    WEBCLIENT_CONTEXT_STATE_IN_PROGRESS,
+    WEBCLIENT_CONTEXT_STATE_ABORTED,
+    WEBCLIENT_CONTEXT_STATE_DONE,
+  } state;
+#endif
 };
 
 struct webclient_poll_info
--- a/include/netutils/webclient.h
+++ b/include/netutils/webclient.h
@@ -235,6 +235,74 @@ struct webclient_tls_ops
                             FAR struct webclient_poll_info *info);
 };
 
+/* Note on webclient_client lifetime
+ *
+ * (uninitialized)
+ *      |
+ * webclient_set_defaults
+ *      |
+ *      v
+ *  INITIALIZED
+ *      |
+ *      |      IN-PROGRESS
+ *      |        |
+ *      |        +-------------+
+ *      |        |             |
+ *      |        |         webclient_abort
+ *      |        |             |
+ *      |        |             v
+ *      |        |         ABORTED
+ *      |        |
+ * webclient_perform
+ *      |
+ *      +---------------+
+ *      |               |
+ *      |             non-blocking mode,
+ *      |             returns -EAGAIN
+ *      |               |
+ *      v               v
+ *     DONE           IN-PROGRESS
+ *
+ * (uninitialized):
+ *   After the memory for webclient_context is allocated,
+ *   it should be initialized with webclient_set_defaults() before
+ *   feeding it to other functions taking a webclient_context.
+ *
+ *   webclient_abort() makes the state back to this state.
+ *   If the application wants to reuse the context for another request,
+ *   it should initialize it with webclient_set_defaults() again.
+ *
+ * INITIALIZED:
+ *   After calling webclient_set_defaults(), the application can set up
+ *   the request parameters by setting the struct fields before the first
+ *   call of webclient_perform().
+ *   E.g. url, method, buffers, and callbacks.
+ *
+ *   webclient_set_static_body() can only be used in this state.
+ *
+ * IN-PROGRESS:
+ *   This state only exists for the non-blocking mode.
+ *
+ *   webclient_get_poll_info() can only be used in this state.
+ *
+ * ABORTED:
+ *   The HTTP operation has been aborted by webclient_abort().
+ *
+ * DONE:
+ *   The HTTP operation has been completed. (Either successfully or not.)
+ *
+ *   The application can examine the struct fields to see the result.
+ *   E.g. http_status and http_reason.
+ *
+ * ABORTED, DONE:
+ *   The application can now dispose the resources associated to the
+ *   context.
+ *   E.g. buffers
+ *
+ *   If the application wants to reuse the context for another request,
+ *   it should initialize it with webclient_set_defaults() again.
+ */
+
 struct webclient_context
 {
   /* request parameters
--- a/include/netutils/webclient.h
+++ b/include/netutils/webclient.h
@@ -77,6 +77,39 @@
 #  endif
 #endif
 
+/* The following WEBCLIENT_FLAG_xxx constants are for
+ * webclient_context::flags.
+ */
+
+/* WEBCLIENT_FLAG_NON_BLOCKING tells webclient_perform() to
+ * use non-blocking I/O.
+ *
+ * If this flag is set, webclient_perform() returns -EAGAIN
+ * when it would otherwise block for network I/O. In that case,
+ * the application should either retry the operation later by calling
+ * webclient_perform() again, or abort it by calling webclient_abort().
+ * It can also use webclient_get_poll_info() to avoid busy-retrying.
+ *
+ * If this flag is set, it's the application's responsibility to
+ * implement a timeout.
+ *
+ * If the application specifies tls_ops, it's the application's
+ * responsibility to make the TLS implementation to use non-blocking I/O
+ * in addition to specifying this flag.
+ *
+ * Caveat: Even when this flag is set, the current implementation performs
+ * the name resolution in a blocking manner.
+ */
+
+#define	WEBCLIENT_FLAG_NON_BLOCKING	1U
+
+/* The following WEBCLIENT_FLAG_xxx constants are for
+ * webclient_poll_info::flags.
+ */
+
+#define	WEBCLIENT_POLL_INFO_WANT_READ	1U
+#define	WEBCLIENT_POLL_INFO_WANT_WRITE	2U
+
 /****************************************************************************
  * Public types
  ****************************************************************************/
@@ -181,6 +214,7 @@ typedef CODE int (*webclient_body_callback_t)(
     FAR void *ctx);
 
 struct webclient_tls_connection;
+struct webclient_poll_info;
 
 struct webclient_tls_ops
 {
@@ -196,6 +230,9 @@ struct webclient_tls_ops
                        FAR void *buf, size_t len);
   CODE int (*close)(FAR void *ctx,
                     FAR struct webclient_tls_connection *conn);
+  CODE int (*get_poll_info)(FAR void *ctx,
+                            FAR struct webclient_tls_connection *conn,
+                            FAR struct webclient_poll_info *info);
 };
 
 struct webclient_context
@@ -248,6 +285,7 @@ struct webclient_context
    *   tls_ops           - A vector to implement TLS operations.
    *                       NULL means no https support.
    *   tls_ctx           - A user pointer to be passed to tls_ops as it is.
+   *   flags             - OR'ed WEBCLIENT_FLAG_xxx values.
    */
 
   FAR char *buffer;
@@ -261,6 +299,7 @@ struct webclient_context
   FAR void *body_callback_arg;
   FAR const struct webclient_tls_ops *tls_ops;
   FAR void *tls_ctx;
+  unsigned int flags;
 
   /* results
    *
@@ -275,6 +314,16 @@ struct webclient_context
   unsigned int http_status;
   FAR char *http_reason;
   size_t http_reason_len;
+
+  struct wget_s *ws;
+};
+
+struct webclient_poll_info
+{
+  /* A file descriptor to wait for i/o. */
+
+  int fd;
+  unsigned int flags; /* OR'ed WEBCLIENT_POLL_INFO_xxx flags */
 };
 
 /****************************************************************************
@@ -329,9 +378,12 @@ int wget_post(FAR const char *url, FAR const char *posts, FAR char *buffer,
 
 void webclient_set_defaults(FAR struct webclient_context *ctx);
 int webclient_perform(FAR struct webclient_context *ctx);
+void webclient_abort(FAR struct webclient_context *ctx);
 void webclient_set_static_body(FAR struct webclient_context *ctx,
                                FAR const void *body,
                                size_t bodylen);
+int webclient_get_poll_info(FAR struct webclient_context *ctx,
+                            FAR struct webclient_poll_info *info);
 
 #undef EXTERN
 #ifdef __cplusplus
--- a/netutils/webclient/webclient.c
+++ b/netutils/webclient/webclient.c
@@ -131,6 +131,14 @@
 #define CONN_WANT_READ  WEBCLIENT_POLL_INFO_WANT_READ
 #define CONN_WANT_WRITE WEBCLIENT_POLL_INFO_WANT_WRITE
 
+#ifdef CONFIG_DEBUG_ASSERTIONS
+#define	_CHECK_STATE(ctx, s)	DEBUGASSERT((ctx)->state == (s))
+#define	_SET_STATE(ctx, s)		ctx->state = (s)
+#else
+#define	_CHECK_STATE(ctx, s)	do {} while (0)
+#define	_SET_STATE(ctx, s)		do {} while (0)
+#endif
+
 /****************************************************************************
  * Private Types
  ****************************************************************************/
@@ -775,6 +783,12 @@ int webclient_perform(FAR struct webclient_context *ctx)
   int len;
   int ret;
 
+#ifdef CONFIG_DEBUG_ASSERTIONS
+  DEBUGASSERT(ctx->state == WEBCLIENT_CONTEXT_STATE_INITIALIZED ||
+              (ctx->state == WEBCLIENT_CONTEXT_STATE_IN_PROGRESS &&
+               (ctx->flags & WEBCLIENT_FLAG_NON_BLOCKING) != 0));
+#endif
+
   /* Initialize the state structure */
 
   if (ctx->ws == NULL)
@@ -782,6 +796,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
       ws = calloc(1, sizeof(struct wget_s));
       if (!ws)
         {
+          _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_DONE);
           return -errno;
         }
 
@@ -797,6 +812,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
         {
           nwarn("WARNING: Malformed URL: %s\n", ctx->url);
           free(ws);
+          _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_DONE);
           return ret;
         }
 
@@ -827,6 +843,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
             {
               nerr("ERROR: unsupported scheme: %s\n", ws->scheme);
               free(ws);
+              _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_DONE);
               return -ENOTSUP;
             }
 
@@ -848,6 +865,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
                 {
                   nerr("ERROR: TLS on AF_LOCAL socket is not implemented\n");
                   free(ws);
+                  _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_DONE);
                   return -ENOTSUP;
                 }
 #endif
@@ -918,6 +936,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
                 {
                   nerr("ERROR: TLS on AF_LOCAL socket is not implemented\n");
                   free(ws);
+                  _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_DONE);
                   return -ENOTSUP;
                 }
 #endif
@@ -967,6 +986,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
 
                       nwarn("WARNING: Failed to resolve hostname\n");
                       free(ws);
+                      _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_DONE);
                       return -EHOSTUNREACH;
                     }
 
@@ -1284,6 +1304,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
   while (ws->state != WEBCLIENT_STATE_DONE);
 
   free(ws);
+  _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_DONE);
   return OK;
 
 errout_with_errno:
@@ -1295,6 +1316,7 @@ errout_with_errno:
           conn->flags |= CONN_WANT_WRITE;
         }
 
+      _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_IN_PROGRESS);
       return -EAGAIN;
     }
 
@@ -1304,6 +1326,7 @@ errout_with_errno:
     }
 
   free(ws);
+  _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_DONE);
   return ret;
 }
 
@@ -1326,6 +1349,9 @@ errout_with_errno:
 
 void webclient_abort(FAR struct webclient_context *ctx)
 {
+  _CHECK_STATE(ctx, WEBCLIENT_CONTEXT_STATE_IN_PROGRESS);
+  DEBUGASSERT((ctx->flags & WEBCLIENT_FLAG_NON_BLOCKING) != 0);
+
   struct wget_s *ws = ctx->ws;
 
   if (ws == NULL)
@@ -1341,6 +1367,7 @@ void webclient_abort(FAR struct webclient_context *ctx)
     }
 
   free(ws);
+  _SET_STATE(ctx, WEBCLIENT_CONTEXT_STATE_ABORTED);
 }
 
 /****************************************************************************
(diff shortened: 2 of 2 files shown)


include/netutils/webclient.h
```diff
@@ -165,6 +165,7 @@ typedef CODE int (*webclient_body_callback_t)(
     FAR void *ctx);
 
 struct webclient_tls_connection;
+struct webclient_poll_info;
```
gustavonihei, line 168:
```suggestion
struct webclient_tls_connection_s;
struct webclient_poll_info_s;
```
According to the coding style, structure names must end with the suffix `_s`.

include/netutils/webclient.h
```diff
@@ -235,6 +240,29 @@ struct webclient_context
...
+   *
+   * If the application specifies tls_ops, it's the application's
+   * responsibility to make the TLS implementation to use non-blocking I/O
+   * in addition to specifying this flag.
+   *
+   * Caveat: Even when this flag is set, the current implementation performs
+   * the name resolution in a blocking manner.
+   */
+
+#define	WEBCLIENT_FLAG_NON_BLOCKING	1U
```
gustavonihei, line 265: Placement of this preprocessor definition is not compliant to the coding standard.

include/netutils/webclient.h
```diff
@@ -249,6 +277,19 @@ struct webclient_context
...
+
+struct webclient_poll_info
+{
+  /* A file descriptor to wait for i/o. */
+
+  int fd;
+  unsigned int flags; /* OR'ed WEBCLIENT_POLL_INFO_xxx flags */
+
+#define	WEBCLIENT_POLL_INFO_WANT_READ	1U
+#define	WEBCLIENT_POLL_INFO_WANT_WRITE	2U
```
gustavonihei, line 292: Placement of these preprocessor definition is not compliant to the coding standard.

netutils/webclient/webclient.c
```diff
@@ -132,6 +129,37 @@
  * Private Types
  ****************************************************************************/
 
+enum webclient_state
```
gustavonihei, line 132:
```suggestion
enum webclient_state_e
```
According to the coding style, enumeration names must end with the suffix `_e`.

netutils/webclient/webclient.c
```diff
@@ -132,6 +129,37 @@
...
+    WEBCLIENT_STATE_SEND_REQUEST,
+    WEBCLIENT_STATE_SEND_REQUEST_BODY,
+    WEBCLIENT_STATE_STATUSLINE,
+    WEBCLIENT_STATE_HEADERS,
+    WEBCLIENT_STATE_DATA,
+    WEBCLIENT_STATE_CLOSE,
+    WEBCLIENT_STATE_DONE,
+  };
+
+struct conn
```
gustavonihei, line 146:
```suggestion
struct conn_s
```

netutils/webclient/webclient.c
```diff
@@ -132,6 +129,37 @@
...
+{
+  bool tls;
+
+  /* for !tls */
+
+  int sockfd;
+  unsigned int flags;
+
+#define CONN_WANT_READ  WEBCLIENT_POLL_INFO_WANT_READ
+#define CONN_WANT_WRITE WEBCLIENT_POLL_INFO_WANT_WRITE
```
gustavonihei, line 156: Placement of these preprocessor definitions is not compliant to the coding standard.

include/netutils/webclient.h
yamt: ok. but i will not mix style fixes into this PR.

include/netutils/webclient.h
yamt: fixed

include/netutils/webclient.h
yamt: fixed

netutils/webclient/webclient.c
yamt: fixed

netutils/webclient/webclient.c
yamt: fixed

yamt:
@gustavonihei i fixed some of style issues you pointed out. i'm not going to fix the rest of them in this PR because they are in the existing code, not this PR.

yamt:
the last push was a rebase with conflict resolutions.

netutils/webclient/webclient.c
```diff
@@ -674,225 +693,336 @@ int webclient_perform(FAR struct webclient_context *ctx)
...
-                      sizeof(server_un.sun_path));
-#if !defined(__NuttX__) && !defined(__linux__)
-              server_un.sun_len = SUN_LEN(&server_un);
+          if (conn->tls)
+            {
+#if defined(CONFIG_WEBCLIENT_NET_LOCAL)
+              if (ctx->unix_socket_path != NULL)
+                {
+                  nerr("ERROR: TLS on AF_LOCAL socket is not implemented\n");
+                  free(ws);
```
gustavonihei, line 766:
```suggestion
                  free(ws);
                  ctx->ws = NULL;
```
To avoid use-after-free errors, the freed pointer should be nullified.

netutils/webclient/webclient.c
```diff
@@ -1044,6 +1227,38 @@ int webclient_perform(FAR struct webclient_context *ctx)
...
+    }
+
+  if (ws->need_conn_close)
+    {
+      struct conn *conn = &ws->conn;
+
+      conn_close(ctx, conn);
+    }
+
+  free(ws);
```
gustavonihei, line 1259:
```suggestion
  free(ws);
  ctx->ws = NULL;
```
To avoid use-after-free errors, the freed pointer should be nullified.

netutils/webclient/webclient.c
```diff
@@ -674,225 +693,336 @@ int webclient_perform(FAR struct webclient_context *ctx)
...
-  if (ret != 0)
-    {
-      nwarn("WARNING: Malformed URL: %s\n", ctx->url);
-      free(ws);
-      return ret;
+      ret = parseurl(ctx->url, ws);
+      if (ret != 0)
+        {
+          nwarn("WARNING: Malformed URL: %s\n", ctx->url);
+          free(ws);
```
gustavonihei, line 715:
```suggestion
          free(ws);
          ctx->ws = NULL;
```
To avoid use-after-free errors, the freed pointer should be nullified.
```

File: netutils/webclient/webclient.c [hunk 10]
```diff
@@ -910,131 +1040,184 @@ int webclient_perform(FAR struct webclient_context *ctx)
...
+          else
+            {
+              ws->state = WEBCLIENT_STATE_DONE;
+            }
+        }
     }
-  while (redirected);
+  while (ws->state != WEBCLIENT_STATE_DONE);
 
   free(ws);
```
gustavonihei, line 1202:
```suggestion
  free(ws);
  ctx->ws = NULL;
```
To avoid use-after-free errors, the freed pointer should be nullified.
```

File: netutils/webclient/webclient.c [hunk 11]
```diff
@@ -910,131 +1040,184 @@ int webclient_perform(FAR struct webclient_context *ctx)
...
+
+      return -EAGAIN;
+    }
+
+  if (ws->need_conn_close)
+    {
+      conn_close(ctx, conn);
+    }
 
   free(ws);
```
gustavonihei, line 1222:
```suggestion
  free(ws);
  ctx->ws = NULL;
```
To avoid use-after-free errors, the freed pointer should be nullified.

netutils/webclient/webclient.c
yamt: are you saying that there is a bug?
or is it just about a coding style?

netutils/webclient/webclient.c
acassis: @yamt https://en.wikipedia.org/wiki/Dangling_pointer

netutils/webclient/webclient.c
gustavonihei: You may call it **defensive coding**.
Even after `free`, the `ctx->ws` pointer will still point to a memory address and since it is a member of a context structure it won't go out-of-scope. `ctx->ws` is now called a **dangling pointer**.
So every other check against `NULL` is simply useless if `ctx->ws` is not nullified, making the software prone to _user-after-free_ and _double-free_ kinds of errors.

yamt:
@gustavonihei 
@acassis 
i know some people like those construct.
but i tend to think it "loose", not "defensive".

gustavonihei:
> @gustavonihei
> @acassis
> i know some people like those construct.
> but i tend to think it "loose", not "defensive".

Sorry, I couldn't follow you. What do you mean by "loose"? Could you please expand on this?

yamt:
> > @gustavonihei
> > @acassis
> > i know some people like those construct.
> > but i tend to think it "loose", not "defensive".
> 
> Sorry, I couldn't follow you. What do you mean by "loose"? Could you please expand on this?

in my view:
assign only when necessary. fix bugs if any. -> strict coding
make unnecessary assignments to hide potential bugs. -> loose coding

yamt:
please don't get me wrong. being loose/defensive is not necessarily a bad thing. actually i myself sometimes do that. but in this particular case, i don't feel it's an improvement.

gustavonihei:
> in my view:
> assign only when necessary. fix bugs if any. -> strict coding
> make unnecessary assignments to hide potential bugs. -> loose coding

Sorry, I cannot agree with your view.
It is not unnecessary when the motivation for it is to prevent (not hide) a potential bug. And depending on how this opensource code is integrated into a product, this potential bug could escalate into a security vulnerability.

> but in this particular case, i don't feel it's an improvement.

It is not an improvement, this is the recommended method for avoiding **use-after-free** and **double-free** errors, both listed in the CWE:
https://cwe.mitre.org/data/definitions/415.html
https://cwe.mitre.org/data/definitions/416.html

yamt:
> > in my view:
> > assign only when necessary. fix bugs if any. -> strict coding
> > make unnecessary assignments to hide potential bugs. -> loose coding
> 
> Sorry, I cannot agree with your view.

no problem with having different views.
actually it's a good thing of a community.

> It is not unnecessary when the motivation for it is to prevent (not hide) a potential bug. And depending on how this opensource code is integrated into a product, this potential bug could escalate into a security vulnerability.

the "defensive" style can make it difficult to find such bugs.

i know some people like constructs like the following.
while i can understand the motivation, my honest impression is that using C was
a bad idea for those people in the first place.
```
#define FREE(a) (free(a); a = NULL)
#define MALLOC(a) calloc(1, a)
```

> 
> > but in this particular case, i don't feel it's an improvement.
> 
> It is not an improvement, this is the recommended method for avoiding **use-after-free** and **double-free** errors, both listed in the CWE:
> https://cwe.mitre.org/data/definitions/415.html
> https://cwe.mitre.org/data/definitions/416.html

i certainly don't recommend to apply the method blindly.
(i feel it "blindly" because you are suggesting to add "ctx->ws = NULL" even in a place where it's already known to be NULL.)

netutils/webclient/webclient.c
gustavonihei: This is one is not actually an issue since it `ctx->ws` is already NULL due to checking on line 696.

gustavonihei:
> the "defensive" style can make it difficult to find such bugs.
@yamt It doesn't make sense. If you are following the recommended mitigation methods and therefore preventing these bugs from happening, which bugs are you expecting to find then?

> i certainly don't recommend to apply the method blindly.
> (i feel it "blindly" because you are suggesting to add "ctx->ws = NULL" even in a place where it's already known to be NULL.)

Indeed, one of the issues I've raised is actually wrong and I acknowledge my mistake. However, my review process is far from being considered "blind", and that was an unfortunate statement of yours.
By the way, instead of undervaluing the reviewer's methods, please explain clearly on why an issue is not valid. If I am still wrong on the other raised issues, I'd like to learn from them.

yamt:
> > the "defensive" style can make it difficult to find such bugs.
> 
> @yamt It doesn't make sense. If you are following the recommended mitigation methods and therefore preventing these bugs from happening, which bugs are you expecting to find then?

unintended uses of values.

> 
> > i certainly don't recommend to apply the method blindly.
> > (i feel it "blindly" because you are suggesting to add "ctx->ws = NULL" even in a place where it's already known to be NULL.)
> 
> Indeed, one of the issues I've raised is actually wrong and I acknowledge my mistake. However, my review process is far from being considered "blind", and that was an unfortunate statement of yours.

sorry if you felt that way. i have no intention to attack you.
"blind" might have been an inappropriate word. maybe "mechanical"?

> By the way, instead of undervaluing the reviewer's methods, please explain clearly on why an issue is not valid. If I am still wrong on the other raised issues, I'd like to learn from them.

if the code is bug-free, the value doesn't matter.
they should not be used by anyone.
thus those NULLify should not be necessary.

if there is a bug (i guess there is :-) i prefer it fails loudly.
that's why i generally don't like the "defensive" methods.

yamt:
@gustavonihei 
i pushed a counter-proposal to the defensive style. https://github.com/apache/incubator-nuttx-apps/pull/695/commits/219356ce7d1b8a26140c98d303d493c6f04b19e9
(it has not been tested and thus i marked this PR a draft)

gustavonihei:
> > > the "defensive" style can make it difficult to find such bugs.
> > 
> > 
> > @yamt It doesn't make sense. If you are following the recommended mitigation methods and therefore preventing these bugs from happening, which bugs are you expecting to find then?
> 
> unintended uses of values.

Fine. So as long the recommended mitigation methods are being followed, you won't find them. Better safe than sorry.

> 
> > > i certainly don't recommend to apply the method blindly.
> > > (i feel it "blindly" because you are suggesting to add "ctx->ws = NULL" even in a place where it's already known to be NULL.)
> > 
> > 
> > Indeed, one of the issues I've raised is actually wrong and I acknowledge my mistake. However, my review process is far from being considered "blind", and that was an unfortunate statement of yours.
> 
> sorry if you felt that way. i have no intention to attack you.
> "blind" might have been an inappropriate word. maybe "mechanical"?
> 

I accept your apologies, but it is important for you to know that either by saying "blind" or "mechanical" you seem to disregard the fact that I've indeed evaluated the scenario before raising the issue. It is not like I had copied some text and pasted it. Otherwise I wouldn't even waste my time reviewing PRs.

> > By the way, instead of undervaluing the reviewer's methods, please explain clearly on why an issue is not valid. If I am still wrong on the other raised issues, I'd like to learn from them.
> 
> if the code is bug-free, the value doesn't matter.
> they should not be used by anyone.
> thus those NULLify should not be necessary.

The value wouldn't matter if it were stored in a local variable, but it is not the case. Since the lifetime of this dangling pointer is extended, the value does matter.
And all it takes for a software to change from "Bug-free" to "Vulnerable" is one single commit.
And if this single commit naively operates on this dangling pointer, the system will crash.

> 
> if there is a bug (i guess there is :-) i prefer it fails loudly.
> that's why i generally don't like the "defensive" methods.

NuttX is embedded in several types of products, from consumer electronics goods to industrial assets. For the majority of applications (if not all of them) the goal is to maximize the mean time between failures.
So, considering this, your preference for "failing loudly" and "fix bugs if any" does not seem much fit.

> i pushed a counter-proposal to the defensive style. 219356c

Good, this commit may be a proof that this code is not **currently** buggy. But that's not the issue I raised.
Now, here it is a simple unit test for making this potential bug evident:
https://godbolt.org/z/f4dzW8M44

yamt:
> > > > the "defensive" style can make it difficult to find such bugs.
> > > 
> > > 
> > > @yamt It doesn't make sense. If you are following the recommended mitigation methods and therefore preventing these bugs from happening, which bugs are you expecting to find then?
> > 
> > 
> > unintended uses of values.
> 
> Fine. So as long the recommended mitigation methods are being followed, you won't find them. Better safe than sorry.

i don't understand your logic here.

> 
> > > > i certainly don't recommend to apply the method blindly.
> > > > (i feel it "blindly" because you are suggesting to add "ctx->ws = NULL" even in a place where it's already known to be NULL.)
> > > 
> > > 
> > > Indeed, one of the issues I've raised is actually wrong and I acknowledge my mistake. However, my review process is far from being considered "blind", and that was an unfortunate statement of yours.
> > 
> > 
> > sorry if you felt that way. i have no intention to attack you.
> > "blind" might have been an inappropriate word. maybe "mechanical"?
> 
> I accept your apologies, but it is important for you to know that either by saying "blind" or "mechanical" you seem to disregard the fact that I've indeed evaluated the scenario before raising the issue. It is not like I had copied some text and pasted it. Otherwise I wouldn't even waste my time reviewing PRs.

ok.

> 
> > > By the way, instead of undervaluing the reviewer's methods, please explain clearly on why an issue is not valid. If I am still wrong on the other raised issues, I'd like to learn from them.
> > 
> > 
> > if the code is bug-free, the value doesn't matter.
> > they should not be used by anyone.
> > thus those NULLify should not be necessary.
> 
> The value wouldn't matter if it were stored in a local variable, but it is not the case. Since the lifetime of this dangling pointer is extended, the value does matter.
> And all it takes for a software to change from "Bug-free" to "Vulnerable" is one single commit.
> And if this single commit naively operates on this dangling pointer, the system will crash.

why doesn't a local variable matter?
with your logic, i think the value left on stack might be unintentionally used by a buggy software.

> 
> > if there is a bug (i guess there is :-) i prefer it fails loudly.
> > that's why i generally don't like the "defensive" methods.
> 
> NuttX is embedded in several types of products, from consumer electronics goods to industrial assets. For the majority of applications (if not all of them) the goal is to maximize the mean time between failures.
> So, considering this, your preference for "failing loudly" and "fix bugs if any" does not seem much fit.

i don't agree.
running with a wrong state is often worse than a crash. it depends.

> 
> > i pushed a counter-proposal to the defensive style. [219356c](https://github.com/apache/incubator-nuttx-apps/commit/219356ce7d1b8a26140c98d303d493c6f04b19e9)
> 
> Good, this commit may be a proof that this code is not **currently** buggy. But that's not the issue I raised.
> Now, here it is a simple unit test for making this potential bug evident:
> https://godbolt.org/z/f4dzW8M44

the unit test is just broken.
it doesn't seem to support your argument.
(btw, the website looks cool.)

gustavonihei:
> > > > By the way, instead of undervaluing the reviewer's methods, please explain clearly on why an issue is not valid. If I am still wrong on the other raised issues, I'd like to learn from them.
> > > 
> > > 
> > > if the code is bug-free, the value doesn't matter.
> > > they should not be used by anyone.
> > > thus those NULLify should not be necessary.
> > 
> > 
> > The value wouldn't matter if it were stored in a local variable, but it is not the case. Since the lifetime of this dangling pointer is extended, the value does matter.
> > And all it takes for a software to change from "Bug-free" to "Vulnerable" is one single commit.
> > And if this single commit naively operates on this dangling pointer, the system will crash.
> 
> why doesn't a local variable matter?
> with your logic, i think the value left on stack might be unintentionally used by a buggy software.
> 
Well, I was referring to local variables that are about to go out-of-scope, that's why I mentioned the extended lifetime for the other cases that the value does matter.
If a local variable goes out-of-scope (i.e. it's lifetime is over), it cannot be referenced anymore, so its pointer value is harmless. But I won't bother if you still want to nullify it.

> > > i pushed a counter-proposal to the defensive style. [219356c](https://github.com/apache/incubator-nuttx-apps/commit/219356ce7d1b8a26140c98d303d493c6f04b19e9)
> > 
> > 
> > Good, this commit may be a proof that this code is not **currently** buggy. But that's not the issue I raised.
> > Now, here it is a simple unit test for making this potential bug evident:
> > https://godbolt.org/z/f4dzW8M44
> 
> the unit test is just broken.
> it doesn't seem to support your argument.

Oh, really? Guess why it is broken?
```
Program stderr
free(): double free detected in tcache 2
```
I rest my case.

yamt:
> > > > if the code is bug-free, the value doesn't matter.
> > > > they should not be used by anyone.
> > > > thus those NULLify should not be necessary.
> > > 
> > > 
> > > The value wouldn't matter if it were stored in a local variable, but it is not the case. Since the lifetime of this dangling pointer is extended, the value does matter.
> > > And all it takes for a software to change from "Bug-free" to "Vulnerable" is one single commit.
> > > And if this single commit naively operates on this dangling pointer, the system will crash.
> > 
> > 
> > why doesn't a local variable matter?
> > with your logic, i think the value left on stack might be unintentionally used by a buggy software.
> 
> Well, I was referring to local variables that are about to go out-of-scope, that's why I mentioned the extended lifetime for the other cases where the value does matter.
> If a local variable goes out-of-scope (i.e. its lifetime is over), it cannot be referenced anymore, so its pointer value is harmless. But I won't bother if you still want to nullify it.

buggy software can happily access the value by reading an uninitialized variable.
anyway, i'm not a fan of nullify.

> 
> > > > i pushed a counter-proposal to the defensive style. [219356c](https://github.com/apache/incubator-nuttx-apps/commit/219356ce7d1b8a26140c98d303d493c6f04b19e9)
> > > 
> > > 
> > > Good, this commit may be a proof that this code is not **currently** buggy. But that's not the issue I raised.
> > > Now, here it is a simple unit test for making this potential bug evident:
> > > https://godbolt.org/z/f4dzW8M44
> > 
> > 
> > the unit test is just broken.
> > it doesn't seem to support your argument.
> 
> Guess why it is broken?
> 
> ```
> Program stderr
> free(): double free detected in tcache 2
> ```
> 
> I rest my case.

it's something like having a unit test like the following and claiming that free() has a problem.
it doesn't make much sense.
```
p = malloc(1);
free(p);
free(p);
```

gustavonihei:
> > Well, I was referring to local variables that are about to go out-of-scope, that's why I mentioned the extended lifetime for the other cases where the value does matter.
> > If a local variable goes out-of-scope (i.e. its lifetime is over), it cannot be referenced anymore, so its pointer value is harmless. But I won't bother if you still want to nullify it.
> 
> buggy software can happily access the value by reading an uninitialized variable.

We are not dealing with "uninitialized variables" here.
And no, the compiler will happily emit an error if the software tries to reference a variable that went out-of-scope.

> anyway, i'm not a fan of nullify.
> 

You are completely free to have your personal preferences, but for employing them on an open-source project you are subject to a review process.
And so far, on my judgement, your argumentation is not reasonable.

> > > > > i pushed a counter-proposal to the defensive style. [219356c](https://github.com/apache/incubator-nuttx-apps/commit/219356ce7d1b8a26140c98d303d493c6f04b19e9)
> > > > 
> > > > 
> > > > Good, this commit may be a proof that this code is not **currently** buggy. But that's not the issue I raised.
> > > > Now, here it is a simple unit test for making this potential bug evident:
> > > > https://godbolt.org/z/f4dzW8M44
> > > 
> > > 
> > > the unit test is just broken.
> > > it doesn't seem to support your argument.
> > 
> > 
> > Guess why it is broken?
> > ```
> > Program stderr
> > free(): double free detected in tcache 2
> > ```
> > 
> > 
> >   
> >     
> >     
> > 
> >     
> >     
> > 
> >   
> > 
> > 
> > I rest my case.
> 
> it's something like having a unit test like the following and claiming that free() has a problem.
> it doesn't make much sense.
> 
> ```
> p = malloc(1);
> free(p);
> free(p);
> ```

It's completely different.
If your comparison is based on the fact that I've called the `webclient_abort` twice, feel free to create another test that calls `webclient_perform` after a single `webclient_abort`.
The check here https://github.com/apache/incubator-nuttx-apps/blob/02bf1583d55cc8d0161866ee40c86cee9b97af46/netutils/webclient/webclient.c#L696 will not pass because you did not nullify `ctx->ws` and from https://github.com/apache/incubator-nuttx-apps/blob/02bf1583d55cc8d0161866ee40c86cee9b97af46/netutils/webclient/webclient.c#L723 onwards is undefined behavior because `ctx->ws` points to an already freed memory region.

gustavonihei (changes requested):
This PR introduces some **use-after-free** and **double-free** potential bugs that should be addressed before merging.

yamt:
> > it's something like having a unit test like the following and claiming that free() has a problem.
> > it doesn't make much sense.
> > ```
> > p = malloc(1);
> > free(p);
> > free(p);
> > ```
> 
> It's completely different.
> If your comparison is based on the fact that I've called the `webclient_abort` twice, feel free to create another test that calls `webclient_perform` after a single `webclient_abort`.

it's same.

webclient_perform after webclient_abort is also same.
they are all wrong use of apis.

you can invent more and more "problematic" sequences. eg. realloc after free.
but it doesn't mean these apis has problems.

netutils/webclient/webclient.c
mlyszczek: This is kinda pointless in situations like this. Dangling pointers are a thing if we are talking about global variables or library objects that are passed to library functions. But in this case? 1) return is called every time free() is called - thus ws is lost anyway 2) ws is accessed in this function directly via ws->, so even if you nullify this, you will still crash when calling ws->status or dereference ws in any way. Nullifying local variables before return is superflous.

yamt: @mlyszczek 
my understanding is that his concern is about ctx->ws, not ws.

mlyszczek: @yamt
Ah yes, didn't look close enough, my bad. Yeah I think in this particular case nulling ctx->ws seems like a good idea. But I didn't check it good enough to be sure if that is necessary.

yamt:
> > > it's something like having a unit test like the following and claiming that free() has a problem.
> > > it doesn't make much sense.
> > > ```
> > > p = malloc(1);
> > > free(p);
> > > free(p);
> > > ```
> > 
> > 
> > It's completely different.
> > If your comparison is based on the fact that I've called the `webclient_abort` twice, feel free to create another test that calls `webclient_perform` after a single `webclient_abort`.
> 
> it's same.
> 
> webclient_perform after webclient_abort is also same.
> they are all wrong use of apis.
> 
> you can invent more and more "problematic" sequences. eg. realloc after free.
> but it doesn't mean these apis has problems.

i added a dedicated comment block to explain webclient_context lifetime
to explain the expected sequence of API calls.

yamt:
the last push was a conflict resolution.

yamt:
the last push fixed doc and debug code a bit.

yamt:
> This PR introduces some **use-after-free** and **double-free** potential bugs that should be addressed before merging.

i don't agree and explained why not.

yamt:
@gustavonihei is the nullification thing the only thing you are unhappy with this PR?

gustavonihei:
> @gustavonihei is the nullification thing the only thing you are unhappy with this PR?

Yes.
Regarding the styling issues, I am fine with doing it in a separate PR.

gustavonihei:
@xiaoxiang781216 Could you please review this PR?
Also, please evaluate whether the issue I've raised is indeed valid. I am not comfortable in merging this PR due to that issue, so a third opinion might be helpful here.
This PR has been in limbo too long, let's try to move it forward.

yamt:
i marked this a draft because i haven't tested after the last rebase, which was not trivial.

yamt:
i marked this ready again as i tested the latest version with my local app.

xiaoxiang781216 (approved):

PR #699, opened 2021-05-02
netutils/webclient: Fixed socket descriptor leak.

tunefs:
## Summary
If a connection to the HTTP server cannot be established, the socket descriptor will leak.

## Testing
I tested my local app.


Ouss4 (approved):
LGTM.

PR #706, opened 2021-05-07
examples/hdc1008_demo: fix wrong printf parameter

diegoherranz:
## Summary
An extra "%d" that shouldn't be there.

## Impact
Only the hdc1008_demo app. Warning removed.

## Testing
Tested on custom board and it works the same way, but the warning is gone.

Thanks!

### DIFF
--- a/examples/hdc1008_demo/hdc1008_main.c
+++ b/examples/hdc1008_demo/hdc1008_main.c
@@ -1,5 +1,5 @@
 /****************************************************************************
- * examples/hdc1008_demo/hdc1008_main.c
+ * apps/examples/hdc1008_demo/hdc1008_main.c
  *
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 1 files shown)


diegoherranz:
Check / check (pull_request) complains:
> Error: /home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/examples/hdc1008_demo/hdc1008_main.c:2:1: error: 
Path relative to repository other than "nuttx" must begin with the root directory

Looking at other example apps, I'm not sure which is the proper way of doing that?

Thanks.

gustavonihei:
> Check / check (pull_request) complains:
> 
> > Error: /home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/examples/hdc1008_demo/hdc1008_main.c:2:1: error:
> > Path relative to repository other than "nuttx" must begin with the root directory
> 
> Looking at other example apps, I'm not sure which is the proper way of doing that?
> 
> Thanks.

Just prefix the path with "apps/"

diegoherranz:
Ah, I see. Done.

Thanks.

xiaoxiang781216 (approved):

PR #710, opened 2021-05-12
system/popen: Follow up task_spawn change from kernel

xiaoxiang781216:
## Summary
Depend on https://github.com/apache/incubator-nuttx/pull/3698

## Impact
Minor

## Testing
Pass CI


slorquet:
I disagree that the impact in minor.

This breaks ALL users that use their own apps that copied the builitin mechanism from the official app repo.

PR #742, opened 2021-06-01
Change NSH kill cmd to accept SIGTERM as default

RoCorbera:
## Summary
nsh kill cmd can be executed with no signal option.
SIGTERM is the default signal, as in unix kill command.
**nsh>** kill [-\<signal\>] \<pid\>

## Impact
Using SIGTERM as default signal makes nsh kill command to get closer to the unix version.
Given that NuttX allows to change the signal number of any signal and that default signal# for SIGTERM in NuttX is 12 (and 15 for POSIX signal), it is hard to guess what signal should be used to nicely kill a task.
Thus by using a default SIGTERM signal in kill command, the user just has to use the "kill \<pid\>" to always terminate a task.

## Testing
Tested in flat mode with CONFIG_SIG_DEFAULT=y as well as undefined.

### DIFF
--- a/nshlib/nsh_proccmds.c
+++ b/nshlib/nsh_proccmds.c
@@ -48,6 +48,7 @@
 #include <fcntl.h>
 #include <dirent.h>
 #include <errno.h>
+#include <signal.h>
 
 #include "nsh.h"
 #include "nsh_console.h"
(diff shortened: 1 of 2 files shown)


nshlib/nsh_proccmds.c
```diff
@@ -613,29 +614,54 @@ int cmd_ps(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
 #ifndef CONFIG_NSH_DISABLE_KILL
 int cmd_kill(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
 {
+/* kill will send SIGTERM to the task in case no signal is selected by
```
acassis, line 617: You need to space in the line bellow "{" to align correctly

nshlib/nsh_proccmds.c
```diff
@@ -613,29 +614,54 @@ int cmd_ps(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
...
     {
-      goto invalid_arg;
-    }
+      /* Check incoming parameters.
+       * The first parameter should be "-<signal>"
+       */
 
-  /* Extract the signal number */
+      ptr = argv[1];
+      if (*ptr != '-' || ptr[1] < '0' || ptr[1] > '9')
```
acassis, line 633: In the future to be more compatible with Unix and Linux, it should also accept the signal name, ie:
$ kill -SIGTERM 4297

RoCorbera: Yes, it will be a good improvement for a next PR.

nshlib/nsh_proccmds.c
RoCorbera: done. file passed checkpatch.sh

nshlib/nsh_proccmds.c
acassis: Nice! Please squash these 4 commits in a single commit.

xiaoxiang781216 (approved):

PR #743, opened 2021-06-01
add SIGTERM handling in leds example

RoCorbera:
## Summary
Based on PR #742 for having a kill \<pid\> that will send SIGTERM default signal to the task,
this example of a "daemon leds" will handle SIGTERM in order to gracefully terminate itself.

## Impact
It allows to restart this daemon after correctly terminating it. No need to reset the board to reestablish a good state.

## Testing
Tested in flat mode with CONFIG_SIG_DEFAULT=y as well as undefined.

### DIFF
--- a/examples/leds/leds_main.c
+++ b/examples/leds/leds_main.c
@@ -1,35 +1,20 @@
 /****************************************************************************
- * examples/leds/leds_main.c
+ * apps/examples/leds/leds_main.c
  *
- *   Copyright (C) 2016 Gregory Nutt. All rights reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
  *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
+ *   http://www.apache.org/licenses/LICENSE-2.0
  *
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in
- *    the documentation and/or other materials provided with the
- *    distribution.
- * 3. Neither the name NuttX nor the names of its contributors may be
- *    used to endorse or promote products derived from this software
- *    without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
- * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
- * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
- * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
- * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
- * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
- * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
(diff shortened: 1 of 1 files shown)


saramonteiro:
LGTM

examples/leds/leds_main.c
```diff
@@ -59,6 +45,26 @@ static bool g_led_daemon_started;
  * Private Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: sigterm_action
+ ****************************************************************************/
+
+void sigterm_action(int signo, siginfo_t *siginfo, void *arg)
```
xiaoxiang781216, line 52: add static

examples/leds/leds_main.c
```diff
@@ -170,16 +197,25 @@ static int led_daemon(int argc, char *argv[])
...
 
-errout_with_fd:
-  close(fd);
+  errout_with_fd:
+    close(fd);
 
-errout:
-  g_led_daemon_started = false;
+  errout:
+    g_led_daemon_started = false;
```
xiaoxiang781216, line 216: why change from line 176-182

RoCorbera: Actually there is no change except for indentation. 
The change is due to tools/checkpatch.sh message.

examples/leds/leds_main.c
RoCorbera: Yes, it will better. I'll change it and commit again.

examples/leds/leds_main.c
xiaoxiang781216: But the old code looks good, what warning do you see?

examples/leds/leds_main.c
xiaoxiang781216: Please merge the two patch into one, thanks.

examples/leds/leds_main.c
acassis: These lines:
  errout:
    g_led_daemon_started = false;

now are redundant, you can remove it

RoCorbera: In fact there were two odd error messages from checkpatch.sh.
One that related to the indentation of the very last "return" on lines 176-182.
The other saying that the path in line 2 (original source code) had to be related to "/" as it was not a nuttx path (or something like that). The point is that this error was related to a commentary block... 
In order to fix I had to change the commentaries to the APACHE message... even with a path, but after that the error was gone.

RoCorbera: > These lines:
> errout:
> g_led_daemon_started = false;
> 
> now are redundant, you can remove it

Not sure because there is a "goto errout;" in the code. The lines of code before that are executed when the while() loop exits due to change on g_led_daemon_started value change, in which case the function exits. Thus the errout: code will only be executed when an error occurs.

examples/leds/leds_main.c
```diff
@@ -170,16 +197,25 @@ static int led_daemon(int argc, char *argv[])
...
+      usleep(500 * 1000L);
+    }
+
+  /* treats signal termination of the task */
+
+  if (g_led_daemon_started == false)
+    {
+      /* task terminated by a SIGTERM */
+
+      exit(EXIT_SUCCESS);
```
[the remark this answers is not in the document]
RoCorbera: The idea was to force the execution of a exit(EXIT_SUCCESS) instead of exiting with error code (last line if normal flow).
POSIX exit() terminates the task, closes all fds, etc. I think it is safer.
I can remove the if(...) { ... }, but the not the exit(EXIT_SUCCESS) given it was correctly terminated by a SIGTERM at that point.

examples/leds/leds_main.c
acassis: right, you could just see the "errout:", although this redundant code is not an issue.

RoCorbera: Acctually it is related to exiting successfully after a normal termination with SIGTERM, or exiting with error code (goto errout: and other labels).

acassis: I see! You are right! It needs to kept in case of failure during the start process!

acassis: When exiting with SIGTERM it will be redundant, because g_led_daemon_started is already false (assuming you remove the "if (g_led_daemon_started == false)" block that I suggested.

RoCorbera: OK. I'll remove the "if (g_led ...)" 
I put it there actually  to make it explicit for others when reading the code, thus indeed redundant.
Most probably the compiler would optimize it anyway. I'll leave just the commentary for clearness.

examples/leds/leds_main.c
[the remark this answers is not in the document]
acassis: Ok, just remove the if() will be fine

xiaoxiang781216 (approved):

acassis:
Very good! Kudos!!!

PR #749, opened 2021-06-07
Add initial support libtommath and libtomcrypt libraries

btashton:
## Summary
This adds the initial support for the libtommath and libtomcrypt libraries.  These are setting the foundation for being able to run the dropbear ssh server/client on NuttX.

## Testing

Simple crypto test
```
nsh> echo "hello world" > /tmp/pt.txt
nsh> ltcrypt aes /tmp/pt.txt /tmp/ct.bin

Enter key: nsh> hexdump /tmp/ct.bin
/tmp/ct.bin at 00000000:
0000: e3 89 05 22 d7 c1 cb 54 b5 2a b2 40 a9 c1 b6 22 ..."...T.*.@..."
0010: d9 83 ac 24 09 79 ca 47 db 3c c9 9e             ...$.y.G.<..
nsh> ltcrypt -d aes /tmp/ct.bin /dev/console

Enter key:
hello world
nsh> 
```

```
nsh> tommath_test
Digit size 64 Bit 
Size of mp_digit: 8
Size of mp_word: 16
MP_DIGIT_BIT: 60
MP_PREC: 32
SEED: 0x11a

TEST feature_detection



TEST trivial_stuff



TEST mp_get_set_i32



TEST mp_get_set_i64



TEST mp_and



TEST mp_cnt_lsb



TEST mp_complement



TEST mp_decr



TEST mp_div_3

     9984
Passed div_3 testing

TEST mp_dr_reduce

31 digit modulus.... passed

TEST mp_pack_unpack



TEST mp_fread_fwrite



TEST mp_get_u32



TEST mp_get_u64

 r = 0x7fffffffffffffff i = 62

TEST mp_get_ul

 t = 0x7fffffffffffffff i = 62

TEST mp_log_u32



TEST mp_incr



TEST mp_invmod



TEST mp_is_square

   999



TEST mp_kronecker



TEST mp_montgomery_reduce

 digit size: 1000



TEST mp_root_u32



TEST mp_or



TEST mp_prime_is_prime

Testing mp_prime_is_prime() with Arnault's pseudoprime  803...901 

Testing mp_prime_is_prime() with certified prime 2^1119 + 53

Testing (    safe-prime):       127 bits    



TEST mp_prime_next_prime



TEST mp_prime_rand

Testing (not safe-prime):       127 bits    


TEST mp_rand



TEST mp_read_radix

 '123456' a == 123456, length = 7
 '-123456' a == -123456, length = 8
 '0' a == 0, length = 2


TEST mp_read_write_ubin

mp_to_ubin_size  113
mp_to_ubin len = 113


TEST mp_read_write_sbin

mp_to_sbin_size  114
mp_to_sbin len = 114


TEST mp_reduce_2k

  128 bits(1)(2)........

TEST mp_reduce_2k_l



TEST mp_set_double



TEST mp_signed_rsh



TEST mp_sqrt

   999

TEST mp_sqrtmod_prime



TEST mp_xor



TEST s_mp_balance_mul



TEST s_mp_karatsuba_mul



TEST s_mp_karatsuba_sqr



TEST s_mp_toom_mul



TEST s_mp_toom_sqr



Tests OK/NOP/FAIL: 42/0/0
```

### DIFF
--- a/crypto/libtomcrypt/Kconfig
+++ b/crypto/libtomcrypt/Kconfig
@@ -1,0 +1,57 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+menuconfig CRYPTO_LIBTOMCRYPT
+	bool "LibTomCrypt CrypographyLibrary"
+	default n
+	select MATH_LIBTOMMATH
+	---help---
+		LibTomCrypt is a fairly comprehensive, modular and portable
+		cryptographic toolkit that provides developers with a vast array
+		of well known published block ciphers, one-way hash functions,
+		chaining modes, pseudo-random number generators, public key
+		cryptography and a plethora of other routines. 
+
+if CRYPTO_LIBTOMCRYPT
+
+config LIBTOMCRYPT_VERSION
+	string "LibTomCrypt Version"
+	default "1.18.2"
+
+menuconfig LIBTOMCRYPT_DEMOS
+	bool "LibTomCrypt Library Demos"
+	default n
+	---help---
+		LibTomCrypt demo and test applications.
+
+if LIBTOMCRYPT_DEMOS
+
+config LIBTOMCRYPT_LTCRYPT
+        tristate "LibTomCrypt ltcrypt"
+        default n
+        ---help---
+                Demo encrypt/decrypt application
+
+if LIBTOMCRYPT_LTCRYPT
+
+config LIBTOMCRYPT_LTCRYPT_PROGNAME
+        string "ltcrypt program name"
+        default "ltcrypt"
+        ---help---
+                LibTomMath ltcrypt application name
+
+config LIBTOMCRYPT_LTCRYPT_PRIORITY
+        int "ltcrypt application priority"
+        default 100
+
+config LIBTOMCRYPT_LTCRYPT_STACKSIZE
+        int "ltcrypt application stack size"
+        default DEFAULT_TASK_STACKSIZE
+
+endif # LIBTOMCRYPT_LTCRYPT
+
+endif # LIBTOMCRYPT_DEMOS
+
+endif # CRYPTO_LIBTOMCRYPT
--- a/crypto/libtomcrypt/Make.defs
+++ b/crypto/libtomcrypt/Make.defs
@@ -1,0 +1,28 @@
+############################################################################
+# apps/crypto/libtomcrypt/Make.defs
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifeq ($(CONFIG_CRYPTO_LIBTOMCRYPT),y)
+CONFIGURED_APPS += $(APPDIR)/crypto/libtomcrypt
+
+# It allows `<tomcrypt.h>` import.
+
+CFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" $(APPDIR)/crypto/libtomcrypt/libtomcrypt/src/headers/}
+CXXFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" $(APPDIR)/crypto/libtomcrypt/libtomcrypt/src/headers/}
+endif
--- a/crypto/libtomcrypt/Makefile
+++ b/crypto/libtomcrypt/Makefile
@@ -1,0 +1,224 @@
+############################################################################
+# apps/crypto/libtomcrypt/Makefile
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+
+CSRCS += libtomcrypt/src/ciphers/aes/aes.c libtomcrypt/src/ciphers/aes/aes_tab.c libtomcrypt/src/ciphers/anubis.c libtomcrypt/src/ciphers/blowfish.c \
+libtomcrypt/src/ciphers/camellia.c libtomcrypt/src/ciphers/cast5.c libtomcrypt/src/ciphers/des.c libtomcrypt/src/ciphers/kasumi.c libtomcrypt/src/ciphers/khazad.c \
+libtomcrypt/src/ciphers/kseed.c libtomcrypt/src/ciphers/multi2.c libtomcrypt/src/ciphers/noekeon.c libtomcrypt/src/ciphers/rc2.c libtomcrypt/src/ciphers/rc5.c \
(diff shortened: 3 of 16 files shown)


crypto/libtomcrypt/Kconfig
```diff
@@ -0,0 +1,57 @@
...
+	string "LibTomCrypt Version"
+	default "1.18.2"
+
+menuconfig LIBTOMCRYPT_DEMOS
+	bool "LibTomCrypt Library Demos"
+	default n
+	---help---
+		LibTomCrypt demo and test applications.
+
+if LIBTOMCRYPT_DEMOS
```
xiaoxiang781216, line 29: can we remove LIBTOMCRYPT_DEMOS? LIBTOMCRYPT_LTCRYPT is enough.

crypto/libtomcrypt/Make.defs
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifeq ($(CONFIG_CRYPTO_LIBTOMCRYPT),y)
```
xiaoxiang781216, line 21: let other Make.defs
ifneq ($(CONFIG_CRYPTO_LIBTOMCRYPT),)

crypto/libtomcrypt/Makefile
```diff
@@ -0,0 +1,224 @@
...
+libtomcrypt/src/pk/rsa/rsa_import.c libtomcrypt/src/pk/rsa/rsa_import_pkcs8.c libtomcrypt/src/pk/rsa/rsa_import_x509.c \
+libtomcrypt/src/pk/rsa/rsa_make_key.c libtomcrypt/src/pk/rsa/rsa_set.c libtomcrypt/src/pk/rsa/rsa_sign_hash.c \
+libtomcrypt/src/pk/rsa/rsa_sign_saltlen_get.c libtomcrypt/src/pk/rsa/rsa_verify_hash.c libtomcrypt/src/prngs/chacha20.c libtomcrypt/src/prngs/fortuna.c \
+libtomcrypt/src/prngs/rc4.c libtomcrypt/src/prngs/rng_get_bytes.c libtomcrypt/src/prngs/rng_make_prng.c libtomcrypt/src/prngs/sober128.c \
+libtomcrypt/src/prngs/sprng.c libtomcrypt/src/prngs/yarrow.c libtomcrypt/src/stream/chacha/chacha_crypt.c libtomcrypt/src/stream/chacha/chacha_done.c \
+libtomcrypt/src/stream/chacha/chacha_ivctr32.c libtomcrypt/src/stream/chacha/chacha_ivctr64.c \
+libtomcrypt/src/stream/chacha/chacha_keystream.c libtomcrypt/src/stream/chacha/chacha_setup.c libtomcrypt/src/stream/chacha/chacha_test.c \
+libtomcrypt/src/stream/rc4/rc4_stream.c libtomcrypt/src/stream/rc4/rc4_test.c libtomcrypt/src/stream/sober128/sober128_stream.c \
+libtomcrypt/src/stream/sober128/sober128_test.c
```
xiaoxiang781216, line 178: can we use $(wildcard ...?

crypto/libtomcrypt/Makefile
```diff
@@ -0,0 +1,224 @@
...
+
+CFLAGS += -DLTC_SOURCE -DLTM_DESC
+CFLAGS += -Wno-deprecated-declarations
+
+#CFLAGS += -Wno-format -DLTC_SOURCE
```
xiaoxiang781216, line 183: remove

crypto/libtomcrypt/digit-bit.patch
```diff
@@ -0,0 +1,11 @@
+--- libtomcrypt/src/math/ltm_desc.c	2018-07-01 13:49:01.000000000 -0700
++++ libtomcrypt/src/math/ltm_desc.c.new	2021-06-06 22:42:12.877254274 -0700
```
xiaoxiang781216, line 2: should we upstream both patch to mainline

math/libtommath/Makefile
```diff
@@ -0,0 +1,112 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+CFLAGS += -Wno-format
+
+CSRCS = bn_cutoffs.c bn_deprecated.c bn_mp_2expt.c bn_mp_abs.c bn_mp_add.c bn_mp_add_d.c bn_mp_addmod.c \
```
xiaoxiang781216, line 25: $(wildcard ...)?

crypto/libtomcrypt/digit-bit.patch
btashton: I can upstream the flush, the other is already fixed is the development branch but there has not been a release created yet (there is a ticket discussing it).

acassis:
@btashton did you get Dropbear working on NuttX?

crypto/libtomcrypt/Makefile
btashton: It's a subset of the files which is why I went with the list which is also what the upstream project does.

btashton:
> @btashton did you get Dropbear working on NuttX?

I will provide a separate PR for it. But yes it is partially working in the sim.

crypto/libtomcrypt/Makefile
Ouss4: Maybe we can `wildcard` and then exclude files that we don't want to build (if it's a smaller list and easier to manage).

math/libtommath/Makefile
Ouss4: This is also a subset as far as I can tell.  Some files are conditioned.  I think here too it's better if we `wildcard` and then filter out those that we don't build or build conditionally.

Ouss4: This is also a subset as far as I can tell.  Some files are conditioned.  I think here too it's better if we `wildcard` and then filter out those that we don't build or build conditionally.

Ouss4: BTW, are the source extracted to this path directly?  No parent directory?

crypto/libtomcrypt/Kconfig
gustavonihei: I believe @btashton let it in a menuconfig in case the other demos from https://github.com/libtom/libtomcrypt/tree/develop/demos may be included in the future.

crypto/libtomcrypt/Makefile
gustavonihei: I think the benefits of not using a wildcard outweigh using it.
We'll never be surprised by unwanted files being compiled after a version bump. And if some file might be removed, the build will also fail.
Furthermore, each time a source file is added/removed to the project, the change to the file list will make it explicit.

I would just prefer to see these files in a single column, since it will be easier to distinguish between the components/algorithms. Or least some kind of grouping.

Ouss4: A version bump won't happen automatically. We will have to update the Kconfig and Makefile for that and do the necessary testing.  Which means, if there is a new file to add or ignore we should be aware of that.

The wildcard can still be somewhat controlled.  For repositories where source files are grouped in multiple directories, we can wildcard pick those that are less likely to change and leave those that we need to condition their build for instance.

btashton: I feel fairly strongly about this, it's so much easier to just grab the file list from upstream for the build than guess about what wildcards we should be using or excluding. It's literally a 30sec update. It's also what the upstream project is doing so it's a much safer pattern to be using.

math/libtommath/Makefile
btashton: There is a parent directory but we can just use the VPATH since it is flat.

xiaoxiang781216:
> > @btashton did you get Dropbear working on NuttX?
> 
> I will provide a separate PR for it. But yes it is partially working in the sim.

It's nice addition, what's feature miss from the porting?

math/libtommath/Makefile
```diff
@@ -0,0 +1,112 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+CFLAGS += -Wno-format
```
yamt, line 23: can you add a comment? eg. "these warnings are safe to ignore because..."

btashton:
Yes I will address the issues noted here this weekend.  Thanks for the review.

I also have not forgotten about that mbedtls one, I will bring that back up as well, but not sure if I will get to it this weekend.

acassis:
> Yes I will address the issues noted here this weekend. Thanks for the review.
> 
> I also have not forgotten about that mbedtls one, I will bring that back up as well, but not sure if I will get to it this weekend.
> 
> There is also some extra stuff we need to do around sending out notices for including cryptographic libraries in the project required by Apache so I will also take care of those.

@btashton are both libtomcrypt and libmcrypt (https://github.com/winlibs/libmcrypt) equivalent?

btashton:
> @btashton are both libtomcrypt and libmcrypt (https://github.com/winlibs/libmcrypt) equivalent?

Totally different.

acassis:
@btashton if you agree I will merge it and let you apply the improvements when you get the chance, is it ok for you?

btashton:
Unfortunate that this was merged with outstanding issues. This PR should not have been merged without the security export work completed...

PR #763, opened 2021-06-13
Call posix_spawn with filename as the first argv entry

xiaoxiang781216:
## Summary
pair with the kernel side change(https://github.com/apache/incubator-nuttx/pull/3916) to follow the standard defintion:
https://pubs.opengroup.org/onlinepubs/009695399/functions/posix_spawn.html

## Impact
More confirm to the standard


yamt (approved):
lgtm

patacongo (approved):

PR #765, opened 2021-06-14
netutils/webclient: fix the fd leak during connect(2) fail

xiaoxiang781216:
## Summary
Minor change in the fail path

## Impact
Fix the resource leak in webclient

## Testing
run webclient

### DIFF
--- a/netutils/webclient/webclient.c
+++ b/netutils/webclient/webclient.c
@@ -903,8 +903,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
       if (ret < 0)
         {
           nerr("ERROR: connect failed: %d\n", errno);
-          free(ws);
-          return ret;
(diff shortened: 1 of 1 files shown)


netutils/webclient/webclient.c
```diff
@@ -903,8 +903,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
       if (ret < 0)
         {
           nerr("ERROR: connect failed: %d\n", errno);
-          free(ws);
-          return ret;
+          goto errout_with_errno;
```
gustavonihei, line 906: Won't this end up closing the socket twice?
In case `connect` fails, line 899 already closes the socket.

xiaoxiang781216: The code is changed recently, look like the leak is fixed.

PR #767, opened 2021-06-14
examples: Ignore the default action if app call sigwait

xiaoxiang781216:
## Summary
Because the configured signo may have the default action(e.g. SIGPIPE),
and then will generate the bad side effect before the caller wakeup.

## Impact
Avoid the default action kill the program


acassis:
Hi @xiaoxiang781216 why ignore the signal action twice?

xiaoxiang781216:
Oh, the merge in our internal git looks wrong. Sorry, let's close this PR.

PR #768, opened 2021-06-14
nshlib/nsh_session: Handle the command arguments in any order

xiaoxiang781216:
## Summary
Many ssh server launch shell with -l option, but NuttX hasn't the session concept, so it's reasonable to ignore it for the better compatibility.

## Impact
Improve the compatiblity with ssh server

### DIFF
--- a/nshlib/nsh_session.c
+++ b/nshlib/nsh_session.c
@@ -70,6 +70,7 @@ int nsh_session(FAR struct console_stdio_s *pstate,
 {
   FAR struct nsh_vtbl_s *vtbl;
   int ret = EXIT_FAILURE;
+  int i;
 
   DEBUGASSERT(pstate);
   vtbl = &pstate->cn_vtbl;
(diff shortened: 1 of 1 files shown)


davids5:
Will this hide errors and mislead users?

xiaoxiang781216:
> Will this hide errors and mislead users?

Many options specify below are either minor adjustion or concept missed in NuttX:
https://pubs.opengroup.org/onlinepubs/9699919799.2013edition/utilities/sh.html
So it's better to run the command or script instead of failing directly.

xiaoxiang781216:
@davids5 do you think the relax is reasonable? Since most options can be safe to ignore.

davids5:
@xiaoxiang781216 - I can see this argument from both sides. It is reasonable to allow ports and not have to work hard to do the porting. On the other hand not error-ing out will mislead a user, if they assume it is supported.  

I would suggest you to ask the question of the group on dev and see what you get for feedback.

xiaoxiang781216:
@davids5 the code change to ignore -l option only, please review it.

davids5:
> I would suggest you to ask the question of the group on dev and see what you get for feedback.

Did you ask and get any feed back?

davids5:
@xiaoxiang781216 - Please close and open a new PR when you change everything (code and title).  Reference the old PR if need be.

xiaoxiang781216:
> > I would suggest you to ask the question of the group on dev and see what you get for feedback.
> 
> Did you ask and get any feed back?

This is minor change now, why need to vote on dev list? anyone can add the comment on PR directly.

> @xiaoxiang781216 - Please close and open a new PR when you change everything (code and title). Reference the old PR if need be.

The old PR doesn't make sense anymore, we don't reference the stale information. All info this PR already update to reflect the change.

xiaoxiang781216:
@davids5 do you have any more concern with this PR?

PR #770, opened 2021-06-14
system/ping[6]: correct the ping return value

xiaoxiang781216:
## Impact
Return EXIT_FAILURE in the error case

### DIFF
--- a/system/ping/ping.c
+++ b/system/ping/ping.c
@@ -42,6 +42,15 @@
 #define ICMP_POLL_DELAY    1000  /* 1 second in milliseconds */
 
 /****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+struct ping_priv_s
+{
+  int code;                      /* Notice code ICMP_I/E/W_XXX */
+};
+
+/****************************************************************************
  * Private Functions
  ****************************************************************************/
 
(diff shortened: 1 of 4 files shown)


system/ping6/ping6.c
```diff
@@ -92,8 +101,11 @@ static void show_usage(FAR const char *progname, int exitcode)
 
 static void ping6_result(FAR const struct ping6_result_s *result)
 {
+  FAR struct ping6_priv_s *priv = result->info->priv;
   char strbuffer[INET6_ADDRSTRLEN];
 
+  priv->code = result->code;
```
yamt, line 107: i guess this way many of errors will be overwritten by ICMP_I_FINISH.
is it intended?

xiaoxiang781216: Yes, you are right. Please review the updated version again.

system/ping/ping.c
```diff
@@ -224,6 +241,8 @@ int main(int argc, FAR char *argv[])
   info.delay     = ICMP_POLL_DELAY;
   info.timeout   = ICMP_POLL_DELAY;
   info.callback  = ping_result;
+  info.priv      = &priv;
+  priv.code      = ICMP_I_BEGIN;
```
yamt, line 245: i feel it's cleaner to introduce a dedicated constant than using ICMP_I_BEGIN

yamt (approved):

system/ping/ping.c
xiaoxiang781216: Fix

PR #771, opened 2021-06-14
apps/system: fix command history behavior

xiaoxiang781216:
## Summary
Fixing "last command" misbehavior when one command in history is a substring of another.

## Impact
Fix the minor corner case


system/readline/readline_common.c
```diff
@@ -601,6 +601,8 @@ ssize_t readline_common(FAR struct rl_common_s *vtbl, FAR char *buf,
                           buf[nch++] = g_cmdhist.buf[idx][i];
                           RL_PUTC(vtbl, g_cmdhist.buf[idx][i]);
                         }
+
+                      buf[nch] = '\0';
```
normanr, line 605: This should probably be just before the strcmp, otherwise if you run `foobar` and `fooba`, then the latter doesn't get saved.

PR #775, opened 2021-06-14
testing/fstest: Add O_TRUNC flag in fstest_wrfile

xiaoxiang781216:
## Summary
to avoid the interference from the lasting run


acassis:
@xiaoxiang781216 what kind of interference was you getting? Is there some way to force the issue to happen?

xiaoxiang781216:
Run the test twice if the first run fail, the second run will see the partial file generated by the first run.

PR #778, opened 2021-06-15
example/posix_spawn: Ensure argv has filename and NULL terminator

xiaoxiang781216:
## Summary
Report here: https://github.com/apache/incubator-nuttx/pull/3928

## Impact
Fix the non-standard usage.

### DIFF
--- a/examples/posix_spawn/spawn_main.c
+++ b/examples/posix_spawn/spawn_main.c
@@ -117,14 +117,18 @@ static unsigned int g_mmstep;     /* Memory Usage at beginning of test step */
 static const char delimiter[] =
   "**************************************"
   "**************************************";
-static const char g_redirect[] = "redirect";
 static const char g_data[]     = "testdata.txt";
 
 static char fullpath[128];
 
-static char * const g_argv[5] =
+static char * const g_hello_argv[5] =
 {
   "hello", "Argument 1", "Argument 2", "Argument 3", NULL
+};
+
+static char * const g_redirect_argv[2] =
+{
+  "redirect", NULL
 };
 
 /****************************************************************************
@@ -270,7 +274,7 @@ int main(int argc, FAR char *argv[])
    * this program from the others.
    */
 
-  testheader(g_argv[0]);
+  testheader(g_hello_argv[0]);
 
   /* Initialize the attributes file actions structure */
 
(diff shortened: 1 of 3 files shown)


examples/posix_spawn/spawn_main.c
```diff
@@ -117,16 +117,20 @@ static unsigned int g_mmstep;     /* Memory Usage at beginning of test step */
...
 
 static char fullpath[128];
 
-static char * const g_argv[5] =
+static char * const g_hello_argv[5] =
 {
   "hello", "Argument 1", "Argument 2", "Argument 3", NULL
 };
 
+static char * const g_redirect_argv[5] =
```
patacongo, line 129:
```suggestion
static char * const g_redirect_argv[2] =
```

patacongo:
Won't your change will fail in these cases:

examples/elf/elf_main.c:

    args[0] = NULL;
    ret = exec(filename, args, g_elf_exports, g_elf_nexports);

examples/nxflat/nxflat_main.c:

    args[0] = NULL;
    ret = exec(filename, args, g_nxflat_exports, g_nxflat_nexports);

netutils/thttpd/thttpd_cgi.c:

    #ifdef CONFIG_THTTPD_NXFLAT
      child = exec(hc->expnfilename, (FAR char * const *)argp,
                   g_thttpdsymtab, g_thttpdnsymbols);
    #else
      child = exec(hc->expnfilename, (FAR char * const *)argp, NULL, 0);
    #endif

In the above, argv[0] is NOT the file name but argv will not be NULL either.

examples/posix_spawn/spawn_main.c
xiaoxiang781216: Done.

xiaoxiang781216:
> Won't your change will fail in these cases:
> 
> examples/elf/elf_main.c:
> 
> ```
> args[0] = NULL;
> ret = exec(filename, args, g_elf_exports, g_elf_nexports);
> ```
> 

Done.

> examples/nxflat/nxflat_main.c:
> 
> ```
> args[0] = NULL;
> ret = exec(filename, args, g_nxflat_exports, g_nxflat_nexports);
> ```
>

Done.
 
> netutils/thttpd/thttpd_cgi.c:
> 
> ```
> #ifdef CONFIG_THTTPD_NXFLAT
>   child = exec(hc->expnfilename, (FAR char * const *)argp,
>                g_thttpdsymtab, g_thttpdnsymbols);
> #else
>   child = exec(hc->expnfilename, (FAR char * const *)argp, NULL, 0);
> #endif
> ```
> 

argp contain filename arleady, we don't need change:
https://github.com/apache/incubator-nuttx-apps/blob/d6d458c60be20d1cc4954552f66989330d63a0ee/netutils/thttpd/thttpd_cgi.c#L323-L331

> In the above, argv[0] is NOT the file name but argv will not be NULL either.
> 
> Hmm.. the first two should be okay because you check (argv && argv[0]), but I am not sure about the last one. That check, (argv && argv[0]), is NOT foolproof.

patacongo (approved):

PR #779, opened 2021-06-15
examples/qencoder: fix nxstyle warning and errors

michallenc:
## Summary
FIx compile error discovered in automatic control to [NuttX PR 3929](https://github.com/apache/incubator-nuttx/pull/3929) (unused label).

### DIFF
--- a/examples/qencoder/qe_main.c
+++ b/examples/qencoder/qe_main.c
@@ -242,7 +242,7 @@ int main(int argc, FAR char *argv[])
     {
       printf("qe_main: open %s failed: %d\n", g_qeexample.devpath, errno);
       exitval = EXIT_FAILURE;
-      goto errout_with_dev;
+      goto errout;
     }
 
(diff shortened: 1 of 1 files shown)


examples/qencoder/qe_main.c
```diff
@@ -242,7 +242,7 @@ int main(int argc, FAR char *argv[])
     {
       printf("qe_main: open %s failed: %d\n", g_qeexample.devpath, errno);
       exitval = EXIT_FAILURE;
-      goto errout_with_dev;
+      goto errout;
```
btashton, line 245: I think this line is actually the only change needed. The open fails we should just take the error path with no close. In the other cases we should take the path with the close.

btashton (changes requested):

examples/qencoder/qe_main.c
michallenc: Done, thanks.

PR #784, opened 2021-06-18
examples: Adds Timer gpout example

saramonteiro:
## Summary

This PR adds one example.
This example uses the timer interrupt to periodically
change the state of a digital output.
The digital output may be a relay, a led or anything else.
This example can be very useful to validate timer drivers
by using a logic analyzer connected to the digital output.
This example, mainly differs from the timer example because it
waits on a sigwaitinfo() instead of using a signal handler.
This approach ensures a deterministic wake-up time when
the signal occurs.

## Impact
N/A

## Testing
Tested and published a video in NuttX Brasil channel running this example with a LA:
https://youtu.be/v4bSGF3b3v4

### DIFF
--- a/examples/timer_gpout/Make.defs
+++ b/examples/timer_gpout/Make.defs
@@ -1,0 +1,23 @@
+############################################################################
+# apps/examples/timer_gpout/Make.defs
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifeq ($(CONFIG_EXAMPLES_TIMER_GPOUT),y)
+CONFIGURED_APPS += $(APPDIR)/examples/timer_gpout
+endif
--- a/examples/timer_gpout/Makefile
+++ b/examples/timer_gpout/Makefile
@@ -1,0 +1,34 @@
+############################################################################
+# apps/examples/timer_gpout/Makefile
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+# Timer built-in application info
+
+PROGNAME  = $(CONFIG_EXAMPLES_TIMER_GPOUT_PROGNAME)
+PRIORITY  = $(CONFIG_EXAMPLES_TIMER_GPOUT_PRIORITY)
+STACKSIZE = $(CONFIG_EXAMPLES_TIMER_GPOUT_STACKSIZE)
+MODULE    = $(CONFIG_EXAMPLES_TIMER_GPOUT)
+
+# Timer example
+
+MAINSRC = timer_gpout_main.c
+
+include $(APPDIR)/Application.mk
--- a/examples/timer_gpout/timer_gpout_main.c
+++ b/examples/timer_gpout/timer_gpout_main.c
@@ -1,0 +1,306 @@
+/****************************************************************************
+ * apps/examples/timer_gpout/timer_gpout_main.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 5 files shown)


examples/timer_gpout/Make.defs
```diff
@@ -0,0 +1,39 @@
...
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+# OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+# AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+############################################################################
```
patacongo, line 35: Shouldn't this be an ASF Apache 2.0 header?

examples/timer_gpout/Makefile
```diff
@@ -0,0 +1,49 @@
...
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+# OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+# AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+############################################################################
+
```
patacongo, line 35: Shouldn't this be an ASF Apache 2.0 header?

examples/timer_gpout/timer_gpout_main.c
```diff
@@ -0,0 +1,306 @@
...
+ * Private Types
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static char devtim_g[DEVNAME_SIZE];
+static char devgpout_g[DEVNAME_SIZE];
+static bool timer_gpout_daemon_started_g = false;
```
patacongo, line 53: The NuttX coding standard requires that all variables defined at global scope must begin with g_.  Putting _g at the end violates the coding standard.  These need to be changed to g_devtim, g_devgpout, and g_timer_gpout_daemon_started.

examples/timer_gpout/timer_gpout_main.c
```diff
@@ -0,0 +1,306 @@
...
+        {
+          int errcode = errno;
+          printf("timer_gpout_daemon: ERROR: sigwaitinfo() failed: %d\n",
+                 errcode);
+          goto errout;
+        }
+
+      /* Change the gpout state */
+
+      state = ~state;
```
patacongo, line 219:
```suggestion
      state = !state;
```
The tilde is a bit operation of a boolean value.  It will toggle from 0 (false) and 0xffffffff (invalid boolean value).  The exclamation point is the correct operator for a boolean and will toggle from 0 (false) to 1 (true).

examples/timer_gpout/Make.defs
saramonteiro: Yes.
Solved.
Thanks.

examples/timer_gpout/Makefile
saramonteiro: Yes.
Solved.
Thanks.

examples/timer_gpout/timer_gpout_main.c
saramonteiro: I inverted. Fixed.

examples/timer_gpout/timer_gpout_main.c
saramonteiro: Ok.
Fixed.

saramonteiro:
@patacongo I fixed the found issues. PTAL.
Note: I'll squash the commits once it's approved.
Do not merge without it.

patacongo (approved):
Thanks!

patacongo:
A question.  This is not essential to the PR, but is it really necessary to limit the number of daemons waiting for the signal?  Is that a limitation of the timer? What would happen if you removed g_timer_gpout_daemon_started so that there could be multiple active timers?

saramonteiro:
> A question. This is not essential to the PR, but is it really necessary to limit the number of daemons waiting for the signal? Is that a limitation of the timer? What would happen if you removed g_timer_gpout_daemon_started so that there could be multiple active timers?

This a good question. Although I haven't tested, I infer that we can't have multiple deamons with one timer. Because each time one daemon was launched it would rewrite the `timer_notify_s` structure which is used within the timer callback to send a signal for a specific process id (task).  As a consequence, I believe we should see the priors daemons starve. To achieve this purpose, I believe we would have to create a Task Group and set up signaling for Multi-threaded Task Group. (What's is actually a good idea for a next example...).
But I see no reasons we shouldn't have multiple daemons (each one associated with a specific timer).
I'll perform both test and I come back here to tell you.

patacongo:
I think I should go ahead a merge this version.  It would be a good extension to testing to prove that there is no issue with multiple concurrent timers.  It would be a bug in the driver if it could not support more than one client task.

But, I think you can extend this testing in any way that you wish in the future.  Let's merge this well verified PR as it is now.

patacongo:
The driver really should handle multiple clients like all other drivers.  That should be a requirement of all drivers so I would personally consider that limitation a bug.

For a single timer, this would be handled by keeping a list of relative timer delays.  This is done, for example, for the wdog dog timers.  See https://github.com/apache/incubator-nuttx/blob/master/sched/wdog/wd_start.c#L226

patacongo:
> The driver really should handle multiple clients like most other drivers. That should be a requirement of all drivers, if possible, so I would personally consider that limitation a bug in this case.

Ha ha... looks like I wrote the driver/timer/timer.c driver way back in 2012.  So I guess if it is a bug, it is my bug.

saramonteiro:
@patacongo I performed some tests and here are the tests with the results:

**1st Test:**
Simply removed the part that checks for `g_timer_gpout_daemon_started`.
Result:
Simply doesn't work, because when it calls the timer start ioctl, it returns EBUSY.

**2nd Test**
Kept the changes from the 1st test, but after opening timer driver, I called the status ioctl to use the status flags and check the TCFLAGS_ACTIVE. Then I checked If the timer was running, in case it was, I set it to perform only the notification ioctl and not to perform settimeout and start.
Result:
It works, but as I expected, `timer_notify_s` structure was subscribed and the previous daemon starved. Take a look:
image
Conclusion for this: For one timer, we should have only one daemon.

**3rd Test**
I kept _only_ the changes from the 1st test and ran multiple daemons. Each one using one timer and a specific gpout. Ex:
`timer_gpout -t /dev/timer0 -g /dev/gpout0`
`timer_gpout -t /dev/timer1 -g /dev/gpout1`
Result:
It works perfectly: 
image

PR #785, opened 2021-06-22
canutils: Fix error: unknown type name 'fd_set'

xiaoxiang781216:
## Summary
Replace sys/time.h with sys/select.h.
Please ignore nxstyle warning since canutils keep the original coding style to help merging.


xiaoxiang781216:
Close since the standard require to include sys/time.h:
https://pubs.opengroup.org/onlinepubs/7908799/xsh/select.html

PR #793, opened 2021-07-02
Update testing/mm testcase, add stress test

GUIDINGLI:
## Summary
Update testing/mm testcase, add stress test

### DIFF
--- a/testing/mm/Kconfig
+++ b/testing/mm/Kconfig
@@ -20,7 +20,7 @@ config TESTING_MM_PROGNAME
 
 config TESTING_MM_PRIORITY
 	int "Task priority"
-	default 100
+	default 250
 
 config TESTING_MM_STACKSIZE
 	int "Stack size"
--- a/testing/mm/mm_main.c
+++ b/testing/mm/mm_main.c
@@ -22,6 +22,7 @@
  * Included Files
  ****************************************************************************/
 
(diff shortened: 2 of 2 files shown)


testing/mm/Kconfig
```diff
@@ -20,7 +20,7 @@ config TESTING_MM_PROGNAME
 
 config TESTING_MM_PRIORITY
 	int "Task priority"
-	default 100
+	default 250
```
acassis, line 23: Why do you need to increase the priority? Does it fail when using priority 100?

xiaoxiang781216:
Fix here: https://github.com/apache/incubator-nuttx/pull/4051

testing/mm/Kconfig
xiaoxiang781216: We run memtest in the background. The high prirority could avoid the bad high prority task block memtest.

acassis: Ok! BTW it still be possible to have higher priority tasks (251-255)

xiaoxiang781216: Yes, but it's rare:).

yamt (commented):
why do you add this to testing/mm, rather then introducing a new command?
do you plan to add more and more later?

xiaoxiang781216:
Let's close this PR.

PR #796, opened 2021-07-02
apps: add variable to identify main shell

GUIDINGLI:
## Summary

apps: add variable to identify main shell

This should be with
https://github.com/apache/incubator-nuttx/pull/4042

Root cause:

These modify for shell ctrl + C usefulness when system() execute.

1. execute 'cu' in shell front end, uart record 'cu' pid for later killing.
2. one thread use system() to do sth anything, uart record new pid for later killing.
3. ctrl +c can't kill 'cu' thread

FIX:
add param to nsh_newconsole() to identify the main shell & system cmd shell.

### DIFF
--- a/nshlib/nsh_altconsole.c
+++ b/nshlib/nsh_altconsole.c
@@ -245,7 +245,7 @@ static int nsh_wait_inputdev(FAR struct console_stdio_s *pstate,
 
 int nsh_consolemain(int argc, FAR char *argv[])
 {
-  FAR struct console_stdio_s *pstate = nsh_newconsole();
+  FAR struct console_stdio_s *pstate = nsh_newconsole(true);
   FAR const char *msg;
   int ret;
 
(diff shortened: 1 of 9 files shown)


davids5:
@GUIDINGLI @xiaoxiang781216  is incubator-nuttx/boards/z16/z16f/z16f2800100zcog/tools/zneo-zdsii-5_0_1-variadic-func-fix.patch tested?

xiaoxiang781216:
zneo-zdsii-5_0_1-variadic-func-fix.patch add voutput callback, this patch add control tty. The chagne is unrelated and orthogonal.

davids5:
Is that target built? Does the patch still apply cleanly?

xiaoxiang781216:
Yes, it still can be applied correctly.

davids5 (approved):

PR #800, opened 2021-07-05
Revert "apps/Application.mk: support compile same prefix file in one folder"

yamt:
Reverts apache/incubator-nuttx-apps#795

because:

* it broke Make.dep
* the use case is not clear

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -73,15 +73,15 @@ PROGNAME := $(shell echo $(PROGNAME))
 RASRCS = $(filter %.s,$(ASRCS))
 CASRCS = $(filter %.S,$(ASRCS))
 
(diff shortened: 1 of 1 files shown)


yamt:
@GUIDINGLI @xiaoxiang781216
how do you think?
my colleagues are suffering from this Make.dep breakage.
and IMO the change itself seems rather controversial. it's more common to produce the similar-looking .o regardless of which language the source code is written.

GUIDINGLI:
Here is the fix:
https://github.com/apache/incubator-nuttx-apps/pull/801

so we don't need revert.

xiaoxiang781216:
After the discussion in #801, let's close this PR.

PR #801, opened 2021-07-06
Update Application.mk

GUIDINGLI:
## Summary

Update Application.mk

1. Generate c source code dependence with c compiler and flags, It's wrong to invoke the c++ compiler with c++ flags
2. Fix https://github.com/apache/incubator-nuttx-apps/pull/795 caused depends error.
    wrong: file.c => file.dir.o
    right: file.c => file.c.dir.o

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -92,7 +92,6 @@ endif
 
 DEPPATH += --dep-path .
 DEPPATH += --obj-path .
-DEPPATH += --obj-suffix $(SUFFIX)$(OBJEXT)
 
 VPATH += :.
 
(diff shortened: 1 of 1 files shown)


yamt (changes requested):
i still want this reverted unless a convincing use case for this unusual setup is provided.

xiaoxiang781216:
> i still want this reverted unless a convincing use case for this unusual setup is provided.

There are projects which have source files only different with suffix.
Here is one(NE10_fft_float32.neon.c/NE10_fft_float32.neon.s):
https://github.com/projectNe10/Ne10/tree/master/modules/dsp
This happen frequently in the algorithm library(like Ne10). We need provide a soluton to support the type of project.

yamt:
> > i still want this reverted unless a convincing use case for this unusual setup is provided.
> 
> There are projects which have source files only different with suffix.
> Here is one(NE10_fft_float32.neon.c/NE10_fft_float32.neon.s):
> https://github.com/projectNe10/Ne10/tree/master/modules/dsp
> This happen frequently in the algorithm library(like Ne10: c v.s. neon). We need provide a soluton to support the type of project.

ok.
it seems like what cmake does by default. i was not aware of the convention. thank you.

yamt (approved):
lgtm

xiaoxiang781216 (approved):

PR #807, opened 2021-07-12
netutils/dhcpd: bind socket to the interface

xiaoxiang781216:
## Summary
to handle the multiple netcard correctly.
## Testing
boot device with two net device, and run dhcpd to wait the client.


xiaoxiang781216:
@acassis could you help review this patch?

acassis (approved):

acassis:
@xiaoxiang781216 I don't have a way to test it here. BTW I approved it because the modifications are self-contained and are under #ifdef symbol

PR #809, opened 2021-07-21
Revert "webclient: Don't call the sink callback if no data is available"

gustavonihei:
## Summary
This PR intends to revert commit ee1f4fdcdbbfbb02f4e9e9970b27e45a95a724e7.

After the mentioned commit an application no longer receives the HTTP header via the `sink_callback` when the amount of received bytes equals the size of the HTTP header (`ws->offset == ws->datend`).
This creates an issue for applications that requires information contained on the HTTP header (e.g., the value of `Content-Length`).
On the other hand, the same filtering from ee1f4fdcdbbfbb02f4e9e9970b27e45a95a724e7 could be performed by the application during the handling of the `sink_callback`, maintaining the same current behavior.

## Impact
Users of `webclient` application.

## Testing
Custom application.

### DIFF
--- a/netutils/webclient/webclient.c
+++ b/netutils/webclient/webclient.c
@@ -1063,11 +1063,7 @@ int webclient_perform(FAR struct webclient_context *ctx)
                    * received file.
                    */
 
-                  if (ws->offset == ws->datend)
-                    {
-                      /* We don't have data to give to the client yet. */
-                    }
-                  else if (ctx->sink_callback)
+                  if (ctx->sink_callback)
                     {
                       ret = ctx->sink_callback(&ws->buffer, ws->offset,
                                                ws->datend, &ws->buflen,


yamt:
i'm not sure if i understand what your app does.
sink_callback is for the response body.
as far as i know, right now there is no api for applications to investigate response headers.
can you explain a bit?

gustavonihei:
> i'm not sure if i understand what your app does.
> sink_callback is for the response body.
> as far as i know, right now there is no api for applications to investigate response headers.
> can you explain a bit?

Sure. I want to log the download progress. I intended to retrieve the file size by parsing the value of Content-Length from the header.
Currently it suceeds most of the time. But occasionally it meets that condition from the commit and fails to retrieve the file size, messing the download progress log.

yamt:
> > i'm not sure if i understand what your app does.
> > sink_callback is for the response body.
> > as far as i know, right now there is no api for applications to investigate response headers.
> > can you explain a bit?
> 
> Sure. I want to log the download progress. I intended to retrieve the file size by parsing the value of Content-Length from the header.
> Currently it suceeds most of the time. But occasionally it meets that condition from the commit and fails to retrieve the file size, messing the download progress log.

how does your app get the value of Content-Length?
afaik, the current api doen't provide an access to response headers.

gustavonihei:
> > > i'm not sure if i understand what your app does.
> > > sink_callback is for the response body.
> > > as far as i know, right now there is no api for applications to investigate response headers.
> > > can you explain a bit?
> > 
> > 
> > Sure. I want to log the download progress. I intended to retrieve the file size by parsing the value of Content-Length from the header.
> > Currently it suceeds most of the time. But occasionally it meets that condition from the commit and fails to retrieve the file size, messing the download progress log.
> 
> how does your app get the value of Content-Length?
> afaik, the current api doen't provide an access to response headers.

Actually it does via the `sink_callback`, see the API documentation:
https://github.com/apache/incubator-nuttx-apps/blob/master/include/netutils/webclient.h#L93-L95

What my app does is check whether `offset > 0`, which means the buffer contains the header data.
Then I parse the Content-Length field using standard library functions:
https://godbolt.org/z/TxYe3fzP6

yamt:
> > > > i'm not sure if i understand what your app does.
> > > > sink_callback is for the response body.
> > > > as far as i know, right now there is no api for applications to investigate response headers.
> > > > can you explain a bit?
> > > 
> > > 
> > > Sure. I want to log the download progress. I intended to retrieve the file size by parsing the value of Content-Length from the header.
> > > Currently it suceeds most of the time. But occasionally it meets that condition from the commit and fails to retrieve the file size, messing the download progress log.
> > 
> > 
> > how does your app get the value of Content-Length?
> > afaik, the current api doen't provide an access to response headers.
> 
> Actually it does via the `sink_callback`, see the API documentation:
> https://github.com/apache/incubator-nuttx-apps/blob/master/include/netutils/webclient.h#L93-L95
> 
> What my app does is check whether `offset > 0`, which means the beginning of the buffer contains the header data.
> Then I parse the Content-Length field using standard library functions:
> https://godbolt.org/z/TxYe3fzP6

ok. i understand what your app does. thank you for explanation.
honestly speaking, i think it's an abuse of the api.

btw, i have some use cases related to Content-Length too. i had a plan to introduce a separate callback for headers. but it has been blocked because https://github.com/apache/incubator-nuttx-apps/pull/695 has been blocked.

yamt (changes requested):
what your app does won't work reliably even if this revert is merged.
depending on how much data each recv() returns, the buffer before the offset might not contain the entire header. (in the worse case, sink_callback will never see offset > 0 at all.)
it's better to introduce an explicit api to get headers.

gustavonihei:
> what your app does won't work reliably even if this revert is merged.
> depending on how much data each recv() returns, the buffer before the offset might not contain the entire header. (in the worse case, sink_callback will never see offset > 0 at all.)
> it's better to introduce an explicit api to get headers.

Ok, it makes sense.
I'll create another callback specific for returning header data.
Thanks.

yamt:
> > what your app does won't work reliably even if this revert is merged.
> > depending on how much data each recv() returns, the buffer before the offset might not contain the entire header. (in the worse case, sink_callback will never see offset > 0 at all.)
> > it's better to introduce an explicit api to get headers.
> 
> Ok, it makes sense.
> I'll create another callback specific for returning header data.
> Thanks.

can you share the rough api before implementing it?
i have some use cases and want to give early feedbacks.

gustavonihei:
> > > what your app does won't work reliably even if this revert is merged.
> > > depending on how much data each recv() returns, the buffer before the offset might not contain the entire header. (in the worse case, sink_callback will never see offset > 0 at all.)
> > > it's better to introduce an explicit api to get headers.
> > 
> > 
> > Ok, it makes sense.
> > I'll create another callback specific for returning header data.
> > Thanks.
> 
> can you share the rough api before implementing it?
> i have some use cases and want to give early feedbacks.

My initial idea was to make the webclient notify the HTTP header to the application line-by-line, by invoking the callback between those lines:
https://github.com/apache/incubator-nuttx-apps/blob/master/netutils/webclient/webclient.c#L561-L562
Then the application may parse the desired information.

yamt:
> > > > what your app does won't work reliably even if this revert is merged.
> > > > depending on how much data each recv() returns, the buffer before the offset might not contain the entire header. (in the worse case, sink_callback will never see offset > 0 at all.)
> > > > it's better to introduce an explicit api to get headers.
> > > 
> > > 
> > > Ok, it makes sense.
> > > I'll create another callback specific for returning header data.
> > > Thanks.
> > 
> > 
> > can you share the rough api before implementing it?
> > i have some use cases and want to give early feedbacks.
> 
> My initial idea was to make the webclient notify the HTTP header to the application line-by-line, by invoking the callback between those lines:
> https://github.com/apache/incubator-nuttx-apps/blob/master/netutils/webclient/webclient.c#L561-L562
> Then the application may parse the desired information.

it sounds like a plan.
how about the "truncated" case?

gustavonihei:
> > > > > what your app does won't work reliably even if this revert is merged.
> > > > > depending on how much data each recv() returns, the buffer before the offset might not contain the entire header. (in the worse case, sink_callback will never see offset > 0 at all.)
> > > > > it's better to introduce an explicit api to get headers.
> > > > 
> > > > 
> > > > Ok, it makes sense.
> > > > I'll create another callback specific for returning header data.
> > > > Thanks.
> > > 
> > > 
> > > can you share the rough api before implementing it?
> > > i have some use cases and want to give early feedbacks.
> > 
> > 
> > My initial idea was to make the webclient notify the HTTP header to the application line-by-line, by invoking the callback between those lines:
> > https://github.com/apache/incubator-nuttx-apps/blob/master/netutils/webclient/webclient.c#L561-L562
> > Then the application may parse the desired information.
> 
> it sounds like a plan.
> how about the "truncated" case?

Haven't thought about it yet. Which restrictions should this API be aware of?

yamt:
> > > > > > what your app does won't work reliably even if this revert is merged.
> > > > > > depending on how much data each recv() returns, the buffer before the offset might not contain the entire header. (in the worse case, sink_callback will never see offset > 0 at all.)
> > > > > > it's better to introduce an explicit api to get headers.
> > > > > 
> > > > > 
> > > > > Ok, it makes sense.
> > > > > I'll create another callback specific for returning header data.
> > > > > Thanks.
> > > > 
> > > > 
> > > > can you share the rough api before implementing it?
> > > > i have some use cases and want to give early feedbacks.
> > > 
> > > 
> > > My initial idea was to make the webclient notify the HTTP header to the application line-by-line, by invoking the callback between those lines:
> > > https://github.com/apache/incubator-nuttx-apps/blob/master/netutils/webclient/webclient.c#L561-L562
> > > Then the application may parse the desired information.
> > 
> > 
> > it sounds like a plan.
> > how about the "truncated" case?
> 
> Haven't thought about it yet. Which restrictions should this API be aware of?

i guess we should either of the following:
* remove CONFIG_WEBCLIENT_MAXHTTPLINE restriction and support arbitrary length.
* or, report some headers are incomplete.

i think the latter is good enough. (at least my use cases)
how about:
```
int (*header_callback)(const char *header_line, bool truncated, void *arg);
void *header_callback_arg;
```
if the callback returns -errno, it makes the entire webclient_perform fail.

gustavonihei:
> i guess we should either of the following:
> 
> * remove CONFIG_WEBCLIENT_MAXHTTPLINE restriction and support arbitrary length.
> * or, report some headers are incomplete.
> 
> i think the latter is good enough. (at least my use cases)
> how about:
> 
> ```
> int (*header_callback)(const char *header_line, bool truncated, void *arg);
> void *header_callback_arg;
> ```
> 
> if the callback returns -errno, it makes the entire webclient_perform fail.

Looks good, this is similar to what I had in mind.
My use case is quite simple, so this proposal would also fit.
The API just needs to ensure that `header_line` is either NULL-terminated, otherwise it may need to provide a `line_length` parameter.

gustavonihei:
> The API just needs to ensure that `header_line` is NULL-terminated, otherwise it may need to provide a `line_length` parameter.

I think we my invoke the callback from here: https://github.com/apache/incubator-nuttx-apps/blob/master/netutils/webclient/webclient.c#L586
Then the line string will be NULL-terminated.

yamt:
> > The API just needs to ensure that `header_line` is NULL-terminated, otherwise it may need to provide a `line_length` parameter.
> 
> I think we may invoke the callback from here: https://github.com/apache/incubator-nuttx-apps/blob/master/netutils/webclient/webclient.c#L599
> Then the line string will be NULL-terminated.

sounds good to me.

PR #824, opened 2021-08-06
boot/mcuboot: Add support for MCUboot bootloader

gustavonihei:
## Summary
This PR intends to add support for **MCUboot**, a secure boot solution for 32-bit microcontrollers.
https://www.mcuboot.com/
https://github.com/mcu-tools/mcuboot/

This port enables the following usages for MCUboot on Apache NuttX:
- MCUboot secure boot solution as a Apache NuttX application.
- Apache NuttX application images capable of accessing MCUboot public interfaces for providing support for secure firmware OTA update solutions.

The Apache NuttX port of MCUboot (https://github.com/mcu-tools/mcuboot/pull/1093) consists of the implementation on the following MCUboot interfaces:
- `<flash_map_backend/flash_map_backend.h>`
- `<mcuboot_config/mcuboot_config.h>`
- `<mcuboot_config/mcuboot_logging.h>`
- `<os/os_malloc.h>`
- `<sysflash/sysflash.h>`

The usage of MCUboot as a complete secure boot solution on top of Apache NuttX relies on the loader application.
By relying on Apache NuttX modular subsystems and its consolidated interfaces, it is possible to achieve the MCUboot integration in a platform-agnostic approach, which contributes to extending the MCUboot support base to basically every Apache NuttX supported platform with minimal implementation effort.

## Impact
New feature, no impact if not used.

## Testing
Tested with `esp32-wrover-kit` using the MCUboot application on both supported scenarios described above.

## Note 1
~~This PR will be left as **DRAFT** until https://github.com/mcu-tools/mcuboot/pull/1093 gets merged.~~
~~This PR is in **DRAFT** for further investigation on the crypto algorithms used by MCUboot, as brought up in this comment https://github.com/apache/incubator-nuttx-apps/pull/824#issuecomment-901468664.~~
PR is now **ready for review**.

## Note 2
**MCUboot** ~~requires~~ may rely on the **MbedTLS** library.
When https://github.com/apache/incubator-nuttx-apps/pull/566 gets merged, **MCUboot** Kconfig will be updated to add the proper dependencies on the library and required algorithms.

### DIFF
--- a/boot/.gitignore
+++ b/boot/.gitignore
@@ -1,0 +1,1 @@
+/Kconfig
--- a/boot/Make.defs
+++ b/boot/Make.defs
@@ -1,0 +1,21 @@
+############################################################################
+# apps/boot/Make.defs
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(wildcard $(APPDIR)/boot/*/Make.defs)
--- a/boot/Makefile
+++ b/boot/Makefile
@@ -1,0 +1,23 @@
+############################################################################
+# apps/boot/Makefile
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+MENUDESC = "Bootloader Utilities"
+
+include $(APPDIR)/Directory.mk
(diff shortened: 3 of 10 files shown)


utzig:
> **MCUboot** requires the **MbedTLS** library.

It only requires MbedTLS for RSA signature and RSA-OAEP image encryption; elliptic curve alternatives can use Tinycrypt, for secp256r1, or fiat for curve25519 based algos, both of which are already bundled with MCUboot.

gustavonihei:
> > **MCUboot** requires the **MbedTLS** library.
> 
> It only requires MbedTLS for RSA signature and RSA-OAEP image encryption; elliptic curve alternatives can use Tinycrypt, for secp256r1, or fiat for curve25519 based algos, both of which are already bundled with MCUboot.

Thanks for this clarification, @utzig!
This means that, if image encryption is not enabled, MCUboot can be built without these libraries, right?
I believed Mbed TLS was still mandatory, as stated by the MCUboot Porting documentation here: https://github.com/mcu-tools/mcuboot/blob/main/docs/PORTING.md. Is it deprecated?

utzig:
> > > **MCUboot** requires the **MbedTLS** library.
> > 
> > 
> > It only requires MbedTLS for RSA signature and RSA-OAEP image encryption; elliptic curve alternatives can use Tinycrypt, for secp256r1, or fiat for curve25519 based algos, both of which are already bundled with MCUboot.
> 
> Thanks for this clarification, @utzig!
> This means that, if image encryption is not enabled, MCUboot can be built without these libraries, right?

Image encryption can use both RSA-OAEP, which requires Mbed TLS, or ECIES with some elliptic curve, which does not require Mbed TLS (but could optionally use it).

I recommend you go with the options `MCUBOOT_SIGN_EC256` for signing, and `MCUBOOT_ENCRYPT_EC256` for encryption, because those options can work with the bundled Tinycrypt.

> I believed Mbed TLS was still mandatory, as stated by the MCUboot Porting documentation here: https://github.com/mcu-tools/mcuboot/blob/main/docs/PORTING.md. Is it deprecated?

Yes it is, Mbed TLS was required a long time ago solely because of its ASN1 parser, but we broke it apart from the main library, I think with: https://github.com/mcu-tools/mcuboot/pull/162, and this now lives under: https://github.com/mcu-tools/mcuboot/tree/main/ext/mbedtls-asn1

gustavonihei:
> Image encryption can use both RSA-OAEP, which requires Mbed TLS, or ECIES with some elliptic curve, which does not require Mbed TLS (but could optionally use it).
> 
> I recommend you go with the options `MCUBOOT_SIGN_EC256` for signing, and `MCUBOOT_ENCRYPT_EC256` for encryption, because those options can work with the bundled Tinycrypt.

Once again, thanks for the clarification! I'll take a deeper look into this and reflect these configuration options on NuttX Kconfig.
Thanks for the support.

normanr:
How is the image hash verification supposed to work? The hash to verify the download against is compiled into the agent example, so how can "the hash for firmware v2" be known (ahead of time) when compiling "firmware v1"?

PR #833, opened 2021-08-17
canutils: libcanutils: add GPL build barrier

jerpelea:
## Summary
Exclude libcanutils from build unless the license is set to
allow GPL code in the build.

NOTE
Depends on
https://github.com/apache/incubator-nuttx/pull/4352

## Impact
LICENSE

## Testing
Compiled

### DIFF
--- a/canutils/libcanutils/Kconfig
+++ b/canutils/libcanutils/Kconfig
@@ -6,7 +6,7 @@
 config CANUTILS_LIBCANUTILS
 	bool "CAN-utils support library"
 	default n
-	depends on NET_CAN
+	depends on NET_CAN && ALLOW_GPL_COMPONENTS
 	---help---
 		Enable the CAN-utils support library ported from
 		https://github.com/linux-can/can-utils


xiaoxiang781216 (approved):

gustavonihei (approved):

jerpelea:
@gustavonihei  @xiaoxiang781216  can you please merge the commit

PetervdPerk-NXP:
@jerpelea May I ask why libcanutils is identified as GPL2.0?

Checking the license header it indicates it's dual-licensed.
https://github.com/apache/incubator-nuttx-apps/blob/1c3d08a738cf395f99f211fad0b3a3ed2c19cc28/canutils/libcanutils/lib.c#L1

This it can be interpreted as either BSD-3-clause OR GPL2.0, I would prefer we see it's as BSD-3-clause since the license is more permissive.

jerpelea:
in my opinion it can be threated both but since GPL v2 is more restrictive I think that it is better to handle it this way

PetervdPerk-NXP:
Sure but let's say we treat it as GPL v2 and the code itself in incubator-nuttx-apps repo/project. Don't you get the problem we've to deal with copyleft?

I've added the code myself interpreting it as BSD3 so it's compatible with the incubator-nuttx-apps repo

jerpelea:
In my opinion if it is not, compiled (due to the build barrier), or used in the final binary for the product, it can be safely treated as GPL
I think that you can consider that it does not influence your copylefts in any way since the code is isolated

PetervdPerk-NXP:
Okay thanks for explanation, makes sense to interpret it that way.

Once last question would the following also work since it's dual licensed?
```
depends on NET_CAN && (ALLOW_GPL_COMPONENTS || ALLOW_BSD_COMPONENTS)
```

jerpelea:
unfortunately if we do that the code will be included if the BSD flag is selected

acassis:
@jerpelea but if we assume we are using libcanutils as GPL and the user want to use it, he/she need to enable GPL components that will make his final firmware GPL, but he doesn't want it as GPL because he is using it for a commercial product that cannot be GPL. If a library is dual license we can choose which license we want to use. I think in this case it is safe to use BSD instead of GPL.

acassis:
Ah ok, I saw the issue: "SPDX-License-Identifier: (GPL-2.0-only OR BSD-3-Clause)", I'm not sure if we can remove the "GPL-2.0-only" info. Probably this is a piece of code that we should move to apps_extra

jerpelea:
when we will move the dual license components to a new git we can consider this option. In the current case it will be included in the main build.

PR #839, opened 2021-08-22
system/termcurses: Add terminate operation into the termcurses_ops_s interface

meerd:
## Summary
system/termcurses: Add terminate operation into the termcurses_ops_s interface
    
The existing implementation of the termcurses_deinitterm calls termcurses_setcolors with some hard-coded values, leading to different terminal emulators' behavior. Instead, a reset call needs to be made to the implementation to roll back changes.
    
A new operation (terminate) has been added to the termcurses_ops interface to achieve this functionality.
    
The vt100's terminate implementation ensures that default foreground and background colors are applied portably across different terminals emulators.

## Impact
Please see the attached screen recordings to see how different terminal emulators behave without fix:
minicom
picocom

## Test
The changes have been tested on minicom, picocom, and gtkterm. All worked well with these changes.

### DIFF
--- a/include/system/termcurses.h
+++ b/include/system/termcurses.h
@@ -110,6 +110,10 @@ struct termcurses_ops_s
   /* Check for cached keycode value */
 
   CODE bool (*checkkey)(FAR struct termcurses_s *dev);
+
+  /* Terminate  */
+
+  CODE int (*terminate)(FAR struct termcurses_s *dev);
 };
 
 struct termcurses_dev_s
@@ -157,7 +161,7 @@ extern "C"
  ****************************************************************************/
 
 int termcurses_initterm(FAR const char *term_type, int in_fd, int out_fd,
-                        FAR struct termcurses_s **dev);
+                        FAR struct termcurses_s **term);
 
 /****************************************************************************
  * Name: termcurses_deinitterm
@@ -168,7 +172,7 @@ int termcurses_initterm(FAR const char *term_type, int in_fd, int out_fd,
  *
  ****************************************************************************/
 
-int termcurses_deinitterm(FAR struct termcurses_s *dev);
+int termcurses_deinitterm(FAR struct termcurses_s *term);
 
 /****************************************************************************
  * Name: termcurses_moveyx
--- a/system/termcurses/tcurses_vt100.c
+++ b/system/termcurses/tcurses_vt100.c
@@ -93,6 +93,7 @@ static int tcurses_vt100_setattributes(FAR struct termcurses_s *dev,
 static int tcurses_vt100_getkeycode(FAR struct termcurses_s *dev,
               FAR int *specialkey, FAR int *keymodifers);
 static bool tcurses_vt100_checkkey(FAR struct termcurses_s *dev);
+static int tcurses_vt100_terminate(FAR struct termcurses_s *dev);
 
 /************************************************************************************
  * Private Data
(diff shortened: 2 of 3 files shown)


system/termcurses/tcurses_vt100.c
```diff
@@ -112,22 +112,22 @@ static const struct termcurses_ops_s g_vt100_ops =
 
 /* VT100 terminal codes */
 
-static const char *g_clrscr         = "\033[2J";      /* Clear screen */
-static const char *g_clreos         = "\033[J";       /* Clear to end of screen */
-static const char *g_clreol         = "\033[K";       /* Clear to end of line */
-
-static const char *g_movecurs       = "\033[%d;%dH";  /* Move cursor to x,y */
-static const char *g_getwinsize     = "\x1b[s\x1b[999;999H\x1b[6n\x1bu";
-static const char *g_setfgcolor     = "\x1b[38;5;%dm";
```
xiaoxiang781216, line 121: it's more simple to change %d to %c?

meerd: Yes, you are right. @xiaoxiang781216. This happens when you try to fix something at very late hours. :) I must have hesitated signed char formatting with an unsigned char value. 

However, we still have a problem with the sequence. On minicom, the 'm' character gets printed out and becomes visible.

meerd:
Having looked at more examples on the web, my assumption (single-byte vs. multiple bytes) seems not correct. This issue needs more investigation.

meerd:
I believe the new fix is a better solution to this problem. Please have a look. @xiaoxiang781216

include/system/termcurses.h
```diff
@@ -96,6 +96,10 @@ struct termcurses_ops_s
   CODE int (*setcolors)(FAR struct termcurses_s *dev,
                         FAR struct termcurses_colors_s *colors);
 
+  /* Set default fg/bg colors */
+
+  CODE int (*setdefaultcolors)(FAR struct termcurses_s *dev);
```
xiaoxiang781216, line 101: how about we add deinit function to the restore work?

xiaoxiang781216:
BTW, it's better to squash the two patch into one:
```
git rebase --interactive HEAD~2
change pick to squash in vi
git push -f origin vt100_color_seq_fix
```

include/system/termcurses.h
meerd: setdefaultcolors removed.  the terminate operation has been added.

system/termcurses/tcurses_vt100.c
```diff
@@ -1479,3 +1485,29 @@ FAR struct termcurses_s *tcurses_vt100_initialize(int in_fd, int out_fd)
...
+
+  /* 
+     Set default foreground and background colors. 
+     (Ignore the return result.) 
+  */
+  
+  write(fd, g_setdefcolors, strlen(g_setdefcolors));
+
+  return OK;
+}
```
xiaoxiang781216, line 1513: add newline

include/system/termcurses.h
```diff
@@ -157,7 +161,7 @@ extern "C"
  ****************************************************************************/
 
 int termcurses_initterm(FAR const char *term_type, int in_fd, int out_fd,
-                        FAR struct termcurses_s **dev);
+                        FAR struct termcurses_s **term);
```
xiaoxiang781216, line 164: why change to term? It looks bad to mix dev and term for the same thing.

PR #840, opened 2021-08-24
interpreters/micropython: Add Micropython support to NuttX (WIP)

mransom-campbell:
I could use some feedback on this. It is still a WIP but it is taking longer than expected so I wanted to get this out there.

## Summary
[Micropython](https://github.com/micropython/micropython) is a Python implementation for Microcontrollers. I think it is a perfect fit for NuttX.

This PR adds NuttX as a Unix "Variant" under Micropython. This allows lots of features like files, sockets, etc.. to work without maintaining a separate Micropython port for NuttX.

Notes:
 - Hardest part was integrating the build systems. Currently Micropython is built using the Micropython ports/unix Makefile during the "context" step. Building during the build step would be better but I couldn't get that to work.
 - I previously manually specified all the .c files but that broke as soon as I updated Micropython. I am far from a Make expert so if
anyone has a better solution let me know.
 - Currently clones my fork of Micropython (https://github.com/mransom-campbell/micropython/tree/nuttx). I will upstream those changes once I do some cleanup.
 - Would it be better to download the latest Micropython release and apply patches like the Duktape NuttX support?

TODO:
- [x] Add NuttX support to Micropython Unix port.
- [x] Create Makefile and KConfig for NuttX Apps.
- [ ] Fix warnings when sockets/ussl are enabled (NuttX builds will -Werror)
- [ ] Fix Micropython stack check size.
- [ ] Add Micropython default Heap size as configuration option.
- [ ] Run Micropython test suite

## Impact

Application.mk: changed OBJS to be += so Applications can add their own object files manually.

## Testing

I have tried a few things on a stm32 board and Renesas RX board, everything seems to work. The full Micropython test suite should be run before this PR is merged.


cederom:
Update:
* I am finishing some projects based on MicroPython so I got some experience with it.
* I now need something faster for upcoming projects so I will get into NuttX internals in upcoming months.
* The best scenario for me is to have MicroPython for rapid prototyping and proof-of-concept and then NuttX RTOS as the target efficient solution. 
* Having MP on NuttX would allow invisible switch from MP ptototype to RTOS and this is the goal :-)

nimish:
Actually, https://github.com/adafruit/circuitpython/tree/main/ports/cxd56 since CircuitPython supports the NuttX SPRESENSE board, it could be quicker. The drivers would need to be generalized a bit, but otherwise it looks clean enough.

xiaoxiang781216:
> I've been able to get these rebased to the latest master, including the minor changes to the unix target in upython, but it fails on finding a `micropython_main` despite my best efforts to ensure that symbol exists.
> 

Source code which contain main need to be added to MAINSRC since the build need rename the main symbols to avoid the conflict.

> If you can get a basic interpreter to build, then I think this is doable pretty quickly
> 
> Of course, moving the device drivers might be harder.

nimish:
>nsh> upy
>MicroPython v1.19.1-24-gffb0e674b-dirty on 2022-06-25; nuttx [GCC 11.2.1] version
>Use Ctrl-D to exit, Ctrl-E for paste mode
>hello

Got it to work, with caveats: 
the repl doesn't echo and doesn't really respond to special chars, something to do with readline I think. 

This is on a SPRESENSE CXD5602 board. Can exit via `sys.exit()`

yamt:
> a full port of micropython must not be based on the unix port but on the embed port.

while i have some sympathies, i suspect it depends on what you want to do in your python code.

cederom:
Okay, so we now have full Python ported to NuttX in https://github.com/apache/nuttx-apps/pull/2879 and actively developed by @tmedicci. I was initially also interested in MicroPython.. but turned out writing code in NuttX is smaller faster and better. Should we still care about MicroPython when we have full Python? :-)

raiden00pl:
@cederom Isn't MicroPython much lighter solution than CPython? I don't think these two can be treated equally. I'll change my mind if CPython can run on stm32 :)

nimish:
> @cederom Isn't MicroPython much lighter solution than CPython? I don't think these two can be treated equally. I'll change my mind if CPython can run on stm32 :)

micropython can fit on a microcontroller, CPython cannot

But if CPython can run on NuttX, MicroPython should be able to do so as well, but it'll need some work.

cederom:
> @raiden00pl: @cederom Isn't MicroPython much lighter solution than CPython? I don't think these two can be treated equally. I'll change my mind if CPython can run on stm32 :)

Allright, so we want the MicroPython too :-) Just asking to see if the stuff is still relevant :-)

In upcoming months I have no time to play unfortunately to do any updates, sorry, but in order to have absolutely minimalistic implementation that is board independent:
1. I still think this implementation port should be kept as minimal as possible.
2. The hardware access should happen over NuttX's `/dev` and `ioctl()` subsystem based on exposed board configuration not the existing peripheral drivers for any specific board as part of MicroPython firmware packages.

nimish:
@cederom absolutely. NuttX runs on microcontrollers so micropython would be amazing

Yxuz66 (approved):

linguini1:
Is there anything that can be done with this PR or may I close it?

PR #856, opened 2021-09-23
wapi: add 5g channel map in wapi

liuyang34:
## Summary
scan result only process channel 1~14, and no process for 5g channel

## Impact
wapi scan result can only show 2g channel

## Testing
local test with brcm 43013 chip which can search 5g channel and result is pass


xiaoxiang781216 (approved):

anchao:
LGTM

PR #857, opened 2021-09-24
chat: Replace _info with ninfo to avoid polluting the nsh

acassis:
## Summary
Replace _info with ninfo to avoid polluting the nsh
## Impact
Avoid verbose message in the terminal
## Testing
Building


Ouss4 (approved):
LGTM.

PR #867, opened 2021-10-29
industry/foc: add support for Hall sensor and Qencoder feedback

raiden00pl:
## Summary
The main goal of this PR is to add Hall sensor and Qencoder feedback for FOC. 

## Impact
New functionality.

## Testing
Tested with modified example/foc app.

### DIFF
--- a/include/industry/foc/foc_common.h
+++ b/include/industry/foc/foc_common.h
@@ -71,6 +71,15 @@ enum foc_ramp_mode_e
   RAMP_MODE_NORMAL    = 3,      /* Normal operation */
 };
 
+/* Angle handler type */
+
+enum foc_angle_type_e
+{
+  FOC_ANGLE_TYPE_INVALID = 0,  /* Reserved */
+  FOC_ANGLE_TYPE_ELE     = 1,  /* Electrical angle */
+  FOC_ANGLE_TYPE_MECH    = 2,  /* Mechanical angle */
+};
+
 /****************************************************************************
  * Public Function Prototypes
  ****************************************************************************/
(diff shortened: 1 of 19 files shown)


include/industry/foc/foc_common.h
```diff
@@ -71,6 +71,15 @@ enum foc_ramp_mode_e
   RAMP_MODE_NORMAL    = 3,      /* Normal operation */
 };
 
+/* Angle handler type */
+
+enum foc_angle_type_e
+{
+  FOC_ANGLE_TYPE_INVALID = 0,  /* Reserved */
+  FOC_ANGLE_TYPE_ELE     = 1,  /* Electrical angle */
+  FOC_ANGLE_TYPE_MECH    = 2,  /* Mechanical angle */
```
acassis, line 80: @raiden00pl why do we need to have two separated types of FOC Angles ?

raiden00pl: There are 2 possible motor angle sensors:
1. the electrical angle sensors (which is the angle between motor poles) like Hall-sensor or observer-based angle estimation
2. the mechanical angle sensors (which is the angle of motor shaft vs motor stator) like encoder  
The FOC algorithm requires an electrical angle as input, so we must known if the value obtained requires conversion to an electrical angle or not.

acassis: Got it! Thank you!

uLipe: Just adding some extra information, having the number of poles which is tipically given the foc controller is able to convert the mechanical angle into the electrical one.

The only extra requirement that may have to be considered is a initial park and go alignment to keep the IQ + ID pair aligned to a initial known value and consider this offset to the account when converting mechanical into electrical angle.

PR #870, opened 2021-10-30
examples/foc: replace the requested velocity input with a more general setpoint input

raiden00pl:
## Summary

-  examples/foc: replace the requested velocity input with a more general setpoint input.
This is the inital step to support torque/velocity/position control
- examples/foc: temporarily disable error checking for ADC configuration to pass CI
NOTE: this should be reverted after updating nuttx/boards

## Impact
All example/foc configurations must be updated to use new configuration options

### DIFF
--- a/examples/foc/foc_adc.h
+++ b/examples/foc/foc_adc.h
@@ -40,8 +40,8 @@
 
 /* Velocity source must be specified */
 
-#if defined(CONFIG_EXAMPLES_FOC_VEL_CONST) &&   \
-  defined(CONFIG_EXAMPLES_FOC_VEL_ADC)
+#if defined(CONFIG_EXAMPLES_FOC_SETPOINT_CONST) &&   \
+  defined(CONFIG_EXAMPLES_FOC_SETPOINT_ADC)
 #  error
 #endif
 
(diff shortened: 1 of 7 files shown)


examples/foc/foc_adc.h
```diff
@@ -66,7 +66,7 @@
 #ifdef CONFIG_EXAMPLES_FOC_SETPOINT_CONST
 #  define SETPOINT_ADC_SCALE   (1)
 #  if CONFIG_EXAMPLES_FOC_SETPOINT_CONST_VALUE == 0
-#    error
+/* #    error */
```
xiaoxiang781216, line 69: can we merge the patch in  nuttx first?

raiden00pl: I think that if we merge first the patch in nuttx then CI will be broken until this PR merge.

related PR https://github.com/apache/incubator-nuttx/pull/4744

xiaoxiang781216: Both patch can't pass the CI, should we merge them together?

raiden00pl: It looks like it's the only possibility.
In that case I remove the last commit, due the fact that CI fails for another reason and please merge this PR together with https://github.com/apache/incubator-nuttx/pull/4744

xiaoxiang781216: Ok, let's merge both together.

xiaoxiang781216 (approved):

PR #873, opened 2021-10-31
nshlib/nsh_session: Handle the command arguments in any order

xiaoxiang781216:
## Summary
Make the command sh command line handling more robust

## Testing
run sh with script

### DIFF
--- a/nshlib/nsh_session.c
+++ b/nshlib/nsh_session.c
@@ -70,6 +70,7 @@ int nsh_session(FAR struct console_stdio_s *pstate,
 {
   FAR struct nsh_vtbl_s *vtbl;
   int ret = EXIT_FAILURE;
+  int i;
 
   DEBUGASSERT(pstate);
   vtbl = &pstate->cn_vtbl;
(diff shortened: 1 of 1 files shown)


nshlib/nsh_session.c
```diff
@@ -113,91 +114,112 @@ int nsh_session(FAR struct console_stdio_s *pstate,
...
+        {
+          fprintf(pstate->cn_errstream, g_fmtcmdfailed, "nsh_session",
+                  "cle", NSH_ERRNO_OF(-ret));
+          continue;
+        }
+#else
+      /* Display the prompt string */
+
+      fputs(g_nshprompt, pstate->cn_outstream);
       fflush(pstate->cn_outstream);
```
gustavonihei, line 197: The check for the definition of `CONFIG_FILE_STREAM` has been removed. It seems that the verification is somewhat inconsistent in this function, maybe because it is expected to be always enabled.
If so, the check for `CONFIG_FILE_STREAM` on line 154 could be removed as well.

xiaoxiang781216: > The check for the definition of `CONFIG_FILE_STREAM` has been removed. It seems that the verification is somewhat inconsistent in this function, maybe because it is expected to be always enabled.

Yes, your finding is right. You can see that there are many place call puts/fprintf without the guard of CONFIG_FILE_STREAM. That's why I remove `CONFIG_FILE_STREAM`.

> If so, the check for `CONFIG_FILE_STREAM` on line 154 could be removed as well.

Done.

xiaoxiang781216: > The check for the definition of `CONFIG_FILE_STREAM` has been removed. It seems that the verification is somewhat inconsistent in this function, maybe because it is expected to be always enabled.

Yes, your finding is right. You can see that there are many place call puts/fprintf without the guard of CONFIG_FILE_STREAM. That's why I remove `CONFIG_FILE_STREAM`.

> If so, the check for `CONFIG_FILE_STREAM` on line 154 could be removed as well.

Done.

gustavonihei: I've just found that `nsh_session.c` is only built when `CONFIG_FILE_STREAM=y`:
https://github.com/apache/incubator-nuttx-apps/blob/master/nshlib/Makefile#L29-L39

Weird, though, is that `nsh_stdsession.c` is built otherwise. But portions of the source file still check for `CONFIG_FILE_STREAM`. I wonder if that is being used anywhere.

PR #904, opened 2021-11-26
netutils/dhcpc: add non-blocking interface

coffee809721232:
## Summary
netutils/dhcpc: add non-blocking interface

Signed-off-by: songlinzhang <songlinzhang@xiaomi.com>
## Impact
Add dhcpc_request_async function
## Testing
Replace dhcpc_request

### DIFF
--- a/netutils/dhcpc/dhcpc.c
+++ b/netutils/dhcpc/dhcpc.c
@@ -52,6 +52,7 @@
 #include <unistd.h>
 #include <errno.h>
 #include <debug.h>
+#include <pthread.h>
 
 #include <arpa/inet.h>
 #include <netinet/udp.h>
@@ -129,12 +130,15 @@ struct dhcp_msg
 struct dhcpc_state_s
 {
   FAR const char    *interface;
-  FAR const void    *ds_macaddr;
-  int                ds_maclen;
   int                sockfd;
   struct in_addr     ipaddr;
   struct in_addr     serverid;
   struct dhcp_msg    packet;
+  bool               cancel;
+  pthread_t          thread;              /* Thread ID of the DHCPC thread */
+  dhcpc_callback_t   callback;            /* Thread callback of the DHCPC thread */
+  int                maclen;
+  uint8_t            macaddr[1];
 };
 
 /****************************************************************************
@@ -229,11 +233,10 @@ static int dhcpc_sendmsg(FAR struct dhcpc_state_s *pdhcpc,
   memset(&pdhcpc->packet, 0, sizeof(struct dhcp_msg));
   pdhcpc->packet.op    = DHCP_REQUEST;
   pdhcpc->packet.htype = DHCP_HTYPE_ETHERNET;
-  pdhcpc->packet.hlen  = pdhcpc->ds_maclen;
+  pdhcpc->packet.hlen  = pdhcpc->maclen;
   memcpy(pdhcpc->packet.xid, xid, 4);
-  memcpy(pdhcpc->packet.chaddr, pdhcpc->ds_macaddr, pdhcpc->ds_maclen);
-  memset(&pdhcpc->packet.chaddr[pdhcpc->ds_maclen],
-         0, 16 - pdhcpc->ds_maclen);
+  memcpy(pdhcpc->packet.chaddr, pdhcpc->macaddr, pdhcpc->maclen);
+  memset(&pdhcpc->packet.chaddr[pdhcpc->maclen], 0, 16 - pdhcpc->maclen);
   memcpy(pdhcpc->packet.options, magic_cookie, sizeof(magic_cookie));
 
   /* Add the common header options */
@@ -434,7 +437,7 @@ static uint8_t dhcpc_parsemsg(FAR struct dhcpc_state_s *pdhcpc, int buflen,
   if (buflen >= 44 && pdhcpc->packet.op == DHCP_REPLY &&
       memcmp(pdhcpc->packet.xid, xid, sizeof(xid)) == 0 &&
       memcmp(pdhcpc->packet.chaddr,
-             pdhcpc->ds_macaddr, pdhcpc->ds_maclen) == 0)
+             pdhcpc->macaddr, pdhcpc->maclen) == 0)
     {
       memcpy(&presult->ipaddr.s_addr, pdhcpc->packet.yiaddr, 4);
       return dhcpc_parseoptions(presult, &pdhcpc->packet.options[4],
@@ -443,6 +446,47 @@ static uint8_t dhcpc_parsemsg(FAR struct dhcpc_state_s *pdhcpc, int buflen,
     }
 
   return 0;
+}
+
+/****************************************************************************
+ * Name: dhcpc_run
+ ****************************************************************************/
+
+static void *dhcpc_run(void *args)
+{
+  FAR struct dhcpc_state_s *pdhcpc = (FAR struct dhcpc_state_s *)args;
+  struct dhcpc_state result;
+  int ret;
+
+  while (1)
+    {
+      ret = dhcpc_request(pdhcpc, &result);
+      if (ret == OK)
+        {
+          pdhcpc->callback(&result);
+        }
+      else
+        {
+          pdhcpc->callback(NULL);
+          nerr("dhcpc_request error\n");
+        }
+
+      if (pdhcpc->cancel)
+        {
+          return NULL;
+        }
+
+      while (result.lease_time)
+        {
+          result.lease_time = sleep(result.lease_time);
+          if (pdhcpc->cancel)
+            {
+              return NULL;
+            }
+        }
+    }
+
+  return NULL;
 }
 
 /****************************************************************************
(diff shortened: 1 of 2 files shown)


netutils/dhcpc/dhcpc.c
```diff
@@ -475,8 +520,8 @@ FAR void *dhcpc_open(FAR const char *interface, FAR const void *macaddr,
 
       memset(pdhcpc, 0, sizeof(struct dhcpc_state_s));
       pdhcpc->interface  = interface;
-      pdhcpc->ds_macaddr = macaddr;
-      pdhcpc->ds_maclen  = maclen;
+      pdhcpc->ds_maclen  = maclen > IFHWADDRLEN ? IFHWADDRLEN : maclen;
```
hartmannathan, line 523: If `maclen` too large, should we return error status (NULL) instead of proceeding?

hartmannathan (commented):
LGTM (with just the one question about `maclen` in dhcpc_open())

netutils/dhcpc/dhcpc.c
```diff
@@ -468,15 +512,15 @@ FAR void *dhcpc_open(FAR const char *interface, FAR const void *macaddr,
 
   /* Allocate an internal DHCP structure */
 
-  pdhcpc = (FAR struct dhcpc_state_s *)malloc(sizeof(struct dhcpc_state_s));
+  pdhcpc = malloc(sizeof(struct dhcpc_state_s) + maclen + 1);
```
xiaoxiang781216, line 515: change + 1 to - 1

netutils/dhcpc/dhcpc.c
coffee809721232: I adjusted the storage method of mac address, please review.

xiaoxiang781216 (approved):
LGTM.

hartmannathan:
Looks like I'm late to the party but LGTM.

Just one question:

Are we sure we want to remove the typecast from the (void *) returned by malloc() to (FAR struct dhcpc_state_s *) in this call:

```
-pdhcpc = (FAR struct dhcpc_state_s *)malloc(sizeof(struct dhcpc_state_s));
+pdhcpc = malloc(sizeof(struct dhcpc_state_s) + maclen - 1);
```

Some compilers may warn about that.

Might also benefit from a comment explaining the `maclen - 1` such as:

```
/* maclen - 1: First byte of maclen is included in the structure */
```

So that someone in the future won't think it's a mistake.

xiaoxiang781216:
> Looks like I'm late to the party but LGTM.
> 
> Just one question:
> 
> Are we sure we want to remove the typecast from the (void *) returned by malloc() to (FAR struct dhcpc_state_s *) in this call:
> 
> ```
> -pdhcpc = (FAR struct dhcpc_state_s *)malloc(sizeof(struct dhcpc_state_s));
> +pdhcpc = malloc(sizeof(struct dhcpc_state_s) + maclen - 1);
> ```
> 

It's a normal practice to let compiler auto convert void * to type * in C:
https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

> Some compilers may warn about that.
> 

This implicit conversion is allowed by C standard. But, it forbid by C++.


> Might also benefit from a comment explaining the `maclen - 1` such as:
> 
> ```
> /* maclen - 1: First byte of maclen is included in the structure */
> ```
> 
> So that someone in the future won't think it's a mistake.

Yes, good suggestion.

hartmannathan:
> > Are we sure we want to remove the typecast from the (void *) returned by malloc() to (FAR struct dhcpc_state_s *) in this call:
> > ```
> > -pdhcpc = (FAR struct dhcpc_state_s *)malloc(sizeof(struct dhcpc_state_s));
> > +pdhcpc = malloc(sizeof(struct dhcpc_state_s) + maclen - 1);
> > ```
> 
> It's a normal practice to let compiler auto convert void * to type * in C: https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc
> 
> > Some compilers may warn about that.
> 
> This implicit conversion is allowed by C standard. But, it forbid by C++.

Ah yes, that's what it is. In my own work I code C so it builds as both C and C++, so I forgot that's a C++ warning. Here I guess it's okay since we are coding for C89.

Thanks!

netutils/dhcpc/dhcpc.c
```diff
@@ -445,6 +448,47 @@ static uint8_t dhcpc_parsemsg(FAR struct dhcpc_state_s *pdhcpc, int buflen,
...
+        }
+
+      if (pdhcpc->cancel)
+        {
+          return NULL;
+        }
+
+      while (result.lease_time)
+        {
+          result.lease_time = sleep(result.lease_time);
```
normanr, line 481: Note: This should only sleep half of the lease_time before renewing.

normanr:
Was there a plan to update `netinit_net_bringup` to use `dhcpc_request_async`? and/or add some built-in commands like `ntpc` has to start/stop a dhcpc daemon?

xiaoxiang781216:
Nice idea, @coffee809721232 please take a look.

netutils/dhcpc/dhcpc.c
xiaoxiang781216: @coffee809721232

coffee809721232: > Note: This should only sleep half of the lease_time before renewing.

Yes, it should be half of the time to re-request.
This bug will be fixed later.
thanks. @normanr @xiaoxiang781216

coffee809721232:
> Was there a plan to update `netinit_net_bringup` to use `dhcpc_request_async`? and/or add some built-in commands like `ntpc` has to start/stop a dhcpc daemon?
Plan to add commands to start and stop the dhcpc daemon

netutils/dhcpc/dhcpc.c
normanr: fixed in: https://github.com/apache/incubator-nuttx-apps/pull/958

normanr:
fyi: I tried to use this code (adding a `dhcpc` daemon based on the `ntpc` daemon), but it has a couple of nasty bugs:
- it doesn't renew dhcp leases, but rather does full re-discovery and request (as if it had never seen the network before),
- you can't start `dhcpc_request_async` from `netinit_net_bringup`, and later call `dhcpc_close` from nsh:
  - the socket isn't closed properly, not sure why (it works fine when calling `dhcpc_request_async` from nsh),
- the interface name stored in `dhcpc_state_s` isn't duplicated, so the name is lost when the caller frees the underlying string,
- sometimes it sees a double-offer, it it will decline it even if it's just a duplicate of the offer that it's just received.

xiaoxiang781216:
> fyi: I tried to use this code (adding a `dhcpc` daemon based on the `ntpc` daemon), but it has a couple of nasty bugs:
> 
> * it doesn't renew dhcp leases, but rather does full re-discovery and request (as if it had never seen the network before),

yes, we can optimize in this case.

> * you can't start `dhcpc_request_async` from `netinit_net_bringup`, and later call `dhcpc_close` from nsh:
>   
>   * the socket isn't closed properly, not sure why (it works fine when calling `dhcpc_request_async` from nsh),

does you call dhcpc_request_async and dhcpc_close from the different task?

> * the interface name stored in `dhcpc_state_s` isn't duplicated, so the name is lost when the caller frees the underlying string,

The design want to save one string copy, and is easy to use copy instead.

> * sometimes it sees a double-offer, it it will decline it even if it's just a duplicate of the offer that it's just received.

normanr:
> > * you can't start `dhcpc_request_async` from `netinit_net_bringup`, and later call `dhcpc_close` from nsh:
> >   * the socket isn't closed properly, not sure why (it works fine when calling `dhcpc_request_async` from nsh),
> does you call dhcpc_request_async and dhcpc_close from the different task?

yes I figured that was why it breaks. I assume that all nsh builtins end up running in the same task, so that probably explains why it works. With more work it should be possible to refactor the code so that it's more like the ntpc daemon (with proper state synchronization) so that it can close the socket from the task that opened it.

Unlike the ntpc daemon, the dhcpc daemon should probably be able support multiple interfaces simultaneously. So it might just be easier to write it as a regular application (where the interface name is given on the command line) that can be safely killed with a signal. That would probably be a simpler way to solve the issue with closing the socket too.

> > * the interface name stored in `dhcpc_state_s` isn't duplicated, so the name is lost when the caller frees the underlying string,
> The design want to save one string copy, and is easy to use copy instead.

yea, it was easy to fix by adding `dup()` in `dhcpc_open` (and an extra `free()` before existing `free()` calls).

normanr:
> * sometimes it sees a double-offer, it it will decline it even if it's just a duplicate of the offer that it's just received.

I noticed that this is _not_ how [RFC 2131](https://datatracker.ietf.org/doc/html/rfc2131) intends the client to behave. An `OFFER` is only supposed to be `DECLINED` "If the client detects that the address is already in use (e.g., through the use of ARP)" (and only at the `ACK` stage of processing. If multiple offers are received then the client indicates which server's offer it is accepting through the `server identifier` field.

I also discovered that if any other udp packets arrive at the client while it's sleeping (e.g. from _other_ clients on the same network doing dhcp requests), then they will consume IO buffers, and never give them up, because `dhcpc_request` doesn't expect responses to be queued at the socket. This ends up starving all other network connections on the system. Changing the sleep into a recv & discard loop fixes this.

PR #912, opened 2021-12-05
testing/ostest: Don't include arch/tls.h

xiaoxiang781216:
## Summary
since it doesn't exist any more after PR https://github.com/apache/incubator-nuttx/pull/4941

## Impact
No

## Testing
Pass CI and ostest


gustavonihei (approved):

testing/ostest/tls.c
```diff
@@ -35,8 +35,6 @@
 
 #if CONFIG_TLS_NELEM > 0
 
-#include <arch/tls.h>
```
Ouss4, line 38: Don't we need to include `<nuttx/arch.h>` now?

gustavonihei: Not needed because `include/nuttx/tls.h` already includes it.

PR #917, opened 2021-12-09
timer_gpout example: use /dev/gpio0 and rename to timer_gpio

jarivanewijk:
## Summary
https://github.com/apache/incubator-nuttx/pull/4774 changes the naming of GPIO devices. Instead of separate gpout, gpin and gpint devices they are all named /dev/gpioN now. The timer_gpout example seems to be the only application that has explicit references to /dev/gpout0, so I have made changes so that it now uses /dev/gpio0 instead.

I have also renamed the application itself from timer_gpout to timer_gpio, which seems more appropriate now that there is no explicit gpout devices anymore.

## Impact
See https://github.com/apache/incubator-nuttx/pull/4774 - everybody and everything that was using /dev/gpoutN devices now needs to use /dev/gpioN. With this PR this will already be fixed for the timer_gpio application, which seems to be the only app to directly use a /dev/gpout device.

## Testing
I don't have access to a board/arch that implements both the timer and gpio drivers, so I was not able to test this on real hardware. However, the application builds without issues for a STM32F4Discovery board with timer and gpio drivers enabled.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1716,43 +1716,46 @@ Example configuration:
 - `CONFIG_EXAMPLES_TIMER_PROGNAME` – This is the name of the program that will
   be used when the NSH ELF program is installed. Default: `timer`.
 
-## `timer_gpout`
+## `timer_gpio`
 
-This example uses the timer interrupt to periodically
-change the state of a digital output.
-The digital output may be a relay, a led or anything else.
-This example can be very useful to validate timer drivers
-by using a logic analyzer connected to the digital output.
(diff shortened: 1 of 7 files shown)


xiaoxiang781216 (approved):

jarivanewijk:
This probably should not have been merged before https://github.com/apache/incubator-nuttx/pull/4774 ...

xiaoxiang781216:
Yes, but I think that patch will merge soon.:)

PR #920, opened 2021-12-12
Saving a file downloaded with wget

koliqi:
# Saving a file downloaded with wget

 wget, as is coded, only displays downloaded files on console.
This patch allow saving files on <local-path> as per command syntax:

            wget [-o <local-path>] <url>

# Tested on esp32-devkit.
                   mount -t spiffs /dev/esp/partition/spiffs /var
                   wget -o /var/test4.txt http://192.168.1.43:8000/partitions.csv


xiaoxiang781216:
@koliqi could you remove the merge commit from the PR? and fix the check error here:
https://github.com/apache/incubator-nuttx-apps/runs/4515099082?check_suite_focus=true

xiaoxiang781216:
let's track the change by: https://github.com/apache/incubator-nuttx-apps/pull/925

PR #921, opened 2021-12-13
crypto/mbedtls: Fix occasional failures during Mbed TLS build

gustavonihei:
## Summary
This PR intends to provide a fix to an occasional build failure on Mbed TLS.
Mbed TLS was being added twice to `CONFIGURED_APPS` variable, resulting in it being built more than once. If parallel build is enabled, the build could randomly fail due to unmet dependencies.

## Impact
Fix to Mbed TLS build. After patch is applied, build passes consistently.

## Testing
`esp32-devkitc` with a custom defconfig.


xiaoxiang781216 (approved):
LGTM.

PR #924, opened 2021-12-14
Wget branch2

koliqi:
## Saving a file downloaded with wget
wget, as is coded, only displays downloaded files on console.

## Impact
This patch allow saving files on <local-path> as per command syntax:

       wget [-o <local-path>] <url>

## Tested on esp32-devkit:

       mount -t spiffs /dev/esp/partition/spiffs /var
       wget -o /var/test4.txt http://192.168.1.43:8000/partitions.csv

## Version
v2. corrected checkpatch.sh errors

### DIFF
--- a/examples/wget/wget_main.c
+++ b/examples/wget/wget_main.c
@@ -33,7 +33,8 @@
 
 #include "netutils/netlib.h"
 #include "netutils/webclient.h"
-
+#include <fcntl.h>
+#include <sys/stat.h>
 /****************************************************************************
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@koliqi could you squash into one commit? BTW, you can achieve the same functionality with redirection:
wget <url> > /path/to/save

xiaoxiang781216:
@koliqi should we close this PR and focus on https://github.com/apache/incubator-nuttx-apps/pull/925?

PR #943, opened 2021-12-28
Update PWM example to handle extra channels

normanr:
## Summary
- Increase number of supported PWM channels from 4 to 6 and add an error if the number of channels is greater than 6
- Adjust channel and duty bounds to allow testing sentinel values
- Use macros from `fixedmath.h` for manipulating duty cycle to avoid magic shift logic in code (no change to behavior)

## Impact
The PWM example app can now be used to test more channels and sentinel values for channel and duty.

## Testing
Run the example app via nsh using a board with a bl602 chipset

### DIFF
--- a/examples/pwm/pwm_main.c
+++ b/examples/pwm/pwm_main.c
@@ -330,7 +330,7 @@ static void parse_args(FAR struct pwm_state_s *pwm, int argc,
 #ifdef CONFIG_PWM_MULTICHAN
           case 'c':
             nargs = arg_decimal(&argv[index], &value);
-            if (value < 1 || value > CONFIG_PWM_NCHANNELS)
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@normanr some style issue need fixed:
https://github.com/apache/incubator-nuttx-apps/runs/4657076135?check_suite_focus=true

normanr:
> @normanr some style issue need fixed: https://github.com/apache/incubator-nuttx-apps/runs/4657076135?check_suite_focus=true

Fixed. I'd been working to 78 characters, and that line was exactly that. I removed the extra whitespace before the equals sign to get it to fix. (I also removed some from the lower lines because there was unnecessary extra whitespace there too.)

xiaoxiang781216 (approved):

PR #944, opened 2021-12-29
Add a system tool to get/set hostname

normanr:
## Summary
Add a tool to get/set hostname

## Impact
Userspace can now get and set the system hostname

## Testing
Tested by setting hostname, and getting it back. Checked that the hostname was used in dhcp requests.

### DIFF
--- a/system/hostname/hostname_main.c
+++ b/system/hostname/hostname_main.c
@@ -1,0 +1,69 @@
+/****************************************************************************
+ * apps/system/hostname/hostname_main.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <stdio.h>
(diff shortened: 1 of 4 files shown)


system/hostname/hostname_main.c
```diff
@@ -0,0 +1,173 @@
...
+
+  if (ret == OK)
+    {
+      *val = ifr.ifr_mii_val_out;
+    }
+
+  return ret;
+}
+
+int set_phy_reg(uint16_t phy_id, uint16_t reg_num, uint16_t val)
```
xiaoxiang781216, line 123: should we remove all unrelated function and header file?

normanr: Woops, left over from mdio_main that I copied as a template. Removed.

xiaoxiang781216 (approved):
LGTM.

normanr:
fwiw, `uname -n` can also be used to get the hostname, but it's not as intuitive, and doesn't support setting the hostname (which inspired adding the tool).

normanr:
A related question: I want to automatically include a unique identifier from the board, eg network mac address or something, in the hostname. What do you think the best way to achieve that would be? Obviously writing a small userspace tool to generate the hostname which can be fed to the `hostname` tool via back quotes would work, but maybe an option to append `BOARDCTL_UNIQUEID` to `LIBC_HOSTNAME` would be viable? I'd prefer to avoid it in `board_app_initialize`. (Also I noticed `NETUTILS_DHCPC_HOST_NAME`, maybe it should just be replace with `LIBC_HOSTNAME` these days?)

xiaoxiang781216:
> fwiw, `uname -n` can also be used to get the hostname, but it's not as intuitive, and doesn't support setting the hostname (which inspired adding the tool).

It's fine to add hostname program since it exists on FreeBSD/Linux too.


> A related question: I want to automatically include a unique identifier from the board, eg network mac address or something, in the hostname. What do you think the best way to achieve that would be? Obviously writing a small userspace tool to generate the hostname which can be fed to the `hostname` tool via back quotes would work, but maybe an option to append `BOARDCTL_UNIQUEID` to `LIBC_HOSTNAME` would be viable? 

you can create a PR and see the feedback from other.

> I'd prefer to avoid it in `board_app_initialize`. (Also I noticed `NETUTILS_DHCPC_HOST_NAME`, maybe it should just be replace with `LIBC_HOSTNAME` these days?)

or let `NETUTILS_DHCPC_HOST_NAME` default value to `LIBC_HOSTNAME`.

PR #946, opened 2021-12-29
Fix link error due to missing libapps.a in KERNEL build

pussuw:
## Summary
The import target for kernel build now fails, due to setting libapps.a
as the default value for the BIN variable.

The fail happens when the ELFLD function passes the LDLIBS parameter
(which contains BIN / libapps.a) for the linker. There is no rule to
create libapps.a in the case of the kernel build, so the linker gives
an error due to it being missing.

This commit patches this behavior so that BIN is not appended to
LDLIBS. Another option would be to implement a dummy rule to create
libapps.a, but looking at the git history this is no longer wanted
behavior, thus the error is patched like this.

## Impact
Should fix the ELFLD function / linkage when using CONFIG_BUILD_KERNEL

## Testing
Tested that it builds. Created a ROMFS image with the .elf binaries. Did _not_ test that the elf files themselves work / contain a valid application (elfloader does work however).

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -59,10 +59,14 @@ else
 CWD = $(CURDIR)
 endif
 
-ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-  LDLIBS += "${shell cygpath -w $(BIN)}"
-else
-  LDLIBS += $(BIN)
+# Add the static application library to the linked libraries. Don't do this
+# with CONFIG_BUILD_KERNEL as there is no static app library
+ifneq ($(CONFIG_BUILD_KERNEL),y)
+  ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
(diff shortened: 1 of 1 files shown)


pussuw:
I'm not 100% sure this is the correct way to fix the ELFLD linkage. I tried to follow the git history and the modifications made to the build system but I don't claim to fully understand how this is supposed to work.

The patch itself might just hide a problem I'm not aware of, and the actual solution / fix is implicit.

xiaoxiang781216 (approved):

xiaoxiang781216:
> I'm not 100% sure this is the correct way to fix the ELFLD linkage. I tried to follow the git history and the modifications made to the build system but I don't claim to fully understand how this is supposed to work.
> 
> The patch itself might just hide a problem I'm not aware of, and the actual solution / fix is implicit.

Since the kernel mode is seldom used, let's trust your change and see other user feedback in the future.

PR #958, opened 2022-01-05
netutils/dhcpc: Fix dhcpc lease bug

coffee809721232:
## Summary
netutils/dhcpc: Fix dhcpc lease bug

Signed-off-by: songlinzhang songlinzhang@xiaomi.com
## Impact
dhcpc lease management
## Testing
call dhcpc_request_async

### DIFF
--- a/netutils/dhcpc/dhcpc.c
+++ b/netutils/dhcpc/dhcpc.c
@@ -468,6 +468,7 @@ static void *dhcpc_run(void *args)
       else
         {
           pdhcpc->callback(NULL);
+          memset(&result, 0, sizeof(result));
           nerr("dhcpc_request error\n");
         }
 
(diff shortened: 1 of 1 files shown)


netutils/dhcpc/dhcpc.c
```diff
@@ -458,6 +458,7 @@ static void *dhcpc_run(void *args)
   struct dhcpc_state result;
   int ret;
 
+  memset(&result, 0, sizeof(result));
```
xiaoxiang781216, line 461: why need memset here?

coffee809721232: If the first request fails, there may be a delay before the second request.

xiaoxiang781216: If so, should we move into while block to handle the fail in no first loop?

xiaoxiang781216: @normanr please take a look.

coffee809721232: > If so, should we move into while block to handle the fail in no first loop?

Yes, updated

xiaoxiang781216 (approved):

PR #959, opened 2022-01-05
wireless/gs2200m: replace the preamble scratch to union 

anchao:
## Summary

wireless/gs2200m: replace the preamble scratch to union 

## Impact

fix the preamble scratch buffer issue

## Testing

N/A, I do not have the wifi module....

### DIFF
--- a/wireless/gs2200m/gs2200m_main.c
+++ b/wireless/gs2200m/gs2200m_main.c
@@ -96,6 +96,23 @@ struct gs2200m_s
   struct usock_s sockets[SOCKET_COUNT];
 };
 
+union usrsock_request_u
+{
+  struct usrsock_request_socket_s       socket;
+  struct usrsock_request_close_s        close;
+  struct usrsock_request_connect_s      connect;
+  struct usrsock_request_sendto_s       sendto;
+  struct usrsock_request_recvfrom_s     recvfrom;
+  struct usrsock_request_setsockopt_s   setsockopt;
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@masayuki2009 could you try the change?

wireless/gs2200m/gs2200m_main.c
```diff
@@ -389,17 +406,17 @@ static int usrsock_request(int fd, FAR struct gs2200m_s *priv)
       return -EIO;
     }
 
-  assert(handlers[com_hdr->reqid].hdrlen < sizeof(hdrbuf));
+  assert(handlers[com_hdr->reqid].hdrlen < sizeof(req));
```
masayuki2009, line 409: @anchao 

I found that the assertion happend here.
I confirmed the following condition works.

```
  assert(handlers[com_hdr->reqid].hdrlen <= sizeof(req));
```

anchao: Thanks for your quick reply, I will update here, will the spresense works properly if this change included?

masayuki2009: > Thanks for your quick reply, I will update here, will the spresense works properly if this change included?

Yes. (I'm still checking the ntp client stack corruption https://github.com/apache/incubator-nuttx-apps/pull/957 but the issue does not relate to this PR)

masayuki2009 (approved):

PR #976, opened 2022-01-23
system/adb: Replace printf with syslog since adb is a service

xiaoxiang781216:
## Impact
adb output through syslog instead printf

## Testing
Pass CI


pkarashchenko (commented):
LGTM!

pkarashchenko (approved):

PR #991, opened 2022-01-30
tools: mkimport.sh: Add support relative file path

masayuki2009:
## Summary

- mkimport.sh only accepted absolute file path
- This commit adds support relative file path

## Impact

- None

## Testing

- Build (mkimport, make import) with sama5d4-ek:knsh


pkarashchenko (approved):
LGTM!

PR #992, opened 2022-01-31
Application.mk: Link the compiler runtime for ELF modules

yamt:
## Summary

An alternative is to have symbols like __udivdi3 in
the symbol table explicitly when loading the module.
But I feel it's too cumbersome to maintain.

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -69,6 +69,18 @@ ifneq ($(CONFIG_BUILD_KERNEL),y)
   endif
 endif
 
+# When building a module, link with the compiler runtime.
+# This should be linked after libapps. Consider that mbedtls in libapps
+# uses __udivdi3.
+ifeq ($(BUILD_MODULE),y)
+  # Revisit: This only works for gcc and clang.
+  # Do other compilers have similar?
+  COMPILER_RT_LIB = $(shell $(CC) $(ARCHCPUFLAGS) --print-libgcc-file-name)
+  ifneq ($(COMPILER_RT_LIB),)
+    LDLIBS += $(COMPILER_RT_LIB)
+  endif
+endif
+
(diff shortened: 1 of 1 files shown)


Application.mk
```diff
@@ -69,6 +69,18 @@ ifneq ($(CONFIG_BUILD_KERNEL),y)
...
 
+# When building a module, link with the compiler runtime.
+# This should be linked after libapps. Consider that mbedtls in libapps
+# uses __udivdi3.
+ifeq ($(BUILD_MODULE),y)
+  # Revisit: This only works for gcc and clang.
+  # Do other compilers have similar?
+  COMPILER_RT_LIB = $(shell $(CC) $(ARCHCPUFLAGS) --print-libgcc-file-name)
+  ifneq ($(COMPILER_RT_LIB),)
+    LDLIBS += $(COMPILER_RT_LIB)
```
xiaoxiang781216, line 80: should we add EXTRA_LIB to the link in case of build module?

yamt: iirc EXTRA_LIBS is to build the nuttx image and contains other things like proprietary drivers for esp32.

yamt: maybe we can make make.defs do
```
COMPILER_RT_LIBS += libgcc
EXTRA_LIBS += $(COMPILER_RT_LIBS)
```
instead of
```
EXTRA_LIBS += libgcc
```

how do you think?

xiaoxiang781216: What's different?

yamt: COMPILER_RT_LIBS would not have things like proprietary drivers.

xiaoxiang781216: I am fine with either.

xiaoxiang781216 (approved):

PR #997, opened 2022-02-04
system/cu: Don't select SERIAL_TERMIOS in Kconfig

xiaoxiang781216:
## Summary
since all terminal related code is already guarded by CONFIG_SERIAL_TERMIOS

## Impact
Minor


pkarashchenko (approved):
LGTM!

PR #1015, opened 2022-02-19
apps: hello_rust example program

ptka:
## Summary
Hello, Rust!!

## Impact
nothing as it's new

## Testing
Simulation on MacBook Air M1 with Monterey 12.2.1 using rustc rustc 1.58.1 (stable) via Homebrew.


xiaoxiang781216:
@ptka can we run it on the real device?

ptka:
Yes, I did some first test on a Raspberry Pico RP2040 device and it seems to run.
Once sim:rust is enabled and working (incl. CI), a configuration for a real target can be added, too.

xiaoxiang781216 (approved):

PR #1021, opened 2022-02-20
hello_rust: changes for target support

ptka:
## Summary
Changes for target support (Raspberry PI Pico)

## Impact
nothing as it's new

## Testing
Simulation on MacBook Air M1 with Monterey 12.2.1 using rustc rustc 1.58.1 (stable) via Homebrew.
Raspberry PI Pico (RP2040)


ptka:
Will be re-opened as the dependencies are fulfilled.

examples/hello_rust/hello_rust_main.rs
```diff
@@ -33,6 +40,20 @@ extern "C"
     pub fn printf(format: *const u8, ...) -> i32;
 }
 
+/****************************************************************************
+ * Private functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Panic handler (needed for [no_std] compilation)
```
no1wudi, line 48: Should we keep the same comment style with NuttX ? Rust has its own comment rule for better intergration of codes and docs.

ptka: Definitely a good point. I followed the NuttX style to have a starting point, knowing, that this is most likely no there best solution.

May I ask you to provide more details, so we can go a step ahead on this point:

  - Is there a good official documentation on the Rust style, we can reference in NuttX?
  - Can we check this, similar to checkpatch.sh and can be used in CI?
  - Is the rust style compatible to current NuttX style, checked from checkpatch.sh?

no1wudi: > 
> May I ask you to provide more details, so we can go a step ahead on this point:
> 
> * Is there a good official documentation on the Rust style, we can reference in NuttX?

Please refer to https://doc.rust-lang.org/reference/comments.html, there are doc comments and non-doc comments.

> * Can we check this, similar to checkpatch.sh and can be used in CI?

Please refer https://lore.kernel.org/rust-for-linux/YO3VIjpaCSZnNapB@moria.home.lan/T, may rustfmtcheck will help you.

xiaoxiang781216:
@ptka does the code pass rustfmt check?

xiaoxiang781216 (approved):

PR #1027, opened 2022-02-23
cmd_dd:fix dd will more than file size

anjiahao1:
## Summary
if a file size is 10
use dd if=a of=b bs = 8
b file size will 16, is a error
## Impact
usd dd cmd apps


xiaoxiang781216:
@anjiahao1 please rebase your patch to remove the merge.

PR #1028, opened 2022-02-23
cmd_dd:if read nbyte less than bs,must write nbyte

anjiahao1:
if a file size is 10
use dd if=a of=b bs = 8
b file size will 16, is a error
## Impact
use cmd dd app

### DIFF
--- a/nshlib/nsh_ddcmd.c
+++ b/nshlib/nsh_ddcmd.c
@@ -102,7 +102,7 @@ static int dd_write(FAR struct dd_s *dd)
   written = 0;
   do
     {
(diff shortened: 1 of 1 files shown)


pkarashchenko:
Should we consider an error if file size is not a multiple of sector size?

xiaoxiang781216:
> Should we consider an error if file size is not a multiple of sector size?

No, the new behavior conform the Unix counterpart.

xiaoxiang781216 (approved):

acassis (approved):

PR #1032, opened 2022-02-26
pwm: Fix issue with directing to "> /dev/null"

acassis:
## Summary
Fix issue with directing to "> /dev/null"
## Impact
Solve issue: https://github.com/apache/incubator-nuttx/issues/5626
## Testing
Running:
nsh> pwm -p /dev/pwm1 > /dev/null

### DIFF
--- a/examples/pwm/pwm_main.c
+++ b/examples/pwm/pwm_main.c
@@ -307,6 +307,13 @@ static void parse_args(FAR struct pwm_state_s *pwm, int argc,
   for (index = 1; index < argc; )
     {
       ptr = argv[index];
+      if (ptr[0] == '>')
+        {
+          /* We don't need to parse the remaining (file redirection) */
+
+          return;
+        }
+
       if (ptr[0] != '-')
         {
           printf("Invalid options format: %s\n", ptr);


examples/pwm/pwm_main.c
```diff
@@ -307,6 +307,13 @@ static void parse_args(FAR struct pwm_state_s *pwm, int argc,
   for (index = 1; index < argc; )
     {
       ptr = argv[index];
+      if (ptr[0] == '>')
```
xiaoxiang781216, line 310: should '>' be processed by nsh? Why we need special handle here?

pkarashchenko: Maybe `CONFIG_FILE_STREAM` is not set, so `>` is getting passed to built-in app?

pkarashchenko (commented):
The other option may be to rework code in `nsh_parse.c`
from
```
#ifdef CONFIG_FILE_STREAM
  /* Check if the output was re-directed using > or >> */

  if (argc > 2)
    {
      /* Check for redirection to a new file */

      if (strcmp(argv[argc - 2], g_redirect1) == 0)
        {
          redirect_save        = vtbl->np.np_redirect;
          vtbl->np.np_redirect = true;
          oflags               = O_WRONLY | O_CREAT | O_TRUNC;
          redirfile            = nsh_getfullpath(vtbl, argv[argc - 1]);
          argc                -= 2;
        }

      /* Check for redirection by appending to an existing file */

      else if (strcmp(argv[argc - 2], g_redirect2) == 0)
        {
          redirect_save        = vtbl->np.np_redirect;
          vtbl->np.np_redirect = true;
          oflags               = O_WRONLY | O_CREAT | O_APPEND;
          redirfile            = nsh_getfullpath(vtbl, argv[argc - 1]);
          argc                -= 2;
        }
    }
#endif
```
to
```
  /* Check if the output was re-directed using > or >> */

  if (argc > 2)
    {
      /* Check for redirection to a new file */

      if (strcmp(argv[argc - 2], g_redirect1) == 0)
        {
#ifdef CONFIG_FILE_STREAM
          redirect_save        = vtbl->np.np_redirect;
          vtbl->np.np_redirect = true;
#endif
          oflags               = O_WRONLY | O_CREAT | O_TRUNC;
          redirfile            = nsh_getfullpath(vtbl, argv[argc - 1]);
          argc                -= 2;
        }

      /* Check for redirection by appending to an existing file */

      else if (strcmp(argv[argc - 2], g_redirect2) == 0)
        {
#ifdef CONFIG_FILE_STREAM
          redirect_save        = vtbl->np.np_redirect;
          vtbl->np.np_redirect = true;
#endif
          oflags               = O_WRONLY | O_CREAT | O_APPEND;
          redirfile            = nsh_getfullpath(vtbl, argv[argc - 1]);
          argc                -= 2;
        }
    }
```

xiaoxiang781216:
The better solution is make nsh work with open/read/write/close, not fopen/fread/fwrite/fclose, which could let's disable CONFIG_FILE_STREAM in the most case.

acassis:
> CONFIG_FILE_STREAM

Yes, I agree! The NSH doesn't pass the "> /dev/null" to the application. But I need to double check it on Linux.

examples/pwm/pwm_main.c
acassis: Unfortunately CONFIG_FILE_STREAM is enabled and the issue is happening.

pkarashchenko: Could you please post `defconfig` file to the issue?

acassis: Hi Petro, I discovered that the issue was happening because I was using an old NuttX version:
NuttX:
19beb307dd5023c2bf414dc54c3141bfc95c7251

Apps:
c222043ed1ede9ef8e9e93ffed14ee9c41c4a2b1

This is may .config: https://pastebin.com/raw/duUsX6hX

PR #1046, opened 2022-03-07
webclient: minimum HTTP 1.1 support

yamt:
## Summary
Add minimum HTTP 1.1 support.

My motivation is to be able to communicate with a server which rejects HTTP 1.0.

## Testing
A version ported to linux and macOS has been tested against:

    * http://httpbin.org/stream/10
    
    * Docker API

### DIFF
--- a/netutils/webclient/webclient.c
+++ b/netutils/webclient/webclient.c
@@ -1625,6 +1625,7 @@ int wget_post(FAR const char *url, FAR const char *posts, FAR char *buffer,
 void webclient_set_defaults(FAR struct webclient_context *ctx)
 {
   memset(ctx, 0, sizeof(*ctx));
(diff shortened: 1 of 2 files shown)


netutils/webclient/webclient.c
```diff
@@ -174,7 +179,8 @@ struct conn_s
 
 /* flags for wget_s::internal_flags */
 
-#define	WGET_FLAG_GOT_CONTENT_LENGTH	1
+#define	WGET_FLAG_GOT_CONTENT_LENGTH	1U
+#define	WGET_FLAG_CHUNKED				2U
```
xiaoxiang781216, line 183: let's align

yamt: done

xiaoxiang781216 (approved):

PR #1059, opened 2022-03-13
App: zig example - Hello

kassane:
## Summary

Hello zig.

## Impact

none.

## Testing

Arch Linux - simulation


examples/hello_rust/Makefile
```diff
@@ -1,5 +1,5 @@
 ############################################################################
-# apps/examples/hello_rust/Make.defs
+# apps/examples/hello_zig/Make.defs
```
xiaoxiang781216: why change hello_rust's Makefile

examples/hello_zig/Makefile
```diff
@@ -23,6 +23,7 @@ include $(APPDIR)/Make.defs
 # Hello, Zig! Example
 
 MAINSRC = hello_zig_main.zig
+ZIGFLAGS += -fsingle-threaded
```
xiaoxiang781216, line 26: should it be a common flags?

xiaoxiang781216:
@kassane please reorg your patchset too?

examples/hello_zig/Makefile
kassane: Sorry, this flag is only needed when for example trying to make it compatible with rust_app which requires lib to be single-thread.

pkarashchenko (approved):

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #1061, opened 2022-03-14
apps/wamr: Add wamr memory Kconfig

DS-LK:
## Summary
wamr itself comes with a task, which we usually do not use directly, so closing this configuration can save about 50k code size. In addition, increase the memory configuration, enable the global variables of wamr and the memory information debug configuration

## Impact
It will not affect the original function.

## Testing
We have no problem in the actual project use process.


interpreters/wamr/Makefile
```diff
@@ -40,12 +40,14 @@ WAMR_URL      = $(WAMR_URL_BASE)/$(WAMR_TARBALL)
 
 VPATH += $(WAMR_UNPACK)/product-mini/platforms/nuttx
 
+ifeq ($(CONFIG_INTERPRETERS_IWASM_TASK), y)
```
pkarashchenko, line 43:
```suggestion
ifeq ($(CONFIG_INTERPRETERS_IWASM_TASK),y)
```

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #1063, opened 2022-03-14
testing/fstest: add cleanup to fatutf8

VanFeo:
Signed-off-by: fanzhuyun <fanzhuyun@xiaomi.com>

## Summary
testing/fstest: add cleanup to fatutf8
## Impact
add cleanup to fatutf8
## Testing
local test

### DIFF
--- a/testing/fatutf8/fatutf8_main.c
+++ b/testing/fatutf8/fatutf8_main.c
@@ -60,7 +60,7 @@ int main(int argc, FAR char *argv[])
   int len;
 
   char buf[128];
-  char path[256];
+  char path[PATH_MAX];
 
   char *basepath = NULL;
 
@@ -164,6 +164,34 @@ int main(int argc, FAR char *argv[])
     {
       printf("open failed: %d\n", errno);
       exit(fd);
+    }
+
+  printf("\n");
+
+  ret = remove(path);
+  if (ret == 0)
+    {
+      printf("removed %s\n", path);
+      basepath = strrchr(path, '/');
+      if (basepath != NULL)
+        {
+          *basepath = '\0';
+          ret = remove(path);
+          if (ret == 0)
+            {
+              printf("removed %s\n", path);
+            }
+          else
+            {
+              printf("remove failed: %d\n", errno);
+              exit(ret);
+            }
+        }
+    }
+  else
+    {
+      printf("remove failed: %d\n", errno);
+      exit(ret);
     }
 
   return EXIT_SUCCESS;


testing/fatutf8/fatutf8_main.c
```diff
@@ -61,6 +61,7 @@ int main(int argc, FAR char *argv[])
 
   char buf[128];
   char path[256];
+  char dir_path[256];
```
pkarashchenko, line 64:
```suggestion
  char path[PATH_MAX];
  char dir_path[PATH_MAX];
```

testing/fatutf8/fatutf8_main.c
```diff
@@ -108,6 +109,7 @@ int main(int argc, FAR char *argv[])
 
   printf("\n");
 
+  strcpy(dir_path, path);
```
pkarashchenko, line 112: I do not think that removal of a directory unconditionally is a good idea. What if directory existed already and was intentionally created by user? The removal in such case will be confusing. Can we better check if directory is already created and delete only if it was created by this code?

testing/fatutf8/fatutf8_main.c
```diff
@@ -166,5 +168,28 @@ int main(int argc, FAR char *argv[])
       exit(fd);
     }
 
+  printf("\n");
+
+  ret = remove(path);
+  if (ret == 0)
+    {
+      printf("removed %s\n", path);
+      ret = remove(dir_path);
```
pkarashchenko, line 177: Maybe we can use next code instead of adding new variable to stack?
```
basepath = strrchr(path, '/');
if (basepath != NULL)
  {
    *basepath = '\0';
    ret = remove(path);
    if (ret == 0)
...
  }
```

VanFeo: It is a good idea and has been revised

testing/fatutf8/fatutf8_main.c
```diff
@@ -52,6 +52,7 @@ static void show_usage(FAR const char *progname, int exitcode)
 #define BASEPATH "/mnt/sdcard"
 #define FOLDER_NAME "/ÄÖÜ 测试文件夹 Test Folder"
 #define FILE_NAME "/ÄÖÜ 测试文件 Test File"
+#define PATH_MAX 256
```
xiaoxiang781216, line 55: why define PATH_MAX which is a standard macro?

VanFeo: removed

testing/fatutf8/fatutf8_main.c
xiaoxiang781216: It's impossible since https://github.com/apache/incubator-nuttx-apps/pull/1063/files#diff-c6c45b54fda3ab97655d3becbfe72f45338ed829e17a019b868a6999e62ceb6cR98-R103 report EEXIST.

pkarashchenko: Ok

pkarashchenko (approved):

PR #1064, opened 2022-03-14
feature: wapi add encode parse when scan

zhhyu7:
wapi add encode parse when scan
fix: ap socket can not accept twice
rpmsg: pointer of netdev ioctl support cross-core access via clean dcache
fix: iperf can not exit when stop

Signed-off-by: zhanghongyu <zhanghongyu@xiaomi.com>
Change-Id: Ic9959a7426091ee2f35b68cb41c14a9e343468ac

### DIFF
--- a/netutils/usrsock_rpmsg/usrsock_rpmsg.h
+++ b/netutils/usrsock_rpmsg/usrsock_rpmsg.h
@@ -27,6 +27,11 @@
 
 #include <nuttx/net/usrsock.h>
 
+#ifdef CONFIG_NETDEV_WIRELESS_IOCTL
+#  include <nuttx/wireless/wireless.h>
+#  include <metal/cache.h>
+#endif
+
 /****************************************************************************
  * Pre-processor definitions
  ****************************************************************************/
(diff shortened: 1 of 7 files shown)


netutils/usrsock_rpmsg/usrsock_rpmsg.h
```diff
@@ -27,6 +27,11 @@
 
 #include <nuttx/net/usrsock.h>
 
+#ifdef CONFIG_NETDEV_WIRELESS_IOCTL
+#  include <nuttx/wireless/wireless.h>
+#  include <metal/cache.h>
```
pkarashchenko, line 32: Where this file is located?

netutils/usrsock_rpmsg/usrsock_rpmsg_client.c
```diff
@@ -155,6 +161,41 @@ static int usrsock_rpmsg_ept_cb(struct rpmsg_endpoint *ept, void *data,
...
+  FAR struct iwreq *wlreq;
+
+  if (req->head.reqid == USRSOCK_REQUEST_IOCTL)
+    {
+      if (WL_IS80211POINTERCMD(req->cmd))
+        {
+          wlreq = (FAR struct iwreq *)(req + 1);
+          if (action == USRSOCK_COHERENT_BEFORE)
+            {
+              metal_cache_flush(wlreq->u.data.pointer,
```
pkarashchenko, line 179: Is `metal_` some arch specific API. Maybe we can use `up_invalidate_dcache`/`up_flush_dcache` instead?

netutils/usrsock_rpmsg/usrsock_rpmsg.h
xiaoxiang781216: libmetal is part of OpenAMP framework:
https://github.com/OpenAMP/libmetal/blob/main/lib/cache.h#L33-L49

netutils/usrsock_rpmsg/usrsock_rpmsg_client.c
xiaoxiang781216: usrsock rpmag is built on the top of openamp/libmetal, it's reasonable to call meta_ API instead of NuttX specific API to improve the portability.

netutils/usrsock_rpmsg/usrsock_rpmsg.h
pkarashchenko: Ok. Some time ago I worked with SiFive RISC-V metal SDK, so was a bit confused. Thank you for explanation

PR #1075, opened 2022-03-15
add Lua interpreter

mogenson:
## Summary

Lua is a popular scripting language that is tailored towards embedding in applications and embedded devices. This PR builds Lua as an NSH app. The `apps/interpreters/lua/Makefile` does a few notable things:
- fetches and unpacks a source tarball from the Lua website
- sets defines to use the `readline` library for the Lua REPL
- adopts logic from `apps/builtin/Makefile` to generate headers for including builtin Lua modules

To expand on the last point: Nuttx apps can include `apps/interpreters/lua/LuaModule.mk` to register a C module with the Lua interpreter. This makefile uses the `register` target to write entries to the `apps/interpreters/lua/registry` directory. See `apps/examples/luamod_hello` for an example of how to register a builtin Lua C module. The Lua source file `linit.c` is excluded from the build in favor of `nuttx_linit.c`, which optionally opens core Lua modules in addition to registered builtin modules.

Here's an example of a Lua script to blink an LED using the `userleds` driver:

```lua
-- concatenate strings and print to NSH console
print("hello from " .. _VERSION)

-- open device driver, file is closed when leds is garbage collected
leds = io.open("/dev/userleds", "wb")

-- from 1 to 10, inclusive
for i = 1, 10 do  -- lua convention is to start at 1
    -- pack value into uint32_t and write to file
    leds:write(string.pack("I", i % 2))
    leds:flush()

    -- run sleep command from NSH
    os.execute("sleep 1")
end
```

Required defconfig:
```
CONFIG_INTERPRETERS_LUA=y
CONFIG_NSH_READLINE=y
CONFIG_READLINE_CMD_HISTORY=y
CONFIG_READLINE_TABCOMPLETION=y
CONFIG_SYSTEM_NSH=y
CONFIG_USERLED=y
CONFIG_USERLED_LOWER=y
```

For Lua bindings to the `userleds` `ioctl` interface, checkout this external Lua C module: https://github.com/mogenson/lua_userleds

## Impact

No modifications to Nuttx are required and no changes outside of `apps/interpreters` and `apps/examples` were made.

## Testing

Lua versions 5.2.0 through 5.4.4 have been tested successfully on a Teensy 4.1 board and MacOS/Linux with the NSH simulator config. Lua version 5.1 and lower have a slightly different build setup that is incompatible with these Makefiles.

### DIFF
--- a/examples/lua_module/luamod_hello.c
+++ b/examples/lua_module/luamod_hello.c
@@ -1,0 +1,76 @@
+/****************************************************************************
+ * apps/examples/lua_module/luamod_hello.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <lua.h>
+#include <lauxlib.h>
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+static int say_hello(lua_State *L);
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static const struct luaL_Reg g_hello[] =
+{
+  {"say_hello", say_hello},
+  {NULL, NULL},
+};
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: say_hello
+ *
+ *   Push a "Hello World!" string to the Lua interpreter.
+ *
+ ****************************************************************************/
+
+static int say_hello(lua_State *L)
+{
+  lua_pushstring(L, "Hello World!");
+  return 1;
+}
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: luaopen_hello
+ *
+ *   Open the "hello" Lua module.
+ *
+ ****************************************************************************/
+
+int luaopen_hello(lua_State *L)
+{
+  luaL_newlib(L, g_hello);
+  return 1;
+}
--- a/interpreters/lua/Kconfig
+++ b/interpreters/lua/Kconfig
@@ -1,0 +1,67 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+menuconfig INTERPRETERS_LUA
+	bool "Lua interpreter support"
+	default n
+	select ARCH_SETJMP_H
+	select LIBC_LOCALE
+	---help---
+		Embed Lua language interpreter.
+		Select the Lua version with the INTERPRETER_LUA_VERSION config.
+		A math library is required. Use the LIBM config or a toolchain library.
+		It's suggested to enable the SYSTEM_READLINE and LIBC_FLOATINGPOINT configs.
+
+if INTERPRETERS_LUA
+
+config INTERPRETER_LUA_VERSION
+	string "Lua interpreter version"
+	default "5.4.0"
+	---help---
+		Lua release version to fetch and build.
+		Versions 5.2.0 and up are supported.
+
+config INTERPRETER_LUA_CORELIBS
+	bool "Load core Lua modules"
+	default y
+	select SYSTEM_SYSTEM
+	---help---
+		Load core Lua modules like "os", "string", and "table".
+
+config INTERPRETER_LUA_PATH
+	string "Lua modules search path"
+	---help---
+		Override default package.path search path for Lua modules.
+
+config INTERPRETER_LUA_CPATH
+	string "Lua C modules search path"
+	---help---
+		Override default package.cpath search path for C modules.
+
+config INTERPRETER_LUA_32BIT
+	bool "Use 32-bit integers and floats"
+	default y
+	---help---
+		"Use int and float instead of long and double for Lua numbers."
+
+config INTERPRETER_LUA_PRIORITY
+	int "Lua interpreter priority"
+	default 100
+	---help---
+		Task priority of the Lua interpreter main task.
+
+config INTERPRETER_LUA_STACKSIZE
+	int "Lua interpreter stack size"
+	default 32768
+	---help---
+		Size of the stack allocated for the Lua interpreter main task.
+
+config INTERPRETER_LUA_IOBUFSIZE
+	int "I/O buffer size"
+	default 1024
+	---help---
+		Size of the statically allocated I/O buffer.
+
+endif # INTERPRETERS_LUA
--- a/interpreters/lua/Make.defs
+++ b/interpreters/lua/Make.defs
@@ -1,0 +1,31 @@
+############################################################################
+# apps/interpreters/lua/Make.defs
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifneq ($(CONFIG_INTERPRETERS_LUA),)
+CONFIGURED_APPS += $(APPDIR)/interpreters/lua
+
+# Enable <lua.h> and <lauxlib.h> includes.
+
+CFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" \
+	$(APPDIR)/interpreters/lua/lua-$(CONFIG_INTERPRETER_LUA_VERSION)/src}
+CXXFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" \
+	$(APPDIR)/interpreters/lua/lua-$(CONFIG_INTERPRETER_LUA_VERSION)/src}
+
+endif
--- a/interpreters/lua/Makefile
+++ b/interpreters/lua/Makefile
@@ -1,0 +1,133 @@
+############################################################################
+# apps/interpreters/lua/Makefile
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+# Lua built-in application info
+
+PROGNAME  = lua
+PRIORITY  = $(CONFIG_INTERPRETER_LUA_PRIORITY)
+STACKSIZE = $(CONFIG_INTERPRETER_LUA_STACKSIZE)
+MODULE    = $(CONFIG_INTERPRETERS_LUA)
+
+# Lua library
+
+LUA_VERSION  = $(patsubst "%",%,$(strip $(CONFIG_INTERPRETER_LUA_VERSION)))
+LUA_TARBALL  = lua-$(LUA_VERSION).tar.gz
+LUA_UNPACK   = lua-$(LUA_VERSION)
+LUA_URL_BASE = http://www.lua.org/ftp
+LUA_URL      = $(LUA_URL_BASE)/$(LUA_TARBALL)
+LUA_SRC      = $(LUA_UNPACK)$(DELIM)src
+
+MAINSRC       = $(LUA_SRC)$(DELIM)lua.c
+CORELIBS_SRCS = $(filter-out $(LUA_SRC)$(DELIM)lauxlib.c,$(wildcard $(LUA_SRC)$(DELIM)*lib.c))
+EXCLUDE_SRCS  = $(MAINSRC) $(CORELIBS_SRCS) $(LUA_SRC)$(DELIM)luac.c $(LUA_SRC)$(DELIM)linit.c
+CSRCS         = $(filter-out $(EXCLUDE_SRCS),$(wildcard $(LUA_SRC)$(DELIM)*.c))
+CSRCS        += nuttx_linit.c
+
+CFLAGS += -DLUA_MAXINPUT=$(CONFIG_INTERPRETER_LUA_IOBUFSIZE)
+CFLAGS += -DLUA_PROGNAME=\"$(PROGNAME)\"
+
+ifeq ($(CONFIG_INTERPRETER_LUA_32BITS),y)
+CFLAGS += -DLUA_32BITS
+endif
+
+ifneq ($(CONFIG_INTERPRETER_LUA_PATH),"")
+CFLAGS += -DLUA_PATH_DEFAULT=\"$(CONFIG_INTERPRETER_LUA_PATH)\"
+endif
+
+ifneq ($(CONFIG_INTERPRETER_LUA_CPATH),"")
+CFLAGS += -DLUA_CPATH_DEFAULT=\"$(CONFIG_INTERPRETER_LUA_CPATH)\"
+endif
+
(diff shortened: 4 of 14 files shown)


interpreters/lua/Kconfig
```diff
@@ -0,0 +1,90 @@
...
+#
+
+menuconfig INTERPRETERS_LUA
+	bool "Lua interpreter support"
+	default n
+	select ARCH_SETJMP_H
+	select LIBC_FLOATINGPOINT
+	select LIBC_LOCALE
+	select LIBM
+	select SYSTEM_SYSTEM
```
xiaoxiang781216, line 13: why need system

interpreters/lua/Kconfig
```diff
@@ -0,0 +1,90 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+menuconfig INTERPRETERS_LUA
+	bool "Lua interpreter support"
+	default n
+	select ARCH_SETJMP_H
+	select LIBC_FLOATINGPOINT
```
xiaoxiang781216, line 10: can we let user select LIBC_FLOATINGPOINT?

interpreters/lua/LuaModule.mk
```diff
@@ -0,0 +1,37 @@
+############################################################################
+# apps/interpreters/lua/LuaModule.mk
```
xiaoxiang781216, line 2: LuaModue.mk to Module.mk, the folder name(lua) is enough

interpreters/lua/Make.defs
```diff
@@ -0,0 +1,31 @@
...
+############################################################################
+
+ifneq ($(CONFIG_INTERPRETERS_LUA),)
+CONFIGURED_APPS += $(APPDIR)/interpreters/lua
+
+# Enable <lua.h> and <lauxlib.h> includes.
+
+CFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" \
+	$(APPDIR)/interpreters/lua/lua-$(CONFIG_INTERPRETER_LUA_VERSION)/src}
+CXXFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" \
```
xiaoxiang781216, line 28: it's very easy to introduce the conflict if we expose the whole code base to the global search path.

interpreters/lua/Makefile
```diff
@@ -0,0 +1,133 @@
...
+
+LUA_VERSION  = $(patsubst "%",%,$(strip $(CONFIG_INTERPRETER_LUA_VERSION)))
+LUA_TARBALL  = lua-$(LUA_VERSION).tar.gz
+LUA_UNPACK   = lua-$(LUA_VERSION)
+LUA_URL_BASE = http://www.lua.org/ftp
+LUA_URL      = $(LUA_URL_BASE)/$(LUA_TARBALL)
+LUA_SRC      = $(LUA_UNPACK)/src
+
+MAINSRC      = $(LUA_SRC)/lua.c
+CORELIB_SRCS = $(filter-out $(LUA_SRC)/lauxlib.c,$(wildcard $(LUA_SRC)/*lib.c))
```
xiaoxiang781216, line 40: why exclude lauxlib.c?

interpreters/lua/Makefile
```diff
@@ -0,0 +1,133 @@
...
+EXCLUDE_SRCS = $(MAINSRC) $(CORELIB_SRCS) $(LUA_SRC)/luac.c $(LUA_SRC)/linit.c
+CSRCS        = $(filter-out $(EXCLUDE_SRCS),$(wildcard $(LUA_SRC)/*.c))
+CSRCS       += nuttx_linit.c
+
+CFLAGS += -DLUA_MAXINPUT=$(CONFIG_INTERPRETER_LUA_IOBUFSIZE)
+CFLAGS += -DLUA_PROGNAME=\"$(PROGNAME)\"
+
+ifeq ($(CONFIG_INTERPRETER_LUA_32BITS),y)
+CFLAGS += -DLUA_32BITS
```
xiaoxiang781216, line 49: why not let source code check _FILE_OFFSET_BITS instead?

interpreters/lua/Make.defs
```diff
@@ -0,0 +1,31 @@
...
+#
+############################################################################
+
+ifneq ($(CONFIG_INTERPRETERS_LUA),)
+CONFIGURED_APPS += $(APPDIR)/interpreters/lua
+
+# Enable <lua.h> and <lauxlib.h> includes.
+
+CFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" \
+	$(APPDIR)/interpreters/lua/lua-$(CONFIG_INTERPRETER_LUA_VERSION)/src}
```
pkarashchenko, line 27: Do we need to use `$(DELIM)` instead of `/`?

interpreters/lua/Makefile
```diff
@@ -0,0 +1,133 @@
...
+# Lua library
+
+LUA_VERSION  = $(patsubst "%",%,$(strip $(CONFIG_INTERPRETER_LUA_VERSION)))
+LUA_TARBALL  = lua-$(LUA_VERSION).tar.gz
+LUA_UNPACK   = lua-$(LUA_VERSION)
+LUA_URL_BASE = http://www.lua.org/ftp
+LUA_URL      = $(LUA_URL_BASE)/$(LUA_TARBALL)
+LUA_SRC      = $(LUA_UNPACK)/src
+
+MAINSRC      = $(LUA_SRC)/lua.c
```
pkarashchenko, line 39: Do we need to use `$(DELIM)` instead of `/`?

interpreters/lua/Kconfig
```diff
@@ -0,0 +1,90 @@
...
+	select LIBC_LOCALE
+	select LIBM
```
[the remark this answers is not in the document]
mogenson: Since Lua needs libm to build, how about adding `depends on LIBM || ARCH_MATH_H`?

interpreters/lua/nuttx_linit.c
```diff
@@ -0,0 +1,67 @@
...
+
+#include "luamod_proto.h"
+
+/****************************************************************************
+ * Private Data
+ *****************************************************************************/
+
+static const luaL_Reg loadedlibs[] = {
+#ifdef CONFIG_INTERPRETER_LUA_CORE_LIBS
+  {"_G", luaopen_base},
```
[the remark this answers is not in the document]
mogenson: The names and presence of other builtin libraries differ between Lua releases. These other libraries are parsed from `lualib.h` and added to the generated `luamod_list.h`. The base library is present in all Lua release, but the "_G" name is not defined in `lualib.h` so it's included inside an `#ifdef` guard here.

interpreters/lua/Kconfig
```diff
@@ -0,0 +1,90 @@
...
+
+menuconfig INTERPRETER_LUA_SET_PATH
+	bool "Override default search path for Lua modules"
+	default n
+	---help---
+		Set a custom package.path search path for Lua modules.
+
+if INTERPRETER_LUA_SET_PATH
+
+config INTERPRETER_LUA_PATH
```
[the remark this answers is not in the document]
mogenson: Will do.

interpreters/lua/Kconfig
```diff
@@ -0,0 +1,90 @@
...
+	string "Lua interpreter search path for Lua modules"
+	default "/mnt/?.lua;/mnt/?/init.lua;./?.lua;./?/init.lua"
+	---help---
+		Default package.path search path for Lua modules.
+
+endif # INTERPRETER_LUA_SET_PATH
+
+menuconfig INTERPRETER_LUA_SET_CPATH
```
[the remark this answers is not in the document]
mogenson: Will do.

interpreters/lua/Kconfig
mogenson: For the `os.execute()` method in Lua. For example:
```lua
    -- run sleep command from NSH
    os.execute("sleep 1")
```

interpreters/lua/Make.defs
mogenson: I'm following the Make.defs pattern of other interpreters (quickjs, wasm3, wamr, duktape) and libraries (libuv, mbedtls, cmsis). Is there a recommendation for an alternative approach?  All Lua headers are prefixed with an `l` character, so they should hopefully not cause any conflicts.

interpreters/lua/Kconfig
xiaoxiang781216: If all Lua number represent as float, the selection look reasonable. If Lua support integer number is better to let user make the final decision.

interpreters/lua/Kconfig
xiaoxiang781216: I assume these type of function expose through Lua module, it may better to implement the wrapper function like this:
```
lua_execute(...)
{
#ifdef CONFIG_SYSTEM_SYTEM
  system(...);
#else
  report error
#endif
}
```
so, if user don't need pay out the cost if they never use it. This trick is especially useful to some feature which increase a mount of code/data size.

interpreters/lua/Makefile
mogenson: Can you explain the use of `_FILE_OFFSET_BITS`?

By default, Lua will use `long` for integers and `double` for floats. The (not clearly named) `LUA_32BITS` define sets integers to `int` and floats to `float`. The `CONFIG_INTERPRETER_LUA_32BITS` setting is meant as an optional config to adjust the resources required by the Lua interpreter. It doesn't have to match the word size of the target architecture.

mogenson:
@xiaoxiang781216 Thanks for the review. I'll make some changes and update.

Do you have any suggestions for the `Mixed case identifier` linter errors due to Lua types like `lua_State` and `luaL_Reg`?

interpreters/lua/Kconfig
mogenson: I wanted to avoid applying a patch to the Lua source, since it changes from release to release. Since the `system()` call is only used in `loslib.c`, which is only built if `CONFIG_INTERPRETER_LUA_CORE_LIBS` is enabled, how about having `INTERPRETER_LUA_CORE_LIBS` select `SYSTEM_SYSTEM`?

xiaoxiang781216:
> @xiaoxiang781216 Thanks for the review. I'll make some changes and update.
> 
> Do you have any suggestions for the `Mixed case identifier` linter errors due to Lua types like `lua_State` and `luaL_Reg`?

Since it's 3rd party code and used widely, it's better to add them to white list:
https://github.com/apache/incubator-nuttx/blob/master/tools/nxstyle.c#L183-L208

interpreters/lua/Makefile
xiaoxiang781216: > Can you explain the use of `_FILE_OFFSET_BITS`?
> 

LUA_32BITS look like that is only used here:
https://github.com/lua/lua/blob/master/lprefix.h#L25-L28

Since _FILE_OFFSET_BITS is a standard macro, we can define _FILE_OFFSET_BITS correctly from nuttx instead.

> By default, Lua will use `long` for integers and `double` for floats. The (not clearly named) `LUA_32BITS` define sets integers to `int` and floats to `float`.  The `CONFIG_INTERPRETER_LUA_32BITS` setting is meant as an optional config to adjust the resources required by the Lua interpreter. It doesn't have to match the word size of the target architecture.

interpreters/lua/Make.defs
xiaoxiang781216: If so, it's fine to add into the search path.

interpreters/lua/Makefile
mogenson: `LUA_32BITS` is used in `luaconf.h` to set the size for integers and floats in Lua: https://github.com/lua/lua/blob/v5.4.0/luaconf.h#L139-L149

It looks like Lua respects the `_FILE_OFFSET_BITS` macro. It only sets this value is `LUA_32BITS` is not defined and if `_FILE_OFFSET_BITS` is not already defined.

If you prefer, we can not select `LIBC_FLOATINGPOINT` and instead list the optional requirement in the README. One side effect, is without `LIBC_FLOATINGPOINT` you will not be able to print floating point numbers. For example:
```lua
> print(1)
1
> print(0.5)
*float*
```

interpreters/lua/Kconfig
xiaoxiang781216: Ok, it look fine.

interpreters/lua/Makefile
xiaoxiang781216: > `LUA_32BITS` is used in `luaconf.h` to set the size for integers and floats in Lua: https://github.com/lua/lua/blob/v5.4.0/luaconf.h#L139-L149
> 
> It looks like Lua respects the `_FILE_OFFSET_BITS` macro. It only sets this value is `LUA_32BITS` is not defined and if `_FILE_OFFSET_BITS` is not already defined. I don't think any change is required.
> 

Thanks for explanation, look good to me now.

> If you prefer, we can not select `LIBC_FLOATINGPOINT` and instead list the optional requirement in the README. One side effect, is without `LIBC_FLOATINGPOINT` you will not be able to print floating point numbers. For example:
> 
> ```lua
> > print(1)
> 1
> > print(0.5)
> *float*
> ```

Yes, I understand the side effect. Maybe, we can change the output to CONFIG_LIBC_FLOATINGPOINT, so the newbie can know how to enable float output.

interpreters/lua/Makefile
xiaoxiang781216: If so why not compile it to help the module writer?

mogenson: This line is actually filtering out `lauxlib.c` from the `*lib.c` glob. `lauxlib.c` will now be included in the `*.c` glob for `CSRCS`.

mogenson:
> > @xiaoxiang781216 Thanks for the review. I'll make some changes and update.
> > Do you have any suggestions for the `Mixed case identifier` linter errors due to Lua types like `lua_State` and `luaL_Reg`?
> 
> Since it's 3rd party code and used widely, it's better to add them to white list: https://github.com/apache/incubator-nuttx/blob/master/tools/nxstyle.c#L183-L208

Submitted a `nxstyle` PR here: https://github.com/apache/incubator-nuttx/pull/5805

examples/lua_module/luamod_hello.c
```diff
@@ -0,0 +1,76 @@
...
+ * Private Function Prototypes
+ ****************************************************************************/
+
+static int say_hello(lua_State *L);
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static const struct luaL_Reg hello[] =
```
xiaoxiang781216, line 38: add g_ prefix to follow the coding style

interpreters/lua/Makefile
```diff
@@ -0,0 +1,133 @@
...
+STACKSIZE = $(CONFIG_INTERPRETER_LUA_STACKSIZE)
+MODULE    = $(CONFIG_INTERPRETERS_LUA)
+
+# Lua library
+
+LUA_VERSION  = $(patsubst "%",%,$(strip $(CONFIG_INTERPRETER_LUA_VERSION)))
+LUA_TARBALL  = lua-$(LUA_VERSION).tar.gz
+LUA_UNPACK   = lua-$(LUA_VERSION)
+LUA_URL_BASE = http://www.lua.org/ftp
+LUA_URL      = $(LUA_URL_BASE)$(DELIM)$(LUA_TARBALL)
```
xiaoxiang781216, line 36: $(DELIM)->/? $(DELIM) is only used for the local path

interpreters/lua/nuttx_linit.c
```diff
@@ -0,0 +1,68 @@
...
+
+#include <nuttx/config.h>
+
+#include "luamod_proto.h"
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static const luaL_Reg loadedlibs[] =
```
xiaoxiang781216, line 39: add g_ prefix too.

interpreters/lua/Module.mk
```diff
@@ -0,0 +1,37 @@
...
+
+define LUAMOD_REGISTER
+	$(Q) echo Register Lua Module: $1
+	$(Q) echo { \"$1\", $2 }, > "$(LUAMOD_REGISTRY)$(DELIM)$1.bdat"
+	$(Q) echo "int $2(lua_State *L);" > "$(LUAMOD_REGISTRY)$(DELIM)$1.pdat"
+	$(Q) touch "$(LUAMOD_REGISTRY)$(DELIM).updated"
+endef
+
+ifneq ($(LUAMODNAME),)
+
```
xiaoxiang781216, line 31: remove the blank line?

interpreters/lua/nuttx_linit.c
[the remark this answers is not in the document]
xiaoxiang781216: Thanks for explanation. For "_G", look like LUA_GNAME is defined in lauxlib.h? Can we use this macro directly?

interpreters/lua/Makefile
```diff
@@ -0,0 +1,133 @@
...
+
+$(LUA_UNPACK): $(LUA_TARBALL)
+	$(Q) echo "Unpacking $(LUA_TARBALL) to $(LUA_UNPACK)"
+	$(Q) tar -xvzf $(LUA_TARBALL)
+
+context:: $(LUA_UNPACK)
+
+# Register core modules
+
+ifeq ($(CONFIG_INTERPRETER_LUA_CORE_LIBS),y)
```
xiaoxiang781216, line 82: CONFIG_INTERPRETER_LUA_CORE_LIBS->CONFIG_INTERPRETER_LUA_CORELIB to math CORELIB_SRCS?

xiaoxiang781216:
@mogenson the change looks ready for merging, could you add new config to enable lua for Teensy 4.1 and sim? so the daily ci could ensure any new change doesn't break your work.

interpreters/lua/nuttx_linit.c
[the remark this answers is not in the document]
mogenson: The `LUA_GNAME` define was added in Lua 5.4.0. To support building older versions of Lua we cannot depend on this define. The "_G" module name has been constant through all the Lua releases.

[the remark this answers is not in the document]
xiaoxiang781216: if so, we can add this to improve the compatibility:
```
#ifndef LUA_GNAME
#define LUA_GNAME "_G"
#endif
```

mogenson:
> @mogenson the change looks ready for merging, could you add new config to enable lua for Teensy 4.1 and sim? so the daily ci could ensure any new change doesn't break your work.

I'll submit a PR to incubator-nuttx to add defconfigs.

xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216:
let's ignore the false warning:
``
Warning: /home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/interpreters/lua/nuttx_linit.c:52:1: warning: #include outside of 'Included Files' section
```

PR #1092, opened 2022-03-27
trace: fix pointer cross the border

zhuyanlinzyl:
VELAPLATFO-2058

Change-Id: I4cfa400c90d9421dfb565de3536b8378a28ae3fb
Signed-off-by: zhuyanlin <zhuyanlin1@xiaomi.com>

### DIFF
--- a/system/trace/trace.c
+++ b/system/trace/trace.c
@@ -485,29 +485,32 @@ static int trace_cmd_switch(int index, int argc, FAR char **argv,
 
   /* Parse the setting parameters */
 
-  if (argv[index][0] == '-' || argv[index][0] == '+')
+  if (index < argc)
     {
-      enable = (argv[index][0] == '+');
-      if (enable ==
-          ((mode.flag & NOTE_FILTER_MODE_FLAG_SWITCH) != 0))
+      if (argv[index][0] == '-' || argv[index][0] == '+')
         {
-          /* Already set */
+          enable = (argv[index][0] == '+');
+          if (enable ==
+              ((mode.flag & NOTE_FILTER_MODE_FLAG_SWITCH) != 0))
+            {
+              /* Already set */
 
-          return false;
+              return false;
+            }
+
+          if (enable)
+            {
+              mode.flag |= NOTE_FILTER_MODE_FLAG_SWITCH;
+            }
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

system/trace/trace.c
```diff
@@ -735,27 +735,30 @@ static int trace_cmd_print(int index, int argc, FAR char **argv,
 
   /* Parse the setting parameters */
 
-  if (argv[index][0] == '-' || argv[index][0] == '+')
+  while (argv[index])
```
xiaoxiang781216, line 738: don't need while loop, let's check index < argc instead.

xiaoxiang781216 (changes requested):

system/trace/trace.c
xiaoxiang781216: and let's fix trace_cmd_switch too

system/trace/trace.c
```diff
@@ -485,29 +485,32 @@ static int trace_cmd_switch(int index, int argc, FAR char **argv,
 
   /* Parse the setting parameters */
 
-  if (argv[index][0] == '-' || argv[index][0] == '+')
+  while (index < argc)
```
pkarashchenko, line 488:
```suggestion
  if (index < argc)
```

system/trace/trace.c
```diff
@@ -735,29 +738,32 @@ static int trace_cmd_print(int index, int argc, FAR char **argv,
...
```
pkarashchenko, line 741:
```suggestion
  if (index < argc)
```

pkarashchenko (changes requested):

system/trace/trace.c
```diff
@@ -538,7 +541,7 @@ static int trace_cmd_syscall(int index, int argc, FAR char **argv,
 
   /* Parse the setting parameters */
 
-  while (index < argc)
+ while (index < argc)
```
pkarashchenko, line 544:
```suggestion
  while (index < argc)
```

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #1094, opened 2022-03-28
import/Make.defs: Add wildcard for LIBM and LIBGCC

xiaoxiang781216:
## Summary
https://github.com/apache/incubator-nuttx/pull/5864

## Impact
Minor

### DIFF
--- a/import/Make.defs
+++ b/import/Make.defs
@@ -71,7 +71,7 @@ ifeq ($(CONFIG_BUILD_KERNEL),y)
     LDLIBS += -lxx
   endif
 ifneq ($(CONFIG_LIBM),y)
-  LIBM = ${shell "$(CC)" $(ARCHCPUFLAGS) --print-file-name=libm.a}
+  LIBM = ${wildcard ${shell $(CC) $(ARCHCPUFLAGS) --print-file-name=libm.a}}
 ifneq ($(LIBM),".")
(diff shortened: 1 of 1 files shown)


import/Make.defs
```diff
@@ -71,7 +71,7 @@ ifeq ($(CONFIG_BUILD_KERNEL),y)
     LDLIBS += -lxx
   endif
 ifneq ($(CONFIG_LIBM),y)
-  LIBM = ${shell "$(CC)" $(ARCHCPUFLAGS) --print-file-name=libm.a}
+  LIBM = ${wildcard ${shell $(CC) $(ARCHCPUFLAGS) --print-file-name=libm.a}}
```
pkarashchenko, line 74: I've reexamined the case and think that this changes are not required. But instead we probably need to rework build logic not to use `LDLIBPATH` and `LDLIBS` logic for compiler libraries , but use `EXTRA_LIBS` instead.

I will try to apply the change, but most probably next week.

pkarashchenko: @xiaoxiang781216 let's close this PR. When I was writing comment in https://github.com/apache/incubator-nuttx/pull/5864 I was not sure if changes in apps are needed or not, but now I see that they are not needed.

xiaoxiang781216: Sure.

PR #1100, opened 2022-03-28
rexec: correct af type from int to sa_family_t

GUIDINGLI:
## Summary

rexec: correct af type from int to sa_family_t

## Impact
rexec

## Testing
VELA

### DIFF
--- a/netutils/rexec/rexec.c
+++ b/netutils/rexec/rexec.c
@@ -22,6 +22,7 @@
  * Included Files
  ****************************************************************************/
 
+#include <sys/types.h>
 #include <getopt.h>
 #include <netdb.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


pkarashchenko (approved):

netutils/rexec/rexec.c
```diff
@@ -45,8 +46,8 @@ struct rexec_arg_s
   FAR const char *command;
   FAR const char *user;
   FAR char *host;
+  sa_family_t af;
```
pkarashchenko, line 49: why the position in the structure is changed?

pkarashchenko (approved):

netutils/rexec/rexec.c
xiaoxiang781216: Make it sort by length?

PR #1107, opened 2022-03-29
system/libuv: Support specify stack size for worker thread

no1wudi:
## Summary
Add a new option to configure stack size of worker thread in thread pool
## Impact
libuv only
## Testing
VELA


system/libuv/0001-libuv-port-for-nuttx.patch
```diff
@@ -169,7 +169,24 @@ index 869ae95f..8fabeba5 100644
  }
  
  
-@@ -212,7 +217,7 @@ static void init_threads(void) {
+@@ -188,6 +193,16 @@ static void init_threads(void) {
+   const char* val;
+   uv_sem_t sem;
+ 
++  const uv_thread_options_t params = {
++#ifdef DEF_THREADPOOL_STACKSIZE
```
pkarashchenko, line 177: Seems like `DEF_THREADPOOL_STACKSIZE` is always defined, so why do we need to have `ifdef` here?

no1wudi: It's defined on NuttX but not on all other platforms

pkarashchenko: yes, but this patch is to build the code with NuttX and not with other platforms :)
Anyway it is minor, so we can leave it as is.

xiaoxiang781216: We need ensure the same code still can built on other platform and with the result. And we plan to upstream the change to libuv community, so it's important to make the change as general as possible.

system/libuv/0001-libuv-port-for-nuttx.patch
```diff
@@ -169,7 +169,24 @@ index 869ae95f..8fabeba5 100644
...
++    UV_THREAD_HAS_STACK_SIZE,
++    CONFIG_LIBUV_THREAD_STACKSIZE
```
pkarashchenko, line 179: @xiaoxiang781216 following the comment above let's apply this change then
```suggestion
+    DEF_THREADPOOL_STACKSIZE
```

xiaoxiang781216 (approved):

PR #1108, opened 2022-03-29
quickjs: fix compile warning

Gary-Hobson:
## Summary
optind is a global variable in getopt.h (a macro defined in stdlib.h in nuttx).
Here it is used as a variable, not a variable of getopt


pkarashchenko:
I just briefly searched in the code and see that `optind++;` is used in many places. Taking into account that `#define optind                           (*(getoptindp()))` this becomes a useless and all the code that is relying on `optind` after `optind++` is executed wrongly.

xiaoxiang781216:
> I just briefly searched in the code and see that `optind++;` is used in many places. Taking into account that `#define optind (*(getoptindp()))` this becomes a useless and all the code that is relying on `optind` after `optind++` is executed wrongly.

`optind++` is good usage to access the argument after option. The problem here is that quickjs parse the argument by self and doesn't use getopt at all, but the local variable definition conflict with optind unintentionally.

xiaoxiang781216:
@Gary-Hobson please fix this warning:
```
Error: /home/runner/work/incubator-nuttx-apps/incubator-nuttx-apps/apps/interpreters/quickjs/qjsmini.c:393:10: error: Multiple data definitions
```

pkarashchenko:
> > I just briefly searched in the code and see that `optind++;` is used in many places. Taking into account that `#define optind (*(getoptindp()))` this becomes a useless and all the code that is relying on `optind` after `optind++` is executed wrongly.
> 
> `optind++` is good usage to access the argument after option. The problem here is that quickjs parse the argument by self and doesn't use getopt at all, but the local variable definition conflict with optind unintentionally.

@xiaoxiang781216 I'm talking not about the problem here. The author solves it by renaming of a variable.
The case is that in Linux (https://linux.die.net/man/3/optind) the `optind` is `int optind` and referenced via `extern int optind`, but in NuttX it is `FAR int *getoptindp(void);  /* Index into argv */` and `#define optind (*(getoptindp()))` so `optind++;` does not increment the index holder.
Please try to run a small example locally
```
#include <stdio.h>

#define optind (*(getoptindp()))

int *getoptindp(void)
{
    static int a;
    return &a;
}

int main()
{
    printf("Test: %d, %d", optind++, optind);

    return 0;
}
```
and you will see that zero is printed two times.

pkarashchenko:
I'm just saying that this PR highlights places in `boards/arm/lpc31xx/ea3131/tools/lpchdr.c`, `boards/arm/lpc31xx/ea3152/tools/lpchdr.c`, `boards/arm/lpc31xx/olimex-lpc-h3131/tools/lpchdr.c` and `optind++;` is not usable in NuttX environment.

xiaoxiang781216:
> > > I just briefly searched in the code and see that `optind++;` is used in many places. Taking into account that `#define optind (*(getoptindp()))` this becomes a useless and all the code that is relying on `optind` after `optind++` is executed wrongly.
> > 
> > 
> > `optind++` is good usage to access the argument after option. The problem here is that quickjs parse the argument by self and doesn't use getopt at all, but the local variable definition conflict with optind unintentionally.
> 
> @xiaoxiang781216 I'm talking not about the problem here. The author solves it by renaming of a variable. The case is that in Linux (https://linux.die.net/man/3/optind) the `optind` is `int optind` and referenced via `extern int optind`, but in NuttX it is `FAR int *getoptindp(void); /* Index into argv */` and `#define optind (*(getoptindp()))` so `optind++;` does not increment the index holder. Please try to run a small example locally
> 
> ```
> #include <stdio.h>
> 
> #define optind (*(getoptindp()))
> 
> int *getoptindp(void)
> {
>     static int a;
>     return &a;
> }
> 
> int main()
> {
>     printf("Test: %d, %d", optind++, optind);
> 
>     return 0;
> }
> ```
> 
> and you will see that zero is printed two times.

But your test isn't good since compiler may evaluate the last argument first(compiler can evaluate the expression in any order).  I change your printf to this:
```
printf("Test: %d, %d\n", optind++, optind++);
```
The output show that the last argument evaluate first:
```
Test: 1, 0
```
But of course, it may change in your environment since the spec say this is an undefined behavior and anything could happen.

pkarashchenko:
> > > > I just briefly searched in the code and see that `optind++;` is used in many places. Taking into account that `#define optind (*(getoptindp()))` this becomes a useless and all the code that is relying on `optind` after `optind++` is executed wrongly.
> > > 
> > > 
> > > `optind++` is good usage to access the argument after option. The problem here is that quickjs parse the argument by self and doesn't use getopt at all, but the local variable definition conflict with optind unintentionally.
> > 
> > 
> > @xiaoxiang781216 I'm talking not about the problem here. The author solves it by renaming of a variable. The case is that in Linux (https://linux.die.net/man/3/optind) the `optind` is `int optind` and referenced via `extern int optind`, but in NuttX it is `FAR int *getoptindp(void); /* Index into argv */` and `#define optind (*(getoptindp()))` so `optind++;` does not increment the index holder. Please try to run a small example locally
> > ```
> > #include <stdio.h>
> > 
> > #define optind (*(getoptindp()))
> > 
> > int *getoptindp(void)
> > {
> >     static int a;
> >     return &a;
> > }
> > 
> > int main()
> > {
> >     printf("Test: %d, %d", optind++, optind);
> > 
> >     return 0;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and you will see that zero is printed two times.
> 
> But your test isn't good since compiler may evaluate the last argument first(compiler can evaluate the expression in any order). I change your printf to this:
> 
> ```
> printf("Test: %d, %d\n", optind++, optind++);
> ```
> 
> The output show that the last argument evaluate first:
> 
> ```
> Test: 1, 0
> ```
> 
> But of course, it may change in your environment since the spec say this is an undefined behavior and anything could happen.

Sorry I completely messed the things with pointer. Probably need to have a good rest.

pkarashchenko (approved):

xiaoxiang781216:
> Sorry I completely messed the things with pointer. Probably need to have a good rest.

I think so, look like you work harder than me:).

xiaoxiang781216 (approved):

PR #1112, opened 2022-03-30
testing/mm: add heap boundary realloc test.

CV-Bowen:
## Summary
Add the heap boundary realloc test.

When CONFIG_MM_SMALL enable and CONFIG_DEBUG_MM disable, **SIZEOF_MM_ALLOCNODE = 4, MM_MIN_CHUNK = 16, SIZEOF_MM_FRENODE = 12**.

image
In above figure, **HEAP_SIZE** is not aligned with **MM_MIN_CHUNK** (because **(2 * SIZEOF_MM_ALLOCNODE) != MM_MIN_CHUNK)**), nodeA is a malloced memory node, nodeB is a free memory node and adjacent to heap end node. In this situation, realloc the nodeA may lead the remain size be **(MM_MIN_CHUNK - 2 * SIZEOF_MM_ALLOCNODE) = 8**, which is not large enough to hold a free node, so software will assert fail at mm_heap/mm_realloc.c line: 319.

Related PR: https://github.com/apache/incubator-nuttx/pull/5906

@masayuki2009

## Testing
Test with nucleo-144 stm32f746zg

### DIFF
--- a/testing/mm/Makefile
+++ b/testing/mm/Makefile
@@ -29,4 +29,6 @@ MODULE    = $(CONFIG_TESTING_MM)
 
 MAINSRC = mm_main.c
 
+CFLAGS += -I$(TOPDIR)/mm/mm_heap
+
 include $(APPDIR)/Application.mk
--- a/testing/mm/mm_main.c
+++ b/testing/mm/mm_main.c
@@ -26,6 +26,12 @@
 #include <stdlib.h>
 #include <malloc.h>
 #include <string.h>
+#include <queue.h>
+#include <assert.h>
+
+/* Include nuttx/mm/mm_heap/mm.h */
+
+#include <mm.h>
 
 /****************************************************************************
  * Pre-processor Definitions
@@ -36,20 +42,6 @@
 /* #define STOP_ON_ERRORS do {} while (0) */
 
 #define STOP_ON_ERRORS exit(1)
-
-/* All other definitions derive from these two */
-
-#define MM_MIN_SHIFT      4  /* 16 bytes */
-#define MM_MIN_CHUNK     (1 << MM_MIN_SHIFT)
-#define MM_GRAN_MASK     (MM_MIN_CHUNK - 1)
-#define MM_ALIGN_UP(a)   (((a) + MM_GRAN_MASK) & ~MM_GRAN_MASK)
-#define MM_ALIGN_DOWN(a) ((a) & ~MM_GRAN_MASK)
-
-#ifdef CONFIG_SMALL_MEMORY
-# define SIZEOF_MM_ALLOCNODE   4
-#else
-# define SIZEOF_MM_ALLOCNODE   8
-#endif
 
 /****************************************************************************
  * Private Data
@@ -105,6 +97,8 @@ static const int g_alignment[NTEST_ALLOCS / 2] =
 
 static FAR void       *g_allocs[NTEST_ALLOCS];
 static struct mallinfo g_alloc_info;
+
+static dq_queue_t g_realloc_queue;
 
 /****************************************************************************
  * Private Functions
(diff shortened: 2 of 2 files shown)


pkarashchenko (approved):

masayuki2009:
@CV-Bowen 

How long does it take to finish the testing on the sabre-6quad:smp (1GB RAM configuration)?

CV-Bowen:
@masayuki2009 , sorry, i didn't consider the big memory condition. I will find a better solution.

CV-Bowen:
@masayuki2009 After thinking for a while, i think this question should not avoid by test, because this is clearly a design mistake in memory manager. Maybe adding a DEBUGASSERT in mm_initailize() is more efficient. What about your opinion?

CV-Bowen:
@masayuki2009 @pkarashchenko I found an memory leak question when working on this, please have a look. https://github.com/apache/incubator-nuttx-apps/pull/1113

xiaoxiang781216:
@masayuki2009 the change ensure the heap memory align with MM_MIN_CHUNK, which is a core heap manager assumption. Why this issue isn't reported before? Because the default config(CONFIG_MM_SMALL=n), 2 * SIZEOF_MM_ALLOCNODE(2 * 8)(start + end node) just equals MM_MIN_CHUNK(16). But once CONFIG_MM_SMALL is enabled, 2 * SIZEOF_MM_ALLOCNODE become 2 * 4(8).

masayuki2009:
@xiaoxiang781216 

>@masayuki2009 the change ensure the heap memory align with MM_MIN_CHUNK, which is a core heap manager assumption. Why this issue isn't reported before? Because the default config(CONFIG_MM_SMALL=n), 2 * SIZEOF_MM_ALLOCNODE(2 * 8)(start + end node) just equals MM_MIN_CHUNK(16). But once CONFIG_MM_SMALL is enabled, 2 * SIZEOF_MM_ALLOCNODE become 2 * 4(8).

So if the https://github.com/apache/incubator-nuttx/pull/5906 is not merged, the issue only happens with `CONFIG_MM_SMALL=y` configuration, right?

masayuki2009:
>@masayuki2009 , sorry, i didn't consider the big memory condition. I will find a better solution.

If the issue only happens `CONFIG_MM_SMALL=y`, how about enabling the new test case for it?

xiaoxiang781216:
> @xiaoxiang781216
> 
> > @masayuki2009 the change ensure the heap memory align with MM_MIN_CHUNK, which is a core heap manager assumption. Why this issue isn't reported before? Because the default config(CONFIG_MM_SMALL=n), 2 * SIZEOF_MM_ALLOCNODE(2 * 8)(start + end node) just equals MM_MIN_CHUNK(16). But once CONFIG_MM_SMALL is enabled, 2 * SIZEOF_MM_ALLOCNODE become 2 * 4(8).
> 
> So if the [apache/incubator-nuttx#5906](https://github.com/apache/incubator-nuttx/pull/5906) is not merged, the issue only happens with `CONFIG_MM_SMALL=y` configuration, right?

Yes, or CONFIG_DEBUG_MM==y since both option change mm_allocnode_s layout and then may make the assumption fail.

masayuki2009:
>Yes, or CONFIG_DEBUG_MM==y since both option change mm_allocnode_s layout and then may make the assumption fail.

@xiaoxiang781216 @CV-Bowen 
OK, so the test case should consider the option as well which means that the new test case would run with large heap memory.

CV-Bowen:
@masayuki2009 I have completed a initial version, if you want to recurrent the question with config sabre-6quad:smp, please enable the CONFIG_MM_DEBUG.

CV-Bowen:
I think redefine the marcos that have defined in nuttx/mm/mm_heap/mm.h is not a good method, but i don't find a good solution. Do you have some advices? @masayuki2009 @xiaoxiang781216

CV-Bowen:
@xiaoxiang781216 Done.

xiaoxiang781216 (approved):

PR #1113, opened 2022-03-30
testing/mm: repair the memory leak question when realloc failed.

CV-Bowen:
## Summary
Repair the memory leak question when realloc failed.
realloc return NULL if failed, so we have to use a local variable to store the return value to avoid the missing of old memory pointer.

## Testing
mm test pass, no memory leak.


CV-Bowen:
@xiaoxiang781216

xiaoxiang781216 (approved):

PR #1114, opened 2022-03-30
sched/note: add support of trace section mark 

anchao:
## Summary

sched/note: add support of trace section mark

The implementation of this feature is based on android systrace:

https://source.android.com/devices/tech/debug/ftrace

Application developers are more concerned about the performance of
the specified application section,
added two APIs to implement performance measurement:

```
void sched_note_begin(FAR const char *str);
void sched_note_end(FAR const char *str);
or
SCHED_NOTE_BEGIN();  /* defined to sched_note_begin(__FUNCTION__) */
SCHED_NOTE_END();    /* defined to sched_note_end(__FUNCTION__) */
```

## Impact

Depends on: https://github.com/apache/incubator-nuttx/pull/5915

## Testing

https://source.android.com/devices/tech/debug/ftrace
https://github.com/catapult-project/catapult/

test code:

```
void ccc(void)
{
  SCHED_NOTE_BEGIN();
  usleep(10);
  SCHED_NOTE_END();
}

void bbb(void)
{
  SCHED_NOTE_BEGIN();
  ccc();
  SCHED_NOTE_END();
}

void aaa(void)
{
  SCHED_NOTE_BEGIN();
  bbb();
  SCHED_NOTE_END();
}

{
  while (1)
    {
      SCHED_NOTE_BEGIN();
      aaa();
      SCHED_NOTE_END();
    }
  return 0;
}

```

Screenshot from 2022-03-30 22-34-16

## How To Use

https://nuttx.apache.org/docs/latest/guides/tasktraceuser.html

1. enable sched  instrumentation

```
CONFIG_DRIVER_NOTERAM=y
...
CONFIG_SCHED_INSTRUMENTATION=y
CONFIG_SCHED_INSTRUMENTATION_SWITCH=y
CONFIG_SCHED_INSTRUMENTATION_SYSCALL=y
CONFIG_SCHED_INSTRUMENTATION_IRQHANDLER=y
CONFIG_SCHED_INSTRUMENTATION_DUMP=y
```
2. Capture sched note info

```
nsh> trace switch +
nsh> trace syscall +
nsh> trace print +
nsh> trace start
nsh> trace dump
```

3. convert to systrace

https://source.android.com/devices/tech/debug/ftrace

If you captured your systrace with --no-compress, this is in the html file in the section beginning with BEGIN TRACE.
If not, run html2trace from the [Catapult tree](https://github.com/catapult-project/catapult/tree/master/) (tracing/bin/html2trace) to uncompress the trace.

```
$ git clone git@github.com:catapult-project/catapult.git
$ cd catapult/
$ ./tracing/bin/trace2html trace.txt
trace.html
```
[systrace.txt](https://github.com/apache/incubator-nuttx/files/8381565/systrace.txt)

4. open the trace html via web browser:
`$ google-chrome trace.html`

### DIFF
--- a/system/sched_note/note_main.c
+++ b/system/sched_note/note_main.c
@@ -84,6 +84,36 @@ static FAR const char *g_statenames[] =
  ************************************************************************************/
 
 /************************************************************************************
+ * Name: trace_dump_unflatten
+ ************************************************************************************/
+
+static void trace_dump_unflatten(FAR void *dst,
+                                 FAR uint8_t *src, size_t len)
+{
+  switch (len)
+    {
+#ifdef CONFIG_HAVE_LONG_LONG
+      case 8:
+        *(uint64_t *)dst = ((uint64_t)src[7] << 56)
+                         + ((uint64_t)src[6] << 48)
+                         + ((uint64_t)src[5] << 40)
+                         + ((uint64_t)src[4] << 32);
+#endif
+      case 4:
+        *(uint32_t *)dst = ((uint64_t)src[3] << 24)
+                         + ((uint64_t)src[2] << 16);
+      case 2:
+        *(uint16_t *)dst = ((uint64_t)src[1] << 8);
+      case 1:
+        *(uint8_t *)dst = src[0];
+        break;
+      default:
+        DEBUGASSERT(FALSE);
+        break;
+    }
+}
+
+/************************************************************************************
  * Name: dump_notes
  ************************************************************************************/
 
@@ -103,22 +133,14 @@ static void dump_notes(size_t nread)
   while (offset < nread)
     {
       note    = (FAR struct note_common_s *)&g_note_buffer[offset];
-      pid     =  (pid_t)note->nc_pid[0] +
-                ((pid_t)note->nc_pid[1] << 8);
+      trace_dump_unflatten(&pid, note->note->nc_pid, sizeof(pid));
 #ifdef CONFIG_SCHED_INSTRUMENTATION_HIRES
-      systime_nsec = (uint32_t)note->nc_systime_nsec[0] +
-                     (uint32_t)(note->nc_systime_nsec[1] << 8) +
-                     (uint32_t)(note->nc_systime_nsec[2] << 16) +
-                     (uint32_t)(note->nc_systime_nsec[3] << 24);
-      systime_sec = (uint32_t)note->nc_systime_sec[0] +
-                    (uint32_t)(note->nc_systime_sec[1] << 8) +
-                    (uint32_t)(note->nc_systime_sec[2] << 16) +
-                    (uint32_t)(note->nc_systime_sec[3] << 24);
+      trace_dump_unflatten(&systime_nsec,
+                           note->nc_systime_nsec, sizeof(systime_nsec));
+      trace_dump_unflatten(&systime_sec,
+                           note->nc_systime_sec, sizeof(systime_sec));
 #else
-      systime = (uint32_t) note->nc_systime[0]        +
-                (uint32_t)(note->nc_systime[1] << 8)  +
-                (uint32_t)(note->nc_systime[2] << 16) +
-                (uint32_t)(note->nc_systime[3] << 24);
+      trace_dump_unflatten(&systime, note->nc_systime, sizeof(systime));
 #endif
 
       switch (note->nc_type)
@@ -395,8 +417,7 @@ static void dump_notes(size_t nread)
                   return;
                 }
 
-              count = (uint16_t) note_preempt->npr_count[0] +
-                      (uint16_t)(note_preempt->npr_count[1] << 8);
+              trace_dump_unflatten(&count, note_preempt->npr_count, sizeof(count));
 
               if (note->nc_type == NOTE_PREEMPT_LOCK)
                 {
@@ -452,8 +473,7 @@ static void dump_notes(size_t nread)
                 }
 
 #ifdef CONFIG_SMP
-              count = (uint16_t) note_csection->ncs_count[0] +
-                      (uint16_t)(note_csection->ncs_count[1] << 8);
+              trace_dump_unflatten(&count, note_csection->ncs_count, sizeof(count));
 
               if (note->nc_type == NOTE_CSECTION_ENTER)
                 {
@@ -509,20 +529,9 @@ static void dump_notes(size_t nread)
                   return;
                 }
 
-              spinlock = (FAR void *)
-                            ((uintptr_t)note_spinlock->nsp_spinlock[0]
-                             + ((uintptr_t)note_spinlock->nsp_spinlock[1] << 8)
-#if UINTPTR_MAX > UINT16_MAX
-                             + ((uintptr_t)note_spinlock->nsp_spinlock[2] << 16)
-                             + ((uintptr_t)note_spinlock->nsp_spinlock[3] << 24)
-#if UINTPTR_MAX > UINT32_MAX
-                             + ((uintptr_t)note_spinlock->nsp_spinlock[4] << 32)
-                             + ((uintptr_t)note_spinlock->nsp_spinlock[5] << 40)
-                             + ((uintptr_t)note_spinlock->nsp_spinlock[6] << 48)
-                             + ((uintptr_t)note_spinlock->nsp_spinlock[7] << 56)
-#endif
-#endif
-                            );
+              trace_dump_unflatten(&spinlock,
+                                   note_spinlock->nsp_spinlock,
+                                   sizeof(spinlock));
 
              switch (note->nc_type)
                {
@@ -666,19 +675,9 @@ static void dump_notes(size_t nread)
                         return;
                       }
 
-                    result =    (uintptr_t)note_sysleave->nsc_result[0]
-                             + ((uintptr_t)note_sysleave->nsc_result[1] << 8)
-#if UINTPTR_MAX > UINT16_MAX
-                             + ((uintptr_t)note_sysleave->nsc_result[2] << 16)
-                             + ((uintptr_t)note_sysleave->nsc_result[3] << 24)
-#if UINTPTR_MAX > UINT32_MAX
-                             + ((uintptr_t)note_sysleave->nsc_result[4] << 32)
-                             + ((uintptr_t)note_sysleave->nsc_result[5] << 40)
-                             + ((uintptr_t)note_sysleave->nsc_result[6] << 48)
-                             + ((uintptr_t)note_sysleave->nsc_result[7] << 56)
-#endif
-#endif
-                    ;
+                    trace_dump_unflatten(&result,
+                                         note_sysleave->nsc_result,
+                                         sizeof(result));
 
                     syslog_time(LOG_INFO,
                            "Task %u Leave SYSCALL %d: %" PRIdPTR "\n",
@@ -738,6 +737,7 @@ static void dump_notes(size_t nread)
                   {
                     FAR struct note_binary_s *note_binary =
                       (FAR struct note_binary_s *)note;
+                    uint32_t module;
                     char out[1280];
                     int count;
                     int ret = 0;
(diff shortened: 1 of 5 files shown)


system/sched_note/note_main.c
```diff
@@ -103,22 +103,14 @@ static void dump_notes(size_t nread)
   while (offset < nread)
     {
       note    = (FAR struct note_common_s *)&g_note_buffer[offset];
-      pid     =  (pid_t)note->nc_pid[0] +
-                ((pid_t)note->nc_pid[1] << 8);
+      trace_dump_unflatten(&pid, note->note->nc_pid, sizeof(pid));
```
xiaoxiang781216, line 136: merge into the next patch

system/trace/trace.h
```diff
@@ -37,6 +37,20 @@ extern "C"
...
+ ****************************************************************************/
+
+typedef enum
+{
+  TRACE_TYPE_LTTNG_KERNEL = 0,  /* Common Trace Format : Linux Kernel Trace */
+  TRACE_TYPE_GENERIC_CTF  = 1,  /* Common Trace Format : Generic CTF Trace */
+  TRACE_TYPE_LTTNG_UST    = 2,  /* Common Trace Format : LTTng UST Trace */
+  TRACE_TYPE_CUSTOM_TEXT  = 3,  /* Custom Text :         TmfGeneric */
+  TRACE_TYPE_CUSTOM_XML   = 4,  /* Custom XML :          Custom XML Log */
+  TRACE_TYPE_ANDROID      = 5,  /* Custom Format :       Android ATrace */
```
xiaoxiang781216, line 51: should we remove the unsupported format now?

system/trace/trace.h
```diff
@@ -83,12 +83,23 @@ bool trace_dump_get_overwrite(void);
...
 
+/****************************************************************************
+ * Name: trace_dump_unflatten
+ *
+ * Description:
+ *   unflatten the dump buffer
+ *
+ ****************************************************************************/
+
+void trace_dump_unflatten(FAR void *dst, FAR uint8_t *src, size_t len);
```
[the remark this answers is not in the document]
anchao: Done

system/trace/trace.h
anchao: Done

system/sched_note/note_main.c
anchao: Done

system/trace/trace_dump.c
```diff
@@ -657,6 +633,35 @@ static int trace_dump_one(FAR FILE *out,
  * Public Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: trace_dump_unflatten
+ ****************************************************************************/
+
+void trace_dump_unflatten(FAR void *dst, FAR uint8_t *src, size_t len)
```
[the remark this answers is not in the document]
anchao: Done

system/trace/trace.h
xiaoxiang781216: no change?

system/trace/trace.h
```diff
@@ -37,6 +37,20 @@ extern "C"
 #define EXTERN extern
 #endif
 
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+typedef enum
+{
+  TRACE_TYPE_LTTNG_KERNEL = 0,  /* Common Trace Format : Linux Kernel Trace */
```
xiaoxiang781216, line 46: LTTNG_KERNEL->LINUX？

system/trace/trace.c
```diff
@@ -142,10 +142,22 @@ static int trace_cmd_start(int index, int argc, FAR char **argv,
...
                           int notectlfd)
 {
+  trace_dump_t type = TRACE_TYPE_LTTNG_KERNEL;
   FAR FILE *out = stdout;
-  int ret;
   bool changed = false;
   bool cont = false;
+  int ret;
+
+  /* Usage: trace dump [-t] <dumptype> */
```
xiaoxiang781216, line 151: [-t]->[-l|-a] instead number to help people

xiaoxiang781216, line 151: update the help message

anchao: Done

anchao: Done

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

system/sched_note/note_main.c
```diff
@@ -83,6 +83,36 @@ static FAR const char *g_statenames[] =
...
+                         + ((uint64_t)src[4] << 32);
+#endif
+      case 4:
+        *(uint32_t *)dst = ((uint64_t)src[3] << 24)
+                         + ((uint64_t)src[2] << 16);
+      case 2:
+        *(uint16_t *)dst = ((uint64_t)src[1] << 8);
+      case 1:
+        *(uint8_t *)dst = src[0];
+        break;
```
pkarashchenko, line 109: This code seems to be totally wrong from many perspectives.
We need to use temporary variable here and memcpy at the end:
```
#ifdef CONFIG_HAVE_LONG_LONG
  uint64_t tmp = 0;
#else
  uint32_t tmp = 0;
#endif

  switch (len)
    {
#ifdef CONFIG_HAVE_LONG_LONG
      case 8:
        tmp += ((uint64_t)src[7] << 56);
        tmp += ((uint64_t)src[6] << 48);
        tmp += ((uint64_t)src[5] << 40);
        tmp += ((uint64_t)src[4] << 32);
#endif
      case 4:
        tmp += ((uint32_t)src[3] << 24);
        tmp += ((uint32_t)src[2] << 16);
      case 2:
        tmp += ((uint16_t)src[1] << 8);
      case 1:
        tmp += src[0];
        break;
      default:
        DEBUGASSERT(FALSE);
        break;
    }

  memcpy(dst, &tmp, len);
```
In this way we will keep data in host byte order. But I'm not sure if that is the goal. If I take a look into a scheduler variant it produce little endian always:
```
static inline void sched_note_flatten(FAR uint8_t *dst,
                                      FAR void *src, size_t len)
{
  switch (len)
    {
#ifdef CONFIG_HAVE_LONG_LONG
      case 8:
        dst[7] = (uint8_t)((*(uint64_t *)src >> 56) & 0xff);
        dst[6] = (uint8_t)((*(uint64_t *)src >> 48) & 0xff);
        dst[5] = (uint8_t)((*(uint64_t *)src >> 40) & 0xff);
        dst[4] = (uint8_t)((*(uint64_t *)src >> 32) & 0xff);
#endif
      case 4:
        dst[3] = (uint8_t)((*(uint32_t *)src >> 24) & 0xff);
        dst[2] = (uint8_t)((*(uint32_t *)src >> 16) & 0xff);
      case 2:
        dst[1] = (uint8_t)((*(uint16_t *)src >> 8) & 0xff);
      case 1:
        dst[0] = (uint8_t)(*(uint8_t *)src & 0xff);
        break;
      default:
        DEBUGASSERT(FALSE);
        break;
    }
}
```

pkarashchenko, line 109: This code seems to be totally wrong.
We need to use temporary variable here and memcpy at the end:
```
  uint64_t tmp = 0;

  switch (len)
    {
#ifdef CONFIG_HAVE_LONG_LONG
      case 8:
        tmp += ((uint64_t)src[7] << 56);
        tmp += ((uint64_t)src[6] << 48);
        tmp += ((uint64_t)src[5] << 40);
        tmp += ((uint64_t)src[4] << 32);
#endif
      case 4:
        tmp += ((uint64_t)src[3] << 24);
        tmp += ((uint64_t)src[2] << 16);
      case 2:
        tmp += ((uint64_t)src[1] << 8);
      case 1:
        tmp += src[0];
        break;
      default:
        DEBUGASSERT(FALSE);
        break;
    }

  memcpy(dst, &tmp, len);
```

xiaoxiang781216: > This code seems to be totally wrong from many perspectives. We need to use temporary variable here and memcpy at the end:
> 
> ```
> #ifdef CONFIG_HAVE_LONG_LONG
>   uint64_t tmp = 0;
> #else
>   uint32_t tmp = 0;
> #endif
> 
>   switch (len)
>     {
> #ifdef CONFIG_HAVE_LONG_LONG
>       case 8:
>         tmp += ((uint64_t)src[7] << 56);
>         tmp += ((uint64_t)src[6] << 48);
>         tmp += ((uint64_t)src[5] << 40);
>         tmp += ((uint64_t)src[4] << 32);
> #endif
>       case 4:
>         tmp += ((uint32_t)src[3] << 24);
>         tmp += ((uint32_t)src[2] << 16);
>       case 2:
>         tmp += ((uint16_t)src[1] << 8);
>       case 1:
>         tmp += src[0];
>         break;
>       default:
>         DEBUGASSERT(FALSE);
>         break;
>     }
> 
>   memcpy(dst, &tmp, len);
> ```
> 

it is no difference from cast the dest to uint64_t * and assign the value directly.

> In this way we will keep data in host byte order. But I'm not sure if that is the goal.

The code assume the source is in little endian and convert to the host endian.

> If I take a look into a scheduler variant it produce little endian always:
> 
> ```
> static inline void sched_note_flatten(FAR uint8_t *dst,
>                                       FAR void *src, size_t len)
> {
>   switch (len)
>     {
> #ifdef CONFIG_HAVE_LONG_LONG
>       case 8:
>         dst[7] = (uint8_t)((*(uint64_t *)src >> 56) & 0xff);
>         dst[6] = (uint8_t)((*(uint64_t *)src >> 48) & 0xff);
>         dst[5] = (uint8_t)((*(uint64_t *)src >> 40) & 0xff);
>         dst[4] = (uint8_t)((*(uint64_t *)src >> 32) & 0xff);
> #endif
>       case 4:
>         dst[3] = (uint8_t)((*(uint32_t *)src >> 24) & 0xff);
>         dst[2] = (uint8_t)((*(uint32_t *)src >> 16) & 0xff);
>       case 2:
>         dst[1] = (uint8_t)((*(uint16_t *)src >> 8) & 0xff);
>       case 1:
>         dst[0] = (uint8_t)(*(uint8_t *)src & 0xff);
>         break;
>       default:
>         DEBUGASSERT(FALSE);
>         break;
>     }
> }
> ```

Yes, the dest is in the little endian.

pkarashchenko: @xiaoxiang781216 I think you misunderstood my comment. I sill point the think more clear:
1. The merged code performs cast to `(uint64_t)` even if `CONFIG_HAVE_LONG_LONG` is not set. I think this is wrong
2. In case of 8, 4 or 2 byte values we get a mess based on different endianess at `dst` because `dst` is not update and `=` is used. For for example of 4 byte value the `dst` byte 0, 1, 2 and 3 will be set to `0, 0, src[3], src[4]` or to `src[4], src[3], 0, 0` and then due to fall-thru to `case 2:` and `case 1:` the part will be overwritten by `src[1]` and `src[0]`. So this code seems to be working on little-endian archs only.

xiaoxiang781216: > @xiaoxiang781216 I think you misunderstood my comment. I sill point the think more clear:
> 
> 1. The merged code performs cast to `(uint64_t)` even if `CONFIG_HAVE_LONG_LONG` is not set. I think this is wrong

Ok, I focus on "tmp and memcpy". For "((uint64_t)src[3] << 24)...", it's better to change to "((uint32_t)src[3] << 24)..." @anchao .

> 2. In case of 8, 4 or 2 byte values we get a mess based on different endianess at `dst` because `dst` is not update and `=` is used. For for example of 4 byte value the `dst` byte 0, 1, 2 and 3 will be set to `0, 0, src[3], src[4]` or to `src[4], src[3], 0, 0` and then due to fall-thru to `case 2:` and `case 1:` the part will be overwritten by `src[1]` and `src[0]`. So this code seems to be working on little-endian archs only.

If so, sched_note_flatten has the similar issue, @anchao please take a look.

xiaoxiang781216: Yes, but src use the similar approach as sched_note_unflatten

anchao: @pkarashchenko Thanks for the comments, I have fix this issue, please review:
https://github.com/apache/incubator-nuttx/pull/5992
https://github.com/apache/incubator-nuttx-apps/pull/1126

PR #1117, opened 2022-04-01
open() options: pass file options 0 if only ioctl is used

pkarashchenko:
## Summary
Pass file options 0 if only ioctl is used. if write or read handler is not available then `-EACCES` is returned by `inode_checkflags`

## Impact
Fix cfgdata and pca9635 example operation.

## Testing
Tested with custom board based on SAME70

### DIFF
--- a/examples/pca9635/pca9635_main.c
+++ b/examples/pca9635/pca9635_main.c
@@ -59,7 +59,7 @@ int main(int argc, FAR char *argv[])
   int fd;
   int ret;
 
-  fd = open(CONFIG_EXAMPLES_PCA9635_DEVNAME, O_WRONLY);
+  fd = open(CONFIG_EXAMPLES_PCA9635_DEVNAME, 0);
   if (fd < 0)
     {
       fprintf(stderr, "ERROR: Failed to open %s: %d\n",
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

acassis:
Hi @pkarashchenko I saw your message in the mailing list, but I still thinking that passing 0 to open is wrong.
Please read the man pages: $ man 2 open
...
```
       The argument flags must include one of the  following  access  modes:  O_RDONLY,  O_WRONLY,  or
       O_RDWR.  These request opening the file read-only, write-only, or read/write, respectively
```
The "must" means it is required

pkarashchenko:
So if driver does not have neither read not write methods then opening with flags will fail due to VFS check.

pkarashchenko:
So if driver does not have neither read not write methods then opening with flags will fail due to VFS check.
@acassis I think you are referring Linux man, and not POSIX, right?

pkarashchenko:
BTW, `examples/lvgldemo/lcddev.c` also use `state.fd = open(lcddev, 0);`

pkarashchenko:
More cases:
- `examples/smps/smps_main.c` - `fd = open(CONFIG_EXAMPLES_SMPS_DEVPATH, 0);`
- `examples/powerled/powerled_main.c` - `fd = open(CONFIG_EXAMPLES_POWERLED_DEVPATH, 0);`

PR #1124, opened 2022-04-06
Symtab vs builtin

yamt:
## Summary
Use -fno-builtin for SYMTABOBJ

The code generated by tools/mksymtab.sh uses
```
extern void *var;
```
for ~everything.

After the recent removal of -fno-builtin, [1]
it ends up with warnings:
* -Wbuiltin-requires-header for clang
* -Wbuiltin-declaration-mismatch for gcc

It also generates errors like the following for clang:
```
symtab_apps.c:125:14: error: redefinition of 'strdup' as different kind of symbol
extern void *strdup;
```
I couldn't find a way to disable it.
(it's err_redefinition_different_kind in clang source)

This commit works it around by restoring -fno-builtin
when building SYMTABOBJ.

[1] https://github.com/apache/incubator-nuttx/pull/5476

## Testing

tested with my app + sim/linux + clang


xiaoxiang781216 (approved):

xiaoxiang781216:
syscall may need the similar change too

yamt:
> syscall may need the similar change too

do you mean the table generated from syscall.csv and libc.csv?
i guess they are ok as they use correct prototypes.

examples/thttpd/Makefile
```diff
@@ -31,6 +31,8 @@ ifeq ($(CONFIG_THTTPD_BINFS),y)
 else
   CONTENT_MAKE += -f Makefile.nxflat
   CSRCS += symtab.c
+  symtab.c_CFLAGS = -fno-builtin
```
pkarashchenko, line 34: @xiaoxiang781216 maybe we also can use the similar for memcpy and other functions instead of adding an attribute that was added to compiler.h?

pkarashchenko (approved):

xiaoxiang781216:
> > syscall may need the similar change too
> 
> do you mean the table generated from syscall.csv and libc.csv? i guess they are ok as they use correct prototypes.

Yes, you are right. After search "Wbuiltin-declaration-mismatch", only https://github.com/apache/incubator-nuttx/blob/master/tools/mkallsyms.sh need the similar change.

yamt:
> > > syscall may need the similar change too
> > 
> > 
> > do you mean the table generated from syscall.csv and libc.csv? i guess they are ok as they use correct prototypes.
> 
> Yes, you are right. After search "Wbuiltin-declaration-mismatch", only https://github.com/apache/incubator-nuttx/blob/master/tools/mkallsyms.sh need the similar change.

https://github.com/apache/incubator-nuttx/pull/6247

examples/elf/Makefile
```diff
@@ -30,6 +30,8 @@ CSRCS = cromfs.c
 endif
 CSRCS += dirlist.c
 CSRCS += symtab.c
+tests$(DELIM)symtab.c_CFLAGS = -fno-builtin
```
xiaoxiang781216, line 33: should we expose -fno-builtin with a macro in Toolchain.defs/Make.defs to avoid the hardcode value in the common file?

yamt: ideally, maybe.
we are using `-fno-lto` in the common file as well.

xiaoxiang781216 (approved):

PR #1126, opened 2022-04-06
system/note: correct unflatten format

anchao:
## Summary

system/note: correct unflatten format

## Impact
Fix endian issue reported by @pkarashchenko :
https://github.com/apache/incubator-nuttx-apps/pull/1114#discussion_r842466352

## Testing

sched note trace test

### DIFF
--- a/system/sched_note/note_main.c
+++ b/system/sched_note/note_main.c
@@ -90,27 +90,15 @@ static FAR const char *g_statenames[] =
 static void trace_dump_unflatten(FAR void *dst,
                                  FAR uint8_t *src, size_t len)
 {
-  switch (len)
+#ifdef CONFIG_ENDIAN_BIG
+  FAR uint8_t *end = (FAR uint8_t *)dst + len - 1;
+  while (len-- > 0)
     {
-#ifdef CONFIG_HAVE_LONG_LONG
-      case 8:
-        *(uint64_t *)dst = ((uint64_t)src[7] << 56)
-                         + ((uint64_t)src[6] << 48)
-                         + ((uint64_t)src[5] << 40)
-                         + ((uint64_t)src[4] << 32);
-#endif
-      case 4:
-        *(uint32_t *)dst = ((uint64_t)src[3] << 24)
-                         + ((uint64_t)src[2] << 16);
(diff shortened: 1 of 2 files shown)


system/sched_note/note_main.c
```diff
@@ -94,18 +95,17 @@ static void trace_dump_unflatten(FAR void *dst,
     {
 #ifdef CONFIG_HAVE_LONG_LONG
       case 8:
-        *(uint64_t *)dst = ((uint64_t)src[7] << 56)
-                         + ((uint64_t)src[6] << 48)
-                         + ((uint64_t)src[5] << 40)
-                         + ((uint64_t)src[4] << 32);
+        *(uint64_t *)dst = le64toh(*(uint64_t *)src);
```
pkarashchenko, line 98: I have only one question. Are we sure that `src` is target type aligned? I mean that if `len == 4` then `src` can be promoted to `uint32_t *`. If not, then we will have troubles with archs that require pointers to be aligned.

anchao: Yes, there are potential issues if the hardware do not support aligned access, seems there is no more elegant way, I used the scheme you recommended.

system/sched_note/note_main.c
```diff
@@ -90,27 +90,35 @@ static FAR const char *g_statenames[] =
...
 #ifdef CONFIG_HAVE_LONG_LONG
       case 8:
-        *(uint64_t *)dst = ((uint64_t)src[7] << 56)
-                         + ((uint64_t)src[6] << 48)
-                         + ((uint64_t)src[5] << 40)
-                         + ((uint64_t)src[4] << 32);
+        tmp += ((uint64_t)src[7] << 56);
+        tmp += ((uint64_t)src[6] << 48);
+        tmp += ((uint64_t)src[5] << 40);
+        tmp += ((uint64_t)src[4] << 32);
```
xiaoxiang781216, line 106: let's follow the same change:
https://github.com/apache/incubator-nuttx/pull/5992

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #1139, opened 2022-04-15
wireless: gs2200m: Fix recvfrom_request() in gs2200m_main.c

masayuki2009:
## Summary

- I noticed that wget causes ASSERT in gs2200m_main.c recently.
- Finally, I found that it sends a wrong ack message in
  recvfrom_request() when the socket is closed passively
- This commit fixes this issue.

## Impact

- gs2200m only

## Testing

- Tested with spresense:wifi_smp


masayuki2009:
See https://github.com/apache/incubator-nuttx/pull/6041 as well

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #1144, opened 2022-04-19
system/i2c: Fix fd leak in i2ccmd_reset

no1wudi:
## Summary
Close fd before return.
## Impact
i2ccmd_reset
## Testing
CI

### DIFF
--- a/system/i2c/i2c_reset.c
+++ b/system/i2c/i2c_reset.c
@@ -24,6 +24,8 @@
 
 #include <nuttx/config.h>
 
+#include <unistd.h>
+
 #include <nuttx/i2c/i2c_master.h>
 
 #include "i2ctool.h"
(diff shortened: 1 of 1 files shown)


system/i2c/i2c_reset.c
```diff
@@ -53,7 +55,15 @@ int i2ccmd_reset(FAR struct i2ctool_s *i2ctool, int argc, char **argv)
   ret = i2cdev_reset(fd);
   if (ret == OK)
     {
-      i2ctool_printf(i2ctool, "Reset command sent successfully\n");
+      if (close(fd) == 0)
```
pkarashchenko, line 58: I think we should close even if `i2cdev_reset` fails

pkarashchenko (changes requested):

pkarashchenko (approved):

system/i2c/i2c_reset.c
```diff
@@ -60,6 +62,13 @@ int i2ccmd_reset(FAR struct i2ctool_s *i2ctool, int argc, char **argv)
       i2ctool_printf(i2ctool, "Failed to send the reset command\n");
     }
 
+  ret = close(fd);
+
```
pkarashchenko, line 66:
```suggestion
```

PR #1147, opened 2022-04-21
nshlib/nsh_fileapps.c: Pass global environ instead of NULL for posix_spawnp

pussuw:
Now that the environ pointer is a list of pointers-to-string, it can be
passed to the standard posix_spawn function as-is.

## Testing
icicle:nsh & icicle:knsh

### DIFF
--- a/nshlib/nsh_fileapps.c
+++ b/nshlib/nsh_fileapps.c
@@ -30,6 +30,7 @@
 #endif
 
 #include <stdbool.h>
+#include <stdlib.h>
 #include <spawn.h>
 #include <errno.h>
 #include <string.h>
@@ -126,7 +127,7 @@ int nsh_fileapp(FAR struct nsh_vtbl_s *vtbl, FAR const char *cmd,
    * failure.
    */
 
-  ret = posix_spawnp(&pid, cmd, &file_actions, &attr, argv, NULL);
+  ret = posix_spawnp(&pid, cmd, &file_actions, &attr, argv, environ);
   if (ret == OK)
     {
       /* The application was successfully started with pre-emption disabled.


xiaoxiang781216 (approved):

nshlib/nsh_fileapps.c
```diff
@@ -126,7 +127,7 @@ int nsh_fileapp(FAR struct nsh_vtbl_s *vtbl, FAR const char *cmd,
    * failure.
    */
 
-  ret = posix_spawnp(&pid, cmd, &file_actions, &attr, argv, NULL);
+  ret = posix_spawnp(&pid, cmd, &file_actions, &attr, argv, environ);
```
pkarashchenko, line 130:
```
====================================================================================
Configuration/Tool: lc823450-xgevk/krndis,CONFIG_ARMV7M_TOOLCHAIN_GNU_EABI
------------------------------------------------------------------------------------
  Cleaning...
  Configuring...
  Disabling CONFIG_ARMV7M_TOOLCHAIN_GNU_EABI
  Enabling CONFIG_ARMV7M_TOOLCHAIN_GNU_EABI
  Building NuttX...
arm-none-eabi-ld: /github/workspace/sources/nuttx/staging//libapps.a(nsh_fileapps.c.github.workspace.sources.apps.nshlib.o): in function `nsh_fileapp':
/github/workspace/sources/apps/nshlib/nsh_fileapps.c:252: undefined reference to `get_environ_ptr'
make[1]: *** [Makefile:59: nuttx_user.elf] Error 1
make[1]: Target 'all' not remade because of errors.
make: *** [tools/Unix.mk:498: nuttx] Error 2
```
Should this be
```suggestion
#ifndef CONFIG_DISABLE_ENVIRON
  ret = posix_spawnp(&pid, cmd, &file_actions, &attr, argv, environ);
#else
  ret = posix_spawnp(&pid, cmd, &file_actions, &attr, argv, NULL);
#endif
```
Or maybe we can
```
#ifndef CONFIG_DISABLE_ENVIRON
#  define environ get_environ_ptr()
#else
#  define environ NULL
#endif
```
in `stdlib.h`.
@xiaoxiang781216 what do you think?

PR #1151, opened 2022-04-22
system/ping: Guard against division by zero when calculating the RTT.

Ouss4:
## Summary
Guard against division by zero when calculating the RTT, otherwise system could crash with the delay is 0.
## Impact
ping
## Testing
esp32-devkitc:netnsh pinging the router.

### DIFF
--- a/system/ping/ping.c
+++ b/system/ping/ping.c
@@ -242,10 +242,17 @@ static void ping_result(FAR const struct ping_result_s *result)
                    result->extra / USEC_PER_MSEC);
             if (result->nreplies > 0)
               {
-                long avg = priv->tsum / result->nreplies;
-                long long tempnum = priv->tsum2 / result->nreplies -
-                                    (long long)avg * avg;
-                long tmdev = ub16toi(ub32sqrtub16(uitoub32(tempnum)));
+                long avg = 0;
+                long long tempnum = 0;
+                long tmdev = 0;
+
+                if (priv->tsum > 0)
+                  {
+                    avg = priv->tsum / result->nreplies;
+                    tempnum = priv->tsum2 / result->nreplies -
+                              (long long)avg * avg;
+                    tmdev = ub16toi(ub32sqrtub16(uitoub32(tempnum)));
+                  }
 
                 printf("rtt min/avg/max/mdev = %ld.%03ld/%ld.%03ld/"
                        "%ld.%03ld/%ld.%03ld ms\n",
--- a/system/ping6/ping6.c
+++ b/system/ping6/ping6.c
@@ -239,10 +239,17 @@ static void ping6_result(FAR const struct ping6_result_s *result)
                    result->extra / USEC_PER_MSEC);
             if (result->nreplies > 0)
               {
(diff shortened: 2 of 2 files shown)


system/ping/ping.c
```diff
@@ -242,10 +242,17 @@ static void ping_result(FAR const struct ping_result_s *result)
                    result->extra / USEC_PER_MSEC);
             if (result->nreplies > 0)
               {
-                long avg = priv->tsum / result->nreplies;
-                long long tempnum = priv->tsum2 / result->nreplies -
+                long tmdev = 0;
+                long avg = 0;
+                long long tempnum = 0;
+
+                if (priv->tsum > 0)
```
xiaoxiang781216, line 249: But, I don't see tsum used as divider. BTW, should we modify ping6 too?

Ouss4: tsum is used to calculate the other variables, if it's 0 all of the rest will be 0 including the one passed to `ub32sqrtub16` which will end up crashing due to a division by zero.

Ouss4: Regarding ping6, yes I believe we should guard there too.

xiaoxiang781216: > tsum is used to calculate the other variables, if it's 0 all of the rest will be 0 including the one passed to `ub32sqrtub16` which will end up crashing due to a division by zero.

Is it crash here?
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/fixedmath/lib_ubsqrt.c#L60

system/ping/ping.c
```diff
@@ -242,10 +242,17 @@ static void ping_result(FAR const struct ping_result_s *result)
                    result->extra / USEC_PER_MSEC);
             if (result->nreplies > 0)
```
xiaoxiang781216, line 243: let's merge line 249 to here:
```
if (result->nreplies > 0 && priv->tsum > 0)
```

system/ping/ping.c
xiaoxiang781216: @Ouss4 https://github.com/apache/incubator-nuttx/pull/6124 may fix this issue too, but I think both change should merge.

Ouss4: > Is it crash here?
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/fixedmath/lib_ubsqrt.c#L60

Yes, because the function will receive 0 as argument.

system/ping/ping.c
Ouss4: In this case we won't print anything if the delay is 0.  Is that ok?
I pushed as suggested but I am having a second guess.

xiaoxiang781216: Sorry, I forget the print, either approach is fine. Since you encounter this special situation, and should know which one is best choice:).

Ouss4: I believe we should keep the print since we had a valid number of replies.  I added it back, please take a look.

xiaoxiang781216 (approved):
LGTM.

pkarashchenko (approved):

PR #1157, opened 2022-04-28
import/Make.defs: Fix C++ library linkage with CONFIG_BUILD_KERNEL

pussuw:
lxx must be before lmm at least, because new uses malloc().

## Summary
Fixes C++ library linkage with CONFIG_BUILD_KERNEL
## Testing
icicle:knsh

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -132,7 +132,7 @@ endef
 
 define ELFLD
 	@echo "LD: $2"
-	$(Q) $(LD) $(LDELFFLAGS) $(LDLIBPATH) $(ARCHCRT0OBJ) $1 $(LDLIBS) -o $2
+	$(Q) $(LD) $(LDELFFLAGS) $(LDLIBPATH) $(ARCHCRT0OBJ) $1 $(LDSTARTGROUP) $(LDLIBS) $(LDENDGROUP) -o $2
 endef
 
 $(RAOBJS): %.s$(SUFFIX)$(OBJEXT): %.s


pkarashchenko:
What is the exact error with the current code?

pussuw:
> What is the exact error with the current code? I mean do we use start/end group for the libs? Otherwise I do not see why order should matter here

libcxx refers to e.g. malloc which is in libmm, the ordering matters here, otherwise e.g. malloc will be left undefined in applications that use libxx but don't use malloc directly.

gustavonihei:
I agree with @pkarashchenko. NuttX usually wraps the libraries around `--start-group` and `--end-group` (via `LDSTARTGROUP` and `LDENDGROUP`), so the ordering of libraries should not matter because those libraries within the group will be searched repeatedly until all symbols are found.

pkarashchenko:
@pussuw maybe you can share the command line call that generates an error so we can take a look at the arguments that are passed during the link stage?

pussuw:
There is no cli error, the error comes from the elf loader because it encounters an undefined symbol it cannot resolve / relocate. I'll provide a more detailed error description tomorrow, but the short description is references to symbols in libxx to e.g. symbols in libmm. The kernel mode ldlibs flag does not show start/end group. LDELF is called and there is no encapsulation either. All I know that appending lxx results in undefined symbols in the resulting cxx application elf.

gustavonihei:
> There is no cli error, the error comes from the elf loader because it encounters an undefined symbol it cannot resolve / relocate. I'll provide a more detailed error description tomorrow, but the short description is referencing symbols from libxx to e.g. symbols in libmm. The kernel mode ldlibs flag does not show start/end group. LDELF is called and there is no encapsulation either. All I know that appending lxx results in undefined symbols in the resulting cxx application elf.
> 
> Maybe the fix is to add usage of start/end group to kernel mode ldflags

https://github.com/apache/incubator-nuttx-apps/blob/a5a4c561fb9c9c5c9af0013f2c3b8901a176b956/Application.mk#L133-L136

If adding wrapping `LDLIBS` with `LDSTARTGROUP` and `LDENDGROUP` works, I believe this would be the definitive fix for the issue.

pussuw:
> 
> If wrapping `LDLIBS` with `LDSTARTGROUP` and `LDENDGROUP` works, I believe this would be the definitive fix for the issue.

Agreed, I'll try that tomorrow

pkarashchenko:
I think you are hitting this issue https://github.com/apache/incubator-nuttx/issues/6082

pussuw:
> I think you are hitting this issue [apache/incubator-nuttx#6082](https://github.com/apache/incubator-nuttx/issues/6082)

That issue is different I think, it is on the export side. What I face is that when linking e.g. cxxtest app, the resulting elf file has undefined symbols after ELFLD. Please keep in mind that this is with CONFIG_BUILD_KERNEL, every application is build separately into one .elf file which can be started e.g. via posix_spawn(), so this is quite different from the monolithic build process used by BUILD_FLAT and even BUILD_PROTECTED.

```
/apps/bin_copy$ objdump -t cxxtest_stripped | grep UND
0000000000000000         *UND*	0000000000000000 malloc
0000000000000000         *UND*	0000000000000000 on_exit
0000000000000000         *UND*	0000000000000000 syslog
0000000000000000         *UND*	0000000000000000 free
```

The problem I think is ELFLD does not have the start/end group encapsulation, like the kernel side does:
`$(LDSTARTGROUP) $(LDLIBS) $(EXTRA_LIBS) $(LDENDGROUP)`


```
define ELFLD
	@echo "LD: $2"
	$(Q) $(LD) $(LDELFFLAGS) $(LDLIBPATH) $(ARCHCRT0OBJ) $1 $(LDLIBS) -o $2
endef
```

pkarashchenko:
> $(LDSTARTGROUP) $(LDLIBS) $(LDENDGROUP) fixes the issue, thanks for the help guys

Great!

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #1171, opened 2022-05-24
nxwm: Fix warning about on_exit() dependency

pussuw:
## Summary
Follow up to kernel side https://github.com/apache/incubator-nuttx/pull/6197


pussuw:
Build error is due to new patch that adds SCHED_ONEXIT=y to one defconfig, it is fixed here: https://github.com/apache/incubator-nuttx/pull/6325

PR #1173, opened 2022-05-27
testing/ostest: Replace up_tls_info with tls_get_info

xiaoxiang781216:
## Summary
follow the kernel side change


testing/ostest/tls.c
```diff
@@ -48,17 +47,17 @@ static struct tls_info_s g_save_info;
 
 static void get_tls_info(FAR struct tls_info_s *info)
 {
-  memcpy(info, up_tls_info(), sizeof(struct tls_info_s));
+  memcpy(info, tls_get_info(), sizeof(struct tls_info_s));
```
pkarashchenko, line 50: Maybe we can switch to define instead? Seems like a simple wrapper.

pkarashchenko: Anyway minor

pkarashchenko (approved):

testing/ostest/tls.c
xiaoxiang781216: Done.

pkarashchenko (approved):

pkarashchenko:
Could you please share the upcoming kernel side changes?
Now we have
```
#ifndef up_tls_info
#  if defined(CONFIG_TLS_ALIGNED) && !defined(__KERNEL__)
#    define up_tls_info() TLS_INFO((uintptr_t)up_getsp())
#  else
#    define up_tls_info() tls_get_info()
#  endif
#endif
```
so this change covers only the `else` case of `if defined(CONFIG_TLS_ALIGNED) && !defined(__KERNEL__)`.

xiaoxiang781216:
> Could you please share the upcoming kernel side changes? Now we have
> 
> ```
> #ifndef up_tls_info
> #  if defined(CONFIG_TLS_ALIGNED) && !defined(__KERNEL__)
> #    define up_tls_info() TLS_INFO((uintptr_t)up_getsp())
> #  else
> #    define up_tls_info() tls_get_info()
> #  endif
> #endif
> ```
> 
> so this change covers only the `else` case of `if defined(CONFIG_TLS_ALIGNED) && !defined(__KERNEL__)`.

Here is patch:
https://github.com/apache/incubator-nuttx/pull/6337/commits/c9b5292676feea98bbb1901d14236d1a0dc7e3a2
This patch need merge first otherwise the above PR can't pass CI.

PR #1174, opened 2022-05-28
industry/foc foc_ident: add the flux linkage identification

zouboan:
## Summary
improve resistance measurement method to solve the failure of resistance measurement  and improve the accuracy of resistance measurement, add the flux linkage measurement for angle observer based on flux linkage.
## Impact
Motor parameter identification
## Testing
Tested on FOC controller based on STM32F405RG


xiaoxiang781216:
@raiden00pl do you have time to review this patch?

raiden00pl:
I'll look at this PR tomorrow

raiden00pl:
@zouboan Could you split these changes into two commits ? One improving the measuring of resistance, the other adding flux linkage identification.
And I think that the flux linkage identification should be optional and enabled from Kconfig. 
The motor controller doesn't always need information about flux linkage (it depends on application), so it doesn't make sense to force the user to take this additional step. On the other hand, the resistance and inductance of motor phases is necessary for the correct tuning of the PI current controller

industry/foc/float/foc_ident.c
```diff
@@ -164,6 +164,8 @@ int foc_ident_res_run_f32(FAR struct foc_ident_f32_s *ident,
   int   ret  = FOC_ROUTINE_RUN_NOTDONE;
   float err  = 0.0f;
   float vref = 0.0f;
+  static float curr_sum  = 0.0f;
+  static float volt_sum  = 0.0f;
```
xiaoxiang781216, line 172: why use static here?

industry/foc/float/foc_ident.c
```diff
@@ -343,6 +347,127 @@ int foc_ident_ind_run_f32(FAR struct foc_ident_f32_s *ident,
...
+  int   ret  = FOC_ROUTINE_RUN_NOTDONE;
+  float curr_avg  = 0.0f;
+  float volt_avg  = 0.0f;
+  static float dir    = DIR_CW;
+  static float vq_sum = 0.0f;
+  static float vd_sum = 0.0f;
+  static float iq_sum = 0.0f;
+  static float id_sum = 0.0f;
+  static float vel = 0.0f;
+  static int aveg_sum  = 0;
```
xiaoxiang781216, line 377: why use static here

industry/foc/float/foc_ident.c
zouboan: App calls foc_routine_ident_run_f32 every cycle to calculated resistance, to get the average value by accumulation, we should set curr_sum and volt_sum as static

industry/foc/float/foc_ident.c
zouboan: to get the average value by accumulation

zouboan: vel will pass to ramp function to get and maintain open loop velocity,

xiaoxiang781216: see my previous comment

industry/foc/float/foc_ident.c
zouboan: I think you're right, I will improve the design

xiaoxiang781216 (approved):

PR #1175, opened 2022-05-30
netutils/dhcpc: Change the timeout unit to milliseconds

SunJun8:
## Summary
netutils/dhcpc: Change the timeout unit to milliseconds
## Impact
None
## Testing
bl602evb:wifi


pkarashchenko:
I think this seems to be a "breaking change".

xiaoxiang781216 (approved):

xiaoxiang781216:
> I think this seems to be a "breaking change".

Yes, I add "break change" label.

pkarashchenko (approved):

PR #1176, opened 2022-05-30
Fixed bug in drawing rectangle using 1 bit palette

adamkaliszan:
## Summary
Exemplary fb App bug fix for drawing rectangle in 1 bit palette

## Impact
Writing own driver gives and testing using fb app gives bad result. The problem was in app, not in new implemented driver. 
The frame buffer was filled with a wrong manner.

## Testing
Tested with won implemented driver. Content of frame buffer was verified in order to check if the bits are correctly set.

### DIFF
--- a/examples/fb/fb_main.c
+++ b/examples/fb/fb_main.c
@@ -28,6 +28,7 @@
 #include <sys/mman.h>
 #include <unistd.h>
 #include <stdlib.h>
+#include <string.h>
 #include <stdio.h>
 #include <fcntl.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@adamkaliszan please fix the warning reported here:
https://github.com/apache/incubator-nuttx-apps/runs/6655177954?check_suite_focus=true
you can check locally by:
```
../nuttx/tools/checkpatch.sh -g HEAD~...HEAD
```

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #1177, opened 2022-05-30
system/libuv: Include nuttx/tls.h to call task local storage api

xiaoxiang781216:


xiaoxiang781216:
@pkarashchenko could you merge this simple change? Other patch depends on this patch.

pkarashchenko (approved):

PR #1178, opened 2022-05-31
mlearning: add Darknet (Open Source Neural Networks in C)

jerpelea:
## Summary
Darknet is an open source neural network framework written
in C and CUDA. It is fast, easy to install, and supports
CPU and GPU computation.

You Only Look Once (YOLO) is a state-of-the-art,
real-time object detection system

## Impact
ML

## Testing
ML

### DIFF
--- a/mlearning/darknet/Kconfig
+++ b/mlearning/darknet/Kconfig
@@ -1,0 +1,19 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config DARKNET_YOLO
+	bool "YOLO: Real-Time Object Detection"
+	default n
+	---help---
+		You only look once (YOLO) is a state-of-the-art,
+                real-time object detection system
+
+if DARKNET_YOLO
+
+config DARKNET_YOLO_VER
+	string "DARKNET YOLO version"
+	default "master"
+
+endif # DARKNET_YOLO
--- a/mlearning/darknet/Makefile
+++ b/mlearning/darknet/Makefile
@@ -1,0 +1,90 @@
+############################################################################
+# apps/mlearning/darknet/Makefile
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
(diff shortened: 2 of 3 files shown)


xiaoxiang781216 (approved):

mlearning/darknet/Makefile
```diff
@@ -0,0 +1,90 @@
...
+CSRCS   +=$(SRC)/yolo_layer.c
+CSRCS   +=$(SRC)/iseg_layer.c
+
+
+CFLAGS += -Wno-shadow -Wno-strict-prototypes -Wno-unknown-pragmas
+
+MODULE = $(CONFIG_DARKNET_YOLO)
+
+darknet.zip:
+	$(Q) curl -L https://github.com/pjreddie/darknet/archive/refs/heads/master.zip -o darknet.zip
```
pkarashchenko, line 78: Maybe adding git hash id is a good next step improvement to get better control of what is built

pkarashchenko (approved):

mlearning/darknet/Kconfig
```diff
@@ -0,0 +1,19 @@
...
+	default n
+	---help---
+		You only look once (YOLO) is a state-of-the-art,
+                real-time object detection system
+
+if DARKNET_YOLO
+
+config DARKNET_YOLO_VER
+	string "DARKNET YOLO version"
+	default master
```
xiaoxiang781216, line 17: change to “master”

mlearning/darknet/Makefile
xiaoxiang781216: let's use DARKNET_YOLO_VER here, so user can select different version through defconfig

jerpelea: @pkarashchenko I prefer to use stable versions or tags but there is none for YOLO

pkarashchenko: I mean, that `master` will be different from build to build. I think that using hash has better predictability, so then is can be changed to stable tag once it is available

xiaoxiang781216 (approved):

xiaoxiang781216:
@jerpelea do you have plan to support TensorFlow Lite?

jerpelea:
@xiaoxiang781216 I am looking into it

PR #1182, opened 2022-06-01
webclient: implement https proxy

yamt:
## Summary
webclient: implement https proxy

## Testing
A ported version was tested on macOS with my local app + squid.

### DIFF
--- a/include/netutils/webclient.h
+++ b/include/netutils/webclient.h
@@ -234,6 +234,7 @@ typedef CODE int (*webclient_body_callback_t)(
 
 struct webclient_tls_connection;
 struct webclient_poll_info;
+struct webclient_conn_s;
 
 struct webclient_tls_ops
 {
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

include/netutils/webclient.h
```diff
@@ -420,6 +462,20 @@ struct webclient_poll_info
...
+  bool tls;
+
+  /* for !tls */
+
+  int sockfd;
+  unsigned int flags;
+
+  /* for tls */
+
+  struct webclient_tls_connection *tls_conn;
```
pkarashchenko, line 476: Good to add `FAR` to all the pointers in future

yamt: sure

netutils/webclient/webclient.c
```diff
@@ -270,11 +259,23 @@ static const char g_httpcache[]      = "Cache-Control: no-cache";
  * Private Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: free_ws
+ ****************************************************************************/
+
+static void free_ws(FAR struct wget_s *ws)
+{
+  free(ws->conn);
```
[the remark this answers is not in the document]
yamt: > Previously, the implementation used `free(3)` instead of `free_ws`. Since the latter does not check for `NULL` pointers, this mean the behaviour has been now changed from the older implementation when a `NULL` pointer is given (therefore, a breaking change), with a potential risk for undefined behaviour.
> 
> _Maybe_ all calls to `free_ws` within this repository are currently used with non-NULL pointers, but I would not rely on this assumption, and instead try to follow the same behaviour as `free(3)` in this regard.

as far as i know, all uses of free_ws takes non NULL ws.

PR #1183, opened 2022-06-02
system/libuv: Skip compile strscpy.c when CONFIG_LIBUV_UTILS_TEST enable

xiaoxiang781216:
## Impact
fix the linker error when compile libuv test

## Testing
Pass CI

### DIFF
--- a/system/libuv/Makefile
+++ b/system/libuv/Makefile
@@ -69,7 +69,7 @@ CSRCS += posix-hrtime.c
 CSRCS += posix-poll.c
 CSRCS += uv-data-getter-setters.c
 CSRCS += version.c
-ifneq ($(CONFIG_LIBUV_UTILS_TEST),)
+ifeq ($(CONFIG_LIBUV_UTILS_TEST),)
 CSRCS += idna.c
 CSRCS += strscpy.c
 endif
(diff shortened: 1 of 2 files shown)


system/libuv/Makefile
```diff
@@ -69,12 +69,12 @@ CSRCS += posix-hrtime.c
 CSRCS += posix-poll.c
 CSRCS += uv-data-getter-setters.c
 CSRCS += version.c
-ifneq ($(CONFIG_LIBUV_UTILS_TEST),)
+ifeq ($(CONFIG_LIBUV_UTILS_TEST),)
```
pkarashchenko, line 72: The logic for `idna.c` is changed. Is it expected?

xiaoxiang781216: Yes, both idna.c and strscpy.c will be included directly when test is enabled. Please reference here:
https://github.com/libuv/libuv/blob/master/test/test-idna.c#L23
https://github.com/libuv/libuv/blob/master/test/test-strscpy.c#L27

pkarashchenko: Sure. Just want to confirm that this is "as expected".

pkarashchenko (approved):

PR #1185, opened 2022-06-03
webclient.h: Avoid relying on indirect inclusion

yamt:
## Summary
Include stdint.h for uint16_t explicitly.

This was necessary when I was trying to build this natively
on Ubuntu.  It seems some other headers happen to pull
the uint16_t definition by luck on NuttX and macOS.

### DIFF
--- a/include/netutils/webclient.h
+++ b/include/netutils/webclient.h
@@ -49,6 +49,7 @@
 #include <nuttx/config.h>
 
 #include <stdbool.h>
+#include <stdint.h>
 #include <sys/types.h>
 
 /****************************************************************************


xiaoxiang781216 (approved):

pkarashchenko (approved):

include/netutils/webclient.h
```diff
@@ -49,6 +49,7 @@
 #include <nuttx/config.h>
 
 #include <stdbool.h>
+#include <stdint.h>
```
pkarashchenko, line 52: @yamt  BTW I noticed that `stdint` type is used only in `uint16_t tunnel_target_port;`, so maybe it is reasonable to change it to `unsigned int tunnel_target_port;` and do not include `stdint.h`?

pkarashchenko: Anyway that is minor and current type is used to match `struct wget_target_s`.

yamt: port number is 16 bit.
using a wider type makes it more difficult to understand, both for human beings and tools.

PR #1186, opened 2022-06-03
Remove "| O_BINARY" or " |= O_BINARY" since O_BINARY is 0 now

xiaoxiang781216:
## Summary
follow nuttx kernel side change: https://github.com/apache/incubator-nuttx/pull/6357

## Impact
Minor

## Testing
Pass CI

### DIFF
--- a/netutils/ftpd/ftpd.c
+++ b/netutils/ftpd/ftpd.c
@@ -1809,9 +1809,6 @@ static int ftpd_stream(FAR struct ftpd_session_s *session, int cmdtype)
 #if defined(O_LARGEFILE)
   oflags |= O_LARGEFILE;
 #endif
-#if defined(O_BINARY)
-  oflags |= O_BINARY;
-#endif
 
   /* Are we creating the file? */
 
--- a/canutils/slcan/slcan.c
+++ b/canutils/slcan/slcan.c
@@ -189,7 +189,11 @@ int main(int argc, char *argv[])
 
   /* CAN */
 
-  int s, nbytes, i, ret, reccount;
+  int s;
+  int nbytes;
+  int i;
+  int ret;
+  int reccount;
   struct sockaddr_can addr;
   struct canfd_frame frame;
   struct msghdr msg;
@@ -197,7 +201,8 @@ int main(int argc, char *argv[])
   fd_set rdfs;
   char ctrlmsg[CMSG_SPACE(sizeof(struct timeval) +
                           3 * sizeof(struct timespec) + sizeof(int))];
-  char sbuf[40], *sbp;
+  char sbuf[40];
+  char *sbp;
 
   if (argc != 3)
     {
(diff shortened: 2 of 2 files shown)


davids5:
This change hides the intent the the file is opened in binary mode. Does it make sense to assume the value the flag has?

xiaoxiang781216:
> This change hides the intent the the file is opened in binary mode. Does it make sense to assume the value the flag has?

Most caller of open want the binary mode behavior:
https://github.com/apache/incubator-nuttx-apps/search?q=open
but only these two places add O_BINARY. So the implementation of open need hold the following assumption:
File must be opened in binary mode if both O_BINARY and O_TEXT isn't passed to open.
Actually, other POSIX compliant OS doesn't support O_BINAY and O_TEXT at all:
https://stackoverflow.com/questions/2266992/no-o-binary-and-o-text-flags-in-linux
So, this change just make the code more consistent.

pkarashchenko:
@davids5 are you ok with this PR and PR to the OS side?

pkarashchenko:
In general there is no harm in keeping the binary flag in code even if it is 0. But we can remove it of course.

davids5:
@pkarashchenko Thank you for asking. I see it as a breaking changes, that does not add much value. But at least the compile will fail and trigger the hunt for "What! Why is NuttX not building again?"

xiaoxiang781216:
> @pkarashchenko Thank you for asking. I see it as a breaking changes, that does not add much value. But at least the compile will fail and trigger the hunt for "What! Why is NuttX not building again?"

It isn't a breaking change: both flag still exist. I modify these two files just want to make sure the assumption more clear and consistent with the rest code in the repo.

davids5:
@xiaoxiang781216 Thank you for detailed explanation. It would be appreciated if your PR would include that sort of information when submitted.

netutils/ftpd/ftpd.c
```diff
@@ -1809,9 +1809,6 @@ static int ftpd_stream(FAR struct ftpd_session_s *session, int cmdtype)
 #if defined(O_LARGEFILE)
   oflags |= O_LARGEFILE;
 #endif
-#if defined(O_BINARY)
```
pkarashchenko, line 1812: Let's check the original canutils code. If it contains this code, then it is better to keep it even if the binary flag is zero

xiaoxiang781216: Sure

xiaoxiang781216:
> @xiaoxiang781216 Thank you for detailed explanation. It would be appreciated if your PR would include that sort of information when submitted.

Ok.

netutils/ftpd/ftpd.c
pkarashchenko: Posted comment on a wrong line, but still... I've checked can-utils and do not see any reference to O_BINARY

pkarashchenko (approved):

pkarashchenko (approved):

PR #1193, opened 2022-06-10
serialblaster bugfix

adamkaliszan:
## Summary
Serlalblaster exemplary app bugfixes.

In my opinion there ware many bugs in previous implementation. 
## Impact
None
## Testing
Yes, on Nucleo Wl55JC board

### DIFF
--- a/examples/serialblaster/serialblaster_main.c
+++ b/examples/serialblaster/serialblaster_main.c
@@ -59,7 +59,8 @@ int main(int argc, FAR char *argv[])
   int ret;
   int fd;
   FAR char *devpath;
-  int size = 0;
+  const int slength = sizeof(s)-1;
+  int size = slength;
   int rem;
 
   if (argc == 1)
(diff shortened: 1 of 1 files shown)


examples/serialblaster/serialblaster_main.c
```diff
@@ -59,7 +59,7 @@ int main(int argc, FAR char *argv[])
   int ret;
   int fd;
   FAR char *devpath;
-  int size = 0;
+  int size = sizeof(s)-1;
```
davids5, line 62:
```suggestion
  const int slength  = sizeof(s)-1;
  int size = slength;
```
and replace it throughout the code

davids5 (changes requested):
Might be better to clean it up in this way

examples/serialblaster/serialblaster_main.c
adamkaliszan: OK. I wanted to reduce number of changes. sizeof(s)-1 in many places is annoying.

examples/serialblaster/serialblaster_main.c
```diff
@@ -84,25 +85,25 @@ int main(int argc, FAR char *argv[])
   fd = open(devpath, O_RDWR);
   if (fd < 0)
     {
-      printf("dev_ttyS2: ERROR Failed to open /dev/ttyS2\n");
+      printf("%s: ERROR Failed to open %s\n", devpath, devpath);
```
pkarashchenko, line 88: Why do we need to print `devpath` two times?

adamkaliszan: OK. It is done. Now there is argv[0] and devpath

examples/serialblaster/serialblaster_main.c
adamkaliszan: It is done

pkarashchenko (approved):

PR #1194, opened 2022-06-10
nshlib/nsh_envcmd.c: Change cmd_unset to only unset existing env vari…

jlaitine:
…ables

Correct the problem, where "unset" always gives an error message if the
variable is not exported (promoted to group-wide environment variable)

This fixes the issue by unsetting the exported env variable only if it exists.

This seems coherent with the way how cmd_set and cmd_export work currently.

Signed-off-by: Jukka Laitinen <jukkax@ssrc.tii.ae>

### DIFF
--- a/nshlib/nsh_envcmds.c
+++ b/nshlib/nsh_envcmds.c
@@ -534,28 +534,48 @@ int cmd_unset(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
 #if defined(CONFIG_NSH_VARS) || !defined(CONFIG_DISABLE_ENVIRON)
   int status;
 #endif
+#if defined(CONFIG_NSH_VARS) && !defined(CONFIG_DISABLE_ENVIRON)
+  FAR char *oldvalue;
+#endif
   int ret = OK;
 
 #if defined(CONFIG_NSH_VARS)
-  /* Unset NSH variable */
+#ifndef CONFIG_DISABLE_ENVIRON
+  /* Check if the NSH variable has already been promoted to an group-
+   * wide environment variable. In this case, remove the exported
+   * variable
+   */
 
-  status = nsh_unsetvar(vtbl, argv[1]);
-  if (status < 0 && status != -ENOENT)
+  oldvalue = getenv(argv[1]);
+  if (oldvalue == NULL)
+#endif
     {
-      nsh_error(vtbl, g_fmtcmdfailed, argv[0], "nsh_unsetvar",
-                NSH_ERRNO_OF(-status));
-      ret = ERROR;
+      /* Unset NSH variable */
+
+      status = nsh_unsetvar(vtbl, argv[1]);
+      if (status < 0 && status != -ENOENT)
+        {
+          nsh_error(vtbl, g_fmtcmdfailed, argv[0], "nsh_unsetvar",
+                    NSH_ERRNO_OF(-status));
+          ret = ERROR;
+        }
     }
 #endif
 
 #if !defined(CONFIG_DISABLE_ENVIRON)
-  /* Unset environment variable */
-
-  status = unsetenv(argv[1]);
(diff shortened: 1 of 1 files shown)


nshlib/nsh_envcmds.c
```diff
@@ -533,30 +533,46 @@ int cmd_unset(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
...
 #if defined(CONFIG_NSH_VARS) || !defined(CONFIG_DISABLE_ENVIRON)
   int status;
+  FAR char *oldvalue;
 #endif
   int ret = OK;
 
 #if defined(CONFIG_NSH_VARS)
-  /* Unset NSH variable */
+#ifndef CONFIG_DISABLE_ENVIRON
+      /* Check if the NSH variable has already been promoted to an group-
```
xiaoxiang781216, line 542: why make extra indent?

jlaitine: mistake, will fix

nshlib/nsh_envcmds.c
```diff
@@ -533,29 +533,45 @@ int cmd_unset(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
 {
 #if defined(CONFIG_NSH_VARS) || !defined(CONFIG_DISABLE_ENVIRON)
   int status;
+  FAR char *oldvalue;
```
jlaitine, line 536: this is inside wrong ifdefs. should be:
#if defined(CONFIG_NSH_VARS) && !defined(CONFIG_DISABLE_ENVIRON)

will fix this

nshlib/nsh_envcmds.c
```diff
@@ -533,29 +533,49 @@ int cmd_unset(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
...
+                    NSH_ERRNO_OF(-status));
+          ret = ERROR;
+        }
     }
 #endif
 
 #if !defined(CONFIG_DISABLE_ENVIRON)
-  /* Unset environment variable */
-
-  status = unsetenv(argv[1]);
```
xiaoxiang781216, line 554: it's more simple to skip the error message if  errno set by unsetenv equals ENOENT.

jlaitine: Agreed. Maybe wr can assume that the variable doesn't exist in both nsh_vars and environment variables (i.e. Local variable can't shadow an environment variable). Since in that case the implementation would be wrong; unsetting a local variable (in script) would also delete the global one. This was the thinking behind, why I wrote it like that. But I can change it as you suggested for sure!

xiaoxiang781216: Event so, We can simplify the code like this:
```
status = nsh_unsetvar(vtbl, argv[1]);
#if !defined(CONFIG_DISABLE_ENVIRON)
if (status == -ENOENT)
  {
     status = unsetenv(argv[1]);
      if (status < 0)
        {
          nsh_error(vtbl, g_fmtcmdfailed, argv[0], "unsetenv",
                    NSH_ERRNO);
          ret = ERROR;
        }
  }
else if (status < 0)
  {
      nsh_error(vtbl, g_fmtcmdfailed, argv[0], "nsh_unsetvar",
                NSH_ERRNO_OF(-status));
      ret = ERROR;
  }
#endif
```

jlaitine: Looks good to me, go ahead and make the PR, as you have the gode ready.

jlaitine:
Closing in favor of better solution from @xiaoxiang781216

jlaitine:
Just one more comment on this; unsetenv is a posix interface, which should return "success" if the environment variable is not set (i.e. it already doesn't exist).

So Implementation of "unsetenv" has been broken at some point in NuttX, since it now return error if unsetting an non-existing variable

PR #1195, opened 2022-06-13
nshlib: Don't output error when unsetenv fail with ENOENT

xiaoxiang781216:
## Summary
Fix the issue report here: https://github.com/apache/incubator-nuttx-apps/pull/1194

### DIFF
--- a/nshlib/nsh_envcmds.c
+++ b/nshlib/nsh_envcmds.c
@@ -552,7 +552,7 @@ int cmd_unset(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
   /* Unset environment variable */
 
   status = unsetenv(argv[1]);
-  if (status < 0)
+  if (status < 0 && errno != ENOENT)
     {
       nsh_error(vtbl, g_fmtcmdfailed, argv[0], "unsetenv", NSH_ERRNO);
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@jlaitine here is the change.

nshlib/nsh_envcmds.c
```diff
@@ -552,7 +552,7 @@ int cmd_unset(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
   /* Unset environment variable */
 
   status = unsetenv(argv[1]);
-  if (status < 0)
+  if (status < 0 && errno != ENOENT)
```
pkarashchenko, line 555: Do we need to try `unsetenv` again in case of `ENOENT`? Because now if `unsetenv` returns with `ENOENT` the `cmd_unset` returns `OK`

pkarashchenko (approved):

pussuw:
unsetenv should never return ENOENT

xiaoxiang781216:
> unsetenv should never return ENOENT
> 
> [apache/incubator-nuttx#6423](https://github.com/apache/incubator-nuttx/pull/6423)

So we don't need this patch anymore, let's close it.

PR #1214, opened 2022-07-03
system/hostname: Add an option to read the hostname from a file

normanr:
## Summary
Add an option to read the hostname from a file (eg: in `/etc/init.d/rcS`). Add a check that the hostname is non-empty.

## Impact
Makes it possible to set hostname from a file without `NSH_CMDPARMS` enabled (eg: "hostname \`cat /etc/hostname\`"). Ensures that hostname must be non-blank.

## Testing
```
nsh> hostname -F /data/hostname
nsh> hostname -F /tmp/missing
ERROR: Failed to open '/tmp/missing': No such file or directory
nsh> hostname ""
ERROR: The specified hostname is invalid
nsh> echo > /tmp/newline
nsh> hostname -F /tmp/newline
ERROR: The specified hostname is invalid
nsh> echo -n > /tmp/empty
nsh> hostname -F /tmp/empty
ERROR: The specified hostname is invalid
nsh> 
```

### DIFF
--- a/system/hostname/hostname_main.c
+++ b/system/hostname/hostname_main.c
@@ -36,26 +36,104 @@
  ****************************************************************************/
 
 /****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: show_usage
+ ****************************************************************************/
+
+static void show_usage(FAR const char *progname)
+{
+  printf("Usage: %s [<hostname>|-F <file>]\n", progname);
+}
+
+/****************************************************************************
  * Public Functions
  ****************************************************************************/
 
 int main(int argc, FAR char *argv[])
 {
   int ret;
-  char hostname[HOST_NAME_MAX + 1];
+  bool set = false;
+  FAR FILE *f;
+  char hostname[HOST_NAME_MAX + 2];
 
-  if (argc > 1)
+  while ((ret = getopt(argc, argv, ":F:h")) != ERROR)
     {
-      if (argc > 2)
+      switch (ret)
         {
-          fprintf(stderr, "ERROR: Too many arguments\n");
+          case 'F':
+            set = true;
+            f = fopen(optarg, "r");
+            if (f == NULL)
+              {
+                fprintf(stderr, "ERROR: Failed to open '%s': %s\n", optarg,
+                        strerror(errno));
+                return EXIT_FAILURE;
+              }
+
+            if (fgets(hostname, sizeof(hostname), f) == NULL)
+              {
+                if (errno != 0)
+                  {
+                    fprintf(stderr, "ERROR: Failed to read '%s': %s\n",
+                            optarg, strerror(errno));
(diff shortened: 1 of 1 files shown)


system/hostname/hostname_main.c
```diff
@@ -42,17 +42,64 @@
...
 {
   int ret;
+  FILE *f;
+  int maxarg = 2;
   char hostname[HOST_NAME_MAX + 1];
 
   if (argc > 1)
     {
-      if (argc > 2)
+      if (strcmp(argv[1], "-F") == 0)
```
xiaoxiang781216, line 51: should we add an usage?

xiaoxiang781216 (approved):

system/hostname/hostname_main.c
```diff
@@ -42,17 +42,64 @@
 int main(int argc, FAR char *argv[])
 {
   int ret;
+  FILE *f;
```
pkarashchenko, line 45:
```suggestion
  FAR FILE *f;
```

system/hostname/hostname_main.c
```diff
@@ -42,17 +42,64 @@
...
+          if (f == NULL)
+            {
+              fprintf(stderr, "ERROR: Failed to open '%s': %s\n", argv[2],
+                      strerror(errno));
+              return EXIT_FAILURE;
+            }
+
+          if (fgets(hostname, sizeof(hostname), f) == NULL && (errno != 0))
+            {
+              fclose(f);
```
pkarashchenko, line 69: `fclose` may modify `errno`, `fprintf` may print the value different from one that met `(errno != 0)` to true in the `if`.

system/hostname/hostname_main.c
```diff
@@ -42,17 +42,64 @@
...
+        }
+
+      if (argc > maxarg)
         {
           fprintf(stderr, "ERROR: Too many arguments\n");
           return EXIT_FAILURE;
         }
 
-      ret = sethostname(argv[1], strlen(argv[1]));
+      if (strlen(hostname) == 0)
```
pkarashchenko, line 96: Optimization (optional):
Maybe we can somehow avoid double call of `strlen(hostname)` in case of non-null strings, like
`if (*hostname == '\0')`?

normanr: I considered this, and the resulting assembly code is identical. The compiler optimizes `strlen(hostname) == 0` as `*hostname == '\0'` so I figured to stick with the more readable version.

I'm not sure if we also want a check for `strlen(hostname) > HOST_NAME_MAX` (to prevent the supplied hostname being truncated. I would expect that the compiler should re-use the returned value and not call `strlen(hostname)` twice.

system/hostname/hostname_main.c
normanr: Done

xiaoxiang781216 (approved):

PR #1215, opened 2022-07-07
dhcpc: Rename timeout Kconfig

yamt:
## Summary

NETUTILS_DHCPC_RECV_TIMEOUT -> NETUTILS_DHCPC_RECV_TIMEOUT_MS

Recently its unit has been changed from second to millisecond. [1]
Using the same Kconfig name for a different meaning is a pitfall
for an upgrade. I was using 3ms timeout because of this.

This commit renames it so that at least the default value
is used for a careless user like me.

[1]
```
commit 8fd4b6105be2a535ecea2312858602d9ffa4e670
Author: zrrong <zrrong@bouffalolab.com>
Date:   Fri May 27 11:35:23 2022 +0800

    netutils/dhcpc: Change the timeout unit to milliseconds
```

### DIFF
--- a/netutils/dhcpc/Kconfig
+++ b/netutils/dhcpc/Kconfig
@@ -17,7 +17,7 @@ config NETUTILS_DHCPC_HOST_NAME
 	string "DHCP client host name"
 	default "nuttx"
 
-config NETUTILS_DHCPC_RECV_TIMEOUT
+config NETUTILS_DHCPC_RECV_TIMEOUT_MS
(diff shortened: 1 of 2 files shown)


davids5 (approved):
@yamt - Thank you

pkarashchenko (approved):
I do not see in-tree configs using this option, but for out-of-tree boards might use it.
@davids5 should we add "breaking change" label here?

PR #1216, opened 2022-07-08
apps: Fix make distclean for CONFIG_BUILD_KERNEL=y

masayuki2009:
## Summary

- I noticed that some files (e.g. xxx.tar.gz) which are generated
   by 'make context' are not deleted when we build the kernel only
    then we do 'make distclean' because .depend is not generated.
- This commit fixes this issue by adding the .context file and
    also adding directories that contain the file to be cleaned.

## Impact

- None

## Testing

- Tested with sabre-6quad:netknsh (not merged yet)

### DIFF
--- a/Directory.mk
+++ b/Directory.mk
@@ -26,6 +26,7 @@ SUBDIRS       := $(dir $(wildcard *$(DELIM)Makefile))
 CONFIGSUBDIRS := $(filter-out $(dir $(wildcard *$(DELIM)Kconfig)),$(SUBDIRS))
 CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).depend))
 CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).kconfig))
+CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).context))
 CLEANSUBDIRS  := $(sort $(CLEANSUBDIRS))
 
 all: nothing
--- a/boot/mcuboot/Makefile
+++ b/boot/mcuboot/Makefile
@@ -84,9 +84,11 @@ $(MCUBOOT_UNPACK): $(MCUBOOT_TARBALL)
 	$(Q) touch $(MCUBOOT_UNPACK)
 
 context:: $(MCUBOOT_UNPACK)
+	$(Q) touch .context
 
 distclean::
 	$(call DELFILE, $(MCUBOOT_TARBALL))
 	$(call DELDIR, $(MCUBOOT_UNPACK))
+	$(call DELFILE, .context)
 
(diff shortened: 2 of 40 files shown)


Directory.mk
```diff
@@ -26,6 +26,7 @@ SUBDIRS       := $(dir $(wildcard *$(DELIM)Makefile))
 CONFIGSUBDIRS := $(filter-out $(dir $(wildcard *$(DELIM)Kconfig)),$(SUBDIRS))
 CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).depend))
 CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).kconfig))
+CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM)*.tar.*))
```
Ouss4, line 29: Not sure why is this required.  I see the discussion here https://github.com/apache/incubator-nuttx/pull/6586

`.depend` is created in the argtable3 directory and thus `distclean` should execute there and delete the downloaded files.

masayuki2009: @Ouss4 
'.depend' is NOT created when building the kernel (CONFIG_BUILD_KERNEL=y) only and will be created when we do 'make export'

Ouss4: Ah I see.  Thanks for the explanation.

xiaoxiang781216: But this fix isn't extensible since not all downloaded package contain *.tar.*

masayuki2009: >But this fix isn't extensible since not all downloaded package contain .tar.

@xiaoxiang781216 
How about creating the .context file when downloading a file such as .tar.gz or generating a new file such as config.h
Then .context will be checked when executing 'make distclean'

xiaoxiang781216: Yes, I think it's better to generate .context in kernel mode just like other mode.

masayuki2009:
@xiaoxiang781216 
Please review this PR again.

xiaoxiang781216:
@masayuki2009 how about we invoke "make distclean) in all subfolder like https://github.com/apache/incubator-nuttx-apps/pull/1217?

masayuki2009:
> @masayuki2009 how about we invoke "make distclean in all subfolder like #1217?

I think the PR will take a longer time because the PR always traverse all app directories which contain Makefile even though the nuttx configuration does not use them.

xiaoxiang781216:
Yes, but it could ensure all clean/distclean target get executed. On the other hand, the method used to detect whether the directories is part of the build is fragile.

masayuki2009:
Let me close this PR, since https://github.com/apache/incubator-nuttx/pull/6719 fixes the original issue.

PR #1217, opened 2022-07-09
Trigger clean and distclean for all directory which contain Makefile

xiaoxiang781216:
## Summary
Found by https://github.com/apache/incubator-nuttx/pull/6586

## Impact
make distclean and make clean

## Testing
Pass CI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@
 *~
 .context
 .depend
-.kconfig
 /bin
 /boot_romfsimg.h
 /external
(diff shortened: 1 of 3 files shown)


Ouss4:
This was the old behavior, before filtering based on `.depend` and `.kconfig`.  Do you want to bring it back?  This will increase the execution time of multiple phases in build and configuration.

xiaoxiang781216:
> This was the old behavior, before filtering based on `.depend` and `.kconfig`. Do you want to bring it back? This will increase the execution time of multiple phases in build and configuration.

Here is my thought:

1. Compare with the compile and link, the time to execute clean/distclean is very small
2. We need ensure the result is right before the optimization

xiaoxiang781216:
@Ouss4 either this patch or https://github.com/apache/nuttx-apps/pull/1632 is needed to fix the clean in the general case

pkarashchenko (approved):

PR #1220, opened 2022-07-17
Add printflike to all printf like functions

xiaoxiang781216:
## Summary
Follow the similar kernel change: https://github.com/apache/incubator-nuttx/pull/6632
 
## Impact
Enable more compiler time check

## Testing
Pass CI

### DIFF
--- a/interpreters/bas/bas_statement.c
+++ b/interpreters/bas/bas_statement.c
@@ -1241,7 +1241,7 @@ struct Value *stmt_EDIT(struct Value *value)
 
   return (struct Value *)0;
 #else
-  return Value_new_ERROR(value, NOTAVAILABLE, strerror(ENOSYS));
+  return Value_new_ERROR(value, NOTAVAILABLE);
 #endif
 }
 
@@ -3526,7 +3526,7 @@ struct Value *stmt_MATINPUT(struct Value *value)
 
           if (var->dim != 1 && var->dim != 2)
             {
-              return Value_new_ERROR(value, NOMATRIX, var->dim);
+              return Value_new_ERROR(value, DIMENSION);
             }
 
           columns = var->dim == 1 ? 0 : var->geometry[1];
@@ -5772,7 +5772,7 @@ struct Value *stmt_SHELL(struct Value *value)
 
   return (struct Value *)0;
 #else
-  return Value_new_ERROR(value, NOTAVAILABLE, strerror(ENOSYS));
+  return Value_new_ERROR(value, NOTAVAILABLE);
 #endif
 }
 
--- a/nshlib/nsh_netcmds.c
+++ b/nshlib/nsh_netcmds.c
@@ -377,8 +377,9 @@ static int nsh_foreach_netdev(nsh_netdev_callback_t callback,
   if (dir == NULL)
     {
       nsh_error(vtbl,
-                "nsh: %s: Could not open %s/net (is procfs mounted?): %d\n",
-                cmd, CONFIG_NSH_PROC_MOUNTPOINT, NSH_ERRNO);
+                "nsh: %s: Could not open %s/net (is procfs mounted?)\n",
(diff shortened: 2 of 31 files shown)


xiaoxiang781216:
Please ignore the mix case warning in:
https://github.com/apache/incubator-nuttx-apps/runs/7376553170?check_suite_focus=true

pkarashchenko (approved):

pkarashchenko:
There still are few build errors

xiaoxiang781216:
Fixed.

pkarashchenko:
```
In file included from nsh_syscmds.c:38:
nsh_syscmds.c: In function 'cmd_pmconfig':
Error: nsh_syscmds.c:220:24: error: format '%d' expects argument of type 'int', but argument 3 has type 'char *' [-Werror=format=]
  220 |       nsh_output(vtbl, "Last state %d, Next state %d",
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  221 |                  "PM stay [%d, %d, %d, %d]\n",
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                  |
      |                  char *
nsh_console.h:53:49: note: in definition of macro 'nsh_output'
   53 | # define nsh_output(v, ...)    (v)->output(v, ##__VA_ARGS__)
      |                                                 ^~~~~~~~~~~
```

pkarashchenko (approved):
LGTM!

nshlib/nsh_netcmds.c
```diff
@@ -377,8 +377,8 @@ static int nsh_foreach_netdev(nsh_netdev_callback_t callback,
   if (dir == NULL)
     {
       nsh_error(vtbl,
-                "nsh: %s: Could not open %s/net (is procfs mounted?): %d\n",
-                cmd, CONFIG_NSH_PROC_MOUNTPOINT, NSH_ERRNO);
```
pkarashchenko, line 381: Why printing of NSH_ERRNO is removed?

xiaoxiang781216: NSH_ERRNO may be a string or an integer:
https://github.com/apache/incubator-nuttx-apps/blob/master/nshlib/nsh.h#L434-L444
it's hard to keep it.

xiaoxiang781216: Ok, I remove else keyword, so the errno will output too.

interpreters/bas/bas_statement.c
```diff
@@ -1241,7 +1241,7 @@ struct Value *stmt_EDIT(struct Value *value)
 
   return (struct Value *)0;
 #else
-  return Value_new_ERROR(value, NOTAVAILABLE, strerror(ENOSYS));
+  return Value_new_ERROR(value, NOTAVAILABLE);
```
pkarashchenko, line 1244: Maybe modify `#define NOTAVAILABLE       RUNTIME+22,_("Feature not available")` to `#define NOTAVAILABLE       RUNTIME+22,_("Feature not available %s")` similar to `STRAYSUBEXIT`?
However only `strerror(ENOSYS)` is printed, so does not make sense now.

pkarashchenko (approved):
LGTM!

interpreters/bas/bas_statement.c
xiaoxiang781216: Yes, I think so.

pkarashchenko:
@xiaoxiang781216 do you have any guess about reason of check error at examples/sotest/lib/sotest/sotest.c#L58 ?

xiaoxiang781216:
> @xiaoxiang781216 do you have any guess about reason of check error at examples/sotest/lib/sotest/sotest.c#L58 ?

that's because nxstyle consider `,` in the string as the multiple variables defined in the same line:
```
static const char g_msg3[] = "Yes, don't be a stranger!";
```

PR #1221, opened 2022-07-17
examples/noteprintf: Change CONFIG_LIBC_LONG_LONG to CONFIG_HAVE_LONG_LONG

xiaoxiang781216:
## Summary
follow the kernel side change:
https://github.com/apache/incubator-nuttx/pull/6613

## Impact
Minor

## Testing
Pass CI

### DIFF
--- a/examples/noteprintf/noteprintf_main.c
+++ b/examples/noteprintf/noteprintf_main.c
@@ -59,7 +59,7 @@ int main(int argc, FAR char *argv[])
   short s        = 2;
   int i          = 3;
   long l         = 4;
-#ifdef CONFIG_LIBC_LONG_LONG
+#ifdef CONFIG_HAVE_LONG_LONG
   long long ll   = 5;
 #endif
   intmax_t im    = 6;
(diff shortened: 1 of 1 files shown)


pkarashchenko (approved):

examples/noteprintf/noteprintf_main.c
```diff
@@ -59,7 +59,7 @@ int main(int argc, FAR char *argv[])
   short s        = 2;
   int i          = 3;
   long l         = 4;
-#ifdef CONFIG_LIBC_LONG_LONG
+#ifdef CONFIG_HAVE_LONG_LONG
```
pkarashchenko, line 62: @xiaoxiang781216 is this still a valid change?
Should line 87 be wrapped?

xiaoxiang781216: Yes, done.

pkarashchenko: So print code still depend on HAVE, but not on LIBC?

xiaoxiang781216: Yes, not only print but all code which use long long must depend on CONFIG_HAVE_LONG_LONG since the old compiler mayn't support long long type at all(long long is defined in c99).

PR #1226, opened 2022-07-21
cmd_reset_cause:remove extra space in text ouput

anjiahao1:
can use ouput string set a env to argv
Signed-off-by: anjiahao <anjiahao@xiaomi.com>

## Impact
in a nsh script:
```set flag "cause:0x9,flag:0x4"
set rel `resetcause`
if [ $flag == $rel ]
then 
  echo "do something"
fi
```
but,if `resetcause` output have `space`,  `if [ $flag == $rel ]` will parse error

### DIFF
--- a/nshlib/nsh_syscmds.c
+++ b/nshlib/nsh_syscmds.c
@@ -351,7 +351,7 @@ int cmd_reset_cause(FAR struct nsh_vtbl_s *vtbl, int argc, char **argv)
       return ERROR;
     }
 
-  nsh_output(vtbl, "cause:0x%x, flag:0x%" PRIx32 "\n",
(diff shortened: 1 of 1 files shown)


pkarashchenko:
I'm not sure about the motivation of this change. Could you please expand description with use case?

xiaoxiang781216 (approved):

pkarashchenko:
does nsh support `""` or `''`? So `"$rel"` or `'$rel'` can be used?

anjiahao1:
> does nsh support `""` or `''`? So `"$rel"` or `'$rel'` can be used?

nsh just use space paser args..

pkarashchenko (approved):

PR #1238, opened 2022-07-28
apps/games: Add shift game

acassis:
## Summary
Add shift game
## Impact
Users could use this game with any board with LCD framebuffer
## Testing
https://www.youtube.com/shorts/dsUN4EQ0_JI

### DIFF
--- a/games/shift/shift_input_console.h
+++ b/games/shift/shift_input_console.h
@@ -1,0 +1,133 @@
+/****************************************************************************
+ * apps/games/shift/shift_input_console.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include <termios.h>
+
+#include "shift_inputs.h"
+
+/****************************************************************************
+ * Preprocessor Definitions
+ ****************************************************************************/
+
+/* Termios functions to have getch on Linux/NuttX */
+
+static struct termios g_old;
+static struct termios g_new;
+
+/* Initialize g_new terminal I/O settings */
+
+void init_termios(int echo)
+{
+  tcgetattr(0, &g_old);                 /* grab old terminal i/o settings */
+  g_new = g_old;                        /* use old settings as starting */
+  g_new.c_lflag &= ~ICANON;             /* disable buffered I/O */
+  g_new.c_lflag &= ~ECHO;               /* disable ECHO bit */
+  g_new.c_lflag |= echo ? ECHO : 0;     /* set echo mode if requested */
+  tcsetattr(0, TCSANOW, &g_new);        /* apply terminal I/O settings */
+}
+
+/* Restore g_old terminal i/o settings */
+
+void reset_termios(void)
+{
+  tcsetattr(0, TCSANOW, &g_old);
+}
+
+/* Read 1 character - echo defines echo mode */
+
+char getch_(int echo)
+{
+  char ch;
+
+  init_termios(echo);
+  ch = getchar();
+  reset_termios();
+
+  return ch;
+}
+
+/* Read 1 character without echo getch() function definition. */
+
+char getch(void)
+{
+  return getch_(0);
+}
+
+/****************************************************************************
+ * dev_input_init
+ ****************************************************************************/
+
+int dev_input_init(FAR struct input_state_s *dev)
+{
+  init_termios(0);
+
+  return OK;
+}
+
+/****************************************************************************
+ * dev_read_input
+ ****************************************************************************/
+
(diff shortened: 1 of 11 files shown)


xiaoxiang781216 (approved):

games/shift/shift_input_console.h
```diff
@@ -0,0 +1,133 @@
...
+static struct termios g_new;
+
+/* Initialize g_new terminal I/O settings */
+
+void init_termios(int echo)
+{
+  tcgetattr(0, &g_old);                 /* grab old terminal i/o settings */
+  g_new = g_old;                        /* use old settings as starting */
+  g_new.c_lflag &= ~ICANON;             /* disable buffered I/O */
+  g_new.c_lflag &= echo ? ECHO : ~ECHO; /* set echo mode */
```
pkarashchenko, line 46: is this correct? usually set/unset are done like
```
if (set)
  {
    a |= BIT;
  }
else
  {
    a &= ~BIT;
  }
```

games/shift/shift_input_console.h
```diff
@@ -0,0 +1,133 @@
...
+
+/* Read 1 character - echo defines echo mode */
+
+char getch_(int echo)
+{
+    char ch;
+    init_termios(echo);
+    ch = getchar();
+    reset_termios();
+    return ch;
```
pkarashchenko, line 65: 2 spaces

games/shift/shift_input_console.h
```diff
@@ -0,0 +1,133 @@
...
+  tcgetattr(0, &g_old);                 /* grab old terminal i/o settings */
+  g_new = g_old;                        /* use old settings as starting */
+  g_new.c_lflag &= ~ICANON;             /* disable buffered I/O */
+  g_new.c_lflag &= echo ? ECHO : ~ECHO; /* set echo mode */
+  tcsetattr(0, TCSANOW, &g_new);        /* apply terminal I/O settings */
+}
+
+/* Restore g_old terminal i/o settings */
+
+void reset_termios(void)
```
pkarashchenko, line 52: why all this API implementation is in header file and not in C-file?

games/shift/shift_input_console.h
```diff
@@ -0,0 +1,133 @@
...
+    {
+      if ((ch = getch()) == 91)
+        {
+          ch = getch();
+          if (ch == 65)
+            {
+              dev->dir = DIR_UP;
+            }
+          else
+          if (ch == 66)
```
pkarashchenko, line 106:
```suggestion
          else if (ch == 66)
```
here and other places

games/shift/shift_input_joystick.h
```diff
@@ -0,0 +1,166 @@
...
+              (long)nread, sizeof(djoy_buttonset_t));
+    }
+
+  g_djoylast = newset;
+
+  switch (newset)
+    {
+      case DJOY_UP_BIT:
+          dev->dir = DIR_UP;
+          break;
```
pkarashchenko, line 151: remove 2 spaces here and other case items

games/shift/shift_main.c
```diff
@@ -0,0 +1,927 @@
...
+                            {
+                              {1, 1, 1, 1, 1, 1, 1, 1},
+                              {1, 0, 0, 0, 0, 0, 0, 1},
+                              {1, 0, 0, 0, 0, 0, 0, 1},
+                              {1, 0, 0, 0, 0, 0, 0, 1},
+                              {1, 0, 0, 0, 0, 0, 0, 1},
+                              {1, 0, 0, 0, 0, 0, 0, 1},
+                              {1, 0, 0, 0, 0, 0, 0, 1},
+                              {1, 1, 1, 1, 1, 1, 1, 1}
+                            };
```
pkarashchenko, line 113: please remove extra spaces

games/shift/shift_input_console.h
acassis: you are right, it only works to disable the echo, it needs to be broken in two parts:
```
g_new.c_lflag &= ~ECHO; /* clear ECHO bit */
g_new.c_lflag |= echo ? ECHO : 0;  /* enable echo if requested */
```

games/shift/shift_input_console.h
acassis: Thanks

games/shift/shift_input_console.h
acassis: I want to avoid creating many .c and .h files. So it I prefer to keep it only in these .h files

games/shift/shift_input_console.h
acassis: Ok, I will modify to use "else if" in a single line

games/shift/shift_input_joystick.h
acassis: Ok

games/shift/shift_main.c
acassis: Ok, my plan was to keep it aligned with the end of previous line, but I will move it to the left

games/shift/shift_input_console.h
```diff
@@ -0,0 +1,133 @@
...
+                   dev->dir = DIR_DOWN;
+                }
+               else if (ch == 67)
+                      {
+                        dev->dir = DIR_RIGHT;
+                      }
+                    else if (ch == 68)
+                          {
+                             dev->dir = DIR_LEFT;
+                          }
```
pkarashchenko, line 119: please fix formatting here

acassis: @pkarashchenko when I put the braces just below the space below the space between "if" and "(" the automatic test fails because it is not in an odd column. What do you suggest? Just keep the braces below letter "f" and move "dev->dir" one position to left?

pkarashchenko: Very strange. I do not think this is the only place in the project where `else if` + `{}` is used.  Let me check why this might happen

acassis: Ok, I found a good example at apps/examples/audio_rttl/audio_rttl.cxx

acassis: I think the nxstyle.c needs to be fixed to this case where we move the braces to just below the external (right most) "if (". Also need to be fixed to catch "case" with more than 2 spaces in the next line.

pkarashchenko (approved):
LGTM!

PR #1247, opened 2022-08-02
mkfats make g_bootcodeblob const to save on .data section usage

PetervdPerk-NXP:
## Summary
g_bootcodeblob was allocated in RAM, to be able to patch the offset logic.
This patch makes g_bootcodeblob a constant and patches the offset afterwards to save on ram usage

## Impact
Saves 132 bytes on .data section and moves it to the .rodata section.

## Testing
Tested on MR-CANHUBK3 formatting SD cards as FAT16/FAT32 and checking partition table

### DIFF
--- a/fsutils/mkfatfs/configfat.c
+++ b/fsutils/mkfatfs/configfat.c
@@ -74,7 +74,7 @@ struct fat_config_s
  * offset 3.
  */
 
-static uint8_t g_bootcodeblob[] =
+const static uint8_t g_bootcodeblob[] =
 {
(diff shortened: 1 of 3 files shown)


pkarashchenko (approved):

pkarashchenko:
The PR description states that "Saves 132 bytes on static ram usage.", but that is only in case of executable flash. I would rework PR description to state that it saves 132 bytes of data segment, but not RAM :)

xiaoxiang781216 (approved):

PR #1269, opened 2022-08-15
Some WAPI changes

zhhyu7:
## Summary
Fix coverity warning
Correct return value for scan case
Make the initial scan buffer configurable through Kconfig

### DIFF
--- a/wireless/wapi/src/util.c
+++ b/wireless/wapi/src/util.c
@@ -114,7 +114,7 @@ errout:
       free(buf);
     }
 
-  if (fd > 0)
+  if (fd >= 0)
     {
       close(fd);
     }
@@ -512,7 +512,7 @@ errout:
       free(buf);
     }
 
-  if (fd > 0)
+  if (fd >= 0)
     {
       close(fd);
     }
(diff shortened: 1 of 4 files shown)


wireless/wapi/src/wireless.c
```diff
@@ -1256,14 +1256,14 @@ int wapi_scan_stat(int sock, FAR const char *ifname)
           return 1;
         }
 
-      printf("err[%d]: %s\n", errno, strerror(errno));
-    }
-  else
```
xiaoxiang781216, line 1263: it's wrong?

zhhyu7: > it's wrong?

change printf(error) to WAPI_IOCTL_STRERROR, and correct return value when ioctl return >=0.
Otherwise, when multiple threads switch, errno will be set to someone else's system call return code.

zhhyu7 (commented):
Does the naming check for cJSON need fixing？

xiaoxiang781216 (approved):

xiaoxiang781216:
Let's ignore the Mix case warning.

wireless/wapi/src/util.c
```diff
@@ -114,7 +114,7 @@ static FAR void *wapi_json_load(FAR const char *confname)
       free(buf);
     }
 
-  if (fd > 0)
+  if (fd >= 0)
```
pkarashchenko, line 117: I think this change matters only if we expect fd to be stdin.

PR #1278, opened 2022-08-16
nshlib: Change ELF priority and stack size based on builtin list

baggio63446333:
## Summary
When an application is used as a loadable ELF module like as `CONFIG_EXAMPLES_XXX=m`,
the priority and stack size of that application task is always fixed at the default values.
In other words, `CONFIG_EXAMPLES_XXX_PRIORITY` and `CONFIG_EXAMPLES_XXX_STACKSIZE`
are not referenced at all.

This commit changes to get the priority and stack size of the application from the bulitin list.

If you change the name of the application offline, it will still work with the default values.
If the application name is used as it was built, the priority and stack size set in Kconfig can be applied.

## Impact
Loadable ELF application.

### DIFF
--- a/nshlib/nsh_fileapps.c
+++ b/nshlib/nsh_fileapps.c
@@ -34,6 +34,8 @@
 #include <spawn.h>
 #include <errno.h>
 #include <string.h>
+#include <libgen.h>
+#include <nuttx/lib/builtin.h>
 
 #include "nsh.h"
 #include "nsh_console.h"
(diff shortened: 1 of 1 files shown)


nshlib/nsh_fileapps.c
```diff
@@ -73,6 +75,10 @@ int nsh_fileapp(FAR struct nsh_vtbl_s *vtbl, FAR const char *cmd,
   pid_t pid;
   int rc = 0;
   int ret;
+#ifdef CONFIG_BUILTIN
+  char *appname;
```
pkarashchenko, line 79:
```suggestion
  FAR char *appname;
```

pkarashchenko (approved):

nshlib/nsh_fileapps.c
baggio63446333: Fixed.

xiaoxiang781216 (approved):

PR #1287, opened 2022-08-20
industry/foc foc_ident: improve flux linkage identification method

zouboan:
## Summary
1. change flux linkage identification from V/F openloop to I/F openloop to solve the problem of some motors fail to start on V/F openloop;
2. reset FOC ident routine state machine so that identification can be performed once more without restart FOC app;
## Impact
flux linkage identification
## Testing
Tested on custom FOC control board based on STM32F405RG

### DIFF
--- a/industry/foc/float/foc_ident.c
+++ b/industry/foc/float/foc_ident.c
@@ -756,6 +756,7 @@ int foc_routine_ident_run_f32(FAR foc_routine_f32_t *r,
       case FOC_IDENT_RUN_DONE:
         {
           ret = FOC_ROUTINE_RUN_DONE;
+          i->stage = FOC_IDENT_RUN_INIT;
 
           break;
         }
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

industry/foc/float/foc_ident.c
```diff
@@ -650,12 +653,23 @@ int foc_routine_ident_run_f32(FAR foc_routine_f32_t *r,
...
       case FOC_IDENT_RUN_IDLE3:
 #ifdef CONFIG_INDUSTRY_FOC_IDENT_FLUX
+        if (i->cntr == 0)
+          {
+            ret = i->cfg.cb.kpki(i->cfg.cb.priv_kpki, i->final.res,
+                                 i->final.ind, i->cfg.per);
+            if (ret < 0)
+              {
+                FOCLIBERR("ERROR: ident kpki callback failed %d!\n", ret);
+                goto errout;
```
pkarashchenko, line 665: nitpick: in most of the cases when  `goto errout;` is used the simple `break;` should be enough

pkarashchenko (approved):

PR #1294, opened 2022-08-26
fix use of addroute/delroute

github-liyi:
Signed-off-by: liyi <liyi25@xiaomi.com>

## Summary
fix the use of addroute/delroute
addroute/delroute function's input parameter may be a structure sockadd_in(netlib_set_dripv4addr),
which has only 16B. But structure sockaddr_storage's size is larger than 16B. In fact, its size is 128B. 
So when you pass a structure sockadd_in and try to copy as a structure sockaddr_storage
it will overstep the buffer's bounary. It is the similar condition for IPv6's structure sockaddr_in6.

## Impact
app's code must fit this change 

## Testing
before this modification, the nuttx may crash with "stack buffer overflow", like blow
image
The fault error is caused by the wrong way to use addroute (netlib_set_dripv4addr)

after this, it can work well!
image

### DIFF
--- a/examples/mld/mld_main.c
+++ b/examples/mld/mld_main.c
@@ -258,9 +258,10 @@ int main(int argc, FAR char *argv[])
   FAR char *iobuffer = NULL;
   struct sockaddr_in6 host;
 #ifdef CONFIG_NET_ROUTE
-  struct sockaddr_in6 target;
-  struct sockaddr_in6 router;
-  struct sockaddr_in6 netmask;
+  struct sockaddr_storage target;
+  struct sockaddr_storage router;
+  struct sockaddr_storage netmask;
+  FAR struct sockaddr_in6 *v6_addr;
 #endif
   struct ipv6_mreq mrec;
   int nsec;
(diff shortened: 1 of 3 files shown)


xiaoxiang781216 (approved):

examples/mld/mld_main.c
```diff
@@ -359,31 +360,31 @@ int main(int argc, FAR char *argv[])
 
   /* Set up a routing table entry for the address of the multicast group */
 
-  memset(&target, 0, sizeof(struct sockaddr_in6));
-  target.sin6_family  = AF_INET6;
-  target.sin6_port    = HTONS(0x4321);
-  memcpy(target.sin6_addr.s6_addr16, g_grp_addr, sizeof(struct in6_addr));
+  memset(&target, 0, sizeof(target));
+  v6_addr = (struct sockaddr_in6 *)&target;
```
pkarashchenko, line 364:
```suggestion
  v6_addr = (FAR struct sockaddr_in6 *)&target;
```

examples/mld/mld_main.c
```diff
@@ -359,31 +360,31 @@ int main(int argc, FAR char *argv[])
...
+  v6_addr->sin6_family  = AF_INET6;
+  v6_addr->sin6_port    = HTONS(0x4321);
+  memcpy(v6_addr->sin6_addr.s6_addr16, g_grp_addr, sizeof(struct in6_addr));
 
-  memset(&netmask, 0, sizeof(struct sockaddr_in6));
-  netmask.sin6_family  = AF_INET6;
-  netmask.sin6_port    = HTONS(0x4321);
-  memset(netmask.sin6_addr.s6_addr16, 0xff, sizeof(struct in6_addr));
+  memset(&netmask, 0, sizeof(netmask));
+  v6_addr = (struct sockaddr_in6 *)&netmask;
```
pkarashchenko, line 370:
```suggestion
  v6_addr = (FAR struct sockaddr_in6 *)&netmask;
```

examples/mld/mld_main.c
```diff
@@ -359,31 +360,31 @@ int main(int argc, FAR char *argv[])
...
+  v6_addr->sin6_family  = AF_INET6;
+  v6_addr->sin6_port    = HTONS(0x4321);
+  memset(v6_addr->sin6_addr.s6_addr16, 0xff, sizeof(struct in6_addr));
 
-  memset(&router, 0, sizeof(struct sockaddr_in6));
-  router.sin6_family  = AF_INET6;
-  router.sin6_port    = HTONS(0x4321);
+  memset(&router, 0, sizeof(router));
+  v6_addr = (struct sockaddr_in6 *)&router;
```
pkarashchenko, line 376:
```suggestion
  v6_addr = (FAR struct sockaddr_in6 *)&router;
```

netutils/netlib/netlib_setdripv4addr.c
```diff
@@ -66,18 +66,21 @@ int netlib_set_dripv4addr(FAR const char *ifname,
...
+  target.ss_family  = AF_INET;
 
   memset(&netmask, 0, sizeof(netmask));
-  netmask.sin_family  = AF_INET;
+  netmask.ss_family  = AF_INET;
 
-  router.sin_addr    = *addr;
-  router.sin_family  = AF_INET;
+  memset(&router, 0, sizeof(router));
+  v4_addr = (struct sockaddr_in *)&router;
```
pkarashchenko, line 81:
```suggestion
  v4_addr = (FAR struct sockaddr_in *)&router;
```

examples/mld/mld_main.c
```diff
@@ -258,9 +258,10 @@ int main(int argc, FAR char *argv[])
   FAR char *iobuffer = NULL;
   struct sockaddr_in6 host;
 #ifdef CONFIG_NET_ROUTE
-  struct sockaddr_in6 target;
-  struct sockaddr_in6 router;
-  struct sockaddr_in6 netmask;
+  struct sockaddr_storage target;
```
pkarashchenko, line 261: Changing to `struct sockaddr_storage` in this block leads to quite a significant stack increase. Do we need to change application default stack value or the current value is sufficient?

xiaoxiang781216: The default stack is 2KB which should be enough. BTW, since printf consume the stack around 256bytes, so the increase from sockaddr_storage(~384 bytes) is safe.

PR #1295, opened 2022-08-27
Don't download tarballs if a local git repo found

no1wudi:
## Summary
Don't download tarballs if a local git repo found, to save time or avoid modify the makefile in local, make it easy to intergrate nuttx with customized third party software.
## Impact
Refactor only
## Testing
CI and local machine.

### DIFF
--- a/crypto/libtomcrypt/Makefile
+++ b/crypto/libtomcrypt/Makefile
@@ -476,10 +476,13 @@ $(LIBTOMCRYPT_UNPACKNAME): $(LIBTOMCRYPT_ZIP)
 	$(Q) patch -p0 < ltcrypt-flush.patch # Missing stdout flush in demo app
 	$(Q) touch $(LIBTOMCRYPT_UNPACKNAME)
 
+# Download and unpack tarball if no git repo found
+ifeq ($(wildcard $(LIBTOMCRYPT_UNPACKNAME)/.git),)
 context:: $(LIBTOMCRYPT_UNPACKNAME)
 
 distclean::
 	$(call DELDIR, $(LIBTOMCRYPT_UNPACKNAME))
 	$(call DELFILE, $(LIBTOMCRYPT_ZIP))
+endif
 
 include $(APPDIR)/Application.mk
--- a/interpreters/quickjs/Makefile
+++ b/interpreters/quickjs/Makefile
@@ -77,7 +77,17 @@ $(QUICKJS_UNPACK): $(QUICKJS_TARBALL)
 $(QUICKJS_UNPACK)/.patch: $(QUICKJS_UNPACK)
 	$(Q) touch $(QUICKJS_UNPACK)/.patch
 
+# Download and unpack tarball if no git repo found
+ifeq ($(wildcard $(QUICKJS_UNPACK)/.git),)
+
+distclean::
+	$(call DELDIR, $(QUICKJS_UNPACK))
+	$(call DELFILE, $(QUICKJS_TARBALL))
+
 build_host: $(QUICKJS_UNPACK)/.patch
+else
+build_host:
+endif
 	$(MAKE) -C $(QUICKJS_UNPACK) \
 		CONFIG_BIGNUM=$(CONFIG_INTERPRETERS_QUICKJS_BIGNUM)
 
@@ -85,9 +95,5 @@ context:: build_host
 
 clean::
 	$(Q) test ! -d $(QUICKJS_UNPACK) || $(MAKE) -C $(QUICKJS_UNPACK) clean
-
-distclean::
-	$(call DELDIR, $(QUICKJS_UNPACK))
-	$(call DELFILE, $(QUICKJS_TARBALL))
 
 include $(APPDIR)/Application.mk
--- a/interpreters/wamr/Makefile
+++ b/interpreters/wamr/Makefile
@@ -50,10 +50,13 @@ $(WAMR_UNPACK): $(WAMR_TARBALL)
 	$(Q) mv wasm-micro-runtime-$(WAMR_VERSION) $(WAMR_UNPACK)
 	$(Q) touch $(WAMR_UNPACK)
 
+# Download and unpack tarball if no git repo found
+ifeq ($(wildcard $(WAMR_UNPACK)/.git),)
 context:: $(WAMR_UNPACK)
 
 distclean::
 	$(call DELDIR, $(WAMR_UNPACK))
 	$(call DELFILE, $(WAMR_TARBALL))
+endif
 
 include $(APPDIR)/Application.mk
--- a/mlearning/darknet/Makefile
+++ b/mlearning/darknet/Makefile
@@ -20,7 +20,6 @@
 
 include $(APPDIR)/Make.defs
 
-ifeq ($(CONFIG_DARKNET_YOLO),y)
 SRC = darknet/src
 
 CSRCS   +=$(SRC)/gemm.c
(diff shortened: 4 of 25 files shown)


xiaoxiang781216 (approved):

crypto/libtomcrypt/Makefile
```diff
@@ -476,10 +476,13 @@ $(LIBTOMCRYPT_UNPACKNAME): $(LIBTOMCRYPT_ZIP)
 	$(Q) patch -p0 < ltcrypt-flush.patch # Missing stdout flush in demo app
 	$(Q) touch $(LIBTOMCRYPT_UNPACKNAME)
 
+# Download and unpack tarball if no git repo found
+ifeq ($(wildcard $(LIBTOMCRYPT_UNPACKNAME)/.git/HEAD),)
```
xiaoxiang781216, line 480: it's enough to check .git folder?

no1wudi: Yes, updated now

interpreters/wamr/Makefile
```diff
@@ -48,12 +48,16 @@ $(WAMR_UNPACK): $(WAMR_TARBALL)
 	$(Q) echo "Unpacking $(WAMR_TARBALL) to $(WAMR_UNPACK)"
 	$(Q) unzip $(WAMR_TARBALL)
 	$(Q) mv wasm-micro-runtime-$(WAMR_VERSION) $(WAMR_UNPACK)
-	$(Q) touch $(WAMR_UNPACK)
```
xiaoxiang781216, line 51: why remove

crypto/libtomcrypt/Makefile
```diff
@@ -476,10 +476,13 @@ $(LIBTOMCRYPT_UNPACKNAME): $(LIBTOMCRYPT_ZIP)
 	$(Q) patch -p0 < ltcrypt-flush.patch # Missing stdout flush in demo app
 	$(Q) touch $(LIBTOMCRYPT_UNPACKNAME)
 
+# Download and unpack tarball if no git repo found
+ifeq ($(wildcard $(LIBTOMCRYPT_UNPACKNAME)/.git),)
```
xiaoxiang781216, line 480: should we guard line 466-477 too?

testing/ltp/Makefile
```diff
@@ -132,9 +132,12 @@ $(LTP_UNPACK):
 	$(Q) git -C $(LTP_UNPACK) am < 0001-Fix-static-struct-warning.patch
 	$(Q) git -C $(LTP_UNPACK) am < 0002-Use-ifdef-instead-of-if-for-__linux__.patch
 
+# Download and unpack tarball if no git repo found
+ifeq ($(wildcard $(LTP_UNPACK)/.git),)
```
xiaoxiang781216, line 136: there are other 3rd party project not change(e.g. system/argtable3.

interpreters/wamr/Makefile
no1wudi: Restored

testing/ltp/Makefile
no1wudi: More other project updated.

crypto/libtomcrypt/Makefile
no1wudi: I think no need for that ? Small guard range is clear to understand and these dangling targets have no side effects for system.

mlearning/darknet/Makefile
```diff
@@ -79,12 +79,14 @@ darknet.zip:
...
+# Download and unpack tarball if no git repo found
+ifeq ($(wildcard darknet/.git),)
 context:: darknet.zip
 
 distclean::
 	$(call DELDIR, darknet)
 	$(call DELFILE, darknet.zip)
-
+endif
 endif
```
xiaoxiang781216, line 90: let's remove ifeq ($(CONFIG_DARKNET_YOLO),y)/endif?

netutils/cjson/Makefile
```diff
@@ -54,6 +56,8 @@ $(CJSON_UNPACKNAME): $(CJSON_TARBALL)
 	@echo "Unpacking: $(CJSON_TARBALL) -> $(CJSON_UNPACKNAME)"
 	$(Q) $(UNPACK) $(CJSON_TARBALL)
 	$(Q) touch $(CJSON_UNPACKNAME)
+else
```
xiaoxiang781216, line 59: why add else?

system/embedlog/Makefile
```diff
@@ -218,6 +220,9 @@ $(EMBEDLOG_SOURCES): $(EMBEDLOG_TARBALL)
 	$(Q) $(call DELDIR, $@)
 	$(Q) $(UNPACK) $<
 	$(Q) touch $@
+else
+$(EMBEDLOG_SOURCES):
```
xiaoxiang781216, line 224: don't need

testing/unity/Makefile
```diff
@@ -53,6 +55,9 @@ $(UNITY_UNPACKNAME): $(UNITY_TARBALL)
 	@echo "Unpacking: $(UNITY_TARBALL) -> $(UNITY_UNPACKNAME)"
 	$(Q) $(UNPACK) $(UNITY_TARBALL)
 	$(Q) touch $(UNITY_UNPACKNAME)
+else
+$(UNITY_UNPACKNAME):
```
xiaoxiang781216, line 59: don't need else clause

interpreters/quickjs/Makefile
```diff
@@ -87,7 +92,9 @@ clean::
 	$(Q) test ! -d $(QUICKJS_UNPACK) || $(MAKE) -C $(QUICKJS_UNPACK) clean
 
 distclean::
+ifeq ($(wildcard $(QUICKJS_UNPACK)/.git),)
```
xiaoxiang781216, line 95: move distclean into ifeq clause at line 80

xiaoxiang781216:
@no1wudi should we do the similar change to nuttx repo too?

crypto/libtomcrypt/Makefile
xiaoxiang781216: Ok.

mlyszczek:
Good idea but I think it's a bad approach. Right now, all packages hardcode their download code, they simply call "wget $(URL)", which makes it basically impossible to provide own tarballs (most common use case would be to protect yourself from upstream link going offline).

I think the right approach would be to create function that would take 2 arguments, base URL (like http://domain.com/distfiles) and filename (libpackage-1.0.0.tar.gz). Kconfig would hold cache url (or urls). Then, said download function would first try to get $filename from cache url(s) and then, once file could not be found, upstream URL from 1st argument would be used.

This will unify package download, and will limit clutter in app Makefiles.

xiaoxiang781216:
@mlyszczek This change doesn't forbid we unify the download process to a common place. Let's improve the system step by step.

mlyszczek:
Sure, but I still think proposed implementation is suboptimal, it adds unnecessary clutter and maintanence to all 3rd party apps.

New download function could also support git links with specified hash/tag. Idea is very good, but proposed implementation is simply a workaround, a band-aid, not a solution.

mlearning/libnnablart/Makefile
```diff
@@ -115,11 +115,14 @@ nnabla.zip:
...
 	$(Q) mv nnabla-c-runtime-$(CONFIG_NNABLA_RT_VER) nnabla-c-runtime
 
+# Download and unpack tarball if no git repo found
+ifeq ($(wildcard nnabla-c-runtime/.git),)
 context:: nnabla.zip
 
 distclean::
 	$(call DELDIR, nnabla-c-runtime)
 	$(call DELFILE, nnabla.zip)
+endif
```
xiaoxiang781216, line 125: line 127 can be removed too.

xiaoxiang781216 (approved):

PR #1306, opened 2022-09-03
ftpd: server port made configurable.

fjpanag:
## Summary

The ftpd was previously only working on the hard-coded port 21.

This commit allows the user to specify the desired port when the server is created.

## Impact

ftpd can run on any port.

## Testing

Tested within custom application on an STM32F427 board.  
Builds and runs as expected.


hartmannathan (approved):
LGTM. Thanks!

PR #1307, opened 2022-09-05
netutils/dhcpc:Make DHCP xid random

XinStellaris:
Signed-off-by: 田昕 <tianxin7@xiaomi.com>

## Summary
In this patch, dhcp will try to get a random xid. If that fails, dhcp will use the default xid.

The reason to do so is from RFC2131:
  "The 'xid' field is used by the client to match incoming DHCP messages
   with pending requests.  A DHCP client MUST choose 'xid's in such a
   way as to minimize the chance of using an 'xid' identical to one used
   by another client. For example, a client may choose a different,
   random initial 'xid' each time the client is rebooted, and
   subsequently use sequential 'xid's until the next reboot.  Selecting
   a new 'xid' for each retransmission is an implementation decision.  A
   client may choose to reuse the same 'xid' or select a new 'xid' for
   each retransmitted message."
## Impact
   DHCP
## Testing
  BL602 tested

### DIFF
--- a/netutils/dhcpc/dhcpc.c
+++ b/netutils/dhcpc/dhcpc.c
@@ -44,6 +44,7 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/time.h>
+#include <sys/random.h>
 
 #include <inttypes.h>
 #include <stdlib.h>
@@ -131,6 +132,7 @@ struct dhcpc_state_s
 {
   FAR const char    *interface;
   int                sockfd;
+  uint8_t            xid[4];
   struct in_addr     ipaddr;
   struct in_addr     serverid;
   struct dhcp_msg    packet;
@@ -144,11 +146,6 @@ struct dhcpc_state_s
 /****************************************************************************
  * Private Data
  ****************************************************************************/
-
-static const uint8_t xid[4] =
-{
-  0xad, 0xde, 0x12, 0x23
-};
 
 static const uint8_t magic_cookie[4] =
 {
@@ -234,7 +231,7 @@ static int dhcpc_sendmsg(FAR struct dhcpc_state_s *pdhcpc,
   pdhcpc->packet.op    = DHCP_REQUEST;
   pdhcpc->packet.htype = DHCP_HTYPE_ETHERNET;
   pdhcpc->packet.hlen  = pdhcpc->maclen;
-  memcpy(pdhcpc->packet.xid, xid, 4);
+  memcpy(pdhcpc->packet.xid, pdhcpc->xid, 4);
   memcpy(pdhcpc->packet.chaddr, pdhcpc->macaddr, pdhcpc->maclen);
   memset(&pdhcpc->packet.chaddr[pdhcpc->maclen], 0, 16 - pdhcpc->maclen);
   memcpy(pdhcpc->packet.options, magic_cookie, sizeof(magic_cookie));
@@ -435,7 +432,7 @@ static uint8_t dhcpc_parsemsg(FAR struct dhcpc_state_s *pdhcpc, int buflen,
                               FAR struct dhcpc_state *presult)
 {
   if (buflen >= 44 && pdhcpc->packet.op == DHCP_REPLY &&
-      memcmp(pdhcpc->packet.xid, xid, sizeof(xid)) == 0 &&
+      memcmp(pdhcpc->packet.xid, pdhcpc->xid, 4) == 0 &&
       memcmp(pdhcpc->packet.chaddr,
              pdhcpc->macaddr, pdhcpc->maclen) == 0)
     {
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

netutils/dhcpc/dhcpc.c
```diff
@@ -145,7 +146,7 @@ struct dhcpc_state_s
  * Private Data
  ****************************************************************************/
 
-static const uint8_t xid[4] =
+static uint8_t xid[4] =
```
xiaoxiang781216, line 149: let's move xid to the context

netutils/dhcpc/dhcpc.c
```diff
@@ -665,6 +666,17 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
   int     retries;
   int     state;
 
+  /* RFC2131: A DHCP client MUST choose 'xid's in such a
+   * way as to minimize the chance of using an 'xid' identical to one used
+   * by another client.
+   */
+
+  ret = getrandom(&xid, sizeof(xid), 0);
```
xiaoxiang781216, line 674: move to dhcpc_open

xiaoxiang781216:
Please check whether dhcp6 need the similar change.

XinStellaris:
I think we should put it in dhcp request, so that if we keep the dhcpc thread, xid will change every time just like the rfc requested.



---Original---
From: "Xiang ***@***.***&gt;
Date: Mon, Sep 5, 2022 22:56 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [apache/incubator-nuttx-apps] netutils/dhcpc:Make DHCP xid random(PR #1307)




 
@xiaoxiang781216 commented on this pull request.
 
 
In netutils/dhcpc/dhcpc.c:
 &gt; @@ -145,7 +146,7 @@ struct dhcpc_state_s   * Private Data   ****************************************************************************/   -static const uint8_t xid[4] = +static uint8_t xid[4] =  
let's move xid to the context
 
 
In netutils/dhcpc/dhcpc.c:
 &gt; @@ -665,6 +666,17 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)    int     retries;    int     state;   +  /* RFC2131: A DHCP client MUST choose 'xid's in such a +   * way as to minimize the chance of using an 'xid' identical to one used +   * by another client. +   */ + +  ret = getrandom(&amp;xid, sizeof(xid), 0);  
move to dhcpc_open

netutils/dhcpc/dhcpc.c
XinStellaris: I think we should put it in dhcp request, so that if we keep the dhcpc thread, xid will change every time just like the rfc requested.

xiaoxiang781216: here is my suggestion:
1.Move xid to dhcpc_state_s
2.Initialize xid with random number in dhcp_open
3.update inside dhcp_request

XinStellaris:
> Please check whether dhcp6 need the similar change.

Currently there is no similar change in dhcp6. If there is, I will change that in another patch

XinStellaris:
@anchao please review

netutils/dhcpc/dhcpc.c
```diff
@@ -517,6 +520,18 @@ FAR void *dhcpc_open(FAR const char *interface, FAR const void *macaddr,
...
+      /* RFC2131: A DHCP client MUST choose 'xid's in such a
+       * way as to minimize the chance of using an 'xid' identical to one
+       * used by another client.
+       */
+
+      memcpy(pdhcpc->xid, default_xid, 4);
+      result = getrandom(pdhcpc->xid, 4, 0);
+      if (result < 0)
+        {
+          getrandom(pdhcpc->xid, 4, GRND_RANDOM);
```
xiaoxiang781216, line 532: let's move memcpy to last:
```
result = getrandom(pdhcpc->xid, 4, GRND_RANDOM);
if (result != 4)
  {
    memcpy(pdhcpc->xid, default_xid, 4);
  }
```

netutils/dhcpc/dhcpc.c
```diff
@@ -517,6 +520,18 @@ FAR void *dhcpc_open(FAR const char *interface, FAR const void *macaddr,
   pdhcpc = malloc(sizeof(struct dhcpc_state_s) + maclen - 1);
   if (pdhcpc)
     {
+      /* RFC2131: A DHCP client MUST choose 'xid's in such a
+       * way as to minimize the chance of using an 'xid' identical to one
+       * used by another client.
+       */
+
+      memcpy(pdhcpc->xid, default_xid, 4);
+      result = getrandom(pdhcpc->xid, 4, 0);
```
xiaoxiang781216, line 529: reuse  ret

netutils/dhcpc/dhcpc.c
```diff
@@ -145,7 +147,7 @@ struct dhcpc_state_s
  * Private Data
  ****************************************************************************/
 
-static const uint8_t xid[4] =
```
anchao, line 148: how about move the global array to corresponding function to save the rodata?

XinStellaris: Sure

anchao (approved):
LGTM

xiaoxiang781216 (approved):

PR #1312, opened 2022-09-07
netutils/dhcpc:receiving unexpected packet not leading to dhcp retry.

XinStellaris:
Signed-off-by: 田昕 <tianxin7@xiaomi.com>

## Summary
If Many devices begin DHCPC at the same time. Current DHCP will fail because one device will receive many dhcp packets not for it, and thus exceeds retry count.
Here I made following modification:
1. Receiving unexpected packets does not result in retry count increasing(Only recv timeout will increase retry count)

## Impact
DHCP

## Testing
ESP32C3 tested.

### DIFF
--- a/netutils/dhcpc/dhcpc.c
+++ b/netutils/dhcpc/dhcpc.c
@@ -680,6 +680,7 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
   uint8_t msgtype;
   int     retries;
   int     state;
+  clock_t start;
 
   /* RFC2131: For example, a client may choose a different,
    * random initial 'xid' each time the client is rebooted, and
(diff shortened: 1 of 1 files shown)


netutils/dhcpc/dhcpc.c
```diff
@@ -720,13 +724,18 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
 
       /* Send the DISCOVER command */
 
-      ninfo("Broadcast DISCOVER\n");
-      if (dhcpc_sendmsg(pdhcpc, presult, DHCPDISCOVER) < 0)
+      if (last_send == 0 ||
```
xiaoxiang781216, line 727: don't need changedhcp_send since it is impossible to block for long time..

netutils/dhcpc/dhcpc.c
```diff
@@ -703,6 +705,8 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
    */
 
   retries = 0;
+  last_send = 0;
```
xiaoxiang781216, line 708: remove last_send

netutils/dhcpc/dhcpc.c
```diff
@@ -767,6 +776,14 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
 
           return ERROR;
         }
+
+      current = clock();
+
+      if (TICK2MSEC(current - last_send) >
```
xiaoxiang781216, line 782:
```suggestion
      if (TICK2MSEC(clock() - current ) >
```

netutils/dhcpc/dhcpc.c
```diff
@@ -783,6 +800,8 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
    */
 
   retries = 0;
+  last_send = 0;
```
xiaoxiang781216, line 803: apply the same change in following code.

netutils/dhcpc/dhcpc.c
XinStellaris: The change is to prevent dhcp_send from burst send multiple discover.
If not, every time we come back here we will send a dhcp. 
This is bad when we are receiving un-wanted dhcp packets, then the do-while will  keep sending

xiaoxiang781216: In this case, we should loop in recv until the timeout

XinStellaris: According to RFC2131: "DHCP clients are responsible for all message retransmission.  The
   client MUST adopt a retransmission strategy that incorporates a
   randomized exponential backoff algorithm to determine the delay
   between retransmissions.  The delay between retransmissions SHOULD be
   chosen to allow sufficient time for replies from the server to be
   delivered based on the characteristics of the internetwork between
   the client and the server. "
It is better to let dhcp send delay be different from recv time out. So I think recv time out shouldn't be related to send  delay. That is the reason why I put a last_send here. 
In the future we might need to add an exponential backoff algorithm here

XinStellaris: I will add the exponential backoff now.

XinStellaris: remove exponential backoff

netutils/dhcpc/dhcpc.c
```diff
@@ -804,67 +811,70 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
...
 
-          else if (msgtype == DHCPNAK)
-            {
-              ninfo("Received NAK\n");
-              break;
-            }
+              else if (msgtype == DHCPNAK)
+                {
+                  ninfo("Received NAK\n");
+                  return ERROR;
```
XinStellaris, line 840: Attention here, it was "break" before this patch.
But since it no longer goes back to send DISCOVER, I changed it to "return ERROR"

XinStellaris: The "goes back to send DISCOVER" was invalid since https://github.com/apache/incubator-nuttx-apps/pull/862

netutils/dhcpc/dhcpc.c
```diff
@@ -680,6 +680,7 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
   uint8_t msgtype;
   int     retries;
   int     state;
+  clock_t recv_begin;
```
xiaoxiang781216, line 683: recv_begin->start/begin

xiaoxiang781216 (approved):

PR #1315, opened 2022-09-09
Fix hello_zig build break

no1wudi:
## Summary
* Zig don't allow unused return value, so let's discard it explictly.
* Correct wrong make variable name for Zig main source, and update compile command for latest compiler.

## Impact
Zig only
## Testing
Custom board


Application.mk
```diff
@@ -127,7 +127,8 @@ endef
 
 define ELFCOMPILEZIG
        @echo "ZIG: $1"
-       $(Q) $(ZIG) build-obj $(ZIGELFFLAGS) $($(strip $1)_ZIGELFFLAGS) $1 --name $2
+	   # Remove target suffix here since zig compiler add .o automatically
```
pkarashchenko, line 130:
```suggestion
	# Remove target suffix here since zig compiler add .o automatically
```

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #1330, opened 2022-09-29
Separate NSH ability to mount a romfs at /etc and to run init scripts.

slorquet:
## Summary
This is useful because the board can mount its own filesystem without requiring any particular type of filesystem and mountpoint.

## Impact
Anyone using nsh init scripts need to regenerate their config.

## Testing
Build is OK, Kconfig menus work as expected
New code with an init script on a custom FS runs on my custom stm32f429 "hn70ap" board.

### DIFF
--- a/nshlib/Kconfig
+++ b/nshlib/Kconfig
@@ -699,12 +699,12 @@ config NSH_DISABLE_LOOPS
 endif # !NSH_DISABLESCRIPT
 
 config NSH_ROMFSETC
-	bool "Support ROMFS start-up script"
+	bool "Mount a ROMFS at /etc (eg, to support a start-up script)"
 	default n
 	depends on FS_ROMFS
 	---help---
-		Mount a ROMFS filesystem at /etc and provide a system init
-		script at /etc/init.d/rc.sysinit and a startup script
+		Mount a ROMFS filesystem at /etc. This can be used to and provide
+		a system init script at /etc/init.d/rc.sysinit and a startup script
 		at /etc/init.d/rcS.  The default system init script will mount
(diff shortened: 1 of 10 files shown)


nshlib/Kconfig
```diff
@@ -857,13 +832,75 @@ config NSH_FATMOUNTPT
...
+	---help---
+		This is the path to the sysinit script executed by nsh at boot.
+		The default is init.d/rc.sysinit. This script can be in a ROMFS mounted
+		by NSH or in any other filesystem that was mounted by the board logic.
+
+endif
+
+config NSH_RUNINITSCRIPT
+	bool "Run a startup script"
+    default n
```
xiaoxiang781216, line 856: please change ALL indent in Kconfig  from space to tab

nshlib/Kconfig
```diff
@@ -857,13 +832,75 @@ config NSH_FATMOUNTPT
 		will mount a FAT FS under /tmp. This is the location where the FAT
 		FS will be mounted.  Default is "/tmp".
 
+endif # NSH_ROMFSETC
+
+config NSH_RUNSYSINITSCRIPT
```
xiaoxiang781216, line 837: the new option will break all config

slorquet: Yes it will. Do you have any idea to avoid that?

slorquet: I'm cancelling this PR. I will come up with a new way to do it.

slorquet:
I'm cancelling this PR. I will come up with a new way to do it that limit breakage.

PR #1335, opened 2022-10-04
make export: Fix make export for CONFIG_BUILD_KERNEL=y

pussuw:
The condition:
[ -f "$${f}" ] && cp -f "$${f}" "${EXPORTDIR}"$(DELIM)registry ;

Fails if the first part of the condition returns empty / false, stopping make for no reason due to the error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,9 @@ ifneq ($(EXPORTDIR),)
 ifneq ($(CONFIG_BUILD_KERNEL),y)
 ifneq ($(BUILTIN_REGISTRY),)
 	for f in "${BUILTIN_REGISTRY}"$(DELIM)*.bdat "${BUILTIN_REGISTRY}"$(DELIM)*.pdat ; do \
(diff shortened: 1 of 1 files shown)


pkarashchenko (approved):

Makefile
```diff
@@ -157,8 +157,10 @@ ifneq ($(EXPORTDIR),)
 	$(Q) mkdir -p "${EXPORTDIR}"$(DELIM)registry || exit 1;
 ifneq ($(CONFIG_BUILD_KERNEL),y)
 ifneq ($(BUILTIN_REGISTRY),)
-	for f in "${BUILTIN_REGISTRY}"$(DELIM)*.bdat "${BUILTIN_REGISTRY}"$(DELIM)*.pdat ; do \
```
xiaoxiang781216, line 160: can we keep DELIM

pussuw: Yes, did not realize that has been added. This fix has been in my local repo for a long time.

xiaoxiang781216 (approved):

PR #1336, opened 2022-10-04
ftpd: Fixed warnings and style.

fjpanag:
## Summary

Fixes some compiler warnings.  
Fixes some improperly defined variables.

## Impact

N/A

## Testing

Build test only.


netutils/ftpd/ftpd.c
```diff
@@ -829,7 +830,7 @@ static int ftpd_rxpoll(int sd, int timeout)
     }
   else if (ret < 0)
     {
-      int errval = errno;
+      errval = errno;
```
xiaoxiang781216, line 833: why not define variable here?

jerpelea: I think that we should declare them at the beginning of the function

jerpelea (approved):

netutils/ftpd/ftpd.c
xiaoxiang781216: But I think it's legal to declare variables at the begin of block in C89. There are many places use this coding style.

pkarashchenko: Declaration at the beginning of the scope is legal for C89 as well as is compatible with the NuttX code style guide. Additionally it is used to reduce stack usage if the condition is not met. I do not see any issue in current code.

xiaoxiang781216: > @xiaoxiang781216 we need to get some device with MCU that only has C89 compiler to check these assumptions. Maybe to eZ80 board should be an option, i.e.: https://www.makerlisp.com

We don't need board, it's enough to build some ez80 config in ci.

pkarashchenko (approved):

xiaoxiang781216:
Let ignore the ci issue

PR #1342, opened 2022-10-10
ostest: add semaphore wait case for task restart

zyfeier:
## Summary

Add sem wait test case for task restart.

## Impact

NA

## Testing

sabre-6quad:smp


xiaoxiang781216:
@masayuki2009 could you verify both change?

PR #1346, opened 2022-10-13
Fixes in DHCP client.

fjpanag:
## Summary

Two fixes in DHCP client:

1. Previously, on every request the client was setting the current IP address to `INADDR_ANY`. This may be OK for the initial request, but it is problematic for renewals. If any other task is using the network simultaneously with the DHCP renewal, then it was disrupted. Also setting the address to `INADDR_ANY` has not any real advantage for DHCP, as it uses broadcasts. Thus this is removed. DHCP should change the IP only when a new valid lease is offered.
2. In case of a NAK response from the server, the current IP must not be used. Previously the code just exited, and didn't respect that. Now, when the IP is refused, the interface is set to `INADDR_ANY`, ensuring that it the old IP will not be used.

## Impact

Network communications should be more stable now during DHCP lease renewals.

## Testing

Tested on a custom target, based on the STM32F427.  
So far no issues encountered, the DHCP requests succeed, I saw no side effects during renewal.

### DIFF
--- a/netutils/dhcpc/dhcpc.c
+++ b/netutils/dhcpc/dhcpc.c
@@ -689,7 +689,6 @@ int dhcpc_request(FAR void *handle, FAR struct dhcpc_state *presult)
 {
   FAR struct dhcpc_state_s *pdhcpc = (FAR struct dhcpc_state_s *)handle;
   struct in_addr oldaddr;
-  struct in_addr newaddr;
   ssize_t result;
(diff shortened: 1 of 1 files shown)


fjpanag:
Is CI fixed now? Can you restart it?

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #1358, opened 2022-10-18
ostest:add sem protocl to priority inheritance

anjiahao1:
## Summary
ostest,sem use lock need set protocl `SEM_PRIO_INHERIT`
[patch](https://github.com/apache/incubator-nuttx/pull/5070)

## impact
use sem as lock,need do is change

## Testing
ostest

### DIFF
--- a/testing/ostest/prioinherit.c
+++ b/testing/ostest/prioinherit.c
@@ -530,6 +530,8 @@ void priority_inheritance(void)
   g_medpri = my_pri - 1;
 
   sem_init(&g_sem, 0, NLOWPRI_THREADS);
+  sem_setprotocol(&g_sem, SEM_PRIO_INHERIT);
+
   dump_nfreeholders("priority_inheritance:");
 
   /* Start the low priority threads */


xiaoxiang781216 (approved):

testing/ostest/prioinherit.c
```diff
@@ -530,6 +530,8 @@ void priority_inheritance(void)
   g_medpri = my_pri - 1;
 
   sem_init(&g_sem, 0, NLOWPRI_THREADS);
+  sem_setprotocol(&g_sem, SEM_PRIO_INHERIT);
```
pkarashchenko, line 533: why not change to nxmutex?

masayuki2009:
Do we need such a change if we enable the priority inheritance feature?

anjiahao1:
> @anjiahao1 @xiaoxiang781216 Do we need such a change if we enable the priority inheritance feature?

if sem used as lock,need set protocl to priority inheritance, others do not need

testing/ostest/prioinherit.c
anjiahao1: ostest is app,nxmutex is kernal,so we need add mutex for app?

xiaoxiang781216: since this is used to test the api usage, it's better to call the core api directly instead of wrapper.

xiaoxiang781216 (approved):

xiaoxiang781216:
@masayuki2009 @pkarashchenko do you have more concern about this change?

pkarashchenko:
> @masayuki2009 @pkarashchenko do you have more concern about this change?

This is a case of counting semaphore with priority inheritance enabled. We discussed this  https://github.com/apache/incubator-nuttx/pull/6965#discussion_r985285801 . So the question that I raise long time in the past comes again: Should priority inheritance be an option of semaphore or mutex only. In this case I'll rephrase it: Should we allow enabling of priority inheritance for non-binary semaphores. If yes, then we probably need to update many places in code with similar changes.

xiaoxiang781216:
> > @masayuki2009 @pkarashchenko do you have more concern about this change?
> 
> This is a case of counting semaphore with priority inheritance enabled.

Yes, but ostest is very special one, because it verify the internal detail of OS implementation, and understanding the real behavior change between the enable and disable of priority inheritance.

> We discussed this [apache/incubator-nuttx#6965 (comment)](https://github.com/apache/incubator-nuttx/pull/6965#discussion_r985285801) . So the question that I raise long time in the past comes again: Should priority inheritance be an option of semaphore or mutex only. In this case I'll rephrase it: Should we allow enabling of priority inheritance for non-binary semaphores. If yes, then we probably need to update many places in code with similar changes.

I don't believe that the non-binary semaphore can work well with the priority inheritance in general. The priority inheritance is meaningful only when:

1. The semaphore has the binary value(0 or 1)
2. The semaphore is used as lock

pkarashchenko:
Yeah, currently with `CONFIG_SEM_PREALLOCHOLDERS` > 2 this will work. I mean that the current implementation with some configuration allows priority inheritence on counting semaphores and really do not makes sense to me, but to unblock further changes.
IMO anyway we need to make a kernel implementation of mutex objects instead of making a semaphore wrapper, so application will use pthread_mutex and kernel will use nxmutex, but that may never happen since many things like mutex usage in libc in FLAT build are not clear.

pkarashchenko (approved):

masayuki2009:
Let me merge this PR since https://github.com/apache/incubator-nuttx/pull/5070 has been merged.

PR #1378, opened 2022-10-23
Revert "Minor fixes"

pkarashchenko:
Reverts apache/incubator-nuttx-apps#1377

### DIFF
--- a/examples/ajoystick/ajoy_main.c
+++ b/examples/ajoystick/ajoy_main.c
@@ -59,7 +59,6 @@
 #ifndef MIN
 #  define MIN(a,b) ((a) < (b) ? (a) : (b))
 #endif
-
 #ifndef MAX
 #  define MAX(a,b) ((a) > (b) ? (a) : (b))
 #endif
--- a/examples/djoystick/djoy_main.c
+++ b/examples/djoystick/djoy_main.c
@@ -56,7 +56,6 @@
 #ifndef MIN
 #  define MIN(a,b) ((a) < (b) ? (a) : (b))
 #endif
-
 #ifndef MAX
 #  define MAX(a,b) ((a) > (b) ? (a) : (b))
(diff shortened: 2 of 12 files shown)


xiaoxiang781216:
@pkarashchenko why revert the change?

pkarashchenko:
> @pkarashchenko why revert the change?

I made a mistake in one commit and CI on mainline failed because of that. I did revert because will not have time till evening to fix the broken commit, so just to unblock others.

pkarashchenko:
Let's try to fix compilation failure when GCC can't find `#include <array>`, so we can have reliable feedback from CI.

xiaoxiang781216:
> > @pkarashchenko why revert the change?
> 
> I made a mistake in one commit and CI on mainline failed because of that. I did revert because will not have time till evening to fix the broken commit, so just to unblock others.

which patch? I don't see your change can trigger `#include <array>` error.

pkarashchenko:
> > > @pkarashchenko why revert the change?
> > 
> > I made a mistake in one commit and CI on mainline failed because of that. I did revert because will not have time till evening to fix the broken commit, so just to unblock others.
> 
> which patch? I don't see your change can trigger `#include <array>` error.

https://github.com/apache/incubator-nuttx-apps/actions/runs/3304553082/jobs/5454071840#step:7:1

xiaoxiang781216:
Here is the fix for `#incude <array>`: https://github.com/apache/incubator-nuttx-apps/pull/1379, not your change @pkarashchenko .

pkarashchenko:
> Here is the fix for `#incude <array>`: https://github.com/apache/incubator-nuttx-apps/pull/1379, not your change @pkarashchenko .

Yes. Thank you. My changes still have an issue that was not caught by CI due to array issue. I will resubmit my change later today

PR #1379, opened 2022-10-23
Fix fatal error: array: No such file or directory

xiaoxiang781216:
## Summary
enable test_stl2 only when CONFIG_LIBCXX is enabled since uClibc++/minic++ doesn't have array header file

## Impact
Fixed the build break report here: https://github.com/apache/incubator-nuttx/actions/runs/3305999084/jobs/5456570758

## Testing
Pass CI


pkarashchenko (approved):

xiaoxiang781216:
Let's merge this patch directly to unblock ci.

PR #1404, opened 2022-11-09
interpreter/toywasm: update toywasm version

yamt:
## Summary
interpreter/toywasm: update toywasm version
## Testing
tested on esp32-devkitc


interpreters/toywasm/Makefile
```diff
@@ -51,21 +53,17 @@ CSRCS += vec.c
...
 CSRCS += xlog.c
 
-CFLAGS += -DTOYWASM_USE_SEPARATE_EXECUTE
-CFLAGS += -DTOYWASM_USE_TAILCALL
-CFLAGS += -DTOYWASM_USE_JUMP_BINARY_SEARCH
-CFLAGS += -DTOYWASM_JUMP_CACHE2_SIZE=4
-CFLAGS += -DTOYWASM_USE_SEPARATE_LOCALS
-CFLAGS += -DTOYWASM_USE_SMALL_CELLS
-CFLAGS += -DTOYWASM_ENABLE_WRITER
+CFLAGS += ${shell $(INCDIR) "$(CC)" $(APPDIR)/interpreters/toywasm/include}
```
xiaoxiang781216, line 56: why need this path? I don't find it.

yamt: to include this flie. https://github.com/apache/incubator-nuttx-apps/blob/16e873587ca898e66e749a22c793dfbcdb919515/interpreters/toywasm/include/toywasm_config.h

(in upstream toywasm repo, this header file is generated by cmake. for nuttx apps, i simply added the generated file instead.)

xiaoxiang781216: Ok, I see.

xiaoxiang781216 (approved):

interpreters/toywasm/Makefile
pkarashchenko, line 56: Shouldn't we use `INCDIR_PREFIX` instead? Ref https://github.com/apache/incubator-nuttx-apps/pull/1403

PR #1406, opened 2022-11-10
mm: add memory stress test

GUIDINGLI:
## Summary

mm: add memory stress test

## Impact

mm test

## Testing

VELA

### DIFF
--- a/testing/mm/mm_main.c
+++ b/testing/mm/mm_main.c
@@ -310,6 +310,60 @@ static void do_frees(FAR void **mem, FAR const int *size,
     }
 }
 
+static int mm_stress_test(int argc, FAR char *argv[])
+{
+  FAR unsigned char *tmp;
+  int delay = 1;
+  int prio = 0;
+  int size;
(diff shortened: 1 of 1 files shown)


testing/mm/mm_main.c
```diff
@@ -310,6 +310,60 @@ static void do_frees(FAR void **mem, FAR const int *size,
...
+  while (1)
+    {
+      size = random() % 1024 + 1;
+      tmp = malloc(size);
+      assert(tmp);
+
+      memset(tmp, 0xfe, size);
+      usleep(delay);
+
+      for (i = 0; i < size; i++)
```
pkarashchenko, line 356: Why not memcmp and assert the result?

pkarashchenko (approved):

testing/mm/mm_main.c
GUIDINGLI: memcmp may replace to the asm version. Than kasan can't insert check code.

PR #1407, opened 2022-11-10
Fix various error in Windows native build

zouboan:
## Summary
These patch finally raised NuttX Windows native build
## Impact
Windows native build
## Testing
Tested build of stm32f4discovery on Windows XP/7 + GnuW32 + MinGW + kconfig-frontends-win32
nuttx-win32

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -62,7 +62,8 @@ ifeq ($(BUILD_MODULE),y)
 endif
 
 SUFFIX = $(subst $(DELIM),.,$(CWD))
-PROGNAME := $(shell echo $(PROGNAME))
+
+PROGNAME := $(subst ",,$(PROGNAME))
 
 # Object files
 
--- a/Directory.mk
+++ b/Directory.mk
@@ -22,11 +22,15 @@ include $(APPDIR)/Make.defs
 
 # Sub-directories that have been built or configured.
 
-SUBDIRS       := $(dir $(wildcard *$(DELIM)Makefile))
-CONFIGSUBDIRS := $(filter-out $(dir $(wildcard *$(DELIM)Kconfig)),$(SUBDIRS))
-CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).depend))
-CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).kconfig))
+SUBDIRS       := $(dir $(wildcard */Makefile))
+CONFIGSUBDIRS := $(filter-out $(dir $(wildcard */Kconfig)),$(SUBDIRS))
+CLEANSUBDIRS  += $(dir $(wildcard */.depend))
+CLEANSUBDIRS  += $(dir $(wildcard */.kconfig))
 CLEANSUBDIRS  := $(sort $(CLEANSUBDIRS))
+ifeq ($(CONFIG_WINDOWS_NATIVE),y)
+	CONFIGSUBDIRS  := $(subst /,\,$(CONFIGSUBDIRS))
+	CLEANSUBDIRS  := $(subst /,\,$(CLEANSUBDIRS))
+endif
 
 all: nothing
 
--- a/Make.defs
+++ b/Make.defs
@@ -21,27 +21,22 @@
 TOPDIR ?= $(APPDIR)/import
 include $(TOPDIR)/Make.defs
 
-# The GNU make CURDIR will always be a POSIX-like path with forward slashes
-# as path segment separators.  This is fine for the above inclusions but
-# will cause problems later for the native build.  If we know that this is
-# a native build, then we need to fix up the APPDIR path for subsequent
-# use
-
-ifeq ($(CONFIG_WINDOWS_NATIVE),y)
-APPDIR := ${shell echo %CD%}
-endif
-
 # Application Directories
 
 # BUILDIRS is the list of top-level directories containing Make.defs files
 # CLEANDIRS is the list of all top-level directories containing Makefiles.
 #   It is used only for cleaning.
 
-BUILDIRS   := $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Make.defs))
-BUILDIRS   := $(filter-out $(APPDIR)$(DELIM)import$(DELIM),$(BUILDIRS))
-CONFIGDIRS := $(filter-out $(APPDIR)$(DELIM)builtin$(DELIM),$(BUILDIRS))
-CONFIGDIRS := $(filter-out $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Kconfig)),$(CONFIGDIRS))
-CLEANDIRS  := $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Makefile))
+BUILDIRS   := $(dir $(wildcard $(APPDIR)/*/Make.defs))
+BUILDIRS   := $(filter-out $(APPDIR)/import/,$(BUILDIRS))
+CONFIGDIRS := $(filter-out $(APPDIR)/builtin/,$(BUILDIRS))
+CONFIGDIRS := $(filter-out $(dir $(wildcard $(APPDIR)/*/Kconfig)),$(CONFIGDIRS))
+CLEANDIRS  := $(dir $(wildcard $(APPDIR)/*/Makefile))
+ifeq ($(CONFIG_WINDOWS_NATIVE),y)
+	BUILDIRS  := $(subst /,\,$(BUILDIRS))
+	CONFIGDIRS  := $(subst /,\,$(CONFIGDIRS))
+	CLEANDIRS  := $(subst /,\,$(CLEANDIRS))
+endif
 
 # CONFIGURED_APPS is the application directories that should be built in
 #   the current configuration.
@@ -85,14 +80,24 @@ endif
 BUILTIN_REGISTRY = $(APPDIR)$(DELIM)builtin$(DELIM)registry
 DEPCONFIG = $(TOPDIR)$(DELIM).config
 
+ifeq ($(CONFIG_WINDOWS_NATIVE),y)
 define REGISTER
 	$(Q) echo Register: $1
-	$(Q) echo { \"$1\", $2, $3, $4 }, > "$(BUILTIN_REGISTRY)$(DELIM)$1.bdat"
+	$(Q) echo { "$(subst ",,$(1))", $2, $3, $(subst ",,$(4)) }, > "$(BUILTIN_REGISTRY)$(DELIM)$1.bdat"
+	$(Q) echo int $(subst ",,$(4))(int argc, char *argv[]); > "$(BUILTIN_REGISTRY)$(DELIM)$1.pdat"
+
+	$(Q) touch $(BUILTIN_REGISTRY)$(DELIM).updated"
+endef
+else
+define REGISTER
+	$(Q) echo "Register: $1"
+	$(Q) echo "{ \"$1\", $2, $3, $4 }," > "$(BUILTIN_REGISTRY)$(DELIM)$1.bdat"
 	$(Q) if [ ! -z $4 ]; then \
 	        echo "int $4(int argc, char *argv[]);" > "$(BUILTIN_REGISTRY)$(DELIM)$1.pdat"; \
 	     fi;
 	$(Q) touch "$(BUILTIN_REGISTRY)$(DELIM).updated"
 endef
+endif
 
 # Standard include path
 
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,16 @@
 export APPDIR = $(CURDIR)
 include $(APPDIR)/Make.defs
 
+# The GNU make CURDIR will always be a POSIX-like path with forward slashes
+# as path segment separators.  This is fine for the above inclusions but
+# will cause problems later for the native build.  If we know that this is
+# a native build, then we need to fix up the APPDIR path for subsequent
+# use
+
+ifeq ($(CONFIG_WINDOWS_NATIVE),y)
+export APPDIR = $(subst /,\,$(CURDIR))
+endif
+
 # Symbol table for loadable apps.
 
 SYMTABSRC = symtab_apps.c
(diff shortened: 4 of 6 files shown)


Directory.mk
```diff
@@ -22,11 +22,15 @@ include $(APPDIR)/Make.defs
 
 # Sub-directories that have been built or configured.
 
-SUBDIRS       := $(dir $(wildcard *$(DELIM)Makefile))
-CONFIGSUBDIRS := $(filter-out $(dir $(wildcard *$(DELIM)Kconfig)),$(SUBDIRS))
-CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).depend))
-CLEANSUBDIRS  += $(dir $(wildcard *$(DELIM).kconfig))
+SUBDIRS       := $(dir $(wildcard */Makefile))
```
pkarashchenko, line 25: Why `$(DELIM)` is not working here?

Make.defs
```diff
@@ -21,27 +21,22 @@
...
 # BUILDIRS is the list of top-level directories containing Make.defs files
 # CLEANDIRS is the list of all top-level directories containing Makefiles.
 #   It is used only for cleaning.
 
-BUILDIRS   := $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Make.defs))
-BUILDIRS   := $(filter-out $(APPDIR)$(DELIM)import$(DELIM),$(BUILDIRS))
-CONFIGDIRS := $(filter-out $(APPDIR)$(DELIM)builtin$(DELIM),$(BUILDIRS))
-CONFIGDIRS := $(filter-out $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Kconfig)),$(CONFIGDIRS))
-CLEANDIRS  := $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Makefile))
+BUILDIRS   := $(dir $(wildcard $(APPDIR)/*/Make.defs))
```
pkarashchenko, line 30: ditto

Directory.mk
zouboan: result CONFIGSUBDIRS and CLEANSUBDIRS be empty set in Windows native build

Make.defs
zouboan: result CONFIGDIRS and BUILDIRS be empty set in Windows native build

Application.mk
```diff
@@ -62,7 +62,10 @@ ifeq ($(BUILD_MODULE),y)
 endif
 
 SUFFIX = $(subst $(DELIM),.,$(CWD))
+
+ifeq ($(CONFIG_WINDOWS_NATIVE),)
 PROGNAME := $(shell echo $(PROGNAME))
```
xiaoxiang781216, line 67: can we change `$(subst ",,$(PROGNAME))` to avoid ifeq

Application.mk
```diff
@@ -189,8 +192,23 @@ install:: $(PROGLIST)
 
 else
 
-MAINNAME := $(addsuffix _main,$(PROGNAME))
+MAINNAME := $(subst ",,$(addsuffix _main,$(PROGNAME)))
```
xiaoxiang781216, line 195: Do we still need change this line after https://github.com/apache/incubator-nuttx-apps/pull/1407/files#r1021549899

zouboan: quite right sir!

Makefile
```diff
@@ -21,6 +21,16 @@
 export APPDIR = $(CURDIR)
 include $(APPDIR)/Make.defs
 
+# The GNU make CURDIR will always be a POSIX-like path with forward slashes
+# as path segment separators.  This is fine for the above inclusions but
+# will cause problems later for the native build.  If we know that this is
+# a native build, then we need to fix up the APPDIR path for subsequent
+# use
+
+ifeq ($(CONFIG_WINDOWS_NATIVE),y)
```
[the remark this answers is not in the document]
zouboan: i have no good idea now, perhaps we can improve it in future?

Application.mk
zouboan: tested both on linux and windows native

Makefile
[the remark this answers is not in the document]
xiaoxiang781216: We can try whether windows style work on Linux/macOS and remove the ifeq if it work.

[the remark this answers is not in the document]
zouboan: I'm tried, but failed sadly!

Directory.mk
hartmannathan: Is `$(DELIM)` obsolete? Do all operating systems (including Windows) recognize `/` delimiter now?

zouboan: > Is `$(DELIM)` obsolete? Do all operating systems (including Windows) recognize `/` delimiter now?

Windows is not a POSIX environment,  some makefile command was not support in Windows, As suggest in nuttx/README we use GnuWin32  to supply this windows version Makefile commad(for example: wildcard), some special cases we must use / rather than \  ,for here, we must use /, otherwise CONFIGSUBDIRS and CLEANSUBDIRS would be empty set in Windows native build, and will give fatal error, but use / we can build succesfully.

Make.defs
```diff
@@ -21,27 +21,22 @@
...
 # Application Directories
 
 # BUILDIRS is the list of top-level directories containing Make.defs files
 # CLEANDIRS is the list of all top-level directories containing Makefiles.
 #   It is used only for cleaning.
 
-BUILDIRS   := $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Make.defs))
-BUILDIRS   := $(filter-out $(APPDIR)$(DELIM)import$(DELIM),$(BUILDIRS))
-CONFIGDIRS := $(filter-out $(APPDIR)$(DELIM)builtin$(DELIM),$(BUILDIRS))
-CONFIGDIRS := $(filter-out $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Kconfig)),$(CONFIGDIRS))
```
xiaoxiang781216, line 43: @zouboan what's $(DELIM) on Windows native build environment? I think you can adjust $(DELIM) to the right value instead, which is more simpler than replacing it in many place.

Application.mk
```diff
@@ -191,6 +192,21 @@ else
 
 MAINNAME := $(addsuffix _main,$(PROGNAME))
 
+ifeq ($(CONFIG_WINDOWS_NATIVE),y)
+$(MAINCXXOBJ): %$(CXXEXT)$(SUFFIX)$(OBJEXT): %$(CXXEXT)
+	$(eval $<_CXXFLAGS += ${shell $(DEFINE) "$(CXX)" main -val $(firstword $(MAINNAME))})
```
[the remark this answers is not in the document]
zouboan: because define.bat is different from the define.sh, and usage is also different this is usage of define.bat:
`echo USAGE:%progname% [-h] ^<compiler-path^> [-val ^<^val1^>] [^<def2^> [-val ^<val2^>] [^<def3^> [-val ^<val3^>] ...]]
echo Where:"
echo  ^<compiler-path^>
echo    The full path to your ccpath
echo  ^<def1^> ^<def2^> ^<def3^> ...
echo    A list of pre-preprocesser variable names to be defined.
echo  [-val ^<val1^>] [-val ^<val2^>] [-val ^<val3^>] ...
echo    optional values to be assigned to each pre-processor variable.
echo    If not supplied, the variable will be defined with no explicit value.`

without -val, there will fatal error in Windows native build, for example:
Dnsh_main become Dnsh Dmain

Make.defs
zouboan: In Windows environment `DELIM := \` but` \ `has two role in Windows environment:
first: `\ `as directory separator, and second `\` as Escape character
in Windows environment:
`BUILDIRS   := $(dir $(wildcard $(APPDIR)$(DELIM)*$(DELIM)Make.defs))`
equivalent to
`BUILDIRS   := $(dir $(wildcard $(APPDIR)\*\Make.defs))`
and the `\` of `\*`  was identified to be Escape character, which result the `* ` was identified to be ordinary characters not used for pattern. and then result the BUILDIRS to be empty.
As described in: [https://www.gnu.org/software/make/manual/html_node/Text-Functions.html](url)
we can use double $(DELIM) to solve problem, and i succeed resolved a similar problem in f91c2383d67ff89f694e51bc59d11f226764be46 
but  when i use same method in apps/Make.defs , it give extra problem
@hartmannathan @xiaoxiang781216

Application.mk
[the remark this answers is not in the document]
xiaoxiang781216: @zouboan it's betteer to sync define.bat and define.sh usage to avoid the duplication here.

Make.defs
```diff
@@ -85,14 +80,24 @@ endif
...
-	$(Q) echo { \"$1\", $2, $3, $4 }, > "$(BUILTIN_REGISTRY)$(DELIM)$1.bdat"
+	$(Q) echo { "$(subst ",,$(1))", $2, $3, $(subst ",,$(4)) }, > "$(BUILTIN_REGISTRY)$(DELIM)$1.bdat"
+	$(Q) echo int $(subst ",,$(4))(int argc, char *argv[]); > "$(BUILTIN_REGISTRY)$(DELIM)$1.pdat"
+
+	$(Q) touch $(BUILTIN_REGISTRY)$(DELIM).updated"
+endef
+else
+define REGISTER
+	$(Q) echo "Register: $1"
+	$(Q) echo "{ \"$1\", $2, $3, $4 }," > "$(BUILTIN_REGISTRY)$(DELIM)$1.bdat"
```
xiaoxiang781216, line 94: can we share the same code, the difference look like very small.

Makefile
```diff
@@ -21,6 +21,16 @@
...
+APPDIR = ${shell echo %CD%}
```
xiaoxiang781216, line 31: if so, should we replace line 21 directly

Makefile
```diff
@@ -83,11 +93,15 @@ else
 # symbol table is required.
 
 ifeq ($(CONFIG_BUILD_LOADABLE),)
-
+ifeq ($(CONFIG_WINDOWS_NATIVE),y)
+$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
+	$(Q) for %%G in ($(CONFIGURED_APPS)) do ( $(MAKE) -C %%G archive )
+else
 $(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
 	$(Q) for app in ${CONFIGURED_APPS}; do \
```
xiaoxiang781216, line 101: what's problem is fixed by change app to %%G

zouboan: the usage of `FOR` loop in Windows is:
`FOR %variable IN (set) DO command [command-parameters]`
to use the FOR command in a batch program, specify `%%variable` instead of `%variable`

I tried to use `$(Q) for %%G in ($(CONFIGURED_APPS)) do ( $(MAKE) -C %%G archive )` or I tried to use `$(Q) for %%G in ($(CONFIGURED_APPS)) do ( $(MAKE) -C ${%%G} archive )` in Linux, but failed

Make.defs
zouboan: In Windows:
`echo aaaaaaaaaaa `result `aaaaaaaaaaa`
`echo 'bbbbbb'` result `'bbbbbb'`
`echo "ccccccccc"` result `"ccccccccc"`
so:
`echo "int nsh_main(int argc, char *argv[]);" > "$(BUILTIN_REGISTRY)$(DELIM)nsh.pdat";` result "int nsh_main(int argc, char *argv[]);" in nsh.pdat, and will give an build error.

what's more nsh name generated by Kconfig is "nsh"
in ../apps/system/nsh/Makefile:
`PROGNAME = $(CONFIG_SYSTEM_NSH_PROGNAME) sh`
in ../apps/example/hello/Makefile:
`PROGNAME  = $(CONFIG_EXAMPLES_HELLO_PROGNAME)`
and CONFIG_EXAMPLES_HELLO_PROGNAME in Kconfig is "hello"
which result the parameter $(4) of function REGISTER in ../apps/Make.defs are:
"nsh" sh "hello" ,we have use $(subst ",,$(4)) to strip "" for Windows

I tried to use :
```
	$(Q) echo { "$(subst ",,$(1))", $2, $3, $(subst ",,$(4)) }, > "$(BUILTIN_REGISTRY)$(DELIM)$1.bdat"
	$(Q) echo int $(subst ",,$(4))(int argc, char *argv[]); > "$(BUILTIN_REGISTRY)$(DELIM)$1.pdat"
```
but failed, may be we need more tricks here.

Makefile
zouboan: do you means change Line 31: `APPDIR = ${shell echo %CD%}` to export `APPDIR = ${shell echo %CD%}` ?

Application.mk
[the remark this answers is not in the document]
zouboan: is a good idea! although i'm not expert in .bat and .sh but I wouldn't mind having a try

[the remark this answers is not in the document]
xiaoxiang781216: Ok, let's wait your result.

Makefile
xiaoxiang781216: Yes, you are right.

Makefile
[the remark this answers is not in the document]
xiaoxiang781216: but should we use the same approach(replace / with \?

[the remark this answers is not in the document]
zouboan: Sorry, what do you mean? I don't follow...
Can you give me some hints?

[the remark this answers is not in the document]
xiaoxiang781216: we can replace / with \ like other place, instead `${shell echo %CD%}`.
and it's better to merge with line 31

zouboan:
thanks for your advise，let me try!



---Original---
From: "Xiang ***@***.***&gt;
Date: Tue, Dec 6, 2022 18:45 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [apache/nuttx-apps] Fix various error in Windows native build (PR #1407)




 
@xiaoxiang781216 commented on this pull request.
 
 
In Makefile:
 &gt; @@ -21,6 +21,16 @@  export APPDIR = $(CURDIR)  include $(APPDIR)/Make.defs   +# The GNU make CURDIR will always be a POSIX-like path with forward slashes +# as path segment separators.  This is fine for the above inclusions but +# will cause problems later for the native build.  If we know that this is +# a native build, then we need to fix up the APPDIR path for subsequent +# use + +ifeq ($(CONFIG_WINDOWS_NATIVE),y)  
we can replace / with \ like other place, instead ${shell echo %CD%}.
 and it's better to merge with line 31

Makefile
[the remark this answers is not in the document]
zouboan: Done,thanks!

Makefile
```diff
@@ -21,6 +21,16 @@
 export APPDIR = $(CURDIR)
 include $(APPDIR)/Make.defs
```
xiaoxiang781216, line 22:
```suggestion
include $(CURDIR)/Make.defs
```
and move line 21 as else statement at line 30

zouboan: done

zouboan: sorry, ci failed, we move line 21 as else statement at line 30, because ../apps/Make.defs will use `APPDIR`

PR #1425, opened 2022-11-19
wapi: fix save_config can not update if string prefix is same

zhhyu7:
Signed-off-by: zhanghongyu <zhanghongyu@xiaomi.com>

### DIFF
--- a/wireless/wapi/src/util.c
+++ b/wireless/wapi/src/util.c
@@ -148,7 +148,7 @@ static bool wapi_json_update(FAR cJSON *root,
           int len = strlen(obj->valuestring);
           if (len > 0)
             {
-              if (!strncmp(value, obj->valuestring, len))
+              if (!strcmp(value, obj->valuestring))
                 {
                   return false;
                 }


wireless/wapi/src/util.c
```diff
@@ -148,7 +148,7 @@ static bool wapi_json_update(FAR cJSON *root,
           int len = strlen(obj->valuestring);
           if (len > 0)
             {
-              if (!strncmp(value, obj->valuestring, len))
+              if (!strcmp(value, obj->valuestring))
```
pkarashchenko, line 151: could you please add spaces at line 159?

pkarashchenko (approved):

wireless/wapi/src/util.c
zhhyu7: > could you please add spaces at line 159?

Can you give me an example, Line 159 seems pretty normal to me...am I missing something？

pkarashchenko: Oops. It seems that it was a broken preview from my browser. It looked like 159 line was missing 2 spaces before `{`. I looked there again and it seems like all is fine.

PR #1426, opened 2022-11-19
Fix clang compiler warning

xiaoxiang781216:
## Summary
Found by https://github.com/apache/incubator-nuttx/pull/7391

## Impact
Minor

## Testing
Pass CI

### DIFF
--- a/examples/dac/dac_main.c
+++ b/examples/dac/dac_main.c
@@ -110,7 +110,7 @@ static void print_cmds(FAR const char *header,
                        size_t ncmds,
                        FAR const char *trailer)
 {
-  printf(header);
+  printf("%s", header);
   while (ncmds--)
     {
       printf("  %s %s %c", cmds->name, cmds->args,
@@ -118,7 +118,7 @@ static void print_cmds(FAR const char *header,
       cmds++;
     }
 
-  printf(trailer);
+  printf("%s", trailer);
 }
 
 static const struct command *find_cmd(FAR const char *name,
--- a/examples/foc/foc_fixed16_thr.c
+++ b/examples/foc/foc_fixed16_thr.c
@@ -185,9 +185,10 @@ int foc_fixed16_thr(FAR struct foc_ctrl_env_s *envp)
   struct foc_mq_s         handle;
   struct foc_motor_b16_s  motor;
   struct foc_device_s     dev;
-  int                     time      = 0;
-  int                     ret       = OK;
+  int                     time = 0;
+  int                     ret  = OK;
 
+  UNUSED(time);
   DEBUGASSERT(envp);
 
   PRINTFV("foc_fixed_thr, id=%d\n", envp->id);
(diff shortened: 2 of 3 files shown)


examples/dac/dac_main.c
```diff
@@ -110,15 +110,15 @@ static void print_cmds(FAR const char *header,
                        size_t ncmds,
                        FAR const char *trailer)
 {
-  printf(header);
+  printf("%s", header);
```
pkarashchenko, line 113: Sorry, but what was the warning here? The `header` seems to be `const char*`

examples/dac/dac_main.c
```diff
@@ -110,15 +110,15 @@ static void print_cmds(FAR const char *header,
...
   while (ncmds--)
     {
       printf("  %s %s %c", cmds->name, cmds->args,
                            (ncmds > 0) ? '\n' : ' ');
       cmds++;
     }
 
-  printf(trailer);
+  printf("%s", trailer);
```
pkarashchenko, line 121: ditto

examples/foc/foc_fixed16_thr.c
```diff
@@ -185,9 +185,10 @@ int foc_fixed16_thr(FAR struct foc_ctrl_env_s *envp)
   struct foc_mq_s         handle;
   struct foc_motor_b16_s  motor;
   struct foc_device_s     dev;
-  int                     time      = 0;
-  int                     ret       = OK;
+  int                     time = 0;
```
pkarashchenko, line 188: Maybe
```suggestion
#ifdef FOC_STATE_PRINT_PRE
  int                     time = 0;
#endif
```
instead?

examples/foc/foc_float_thr.c
```diff
@@ -186,9 +186,10 @@ int foc_float_thr(FAR struct foc_ctrl_env_s *envp)
   struct foc_mq_s         handle;
   struct foc_motor_f32_s  motor;
   struct foc_device_s     dev;
-  int                     time      = 0;
-  int                     ret       = OK;
+  int                     time = 0;
```
pkarashchenko, line 189: ditto

examples/dac/dac_main.c
xiaoxiang781216: Here:
```
Error: format string is not a string literal (potentially insecure)
```
In general is dangerous to printf(str) and the content of str can't be determined in compile time since str may contain % and trigger printf do some bad action.

examples/foc/foc_fixed16_thr.c
xiaoxiang781216: Can't since time is also used in other place:
https://github.com/apache/incubator-nuttx-apps/pull/1426/files#diff-df4ad1889816b80fd16d4fa0906888c2f9313974ee14e02acd88231f569c7afcR234
https://github.com/apache/incubator-nuttx-apps/pull/1426/files#diff-df4ad1889816b80fd16d4fa0906888c2f9313974ee14e02acd88231f569c7afcR372

examples/dac/dac_main.c
pkarashchenko: The printf is used below, so let's go with the current change. Otherwise I would recommend to use puts

pkarashchenko (approved):

PR #955, opened 2022-01-04
ostest: sighand.c: add sem_wait in signal handler

zhaoxiu-zeng:
task is blocked by semphore1, signal handler is blocked by semphore2,
after post semphore2, the task must get -EINTR.

Signed-off-by: Zeng Zhaoxiu <zhaoxiu.zeng@gmail.com>

PR #1463, opened 2022-12-20
ostest: remove printf & sem_wait in sigaction

GUIDINGLI:
## Summary

ostest: remove printf & sem_wait in sigaction

As the note said:
Use of printf in a signal handler is NOT safe! It can cause deadlocks!

## Impact

ostest signal test

## Testing

VELA


masayuki2009:
@GUIDINGLI 
Does this PR relate to the following PR?
https://github.com/apache/nuttx-apps/pull/955

GUIDINGLI:
@masayuki2009 
Yes, I think we can't do sem_wait in sigaction, that will caused error.

PR #1480, opened 2022-12-28
NSH Sysinit optional exec

PetervdPerk-NXP:
## Summary
Make the sysinit startup script optional since most existing systems just only use rcS and get this error now.
`nsh: sysinit: fopen failed: No such file or directory`


Donny9:
@PetervdPerk-NXP hello, let's remove this unnecessary error log? if add new config, we need add this config to all defconfig file with this script file, this modify is too large?

xiaoxiang781216:
Since https://github.com/apache/nuttx-apps/pull/1481 is simpler to fix the problem, let's close this one.

PR #1481, opened 2022-12-28
nshlib/script: delete unnecessary log when missing script file

Donny9:
## Summary
nshlib/script: delete unnecessary log when missing script file.
Signed-off-by: dongjiuzhu1 <dongjiuzhu1@xiaomi.com>

## Impact
delete unnecessary log 
## Testing
local test

### DIFF
--- a/nshlib/nsh_script.c
+++ b/nshlib/nsh_script.c
@@ -37,7 +37,8 @@
 #if defined(CONFIG_NSH_ROMFSETC) || defined(CONFIG_NSH_ROMFSRC)
 static int nsh_script_redirect(FAR struct nsh_vtbl_s *vtbl,
                                FAR const char *cmd,
-                               FAR const char *path)
+                               FAR const char *path,
+                               bool log)
 {
   uint8_t save[SAVE_SIZE];
   int fd = -1;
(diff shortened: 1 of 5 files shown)


davids5:
@Donny9 Will this hide an error message for a missing script? If so this needs to be a Kconfig option or not merged.

Donny9:
> @Donny9 Will this hide an error message for a missing script? If so this needs to be a Kconfig option or not merged.

Ok, Control log output only for missing rc.sysinit by add parameter log. what do you think? @davids5

davids5 (commented):
@Donny9 - that works. Should there be comment for the the use and state why sysinit is not logged

xiaoxiang781216 (approved):

nshlib/nsh_script.c
```diff
@@ -194,7 +198,11 @@ int nsh_script(FAR struct nsh_vtbl_s *vtbl, FAR const FAR char *cmd,
 #ifdef CONFIG_NSH_ROMFSETC
 int nsh_sysinitscript(FAR struct nsh_vtbl_s *vtbl)
 {
-  return nsh_script_redirect(vtbl, "sysinit", NSH_SYSINITPATH);
+  /* Closing error log output for missing NSH_SYSINITPATH script file
```
davids5, line 201:
```suggestion
  /* Since most existing systems only use the NSH_INITPATH script file. Do not log error output for a missing 
   * NSH_SYSINITPATH script file.
```

davids5: I think this is clearer wording. But the formatting needs to be fixed

davids5 (changes requested):

nshlib/nsh_script.c
Donny9: Done!

davids5 (approved):

davids5:
@Donny9 - There are real CI failures.

PR #1487, opened 2023-01-04
move embedlog from system to logging

raiden00pl:
## Summary
move embedlog from apps/system to apps/logging
FYI @mlyszczek 

## Impact
Breaking change for projects that use embedlog

## Testing
sim/nsh with examples/embedlog


mlyszczek:
> Breaking change for projects that use embedlog

Why would this change break anything? Path of include file should not change in my opinion. It should stay in path as in `#include <embedlog.h>`. This is how it's on Linux and other Unices, and so should be the same on Nuttx.

raiden00pl:
Kconfig definitions are changed, so project configuration files must be updated. I think it qualifies as 'breaking change' :]

mlyszczek:
This is minor I would say - you just have to change your own project. If include pathing is broken, this will be a bigger issue, as all out of tree project will have to include patch to change the path. Especially when project is built both on Linux (for testing) and Nuttx. If the change does not break `#include` then I say +1.

xiaoxiang781216 (approved):

PR #1492, opened 2023-01-12
Enable TMPFS and NSH_QUOTE to get CONFIG_NSH_CMDPARMS working

acassis:
## Summary
Enable TMPFS and NSH_QUOTE to get CONFIG_NSH_CMDPARMS working
## Impact
Now users will get it working because it selects automatically the required features
## Testing
ESP32-DEVKIT

### DIFF
--- a/nshlib/Kconfig
+++ b/nshlib/Kconfig
@@ -125,8 +125,9 @@ config NSH_QUOTE
 
 config NSH_CMDPARMS
 	bool "Enable commands as parameters"
-	default !DEFAULT_SMALL
+	default !DEFAULT_SMALL && NSH_QUOTE
 	depends on !DISABLE_MOUNTPOINT
+	depends on FS_TMPFS
 	---help---
 		If selected, then the output from commands, from file applications, and
 		from NSH built-in commands can be used as arguments to other
@@ -141,6 +142,8 @@ config NSH_CMDPARMS
 
 		Because this feature commits significant resources, it is disabled by
 		default.
+
+		This features requires TMPFS mounted at /tmp and NSH_QUOTE enabled.
 
 config NSH_MAXARGUMENTS
 	int "Maximum number of command arguments"


nshlib/Kconfig
```diff
@@ -127,6 +127,8 @@ config NSH_CMDPARMS
 	bool "Enable commands as parameters"
 	default !DEFAULT_SMALL
 	depends on !DISABLE_MOUNTPOINT
+	select FS_TMPFS
+	select NSH_QUOTE
```
xiaoxiang781216, line 131: should we add depends on ?

acassis: I don't like the idea of using 'depend on' when the dependent features is not enabled by default. It will hide the current feature from user.

acassis:
Ok, I think new solution is better, I will sent a patch proposing to enable FS_TMPFS by default if !DEFAULT_SMALL

nshlib/Kconfig
acassis: Fixed, doesn't enable it

hartmannathan: Too bad there isn't a feature in the Kconfig menu to show all features, including those with missing dependent features. By default they would be hidden, but the user could show all features to find out what other possibilities exist. This would require opening a PR with Kconfig-frontends. I think that's [here](https://github.com/uvc-ingenieure/kconfig-frontends) now? (Not sure if that's ahead or behind the one we've used in the past.)

Alternately it would be cool if autogenerated docs from the Kconfigs could show everything and explain the dependencies. We could probably do this.

xiaoxiang781216 (approved):

nshlib/Kconfig
btashton: I actually did this a year ago (closed the PR) because it took forever to render all the documentation pages because of how many configs we have :(

btashton: > Alternately it would be cool if autogenerated docs from the Kconfigs could show everything and explain the dependencies. We could probably do this.

I actually did this a year ago (closed the PR) because it took forever to render all the documentation pages because of how many configs we have :(

hartmannathan: Do you have the PR still? Maybe it's possible to speed it up by using a different algorithm. If Kconfig can process it quickly, then a tool to convert it to documentation should be able to do its thing quickly, too. I'm willing to look into it if there's some kind of starting point to put me on the right path :-)

btashton: > Do you have the PR still? Maybe it's possible to speed it up by using a different algorithm. If Kconfig can process it quickly, then a tool to convert it to documentation should be able to do its thing quickly, too. I'm willing to look into it if there's some kind of starting point to put me on the right path :-)

https://github.com/apache/nuttx/pull/3608

xiaoxiang781216:
@acassis some defconfig is still not normalized?

acassis:
> @acassis some defconfig is still not normalized?

@xiaoxiang781216 I did a git pull here in my upstream branch and did:

$ ./tools/refresh.sh --silent --defaults all

$ git diff

It didn't show any defconfig modified. I'll force the github CI to tested, let see!

acassis:
@xiaoxiang781216 could you please merge?

jerpelea (approved):

PR #1493, opened 2023-01-13
wireless/wapi:rework wapi error print in util.h

XinStellaris:
## Summary
I want to control print information through setlogmask()，so I changed wapi err print to wlerr.
I don't get why the original wapi util.h defined its own err print, which seems redundant to me.

Please review, I don't know if I am missing anything. 

## Impact
Only on wapi err print message.

## Testing
Tested on ESP32C3.


pkarashchenko (approved):
LGTM!

PR #1500, opened 2023-01-24
ostest: Add initial support for CONFIG_BUILD_KERNEL

pussuw:
## Summary
Implement task_create and task_delete for use with ostest. The functions are unavailable in kernel build mode but overload them here and create a pthread worker instead. Some tests should be re-visited because the intent is to user another task (in this case another process) to e.g. receive signals. That will require quite a bit of extra work.

Tests that had to be disabled:
- waitpid: pthreads don't work with waitpid, but this can be fixed by starting a dummy process instead and using waitpid on that (NOT IMPLEMENTED)
- restart: task_restart() does not work at all with kernel mode so it is disabled entirely
- fpu: make sure the FPU test is not even attempted, because it will cause ASSERT() and stop the test
- vfork: vfork() does not work for some reason in CONFIG_BUILD_KERNEL, there is something missing on the kernel side, so just disable the test for now

Tests that should be re-visited:
- The signal tests, now they signal the process itself while before the signal was sent to another task. This will require building the part that receives the signal as a separate process
- waitpid: Like stated above, waitpid does not work for pthreads
- user_main: It might be a better idea to just call user_main() because the "test creating a task works"-part has already been tested; if ostest starts, then task starting and creation also works
## Impact
Adds basic support for ostest with CONFIG_BUILD_KERNEL=y
## Testing
icicle:knsh

### DIFF
--- a/testing/ostest/Kconfig
+++ b/testing/ostest/Kconfig
@@ -82,7 +82,8 @@ if ARCH_FPU && SCHED_WAITPID
 
 config TESTING_OSTEST_FPUTESTDISABLE
 	bool "Disable FPU test"
-	default n
+	default y if !BUILD_FLAT
+	default n if BUILD_FAT
 
 if !TESTING_OSTEST_FPUTESTDISABLE
 
--- a/testing/ostest/Makefile
+++ b/testing/ostest/Makefile
@@ -123,4 +123,8 @@ ifeq ($(CONFIG_ARCH_SETJMP_H),y)
 CSRCS += setjmp.c
 endif
 
+ifeq ($(CONFIG_BUILD_KERNEL),y)
+CSRCS += task.c
+endif
+
 include $(APPDIR)/Application.mk
--- a/testing/ostest/ostest.h
+++ b/testing/ostest/ostest.h
@@ -118,11 +118,13 @@ void aio_test(void);
 
 /* restart.c ****************************************************************/
 
+#ifndef CONFIG_BUILD_KERNEL
 void restart_test(void);
+#endif
 
 /* waitpid.c ****************************************************************/
 
-#ifdef CONFIG_SCHED_WAITPID
+#if defined(CONFIG_SCHED_WAITPID) && !defined(CONFIG_BUILD_KERNEL)
 int waitpid_test(void);
 #endif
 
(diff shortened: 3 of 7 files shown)


xiaoxiang781216:
Why not skip the test which involve task_xxx function instead simulating task by pthread?

xiaoxiang781216:
> > Why not skip the test which involve task_xxx function instead simulating task by pthread?
> 
> Of course that is another way of doing it. I thought this would be better, many tests create "workers" via task_create and I thought it would be fine to use pthreads for that.
> 

If the case doesn't depend on task_create, it's better to migrate to pthread or remove task_create calling directly.

> We currently do not have any kind of ostest for CONFIG_BUILD_KERNEL, I made this to have at least something. It is not a full implementation but IMO better than nothing.

If the test invokes the function doesn't exist in kernel mode:

1. Change to the similar function exist in kernel mode
2. Skip the test by `#ifdef CONFIG_BUILD_KERNEL` if it indeed test the specific function

pussuw:
> If the case doesn't really depend on task_create, it's better to migrate to pthread or remove task_create calling directly.
I would have to go them through one by one, because I don't know how each test is supposed to work. Instead, I made a kludge solution that allows the test to compile and run (most) tests.

pussuw:
> If the test invokes the function which is impossible to implement in kernel mode:
> 
>     1. Change to the similar function exist in kernel mode


The similar function is posix_spawn(), which requires:
- The test case is split into another (or perhaps even multiple) process's, compiled into a file, which can be started with posix_spawn()
- Potential communication between the two (or multiple) processes
Achieving this kind of functionality is a lot of work which I'm not prepared to do at this point.

>     2. Skip the test by `#ifdef CONFIG_BUILD_KERNEL` if it indeed test the specific function
Skipping these kind of tests is fine by me, but I don't know which tests could be converted to use pthread and which must be skipped. I already disabled some of the tests that I know won't work in kernel mode.

pussuw:
Btw, I 100% agree with you that this incomplete implementation is not perfect and needs more work + clean up. 

I have been using this version for some time now and have found it useful (better than nothing) and I decided to make this PR to at least offer the solution to the community.

However, if no extra value is seen in this partial implementation, I will just close the PR.

Maybe at some point I will make a proper implementation for this, but as I said, I don't have time to do that now.

xiaoxiang781216:
The patch is valuable. My suggestion is very simple that:

1. Remove user_main, let's start the test from main directly
2. Skip the test case which invoke the task_xxx function in kernel mode
3. Remove the task_xxx simulation layer

PR #1519, opened 2023-01-31
ifconfig: Support `ifconfig interface` and update arg limit

wengzhe:
## Summary
From comment in cmd_ifconfig, it seems that we support `ifconfig [interface]`, but actually not considering the interface param now. After this commit, we only print the information of interface specified in param, just like linux.

Also updates the arg limit of the ifconfig command to meet current function.
## Impact
Support `ifconfig eth0` to only print information on `eth0`

## Testing
Manually on sim.

### DIFF
--- a/nshlib/nsh_command.c
+++ b/nshlib/nsh_command.c
@@ -244,9 +244,10 @@ static const struct cmdmap_s g_cmdmap[] =
 
 #ifdef CONFIG_NET
 # ifndef CONFIG_NSH_DISABLE_IFCONFIG
-  { "ifconfig", cmd_ifconfig, 1, 11,
-    "[interface [<ip-address>|dhcp]] [dr|gw|gateway <dr-address>] "
-    "[netmask <net-mask>] [dns <dns-address>] [hw <hw-mac>]" },
+  { "ifconfig", cmd_ifconfig, 1, 12,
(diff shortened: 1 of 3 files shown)


pkarashchenko (approved):

nshlib/nsh_command.c
```diff
@@ -244,9 +244,10 @@ static const struct cmdmap_s g_cmdmap[] =
 
 #ifdef CONFIG_NET
 # ifndef CONFIG_NSH_DISABLE_IFCONFIG
-  { "ifconfig", cmd_ifconfig, 1, 11,
-    "[interface [<ip-address>|dhcp]] [dr|gw|gateway <dr-address>] "
-    "[netmask <net-mask>] [dns <dns-address>] [hw <hw-mac>]" },
+  { "ifconfig", cmd_ifconfig, 1, 12,
```
pkarashchenko, line 247: BTW, why 11 is changed to 12?

wengzhe: Because previously `[address_family]` is not taken into count (which is needed for IPv6 and optional for IPv4), and we may have the longest command like `ifconfig eth0 inet6 fc00::2 gw fc00::1 prefixlen 64 dns 2001:4860:4860::8888 hw aa:bb:cc:dd:ee:ff`

PR #1539, opened 2023-02-04
build: Remove the annoy warning from distclean

xiaoxiang781216:
## Summary

and nuttx/tools/testbuild.sh already do the same check.

## Impact

Minor

## Testing

CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,20 +205,6 @@ clean: $(foreach SDIR, $(CLEANDIRS), $(SDIR)_clean)
 	$(call CLEAN)
 
 distclean: $(foreach SDIR, $(CLEANDIRS), $(SDIR)_distclean)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -205,20 +205,6 @@ clean: $(foreach SDIR, $(CLEANDIRS), $(SDIR)_clean)
 	$(call CLEAN)
 
 distclean: $(foreach SDIR, $(CLEANDIRS), $(SDIR)_distclean)
-ifeq ($(CONFIG_WINDOWS_NATIVE),y)
```
pkarashchenko, line 208: But does testbuild.sh suitable for windows native build?

xiaoxiang781216: Linux/macOS CI could catch all path since they cover all defined config except a few ones.

GUIDINGLI (approved):

PR #1563, opened 2023-02-13
system/cle.c: Fix configuration for CONFIG_EOL_IS_EITHER_CRLF=y

pussuw:
## Summary
The old flagging was insufficient, only '\r' was accepted as a line terminator if CONFIG_EOL_IS_EITHER_CRLF was set.

I noticed this regression after:
https://github.com/apache/nuttx/pull/8454
## Impact
Now that CR is bucketed into LF, this patches the nsh terminal, if the PC terminal uses CR
## Testing
icicle:nsh

### DIFF
--- a/system/cle/cle.c
+++ b/system/cle/cle.c
@@ -1077,12 +1077,12 @@ static int cle_editloop(FAR struct cle_s *priv)
 
         /* Newline terminates editing.  But what is a newline? */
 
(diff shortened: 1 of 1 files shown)


pussuw:
I think the EOL_IS_XXX macros will be removed at some point, so if this fix is not relevant any longer, I can just abandon.

xiaoxiang781216 (approved):

jerpelea (approved):

PR #1564, opened 2023-02-13
Bugfix on downloading lvgl for each build

spi3ex:
## Summary
While building LVGL #include<lvgl/lvgl.h> the make build system downloads every time from the repository even when the file is present. Tweaked the build to skip downloading again if LVGL makefiles are already Present.
## Impact
 When building using "make clean" and "make", everytime lvgl is downloaded irrespective of whether the source is present.
## Testing
Unit Tests on With lvgl Source and without lvgl source was done.


spi3ex:
modified

PR #1568, opened 2023-02-16
netinit: Improve WiFi configuration

acassis:
## Summary
The current WiFi configuration expects user to look at symbols in the wireless headers to configure it. This patch will improve it. Notice that even those symbols that are bitfields are not allowed to have more than 1 bit set at time, so this modification will not have impact since the bits were already mutually exclusive.
## Impact
Make it easier to users to configure the WiFi
## Testing
ESP32


xiaoxiang781216 (approved):

pkarashchenko:
Some defconfigs need to be adjusted now

pkarashchenko (approved):

PR #1572, opened 2023-02-16
trace: fix get_task_context memory leak

Gary-Hobson:
## Summary
fix get_task_context memory leak
The memory of malloc is not added to the linked list, resulting in a memory leak

### DIFF
--- a/system/trace/trace_dump.c
+++ b/system/trace/trace_dump.c
@@ -223,28 +223,28 @@ static void copy_task_name(FAR char *dst, FAR const char *src)
 FAR static struct trace_dump_task_context_s *get_task_context(pid_t pid,
                                       FAR struct trace_dump_context_s *ctx)
 {
-  FAR struct trace_dump_task_context_s **tctxp;
-  tctxp = &ctx->task;
-  while (*tctxp != NULL)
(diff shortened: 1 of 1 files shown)


system/trace/trace_dump.c
```diff
@@ -223,28 +223,45 @@ static void copy_task_name(FAR char *dst, FAR const char *src)
 FAR static struct trace_dump_task_context_s *get_task_context(pid_t pid,
                                       FAR struct trace_dump_context_s *ctx)
 {
-  FAR struct trace_dump_task_context_s **tctxp;
-  tctxp = &ctx->task;
-  while (*tctxp != NULL)
+  FAR struct trace_dump_task_context_s *tctxp;
```
xiaoxiang781216, line 226: I don't see the benefit to change tctxp from AR struct trace_dump_task_context_s ** to AR struct trace_dump_task_context_s *.

xiaoxiang781216:
> ## Summary
> fix get_task_context memory leak The memory of malloc is not added to the linked list, resulting in a memory leak
> 

The leak is due to the first patch in this PR, the original code doesn't have this problem, I think.

PR #1574, opened 2023-02-17
sync ps/assert output

flyingfish89:
## Summary
sync ps/assert output. Changes to be synchronized https://github.com/apache/nuttx/pull/8556/files

### DIFF
--- a/nshlib/nsh_proccmds.c
+++ b/nshlib/nsh_proccmds.c
@@ -348,7 +348,7 @@ static int ps_callback(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
   nsh_output(vtbl, "%3s ", status.td_cpu);
 #endif
 
-  nsh_output(vtbl, "%3s %-8s %-7s %3s %-8s %-9s ",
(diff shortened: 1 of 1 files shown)


nshlib/nsh_proccmds.c
```diff
@@ -348,7 +348,7 @@ static int ps_callback(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
   nsh_output(vtbl, "%3s ", status.td_cpu);
 #endif
 
-  nsh_output(vtbl, "%3s %-8s %-7s %3s %-8s %-9s ",
+  nsh_output(vtbl, "%3s %-8s %-7s %3s %-8s %-15s ",
```
xiaoxiang781216, line 351: why need extend the width?

xiaoxiang781216: do we still need this patch after we reduce the output length [https://github.com/apache/nuttx/pull/8556](https://github.com/apache/nuttx/pull/8556/files)

flyingfish89: don't need,hhh

PR #1590, opened 2023-02-23
interpreters/bas: Fix the nxstyle warning

xiaoxiang781216:
## Summary

except "Mixed case identifier found"

## Impact

Minor

## Testing

CI

### DIFF
--- a/interpreters/bas/bas_value.c
+++ b/interpreters/bas/bas_value.c
@@ -13,8 +13,8 @@
  * The above copyright notice and this permission notice shall be included in
  * all copies or substantial portions of the Software.
  *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
- * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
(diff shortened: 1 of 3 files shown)


acassis:
Does sim:bastest still working after this PR?

xiaoxiang781216:
> Does sim:bastest still working after this PR?

Yes.
```
NuttShell (NSH) NuttX-12.0.0
nsh> bas
bas 2.4
Copyright 1999-2014 Michael Haardt.
This is free software with ABSOLUTELY NO WARRANTY.
> 
> 
> help
Error: Undeclared function or variable at: help
> ?

> 4+5
> print 4+5
 9 
```

interpreters/bas/bas_value.c
```diff
@@ -2026,9 +2042,21 @@ struct String *Value_toWrite(struct Value *this, struct String *s)
...
+  {
+    V_REAL
+  };
+
+  static struct Value string =
+  {
+    V_STRING
+  };
+
   static char n[] = "";
```
pkarashchenko, line 2060: Can we remove `n` and use `""` directly?

xiaoxiang781216: Done.

interpreters/bas/bas_value.c
```diff
@@ -2026,18 +2042,29 @@ struct String *Value_toWrite(struct Value *this, struct String *s)
...
+  {
+    V_REAL
+  };
+
+  static struct Value string =
+  {
+    V_STRING
+  };
+
   static int init = 0;
```
pkarashchenko, line 2060: Also I do not see where `init` is set to 1. Is that needed at all? Can we go with static initialization for all the structures?

xiaoxiang781216: > Also I do not see where `init` is set to 1. Is that needed at all? Can we go with static initialization for all the structures?

No, struct Value contain union, it's hard to initialize union with c89 syntax. So, I add code to set init to 1

pkarashchenko (approved):

interpreters/bas/bas_value.c
xiaoxiang781216: but the init value point to a string which is an null string, not zero.

pkarashchenko: Yeap. Missed that. So there is no other way.

PR #1597, opened 2023-02-25
add "hello_nim" example application written in Nim.

centurysys:
## Summary

Added sample application in Nim language

## Impact

It allows you to create applications in the Nim language, which can be easily integrated with the C language so that you can use NuttX features in Nim right away.

1.  No need to create HAL, binding modules.
2.  Can use Nim standard libraries, external libraries.

## Testing

Tested on sim, arm Cortex-A5, Cortex-M4.

1.  install choosenim, latest Nim language.

> https://nim-lang.org/install_unix.html
>  `$ choosenim devel --latest`

2. configure NuttX.

This example uses async/await, which requires the following items to be enabled

>CONFIG_SCHED_CHILD_STATUS=y
>CONFIG_NET=y
>CONFIG_NET_SOCKOPTS=y

important:
I am sending out a pull request to clear Nim's asyncdispatch module global dispatcher at the end of the task.

https://github.com/nim-lang/Nim/pull/21432

Until this accepted, the second and subsequent runs will fail.

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -1,0 +1,126 @@
+############################################################################
+# apps/config.nims
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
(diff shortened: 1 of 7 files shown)


xiaoxiang781216:
Please fix the nxstyle error reported here:
https://github.com/apache/nuttx-apps/actions/runs/4272067879/jobs/7438598461

xiaoxiang781216:
Fxing "Mixed case identifier found" is optional since the mixed case can't avoid completely in some context. Anyway, you can add NimMain to the white list here: https://github.com/apache/nuttx/blob/master/tools/nxstyle.c#L500

centurysys:
Thank you very much.
I see that we need to make a change to add it to the whitelist of the repository on the nuttx OS side. Would you be willing to merge this one after I deal with that one?

xiaoxiang781216:
Sure.

centurysys:
Thank you for merging the changes on the OS side.
I have modified the style, is this correct?

xiaoxiang781216:
Let's see the style check report.

xiaoxiang781216 (approved):

jerpelea (approved):

PR #1608, opened 2023-03-02
readline: Don't echo input except VT100 command

no1wudi:
## Summary
Let's echo character by serial driver, this will imporve the experience of telnetd, which will display input command and newline twice.
And also enhance the compatiblity with some third party linux readline based application (like WAMR repl).
## Impact
nsh
## Testing
nsh over serial and telnetd

### DIFF
--- a/system/readline/readline_common.c
+++ b/system/readline/readline_common.c
@@ -708,11 +708,6 @@ ssize_t readline_common(FAR struct rl_common_s *vtbl, FAR char *buf,
           buf[nch++] = '\n';
           buf[nch]   = '\0';
 
-#ifdef CONFIG_READLINE_ECHO
-          /* Echo the newline to the console */
-
-          RL_PUTC(vtbl, '\n');
-#endif
           return nch;
         }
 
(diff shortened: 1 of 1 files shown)


system/readline/readline_common.c
```diff
@@ -708,11 +708,6 @@ ssize_t readline_common(FAR struct rl_common_s *vtbl, FAR char *buf,
           buf[nch++] = '\n';
           buf[nch]   = '\0';
 
-#ifdef CONFIG_READLINE_ECHO
```
xiaoxiang781216, line 711: should we remove CONFIG_READLINE_ECHO?

no1wudi: Maybe should rename it to represent its function now (cmd history/backspace etc)

xiaoxiang781216: understand, let's continue use CONFIG_READLINE_ECHO represent other echo.

xiaoxiang781216 (approved):

masayuki2009:
@no1wudi 
I noticed that nsh does not show input characters when tested with qemu (e.g. rv-virt:smp64)
If I revert this PR, it works again.

no1wudi:
> 

@masayuki2009 Sorry, This PR should works whti https://github.com/apache/nuttx/pull/8691, but it merged late.

masayuki2009:
>@masayuki2009 Sorry, This PR should works whti https://github.com/apache/nuttx/pull/8691, but it merged late.

Thanks for the information. 
I confirmed that the issue has been resolved with the above PR.

PR #1617, opened 2023-03-03
interpreter/toywasm: Update toywasm version

yamt:
## Summary
interpreter/toywasm: Update toywasm version

## Testing
Tested on esp32-devkitc.


xiaoxiang781216:
let's squash the second patch into the first one.

yamt:
> let's squash the second patch into the first one.

i'm not inclined to do so because the file in the first commit is a copy of machine-generated file and the second commit is a manual edit to appease nxstyle.

(while IMO there is no point to run nxstyle on these files, i don't have a better solution right now.)

xiaoxiang781216:
If so, it's better to update the script to generate the code in the well format.

yamt:
> If so, it's better to update the script to generate the code in the well format.

in future, maybe.
it's sad nxstyle doesn't have an auto formatting tool.

pkarashchenko (approved):

xiaoxiang781216:
> > If so, it's better to update the script to generate the code in the well format.
> 
> in future, maybe. it's sad nxstyle doesn't have an auto formatting tool.

So, you have to sqaush the patch manually now.

yamt:
> > > > If so, it's better to update the script to generate the code in the well format.
> > > 
> > > 
> > > in future, maybe. it's sad nxstyle doesn't have an auto formatting tool.
> > 
> > 
> > So, you have to sqaush the patch manually now.
> 
> i still prefer not to squash. but i will.

done

xiaoxiang781216 (approved):

PR #1588, opened 2023-02-22
lvgl: silent download and unpack

Cynerd:
This introduces changes from NuttX commit
03b164f59ce40a3f5677b0588af2aee8d9697bf6.

PR #1630, opened 2023-03-06
treewide: silent download, unpack and others

Cynerd:
This introduces changes from NuttX commit
03b164f59ce40a3f5677b0588af2aee8d9697bf6.

The only left out is 'Register:' as that provides pretty handly list of applications in the firmware.

The download is unified to handle it always the same way.

Follow up on https://github.com/apache/nuttx-apps/pull/1588#event-8649968171.


Make.defs
```diff
@@ -100,6 +100,14 @@ define REGISTER
 endef
 endif
 
+# Download
+
+define DOWNLOAD
```
xiaoxiang781216, line 105: should we move to nuttx/tools/Config.mk? so all place incuding nuttx can use this macro.

Cynerd: That is up to you, but I suspected that it might not be desirable to download in NuttX itself. At the same time, providing it to anyone through NuttX might not do any harm.

For me, the inclusion in applications is enough. I am always using NuttX with applications and adding my software as external applications. Thus I can use this in my external applications.

xiaoxiang781216: > That is up to you, but I suspected that it might not be desirable to download in NuttX itself. At the same time, providing it to anyone through NuttX might not do any harm.
> 
> For me, the inclusion in applications is enough. I am always using NuttX with applications and adding my software as external applications. Thus I can use this in my external applications.

 nuttx repo has many external project downloaded by curl too, we can't reuse the same macro in nuttx if DOWNLOAD is defined in Config.mk.

Cynerd: I haven't check. If so then let's move it to NuttX then.

Cynerd: The updated version now uses `DOWNLOAD` from Nuttix, submitted over here: https://github.com/apache/nuttx/pull/8759.

I had to tweak it to allow different output, and I removed the prefix to the reported output. I think that the correct approach is to use the output (the third argument of the `DOWNLOAD` variable). The nothing saying downloads (those have versions as file names) should be replaced by a custom output name. I just skip doing that as part of this PR and we can decide later if we want that.

xiaoxiang781216 (approved):

Cynerd:
Please, wait with merge. There seems to be issues with downloading wamr. I have to look into it.

acassis:
@Cynerd please let me know: does this PR hides the Downloading and Patching lines? I think it is important to users know that the files are downloaded during the building phase.

acassis:
Hi @Cynerd maybe a better approach is just show: Downloading: filename-x.y.z  and  when it finishes download: done.

What do you think?

xiaoxiang781216:
@Cynerd could you fix the conflict?

Cynerd:
I will look at this next week. This is pretty old and was created only to keep it consistent in my eyes with the rest of the changes introduced at that time. I think that it still makes sense in terms of echoing that the download is happening but not in the way of reporting its progress.

In the meantime, we also modified the build process of our own projects in a way that nothing is being downloaded, and thus, I had less incentive to deal with this.

acassis:
> I will look at this next week. This is pretty old and was created only to keep it consistent in my eyes with the rest of the changes introduced at that time. I think that it still makes sense in terms of echoing that the download is happening but not in the way of reporting its progress.
> 
> In the meantime, we also modified the build process of our own projects in a way that nothing is being downloaded, and thus, I had less incentive to deal with this.

Agree! All user friendly improvements are welcome!

cederom:
ping :-)

Cynerd:
I am kind of busy with other work, and I am failing to get some free time to finish this maintenance work. If anyone has free time on its hands, feel free to finish it. I just can't promise to look at it in a few weeks' time (I have been postponing it already for half a year, and the target now diverged so much that it must be started almost from scratch).

cederom:
ping :-)

PR #1632, opened 2023-03-06
Revert "Don't run preconfig, clean or distclean when it's not necessary."

xiaoxiang781216:
## Summary

to ensure all clean/distclean get run.
This reverts commit 1dae12ba05d30933253afbafccb67cb6ca8b4544.

## Impact

clean/distclean

## Testing

CI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@
 *~
 .context
 .depend
-.kconfig
 /bin
 /boot_romfsimg.h
 /external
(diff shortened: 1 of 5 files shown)


xiaoxiang781216:
@Ouss4 could you take a look? We found that it's hard to trigger clean/distclean with your patch.

pkarashchenko:
@xiaoxiang781216 I see that both your changes are closed. So does it mean that we do not have the issue, or that you are preparing another solution?

xiaoxiang781216:
I will hack the individual Makefile instead.

PR #1638, opened 2023-03-08
wasm: increase WASM_INITIAL_MEMORY if stack size is great than initial memory

anchao:
## Summary

wasm: increase INITMEMORY if stack size is great than initial memory

1. increase INITMEMORY if stack size is great than initial memory
2. Do not pass initial-memory if caller does not specific

Signed-off-by: chao an <anchao@xiaomi.com>

## Impact

N/A

## Testing

sim/wamr

### DIFF
--- a/interpreters/Wasm.mk
+++ b/interpreters/Wasm.mk
@@ -21,9 +21,14 @@
 ifeq ($(WASM_BUILD),y)
   ifneq ($(CONFIG_INTERPRETERS_WAMR)$(CONFIG_INTERPRETERS_WAMR3),)
 
-WASM_INITIAL_MEMORY ?= 65536
-STACKSIZE           ?= $(CONFIG_DEFAULT_TASK_STACKSIZE)
-PRIORITY            ?= SCHED_PRIORITY_DEFAULT
+STACKSIZE  ?= $(CONFIG_DEFAULT_TASK_STACKSIZE)
+PRIORITY   ?= SCHED_PRIORITY_DEFAULT
+
+ifneq ($(INITMEMORY),)
+  ifeq ($(shell test $(INITMEMORY) -lt $(STACKSIZE); echo $$?),0)
+    INITMEMORY := $(shell expr $(STACKSIZE) + $(INITMEMORY))
+  endif
+endif
 
 # Wamr mode:
 # INT: Interpreter (Default)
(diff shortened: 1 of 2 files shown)


interpreters/Wasm.mk
```diff
@@ -25,6 +25,10 @@ WASM_INITIAL_MEMORY ?= 65536
 STACKSIZE           ?= $(CONFIG_DEFAULT_TASK_STACKSIZE)
 PRIORITY            ?= SCHED_PRIORITY_DEFAULT
 
+ifeq ($(shell test $(WASM_INITIAL_MEMORY) -lt $(STACKSIZE); echo $$?),0)
+  WASM_INITIAL_MEMORY := $(shell expr $(STACKSIZE) + $(WASM_INITIAL_MEMORY))
+endif
```
yamt, line 30: i guess it's simpler to stop specifying init memory. (at least for the default)

anchao: Done, I changed INITMEMORY to a selectable configuration

interpreters/Wasm.mk
```diff
@@ -21,9 +21,14 @@
...
 
-WASM_INITIAL_MEMORY ?= 65536
-STACKSIZE           ?= $(CONFIG_DEFAULT_TASK_STACKSIZE)
-PRIORITY            ?= SCHED_PRIORITY_DEFAULT
+STACKSIZE  ?= $(CONFIG_DEFAULT_TASK_STACKSIZE)
+PRIORITY   ?= SCHED_PRIORITY_DEFAULT
+
+ifneq ($(INITMEMORY),)
+  ifeq ($(shell test $(INITMEMORY) -lt $(STACKSIZE); echo $$?),0)
+    INITMEMORY := $(shell expr $(STACKSIZE) + $(INITMEMORY))
```
yamt, line 29: do you still need this?
if a user specifies INITMEMORY smaller than STACKSIZE, it's likely a mistake. i feel it's more user-friendly to just leave the linker complain.

anchao: This is a configurable value. The stack size requested by the developer may vary under different configurations. I think it is better  to solve this issue in the general makefile, otherwise this logic will appear in some application

yamt: sorry, i don't understand your logic.
if a developer specified INITMEMORY < STACKSIZE, i'd assume it was a mistake. do you have counter examples?

anchao: Some stack size configurations may be written like this:

```
config EXAMPLES_HELLO_STACKSIZE
   int "Hello stack size"
   default 16384 if ARCH_SIM
   default DEFAULT_TASK_STACKSIZE
```

It is common to have different stack size on different platforms

yamt: > Some stack size configurations may be written like this:
> 
> ```
> config EXAMPLES_HELLO_STACKSIZE
>    int "Hello stack size"
>    default 16384 if ARCH_SIM
>    default DEFAULT_TASK_STACKSIZE
> ```
> 
> It is common to have different stack size on different platforms

yes.
it seems wrong to use platform DEFAULT_TASK_STACKSIZE for wasm in the first place.

yamt: also, i suspect there is no reasonable way to fix the situation (INITMEMORY < STACKSIZE) automatically.
your solution (INITMEMORY = INITMEMORY + STACKSIZE) seems wrong because INITMEMORY means to include STACKSIZE.
only the user who specified INITMEMORY can know how to fix it.

anchao:
The new wasm build framework no longer needs this mechanism, let us close this PR, thanks a lot for review @yamt

PR #1640, opened 2023-03-08
nshlib: one-time output uname

Gary-Hobson:
## Summary
 one-time output uname，fix "uname -a > /dev/null" output format error

## Testing
sim/nsh

### DIFF
--- a/nshlib/nsh_syscmds.c
+++ b/nshlib/nsh_syscmds.c
@@ -25,6 +25,7 @@
 #include <nuttx/config.h>
 
 #include <nuttx/rptun/rptun.h>
+#include <nuttx/streams.h>
 #include <sys/boardctl.h>
 #include <sys/ioctl.h>
 #include <sys/utsname.h>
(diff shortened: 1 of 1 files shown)


pkarashchenko (approved):

nshlib/nsh_syscmds.c
```diff
@@ -579,6 +581,8 @@ int cmd_uname(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
   /* Process each option */
 
   first = true;
+  lib_memoutstream(&stream, alloca(sizeof(struct utsname)),
```
xiaoxiang781216, line 584: is it enough to allocate struct utsname in the extreme case? since we add more ' ' and '\n'.

Gary-Hobson: It is adequate, an extra 7 characters are added when using “uname -a”
But utsname->sysname is 21 (SYS_NAMELEN) bytes, only 6 bytes ("NuttX") are actually used

PR #1648, opened 2023-03-08
Using poll instead of epoll in NNG library

jturnsek:
## Summary

Switching to poll instead of epoll mechanism in NNG library, because PUB/SUB example doesn't work when epoll is used. This was not an issue a few months ago, thus regression testing on epoll and local socket implementation need to be done.
DIsabled mbedtls functionality in Makefile can be enabled, but then MBEDTLS_X509_CRL_PARSE_C macro must be defined in mbedtls_config.h also.

## Impact

NNG library only.

## Testing

Pub/Sub example

### DIFF
--- a/netutils/nng/Makefile
+++ b/netutils/nng/Makefile
@@ -36,12 +36,19 @@ CFLAGS += -DNNG_TRANSPORT_WS
 CFLAGS += -DNNG_TRANSPORT_WSS
 
 CFLAGS += -DNNG_USE_EVENTFD
+ifeq ($(CONFIG_NETUTILS_NNG_HAVE_EPOLL),y)
 CFLAGS += -DNNG_HAVE_EPOLL
 CFLAGS += -DNNG_HAVE_EPOLL_CREATE1
+endif
+
+ifeq ($(CONFIG_NETUTILS_NNG_USE_MBEDTLS),y)
+CFLAGS += -DNNG_TLS_ENGINE_INIT=nng_tls_engine_init_mbed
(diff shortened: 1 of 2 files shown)


netutils/nng/Makefile
```diff
@@ -88,7 +91,8 @@ CSRCS += $(NNG_SRCDIR)/platform/posix/posix_tcpdial.c
 CSRCS += $(NNG_SRCDIR)/platform/posix/posix_tcplisten.c
 CSRCS += $(NNG_SRCDIR)/platform/posix/posix_thread.c
 CSRCS += $(NNG_SRCDIR)/platform/posix/posix_udp.c
-CSRCS += $(NNG_SRCDIR)/platform/posix/posix_pollq_epoll.c
+#CSRCS += $(NNG_SRCDIR)/platform/posix/posix_pollq_epoll.c
```
xiaoxiang781216, line 94: let's wait @CV-Bowen 's analysis?

jturnsek: I agree that we need to analyse the epoll functionality for possible bugs, but at the same time NNG should work. As NNG supports different mechanisms for event notifications, I suggest we move to poll in order to make example work. AFAIK poll is more efficient in case we have small number of file descriptors. And this is the case here.

jturnsek: I was thinking that too. There is quite a few things we can make configurable.

jerpelea: please do not comment the lines! remove them if they are not needed

xiaoxiang781216 (approved):

PR #1649, opened 2023-03-08
Add another batch of missing headers throughout the repository 

gustavonihei:
## Summary
This PR is a complement to https://github.com/apache/nuttx-apps/pull/1629 and https://github.com/apache/nuttx-apps/pull/1633, all related to the work started on https://github.com/apache/nuttx/pull/8744.

## Impact
No functional changes, should have no impact.

## Testing
CI build pass.

### DIFF
--- a/industry/foc/float/foc_handler.c
+++ b/industry/foc/float/foc_handler.c
@@ -27,6 +27,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <string.h>
 #include <unistd.h>
 
 #include "industry/foc/foc_log.h"
--- a/interpreters/luamodules/cjson/0002-bugfix-Fix-implicit-declaration-warning-for-strncase.patch
+++ b/interpreters/luamodules/cjson/0002-bugfix-Fix-implicit-declaration-warning-for-strncase.patch
@@ -1,0 +1,26 @@
+From e6e8fbe68a95bf6fca463781dbc22e631c4b7b7a Mon Sep 17 00:00:00 2001
+From: Gustavo Henrique Nihei <gustavo.nihei@espressif.com>
+Date: Wed, 8 Mar 2023 17:18:02 -0300
+Subject: [PATCH] bugfix: Fix implicit declaration warning for strncasecmp
+ function
+
+Signed-off-by: Gustavo Henrique Nihei <gustavo.nihei@espressif.com>
+---
+ lua_cjson.c | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/lua_cjson.c b/lua_cjson.c
+index 42672de..f443626 100644
+--- a/lua_cjson.c
++++ b/lua_cjson.c
+@@ -39,6 +39,7 @@
+ #include <assert.h>
+ #include <stdint.h>
+ #include <string.h>
++#include <strings.h>
+ #include <math.h>
+ #include <limits.h>
+ #include <lua.h>
+-- 
+2.37.2
+
--- a/system/nxlooper/nxlooper.c
+++ b/system/nxlooper/nxlooper.c
@@ -24,21 +24,21 @@
 
 #include <nuttx/config.h>
 
-#include <sys/param.h>
-#include <sys/types.h>
-#include <sys/ioctl.h>
-
-#include <stdint.h>
+#include <assert.h>
+#include <debug.h>
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <sched.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <strings.h>
-#include <stdlib.h>
-#include <fcntl.h>
-#include <assert.h>
-#include <errno.h>
-#include <dirent.h>
-#include <debug.h>
+#include <sys/ioctl.h>
+#include <sys/param.h>
+#include <sys/types.h>
 #include <unistd.h>
 
 #include <nuttx/audio/audio.h>
--- a/system/nxplayer/nxplayer.c
+++ b/system/nxplayer/nxplayer.c
@@ -24,28 +24,27 @@
 
 #include <nuttx/config.h>
 
-#include <sys/types.h>
+#include <assert.h>
+#include <debug.h>
+#include <dirent.h>
+#include <errno.h>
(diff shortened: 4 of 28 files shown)


testing/ostest/suspend.c
```diff
@@ -22,14 +22,18 @@
...
+#include <semaphore.h>
+#include <signal.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <unistd.h>
-#include <semaphore.h>
-#include <signal.h>
-#include <sched.h>
-#include <errno.h>
+#include <sys/types.h>
```
pkarashchenko, line 35: Do we really need this include? I think sys/types.h is included in most of the standard headers above. What specific type used here needs this explicit include?
And same question in other similar places

gustavonihei: `sys/types.h` is most probably required due to `size_t`, but I agree that there might be some cases where it is needlessly being included.

In this current endeavor I am just trying to make #8744 pass the CI by including the missing headers. Later we could use a tool-assisted method (e.g. [IWYU](https://include-what-you-use.org/) to remove unused headers.

interpreters/luamodules/cjson/0002-bugfix-Fix-implicit-declaration-warning-for-strncase.patch
```diff
@@ -0,0 +1,26 @@
...
+
+diff --git a/lua_cjson.c b/lua_cjson.c
+index 42672de..f443626 100644
+--- a/lua_cjson.c
++++ b/lua_cjson.c
+@@ -39,6 +39,7 @@
+ #include <assert.h>
+ #include <stdint.h>
+ #include <string.h>
++#include <strings.h>
```
pkarashchenko, line 20:
```suggestion
-#include <string.h>
+#include <strings.h>
```

system/nxlooper/nxlooper.c
```diff
@@ -24,21 +24,21 @@
...
+#include <debug.h>
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <sched.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
```
pkarashchenko, line 37: I think that strings.h includes string.h, so there is no need to include both

system/nxplayer/nxplayer.c
```diff
@@ -24,30 +24,29 @@
 
 #include <nuttx/config.h>
 
+#include <assert.h>
+#include <debug.h>
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <netutils/netlib.h>
```
pkarashchenko, line 32: Maybe this can be moved to the bottom and grouped out of standard includes?

system/nxplayer/nxplayer_mp3.c
```diff
@@ -22,12 +22,14 @@
  * Included Files
  ****************************************************************************/
 
-#include <sys/types.h>
+#include <nuttx/config.h>
```
pkarashchenko, line 25: Not sure about this change as sys/types.h includes nuttx/config.h as the first line

system/nxrecorder/nxrecorder.c
```diff
@@ -24,23 +24,24 @@
...
 #include <string.h>
 #include <strings.h>
-#include <stdlib.h>
-#include <fcntl.h>
-#include <assert.h>
-#include <errno.h>
-#include <dirent.h>
-#include <debug.h>
+#include <sys/types.h>
+#include <sys/ioctl.h>
```
pkarashchenko, line 40: Maybe it can be put after unistd.h?
What was the reason for moving it here?

system/nxplayer/nxplayer.c
gustavonihei: Oh, that's true, this one is from /apps and should be separate, thanks for noticing!

system/nxplayer/nxplayer_mp3.c
gustavonihei: `sys/types.h` is a C library header, so it should be grouped with the others.
`nuttx/config.h` is just for standard practice in NuttX to be always included first.

system/nxrecorder/nxrecorder.c
gustavonihei: Just sorting it alphabetically.

system/nxlooper/nxlooper.c
gustavonihei: Even though NuttX coding standard imposes no restrictions or rules on this, I usually like to follow Google's C Style guide, in this case specifically the [Include What You Use guideline](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use).
It is better to be pedantic and include everything that is being referenced than to rely on transitive inclusions.

system/nxplayer/nxplayer.c
gustavonihei: Fixed!

acassis:
@gustavonihei please check this conflict

gustavonihei:
> @gustavonihei please check this conflict

Thanks for notifying me, I've fixed the conflict.

industry/foc/float/foc_handler.c
```diff
@@ -27,6 +27,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <string.h>
```
pkarashchenko, line 30: I do not fully understand such changes. Is the compilation error happening without this include?

system/nxplayer/nxplayer_mp3.c
pkarashchenko: Ok.

system/nxlooper/nxlooper.c
pkarashchenko: Ok.

industry/foc/float/foc_handler.c
gustavonihei: Yes. I don't recall the exact warning , but here I believe the compiler was reporting the missing declaration for the `memset` function.

It all started after I removed `#include <nuttx/userspace.h>` from `kmalloc.h` when trying to build WARM for ESP32-S3, as described in https://github.com/apache/nuttx/pull/8744.

xiaoxiang781216:
@acassis from the policy, you can't merge the common code from the same company.

PR #1650, opened 2023-03-09
system/cle: Remove cle_getcursor()

no1wudi:
## Summary
Fix https://github.com/apache/nuttx/issues/8731, don't rely on the vt100 to report
the cursor position.

This avoid the implicit input (response of get cursor command) during the user inputs.

Notice: This assume that the nsh prompt is always shown at line start.
## Impact
cle
## Testing
esp32c3-devkit:usbconsole with cle and lua

### DIFF
--- a/system/cle/cle.c
+++ b/system/cle/cle.c
@@ -108,9 +108,8 @@ enum cle_key_e
 
 struct cle_s
 {
-  uint16_t curpos;          /* Current cursor position */
-  uint16_t cursave;         /* Saved cursor position */
-  uint16_t row;             /* This is the row that we are editing in */
+  int16_t  curpos;          /* Current cursor position in buffer */
+  int16_t  realpos;         /* Current cursor position in terminal */
   uint16_t coloffs;         /* Left cursor offset */
   uint16_t linelen;         /* Size of the line buffer */
   uint16_t nchars;          /* Size of data in the line buffer */
@@ -136,9 +135,7 @@ static void     cle_putch(FAR struct cle_s *priv, char ch);
 static int      cle_getch(FAR struct cle_s *priv);
 static void     cle_cursoron(FAR struct cle_s *priv);
 static void     cle_cursoroff(FAR struct cle_s *priv);
-static void     cle_setcursor(FAR struct cle_s *priv, uint16_t column);
-static int      cle_getcursor(FAR struct cle_s *priv, uint16_t *prow,
-                  uint16_t *pcolumn);
+static void     cle_setcursor(FAR struct cle_s *priv, int16_t column);
 static void     cle_clrtoeol(FAR struct cle_s *priv);
 
 /* Editor function */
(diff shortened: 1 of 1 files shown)


system/cle/cle.c
```diff
@@ -82,6 +84,14 @@
 #  define COLOR_OUTPUT  VT100_GREEN
 #endif
 
+/* Poll timeout during getcursor,
+ * 50ms should be enough to fetch a single byte from powerful PC (or telent)
+ */
+
+#define CLE_POLL_TIMEOUT 50
+
+#define CLE_GETCURSOR_RETRY_COUNT 5
```
pkarashchenko, line 93: Maybe 100ms + 10 tries so it will be up to 1s in total?
I'm not sure what the real use case numbers are.

no1wudi: Maybe more, a response have about three bytes, will poll fd before get each byte.

So the 1s in your examples should be atleast 1s instead of up to 1s.

system/cle/cle.c
```diff
@@ -1124,11 +1149,20 @@ int cle_fd(FAR char *line, FAR const char *prompt, uint16_t linelen,
 
   /* Get the current cursor position */
 
-  ret = cle_getcursor(&priv, &priv.row, &column);
-
-  if (ret < 0)
+  for (; ; )
     {
-      return ret;
+      ret = cle_getcursor(&priv, &priv.row, &column);
```
xiaoxiang781216, line 1154: but why the error doesn't happen before ECHO change?

no1wudi: In fact this issue is not due to ECHO but https://github.com/apache/nuttx-apps/commit/95f32fd018be7a66354942647d9e1c778ad5d488 , 
seems c3's usb-serial driver need more time to get ready. detail please refer to https://github.com/apache/nuttx/issues/8731#issuecomment-1461288702

xiaoxiang781216: should we fix the usb serial driver instead

no1wudi: I think usb driver should be fixed, but this patch is also needed since nsh shold not hang here if can't read vt100 command respons.

xiaoxiang781216: but the response is enter by user, you can't expect how fast the user will generate the input.

no1wudi: No, the response here is sended by terminal auto.

xiaoxiang781216: but user may not open terminal yet

system/cle/cle.c
```diff
@@ -329,6 +338,44 @@ static int cle_getch(FAR struct cle_s *priv)
...
+  fd.events = POLLIN;
+
+  ret = poll(&fd, 1, CLE_POLL_TIMEOUT);
+
+  if (ret < 0)
+    {
+      return ret;
+    }
+
+  if (fd.revents & POLLIN)
```
pkarashchenko, line 366: Optional
```suggestion
  if ((fd.revents & POLLIN) != 0)
```

system/cle/cle.c
```diff
@@ -469,16 +517,18 @@ static int cle_getcursor(FAR struct cle_s *priv, FAR uint16_t *prow,
     {
       /* Look for initial ESC */
 
-      ch = cle_getch(priv);
-      if (ch != ASCII_ESC)
+      ret = cle_pollch(priv, &ch);
+
```
pkarashchenko, line 521: why do we need extra newlines?

no1wudi:
@pkarashchenko @xiaoxiang781216 I updated this PR to another way, the new approach is to maintain the cursor position in
cle internal instead of query from terminal, this avoid the query operation before each line read operation.

Then the `cle`'s behavior is much more like `readline`, and also avoid some issue if the connection between nuttx and PC is not so stable.

system/cle/cle.c
```diff
@@ -367,21 +363,49 @@ static void cle_cursoroff(FAR struct cle_s *priv)
...
+  if (off + priv->realpos < priv->coloffs)
+    {
+      off = priv->realpos - priv->coloffs;
+    }
+
+  /* Format the cursor position command.
+   * Move left or right depends on the current cursor position in buffer.
+   */
+
+  len = snprintf(buffer, 16,
```
pkarashchenko, line 398:
```suggestion
  len = snprintf(buffer, sizeof(buffer),
```

pkarashchenko (approved):

PR #1654, opened 2023-03-10
nsh/alias: Add shell alias support for NSH

pussuw:
## Summary
Add support for shell aliases into NSH. As a pre-requisite add support for single ('') and double ("") quotes as well. Note, this changes the parser behavior quite significantly so if ANY issues please let me know ASAP and I'll fix them at once.
## Impact
Quite significant on the parser (the code is simplified somewhat IMO), adds new POSIX feature to the shell.
## Testing
icicle:nsh, tried several happy cases and error cases. Everything seemingly works.

### DIFF
--- a/nshlib/nsh.h
+++ b/nshlib/nsh.h
@@ -715,6 +715,25 @@ struct nsh_parser_s
 #endif
 };
 
+#ifdef CONFIG_NSH_ALIAS
+struct nsh_alias_s
+{
+  FAR struct nsh_alias_s *next;    /* Single link list for traversing */
+  FAR char               *name;    /* Name of the alias */
+  FAR char               *value;   /* Value behind the name */
+  union
+  {
+    struct
+    {
+      uint8_t             exp : 1; /* Already expanded ? */
+      uint8_t             rem : 1; /* Marked for deletion */
+    };
+
+    uint8_t               flags;   /* Raw value */
+  };
+};
+#endif
+
 /* This is the general form of a command handler */
 
 struct nsh_vtbl_s; /* Defined in nsh_console.h */
@@ -791,6 +810,13 @@ int nsh_romfsetc(void);
 int nsh_usbconsole(void);
 #else
 #  define nsh_usbconsole() (-ENOSYS)
+#endif
+
+#ifdef CONFIG_NSH_ALIAS
+FAR struct nsh_alias_s *nsh_aliasfind(FAR struct nsh_vtbl_s *vtbl,
+                                      FAR const char *token);
+void nsh_aliasfree(FAR struct nsh_vtbl_s *vtbl,
+                   FAR struct nsh_alias_s *alias);
 #endif
 
 #ifndef CONFIG_NSH_DISABLESCRIPT
@@ -1191,6 +1217,11 @@ int cmd_pmconfig(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv);
 #  ifndef CONFIG_NSH_DISABLE_URLENCODE
   int cmd_urldecode(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv);
 #  endif
+#endif
+
+#ifdef CONFIG_NSH_ALIAS
+int cmd_alias(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv);
+int cmd_unalias(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv);
 #endif
 
 /****************************************************************************
--- a/nshlib/nsh_alias.c
+++ b/nshlib/nsh_alias.c
@@ -1,0 +1,440 @@
+/****************************************************************************
+ * apps/nshlib/nsh_alias.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+
+#include <nuttx/queue.h>
+
+#include "nsh.h"
+#include "nsh_console.h"
+
+#ifdef CONFIG_NSH_ALIAS
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Macro to get head of alias list */
+
+#define alias_head(list)        (FAR struct nsh_alias_s *)sq_peek(list)
+#define alias_remfirst(list)    (FAR struct nsh_alias_s *)sq_remfirst(list)
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/* Alias message format */
+
+static const char g_aliasfmt[]    = "alias %s='%s'\n";
+static const char g_savefailfmt[] = "alias %s='%s' failed\n";
+
+/* Common for both alias / unalias */
+
+static const char g_noaliasfmt[]  = "%s: %s not found\n";
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: alias_init
+ ****************************************************************************/
+
+void alias_init(FAR struct nsh_vtbl_s *vtbl)
+{
+  int i;
+
+  if (!sq_empty(&vtbl->alist) || !sq_empty(&vtbl->afreelist))
+    {
+      /* If either list is non-empty, we are initialized already */
+
+      return;
+    }
+
+  sq_init(&vtbl->alist);
+  sq_init(&vtbl->afreelist);
+
+  for (i = 0; i < CONFIG_NSH_ALIAS_MAX_AMOUNT; i++)
+    {
+      sq_addlast((FAR struct sq_entry_s *)&vtbl->atab[i], &vtbl->afreelist);
+    }
+
+  return;
+}
+
+/****************************************************************************
+ * Name: alias_find
+ ****************************************************************************/
+
+static FAR struct nsh_alias_s *alias_find(FAR struct nsh_vtbl_s *vtbl,
+                                          FAR const char *name)
+{
+  FAR struct nsh_alias_s *alias;
+
+  for (alias = alias_head(&vtbl->alist); alias; alias = alias->next)
+    {
+      if (strcmp(alias->name, name) == 0)
+        {
+          return alias;
+        }
+    }
+
+  return NULL;
+}
+
+/****************************************************************************
+ * Name: alias_delete
+ ****************************************************************************/
+
+static void alias_delete(FAR struct nsh_vtbl_s *vtbl,
+                         FAR struct nsh_alias_s *alias)
+{
+  if (alias)
+    {
+      if (alias->exp)
+        {
+          /* Mark it for removal, but keep the data intact */
+
+          alias->rem = 1;
+          return;
+        }
+
+      if (alias->name)
+        {
+          free(alias->name);
+        }
+
+      if (alias->value)
+        {
+          free(alias->value);
+        }
+
+      alias->name = NULL;
+      alias->value = NULL;
+      alias->flags = 0;
+
+      sq_rem((FAR sq_entry_t *)alias, &vtbl->alist);
+      sq_addfirst((FAR sq_entry_t *)alias, &vtbl->afreelist);
+    }
+}
+
+/****************************************************************************
+ * Name: alias_save
+ ****************************************************************************/
+
+static int alias_save(FAR struct nsh_vtbl_s *vtbl, FAR const char *name,
+                      FAR const char *value)
+{
+  FAR struct nsh_alias_s *alias;
+  int ret = OK;
+
+  if (!name || *name == '\0' || !value)
+    {
+      return -EINVAL;
+    }
+
+  if ((alias = alias_find(vtbl, name)) != NULL)
+    {
+      /* Update the value */
+
+      free(alias->value);
+      alias->value = strdup(value);
+    }
+  else if ((alias = alias_remfirst(&vtbl->afreelist)) != NULL)
+    {
+      /* Create new value */
+
+      alias->name = strdup(name);
+      alias->value = strdup(value);
+      sq_addlast((FAR sq_entry_t *)alias, &vtbl->alist);
+    }
+
+  if (!alias || !alias->name || !alias->value)
+    {
+      /* Something went wrong, clean up after ourselves */
+
+      alias_delete(vtbl, alias);
+      ret = -ENOMEM;
+    }
+
+  return ret;
+}
+
+/****************************************************************************
+ * Name: alias_printall
+ ****************************************************************************/
+
+static void alias_printall(FAR struct nsh_vtbl_s *vtbl)
+{
+  FAR struct nsh_alias_s *alias;
+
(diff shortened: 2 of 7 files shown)


nshlib/nsh_parse.c
```diff
@@ -1543,111 +1697,78 @@ static FAR char *nsh_argument(FAR struct nsh_vtbl_s *vtbl,
...
-               * Now we can break out of the loop.
-               */
+                  char qterm[2];
 
-              break;
-            }
-        }
-#else
-      backquote = false;
+                  qterm[0] = ptr;
```
pussuw, line 1749: Copy&paste error when fixing code style issue. Will fix this on monday (should be qterm[0] = *pend;)

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,454 @@
...
+{
+  FAR struct console_stdio_s *cn = (FAR struct console_stdio_s *)vtbl;
+  int i;
+
+  sq_init(&cn->alist);
+  sq_init(&cn->afreelist);
+
+  for (i = 0; i < CONFIG_NSH_ALIAS_MAX_AMOUNT; i++)
+    {
+      sq_addlast((FAR struct sq_entry_s *) &cn->atab[i], &cn->afreelist);
```
pkarashchenko, line 102:
```suggestion
      sq_addlast((FAR struct sq_entry_s *)&cn->atab[i], &cn->afreelist);
```

pkarashchenko (approved):

nshlib/nsh_alias.c
pussuw: Done

pussuw:
Seems like argument expansion might not work as expected after this change, I'll investigate

pussuw:
Yep, found it:

```
      if (*pbegin == '"')
        {
          /* A quoted string can only be terminated with another quotation
           * mark.  Set pbegin to point at the character after the opening
           * quote mark.
           */

          pbegin++;
          term = "\"";

          /* If this is an environment variable in double quotes, we don't
           * want it split into multiple arguments. So just invalidate the
           * flag pointer which would otherwise communicate such back up
           * the call tree.
           */

          isenvvar = NULL;
        }
```
The isenvvar = NULL; is significant, as later the parser splits an env string into multiple parameters *if* they are not in quotes. Example:

set FOO "$bar $baz" 

will result in the parser splitting "$bar $baz" into two arguments, even though they are in quotes. Patch coming in a moment.

pussuw:
The var expansion issue should now be fixed.

nshlib/nsh.h
```diff
@@ -715,6 +715,25 @@ struct nsh_parser_s
...
+{
+  FAR struct nsh_alias_s *next;   /* Single link list for traversing */
+  FAR char               *name;   /* Name of the alias */
+  FAR char               *value;  /* Value behind the name */
+  union
+  {
+    struct
+    {
+      uint8_t             exp :1; /* Already expanded ? */
+      uint8_t             rem :1; /* Marked for deletion */
```
pkarashchenko, line 729:
```suggestion
      uint8_t             exp : 1; /* Already expanded ? */
      uint8_t             rem : 1; /* Marked for deletion */
```

pkarashchenko (approved):

nshlib/nsh.h
pussuw: Done

nshlib/nsh.h
```diff
@@ -715,6 +715,25 @@ struct nsh_parser_s
...
+  FAR char               *value;   /* Value behind the name */
+  union
+  {
+    struct
+    {
+      uint8_t             exp : 1; /* Already expanded ? */
+      uint8_t             rem : 1; /* Marked for deletion */
+    };
+
+    uint8_t               flags;   /* Raw value */
```
pkarashchenko, line 732: Do we need raw value? Where is it used?

pussuw: I put it there just for completeness, if for whatever reason the flags need to be cleared as a bunch.

pkarashchenko: Yes. For clearing it makes sense. I was just worried about LE vs BE bit positioning, so just want to make sure that fields are not used as `& flags`.

pkarashchenko (commented):
Most of the comments are optional. Just need to clarify `strdup` question

nshlib/nsh.h
pussuw: Yes, like you said accessing the bitfield via any other way besides the named fields is meaningless. To be precise, it is not even about LE/BE, the relative order of the bits can also change between implementations. Basically the standard says nothing about how the storage of bitfields must be implemented, the compiler can decide everything.

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,454 @@
...
+    {
+      return ERROR;
+    }
+
+  if ((alias = alias_find(vtbl, name)) != NULL)
+    {
+      /* Update the value */
+
+      free(alias->value);
+      alias->value = strdup(value);
```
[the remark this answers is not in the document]
pussuw: Yes, everything is checked on line 195, and if any strdup fails the whole alias is deleted.

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,454 @@
...
+/****************************************************************************
+ * Name: alias_save
+ ****************************************************************************/
+
+static int alias_save(FAR struct nsh_vtbl_s *vtbl, FAR const char *name,
+                      FAR const char *value)
+{
+  FAR struct console_stdio_s *cn = (FAR struct console_stdio_s *)vtbl;
+  FAR struct nsh_alias_s *alias;
+  int ret = 0;
```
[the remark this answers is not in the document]
pussuw: Done

nshlib/nsh_parse.c
```diff
@@ -331,6 +372,57 @@ static void nsh_memlist_free(FAR struct nsh_memlist_s *memlist)
...
+  if (vtbl && alist)
+    {
+      FAR struct nsh_alias_s *alias;
+      int index;
+
+      for (index = 0; index < alist->nallocs; index++)
+        {
+          alias = alist->allocs[index];
+          alias->exp = 0;
+          if (alias->rem == 1)
```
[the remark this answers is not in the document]
pussuw: Done, cleared name and value too so they don't propagate as stale pointers

pkarashchenko (approved):

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,458 @@
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/* Alias message format */
+
+const char g_aliasfmt[]    = "alias %s='%s'\n";
+const char g_savefailfmt[] = "alias %s='%s' failed\n";
```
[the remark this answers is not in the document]
pussuw: Sure, no reason to keep these globally visible

pussuw:
Some sort of resource DL issue:
 curl: (28) Failed to connect to git.busybox.net port 443: Connection timed out
make[1]: *** [uClibc++.defs:26: uClibc++] Error 28
make[1]: Target 'context' not remade because of errors.
make: *** [tools/Unix.mk:436: libs/libxx/.context] Error 2
/github/workspace/sources/nuttx/tools/testbuild.sh: line 338: /github/workspace/sources/nuttx/../nuttx/nuttx.manifest: No such file or directory
  Normalize sim/cxxtest

nshlib/nsh_console.h
```diff
@@ -153,6 +155,15 @@ struct console_stdio_s
   size_t varsz;
 #endif
 
+#ifdef CONFIG_NSH_ALIAS
```
xiaoxiang781216, line 158: it's better to move into nsh_vtbl_s, so all possible console can get alias support automatically

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,458 @@
...
+#define ALIAS_INIT  0x01
+
+/* Macro to get head of alias list */
+
+#define alias_head(list)        (FAR struct nsh_alias_s *)sq_peek(list)
+#define alias_remfirst(list)    (FAR struct nsh_alias_s *)sq_remfirst(list)
+
+/* Initialize the alias shell container (if necessary) */
+
+#define do_alias_init(vtbl) \
```
xiaoxiang781216, line 56: let's merge ALIAS_INIT check to alias_init and remove do_alias_init

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,458 @@
...
+  do \
+    { \
+      if (!(((FAR struct console_stdio_s *)(vtbl))->flags & ALIAS_INIT)) \
```
xiaoxiang781216, line 59: we can check the field in both alist and afreelist is null instead, and remove flags field and ALIAS_INIT macro.

nshlib/nsh.h
```diff
@@ -715,6 +715,25 @@ struct nsh_parser_s
...
+struct nsh_alias_s
+{
+  FAR struct nsh_alias_s *next;    /* Single link list for traversing */
+  FAR char               *name;    /* Name of the alias */
+  FAR char               *value;   /* Value behind the name */
+  union
+  {
+    struct
+    {
+      uint8_t             exp : 1; /* Already expanded ? */
```
xiaoxiang781216, line 728: why need exp?

nshlib/nsh_alias.c
pussuw: Sure, I can merge the check into alias_init.

nshlib/nsh_console.h
pussuw: Sure, I did not really know where to put the aliases. I noticed that NSH_VARS are here, so I thought this would be the place.

Can you explain to me the difference between `nsh_vtbl_s` and `console_stdio_s`, it was not totally clear to me?

xiaoxiang781216: you can think nsh_vtbl_s is a base class and console_stdio_s is a derived class(handle the different I/O), more derived classes may add later. So if this is a common feature for all derived class, nsh_vtbl_s is a better place.

nshlib/nsh_parse.c
```diff
@@ -115,6 +129,14 @@ struct nsh_memlist_s
 };
 #endif
 
+#ifdef CONFIG_NSH_ALIAS
+struct nsh_alist_s
+{
+  int nallocs;
+  FAR struct nsh_alias_s *allocs[CONFIG_NSH_ALIAS_MAX_AMOUNT];
```
xiaoxiang781216, line 136: why need another table to keep the expanded variable if nsh_alias_s already has exp field? I think it's more simple to iterate all alias and clear exp directly.

nshlib/nsh.h
xiaoxiang781216: OK, understand.

nshlib/nsh_console.h
pussuw: Thanks for the explanation.

nshlib/nsh_parse.c
pussuw: > It's more simple to iterate all alias and clear exp directly after the end of expansion.
I thought of this and decided to make the list anyway. 

It is there so you don't need to traverse the whole alias list after every command. I for example have >50 aliases in my use case. In the normal case when using the terminal no alias expansion is performed.

Example:
```
$ alias foo1=bar
$ alias foo2=bar
...
$ alias foo50=bar
$ ls
```
Here after executing the ls command, even though no alias expansion has been performed you would have to go through all 50 aliases to see if they have been expanded. Instead now we have a "expanded aliases list" that is empty and nothing needs to be done.

There is a purpose for the expanded alias list, the exp flag and the rem flag.
1. The expanded alias list contains every expanded alias, so they can be cleared and if necessary, removed from the alias list after a command has been executed
2. The exp flag is already explained above, an alias shall be expanded only once
3. Rem field is also needed as the expanded alias string is not copied onto the parsed command line, simply a reference is returned and this means the alias data must be retained until the command has been executed. 

When is the rem field relevant ? Well, some insane corner cases can appear...

```
$ alias ua='unalias -a'
$ ua
```
Here the string behind the alias removes all aliases, but it is not safe to remove ua before the command unalias -a has been executed.

xiaoxiang781216: If so, look like it's better to manage the expansion list directly:
https://github.com/apache/nuttx-apps/pull/1654/files#diff-7ba973a1f25993d049f08f4b5d54ac41f9a0955f4f2642c0ddb069dde7c4c5e4R124-R130
for example, adding an expansion and move the alias from alist to expansion list vice vesa.

pussuw: I thought of this as well, but it has side effects too.

1. The ordering of the aliases changes when pumping from one list to another. Although the significance of this is mostly cosmetic.
2. The corner case above would not work any more, as executing `$ ua ` would temporarily remove ua from the valid alias list (moving it to the expanded list). So all aliases *besides* ua would be removed.

Another example:

```
$ alias foo='ls;unalias foo'
$ foo
```
This would fail to perform the "unalias foo" part, as it has been (temporarily) plucked from the list of known aliases (it cannot be found anymore).

I know these corner cases are insane usages of aliases, but the standard does not prohibit making such aliases.

I did explore other options for the separate NSH_ALIASLIST storage but all of them had some corner case that did not work. This implementation does at least work.

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,441 @@
...
+  for (arg = argv + 1; *arg; arg++)
+    {
+      /* Look for name=value */
+
+      if ((value = strchr(*arg, '=')) != NULL)
+        {
+          /* Save new / modify existing alias */
+
+          *value++ = '\0';
+          if (alias_save(vtbl, *arg, value) < 0)
```
xiaoxiang781216, line 348:
```
ret = alias_save(vtbl, *arg, value);
if (ret < 0)
```
and remove line 351

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,441 @@
...
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: alias_init
+ ****************************************************************************/
+
+int alias_init(FAR struct nsh_vtbl_s *vtbl)
```
xiaoxiang781216, line 75: int->void

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,441 @@
...
+
+static int alias_save(FAR struct nsh_vtbl_s *vtbl, FAR const char *name,
+                      FAR const char *value)
+{
+  FAR struct nsh_alias_s *alias;
+  int ret = OK;
+
+  if (!name || *name == '\0' || !value)
+    {
+      return ERROR;
```
xiaoxiang781216, line 165: return -EINVAL

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,441 @@
...
+          /* Found it */
+
+          nsh_output(vtbl, g_aliasfmt, alias->name, alias->value);
+        }
+      else
+        {
+          /* Nothing found */
+
+          nsh_error(vtbl, g_noaliasfmt, "alias", *arg);
+          ret = ERROR;
```
xiaoxiang781216, line 365: ret = -ENOENT;

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,441 @@
...
+
+  /* Init, if necessary */
+
+  alias_init(vtbl);
+
+  if (argc < 2)
+    {
+      /* No arguments given, this is an error */
+
+      return ERROR;
```
xiaoxiang781216, line 407: return -EINVAL;

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,441 @@
...
+          /* Found it */
+
+          alias_delete(vtbl, alias);
+        }
+      else
+        {
+          /* Nothing found */
+
+          nsh_error(vtbl, g_noaliasfmt, "unalias", *arg);
+          ret = ERROR;
```
xiaoxiang781216, line 434: ret = -ENOENT;

nshlib/nsh_alias.c
```diff
@@ -0,0 +1,454 @@
...
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int cmd_alias(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
+{
+  FAR struct nsh_alias_s *alias;
+  FAR char **arg;
+  FAR char *value;
+  int ret;
```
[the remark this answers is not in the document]
xiaoxiang781216: But why not change `cmd_unalias` too, the code is clear if follow @pkarashchenko 's suggestion.

nshlib/nsh_parse.c
xiaoxiang781216: Ok.

nshlib/nsh_alias.c
pussuw: Nice catch, I'll fix this and all the rest of the return values.

nshlib/nsh_alias.c
pussuw: Sure

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #1664, opened 2023-03-15
apps: add <assert.h> header file

zyfeier:
## Summary

Add missing assert.h header file.

## Impact

NA

## Testing

CI Pass


pkarashchenko (approved):

testing/ostest/ostest.h
```diff
@@ -27,6 +27,8 @@
 
 #include <nuttx/config.h>
 
+#include <assert.h>
```
pkarashchenko, line 30: In general I do not support indirect including. What from ostest.h needs assert.h?

gustavonihei: Agreed, `assert.h` should be included directly in the files where it is required.

pkarashchenko (approved):

PR #1665, opened 2023-03-16
ostest: Introduce basic work queue test

wengzhe:
## Summary
An ostest for https://github.com/apache/nuttx/pull/8832

The test consists of two parts:
- A tester that tries to trigger wrong states of work queue
- A verifier that checks whether the wqueue is still working properly

The tester is trying to queue and cancel work several times with priority lower/same/higher than the work queue.
Most wrong cases are likely to happen with high priority like:
- https://github.com/apache/nuttx/pull/8321
  - If `cancel` never decreases semcount, the count may keep growing and finally overflow
- https://github.com/apache/nuttx/pull/8832
  - If `cancel` is decreasing semcount too much, the `work_thread` may be waken up less times than expected

The lower/same priority testers are just added for covering other unexpected situations.

The verifier is trying to queue some works and check they are called as expected:
- Frist queue a 'sleep' worker, to let a work queue thread be in busy status and not waiting on sem, while other work queue thread(s) (if any) still waiting for sem. If sem is in wrong state, it may cause wrong behavior in either thread waiting/not waiting on the sem.
- Then queue a few count works, if the work queue(s) are still working properly, these works should finally be all called once.

## Impact
Add an ostest for work queue.

## Testing
Without https://github.com/apache/nuttx/pull/8832, this would fail, but may success after it is fixed.

### DIFF
--- a/testing/ostest/wqueue.c
+++ b/testing/ostest/wqueue.c
@@ -1,0 +1,190 @@
+/****************************************************************************
+ * apps/testing/ostest/wqueue.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <assert.h>
+#include <pthread.h>
+#include <sched.h>
+#include <semaphore.h>
+#include <stdio.h>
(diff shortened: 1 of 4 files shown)


davids5:
@wengzhe Please add to the commit message and the PR summary what conditions the test tests and how it works to perform the tests.

testing/ostest/wqueue.c
```diff
@@ -0,0 +1,188 @@
...
+}
+
+static void run_once(int interval, int priority_test, int priority_verify)
+{
+  pthread_t thread;
+  pthread_attr_t attr;
+  struct sched_param sparam;
+
+  pthread_attr_init(&attr);
+  memset(&sparam, 0, sizeof(sparam));
```
pkarashchenko, line 143: Shouldn't we get param instead of memset?

pkarashchenko: Ack

wengzhe:
> Oops, I noticed that current CI break is not because this test find problem of wqueue, but use too much stack size of `struct work_s work[VERIFY_COUNT + 1]` and it overflowed. After look deep into the `sabre-6quad:citest`, I need to make some changes:
> 
> 1. Set stack size instead of default one.
> 2. This test needs pthread, so check `CONFIG_DISABLE_PTHREAD`
> 3. Reduce the `TEST_COUNT`, since it's a little bit slow on `sabre-6quad:citest` (several minutes in total), now I try to reduce it into less than 1min.

Now it seems more likely to be correct behavior: ostest failed because of wrongly wqueue behavior (kernel not fixed), instead of timed out.

wengzhe:
It seems that "re-run" of our CI didn't re-run the "Fetch-Source" step, then it just got same result with NuttX kernel on same commit(52525796a). So I pushed an identical commit to fully re-trigger the CI.

davids5 (approved):
@wengzhe - Thank you for updating commit message.

Issue #1667, opened 2023-03-19
NSH Login Now Exposes Password


patacongo:
Using stock configuration nsh:sim under ChromeOS (Debian), the login looks like:

    login: admin
    password: Administrator

That is an error.  The password should not be exposed.  Prior to recent breaking changes to the serial driver, the login correctly looked like:

    login: admin
    password: 

The password was not echoed so that it did not remain in the terminal display for anyone to see.

I suppose that the correct solution with the changes to the serial driver would be to disable ECHO via termios before asking for the password, and re-enabling ECHO after the password has been entering.

If TERMIOS is not enabled, then you are just screwed.  The password will always be echoed on the console.

patacongo:
It seems to me that one good solution would be behave just like Linux:  Do not special-case the console but, instead, require that the user application always configure the TTY as needed for correct operation.  This is definitely more consistent with the way that Unix is supposed to handle TTYs.

This would solve another issue:  Occasional a device other than the serial console is used for a console.  For example, a different serial port that is not the one designated as the console.  If the application configured the TTY for its particular use, then this dependence on a magic, non-standard console device would be eliminated.

xiaoxiang781216:
With https://github.com/apache/nuttx/pull/8843, the termios unrelated to hardware setting is always enabled, so the terminal aware application could always switch to/from raw mode as needed.

xiaoxiang781216:
Please try: https://github.com/apache/nuttx-apps/pull/1668 and https://github.com/apache/nuttx/pull/8843 which should fix the issue reported here.

PR #1668, opened 2023-03-19
apps: Remove the check of CONFIG_SERIAL_TERMIOS

xiaoxiang781216:
## Summary

If the code only change c_oflag, c_iflag and c_lflag, not c_cflag in termios.
Follow up the change from kernel: https://github.com/apache/nuttx/pull/8843

## Impact

Avoid the shell password echo to terminal

## Testing

sim:adb


xiaoxiang781216:
Please ignore the remain nxstyle warning.

patacongo:
Failing because of mixed case identifiers in modbus.

patacongo (approved):
Most of this change simply removes logic previously conditioned on CONFIG_SERIAL_TERMIOS but no longer needed because of NuttX PR #8843.  Most of this is straight forward, but easy to review thoroughly.

One consequence of this is that NuttX Issue #1667 should also be resolved.

patacongo:
This can be merged since the only errors are due to mixed case identifiers and include files in the wrong location in modus.  Other reviewers should not hesitate to review this as I may very well have missed something.

PR #1671, opened 2023-03-21
interpreters/wamr: New option to dump call stack

no1wudi:
## Summary
 New option to dump call stack
## Impact
WAMR only
## Testing
QEMU


cederom:
what would be the implementation? :-)

no1wudi:
> what would be the implementation? :-)

Emm, the implementation is in WAMR side https://github.com/bytecodealliance/wasm-micro-runtime/pull/2042 and already done, just need a option in nuttx side to enable it.

anchao (approved):
LGTM

PR #1672, opened 2023-03-21
netutils/ftpc: Cancel wdog on error state in ftpc_reconnect.

wengzhe:
## Summary
The timer is for connect timeout, but is not cancelled on connect error, then it may be triggered after ftpc quit and cause heap-use-after-free.

## Impact
ftpc won't trigger heap-use-after-free on connect failure now.

## Testing
Manually

### DIFF
--- a/netutils/ftpc/ftpc_connect.c
+++ b/netutils/ftpc/ftpc_connect.c
@@ -276,5 +276,10 @@ int ftpc_reconnect(FAR struct ftpc_session_s *session)
 errout_with_socket:
   ftpc_sockclose(&session->cmd);
 errout:
+  if (WDOG_ISACTIVE(&session->wdog))
+    {
+      wd_cancel(&session->wdog);
+    }
+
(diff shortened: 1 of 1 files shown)


netutils/ftpc/ftpc_connect.c
```diff
@@ -276,5 +276,6 @@ int ftpc_reconnect(FAR struct ftpc_session_s *session)
 errout_with_socket:
   ftpc_sockclose(&session->cmd);
 errout:
+  wd_cancel(&session->wdog);
```
pkarashchenko, line 279: Maybe better to add this in an error case of socket connect since watchdog is already canceled in case of other errors?

pkarashchenko: Ok. L165 can be fixed by reordering of wd_start and socket init. But anyway let's keep it with is_active way

pkarashchenko (approved):

PR #1680, opened 2023-03-23
apps: fix nxstyle violations

simbit18:
## Summary
fix error: Relative file path does not match actual file
Remove LF first line
## Impact
none


acassis:
Thank you!

PR #1693, opened 2023-03-30
system/composite_main.c: Hide fprintf behind macro to handle FILE_STR…

pussuw:
…EAMS=n

## Summary
CONFIG_FILE_STREAMS now defaults to 'n' when DEFAULT_SMALL is enabled. This is a good change, but this source file fails to compile when file streams are disabled.

Fix this by using fprintf / dprintf depending on the config value.
## Impact
Fixes build error when DEFAULT_SMALL used
## Testing
Out-of-tree target with: 
CONFIG_DEFAULT_SMALL=y
...
CONFIG_USBDEV=y
CONFIG_USBDEV_COMPOSITE=y

### DIFF
--- a/system/composite/composite_main.c
+++ b/system/composite/composite_main.c
@@ -398,7 +398,7 @@ int main(int argc, FAR char *argv[])
 
   if (g_composite.cmphandle)
     {
-      fprintf(stderr, "conn_main: ERROR: Already connected\n");
+      dprintf(STDERR_FILENO, "conn_main: ERROR: Already connected\n");
       return 1;
     }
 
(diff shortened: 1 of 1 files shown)


system/composite/composite_main.c
```diff
@@ -55,6 +55,12 @@
 #  endif
 #endif
 
+#ifdef CONFIG_FILE_STREAM
+#  define printerr(fmt, ...) fprintf(stderr, (fmt), ##__VA_ARGS__)
+#else
+#  define printerr(fmt, ...) dprintf(STDERR_FILENO, (fmt), ##__VA_ARGS__)
```
xiaoxiang781216, line 61: why not always use dprintf?

pussuw: Good question, I did not write the composite_main.c code.

The problem for me is I'm running out of flash memory

xiaoxiang781216: I think it' more simple to replace fprintf with dprintf.

pussuw: Ok, I can convert this file because this is a blocker issue for me (the SW does not build any more)

xiaoxiang781216: Sure, let's fix it one by one.

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #1698, opened 2023-04-02
fsutils: Add ROMLoader utility

lucasssvaz:
## Summary

Currently, most applications that need to load files duplicate code to generate and mount an ROMFS file system.
This PR aims to provide applications and users an easy way to include files in a ROMFS/CROMFS image by providing a ROMLoader application that creates an image based on what is included in its ```rom``` folder.
It also provides other applications the ```ROMLOADER_COPY``` variable to automatically add files to the image.

Note that this is the best compromise between intrusiveness and easiness of use I've found. Any suggestions on how to improve it are greatly appreciated. The first commit I submitted was less intrusive but had problems when trying to add files that are downloaded/generated during the compilation process.

## Impact

Add an easy way to include files into an ROM file system to avoid code duplication.

## Testing

Tested using a modified version of the `bastest` and `bas` applications and a WIP port of the game ZORK.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ $(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),install)
 $(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),context)))
 $(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),register)))
 $(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),depend)))
+$(foreach SDIR, $(CONFIGURED_APPS), $(eval $(call SDIR_template,$(SDIR),romloader)))
 $(foreach SDIR, $(CLEANDIRS), $(eval $(call SDIR_template,$(SDIR),clean)))
 $(foreach SDIR, $(CLEANDIRS), $(eval $(call SDIR_template,$(SDIR),distclean)))
 
@@ -76,7 +77,7 @@ ifeq ($(CONFIG_BUILD_KERNEL),y)
 
 install: $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_install)
 
-$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
+$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_romloader) $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
 	$(Q) for app in ${CONFIGURED_APPS}; do \
 		$(MAKE) -C "$${app}" archive ; \
 	done
@@ -98,10 +99,10 @@ else
 
 ifeq ($(CONFIG_BUILD_LOADABLE),)
 ifeq ($(CONFIG_WINDOWS_NATIVE),y)
-$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
+$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_romloader) $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
 	$(Q) for %%G in ($(CONFIGURED_APPS)) do ( $(MAKE) -C %%G archive )
 else
-$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
+$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_romloader) $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
 	$(Q) for app in ${CONFIGURED_APPS}; do \
 		$(MAKE) -C "$${app}" archive ; \
 	done
@@ -112,7 +113,7 @@ endif
 
 else
 
-$(SYMTABSRC): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
+$(SYMTABSRC): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_romloader) $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
 	$(Q) for app in ${CONFIGURED_APPS}; do \
 		$(MAKE) -C "$${app}" archive ; \
 	done
--- a/fsutils/romloader/.gitignore
+++ b/fsutils/romloader/.gitignore
@@ -1,0 +1,3 @@
+/romfs.img
+/rom.c
+/rom
--- a/fsutils/romloader/Kconfig
+++ b/fsutils/romloader/Kconfig
@@ -1,0 +1,71 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config FSUTILS_ROMLOADER
+	bool "Enable ROM image creator and loader (ROMLoader)"
+	default n
+	depends on (FS_ROMFS && BOARDCTL_ROMDISK) || FS_CROMFS
+	depends on !DISABLE_MOUNTPOINT
+	---help---
+		Mount a ROM file system containing the files inside the loader
+		directory (apps/fsutils/romloader/rom). This is a convenience
+		feature that allows the user and/or application files to be built into
+		a ROM file system without code duplication.
+
+if FSUTILS_ROMLOADER
+
+choice
+	prompt "ROM file system type"
+	default FSUTILS_ROMLOADER_ROMFS  if   FS_ROMFS && BOARDCTL_ROMDISK
+	default FSUTILS_ROMLOADER_CROMFS if  !FS_ROMFS && FS_CROMFS
+
+config FSUTILS_ROMLOADER_ROMFS
+	bool "ROMFS"
+	depends on FS_ROMFS && BOARDCTL_ROMDISK
+	---help---
+		Automatically generates and mounts an internal ROMFS file
+		system
+
+config FSUTILS_ROMLOADER_CROMFS
+	bool "CROMFS"
+	depends on FS_CROMFS
+	---help---
+		Automatically generates and mounts an internal CROMFS file
+		system
+
+endchoice # ROM file system type
+
+config FSUTILS_ROMLOADER_MOUNTPOINT
+	string "ROMFS mount point"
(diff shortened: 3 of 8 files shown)


fsutils/romloader/Kconfig
```diff
@@ -0,0 +1,47 @@
...
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config FSUTILS_ROMLOADER
+	bool "Enable ROMFS image creator and loader (ROMLoader)"
+	default n
+	depends on FS_ROMFS && BOARDCTL_ROMDISK
+	depends on !DISABLE_MOUNTPOINT
+	---help---
+		Mount a ROMFS file system containing the files inside the	loader
```
davids5, line 12: formatting

davids5 (changes requested):
This looks nice, can it also support the compressed version? 

There is a formatting issue in the Kconfig.

lucasssvaz:
> This looks nice, can it also support the compressed version?
> 
> There is a formatting issue in the Kconfig.

What exactly do you mean by "support the compressed version" ?

fsutils/romloader/romloader_main.c
```diff
@@ -0,0 +1,103 @@
...
+
+#include <nuttx/drivers/ramdisk.h>
+
+#include "romfs.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define NSECTORS(b)  (((b)+CONFIG_FSUTILS_ROMLOADER_SECTORSIZE-1)/\
```
pkarashchenko, line 41:
```suggestion
#define NSECTORS(b)  (((b) + CONFIG_FSUTILS_ROMLOADER_SECTORSIZE - 1) / \
```

lucasssvaz:
Current CI fail was due to an curl timeout and is not related to this MR:
```curl: (28) SSL connection timeout```

File: fsutils/romloader/romloader_main.c [hunk 3]
```diff
@@ -0,0 +1,103 @@
...
+
+  desc.minor    = CONFIG_FSUTILS_ROMLOADER_DEVMINOR;   /* Minor device number of the ROM disk. */
+  desc.nsectors = NSECTORS(romfs_img_len);             /* The number of sectors in the ROM disk */
+  desc.sectsize = CONFIG_FSUTILS_ROMLOADER_SECTORSIZE; /* The size of one sector in bytes */
+  desc.image    = (FAR uint8_t *)romfs_img;            /* File system image */
+
+  ret = boardctl(BOARDIOC_ROMDISK, (uintptr_t)&desc);
+  if (ret < 0)
+    {
+      fprintf(stderr, "ERROR: romdisk_register failed: %s\n",
```
pkarashchenko, line 74: Maybe better to use `dprintf(STDERR_FILENO,` instead?

fsutils/romloader/romloader_main.c
```diff
@@ -0,0 +1,103 @@
...
+
+  ret = mount(CONFIG_FSUTILS_ROMLOADER_DEVPATH,
+              CONFIG_FSUTILS_ROMLOADER_MOUNTPOINT,
+              "romfs",
+              MS_RDONLY,
+              NULL);
+
+  if (ret < 0)
+    {
+      fprintf(stderr,
```
pkarashchenko, line 93: ditto

davids5:
> > This looks nice, can it also support the compressed version?
> > There is a formatting issue in the Kconfig.
> 
> What exactly do you mean by "support the compressed version" ?

gen**c**romfs  CROMFS is a commpressed, in-memory, read-only file system based on LZF.

lucasssvaz:
> > > This looks nice, can it also support the compressed version?
> > > There is a formatting issue in the Kconfig.
> > 
> > 
> > What exactly do you mean by "support the compressed version" ?
> 
> gen**c**romfs CROMFS is a compressed, in-memory, read-only file system based on LZF.

I'll try, shouldn't be that hard

fsutils/romloader/romloader_main.c
lucasssvaz: Idk, all occurrences in NuttX use ```stderr```. If we use ```STDERR_FILENO```, it would be the only one.

lucasssvaz: From what I saw, ```stderr``` should be used with ```fprintf``` and ```STDERR_FILENO``` with ```dprintf```:

```
#ifdef CONFIG_FILE_STREAM
  fprintf(stderr, "%d: %pV: %s\n", _SCHED_GETTID(), &vaf, strerror(error));
#else
  dprintf(STDERR_FILENO, "%d: %pV: %s\n", _SCHED_GETTID(),
                                          &vaf, strerror(error));
#endif
```

lucasssvaz: I could change it to ```dprintf``` to avoid compilation errors when ```CONFIG_DEFAULT_SMALL``` is enabled.

lucasssvaz:
Added support for CROMFS and fixed console output when ```CONFIG_DEFAULT_SMALL``` is enabled.

davids5 (approved):
@lucasssvaz - Thank you for adding CROMFS.  LGTM

fsutils/romloader/.gitignore
```diff
@@ -0,0 +1,3 @@
+/romfs.img
+/rom.c
+/rom
```
pkarashchenko, line 3:
```suggestion
/rom

```

fsutils/romloader/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+nsh> romloader
+Registering romdisk at /dev/ram0
+Mounting ROMFS filesystem at target=/mnt/romloader with source=/dev/ram0
+nsh> ls mnt/romloader
+/mnt/romloader:
+ file_a
+ file_b
+```
```
pkarashchenko, line 56:
```suggestion
```

```

Comment (hartmannathan) [reply fsutils/romloader/README.md:56]: @pkarashchenko Are you sure this line should be removed?

Comment (pkarashchenko) [reply fsutils/romloader/README.md:56]: No, sorry it is wrong edit. It was add new line at the end of file

**Review by hartmannathan — APPROVED** [review]
LGTM thanks!

**Review by pkarashchenko — APPROVED** [review]
LGTM!

File: Makefile [hunk 7]
```diff
@@ -76,7 +77,7 @@ ifeq ($(CONFIG_BUILD_KERNEL),y)
 
 install: $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_install)
 
-$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
+$(BIN): $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_romloader) $(foreach SDIR, $(CONFIGURED_APPS), $(SDIR)_all)
```
xiaoxiang781216, line 80: is it good design to modify the common script to support a special application needs?

fsutils/romloader/Makefile
```diff
@@ -0,0 +1,106 @@
...
+############################################################################
+# Applications Configuration
+############################################################################
+
+PROGNAME = romloader
+PRIORITY = SCHED_PRIORITY_DEFAULT
+STACKSIZE = $(CONFIG_DEFAULT_TASK_STACKSIZE)
+MODULE = $(CONFIG_FSUTILS_ROMLOADER)
+
+MAINSRC = romloader_main.c
```
xiaoxiang781216, line 104: nsh already support mount romfs/cromfs functionality, why not add ROMLOADER_COPY to nsh instead?

xiaoxiang781216 (changes requested):

Makefile
lucasssvaz: I don't like it either but it was the only way I found to make sure that the romloader app is compiled after all the required files are generated (without making it a kconfig dependency)

fsutils/romloader/Makefile
lucasssvaz: What do you mean ?

Makefile
xiaoxiang781216: Let's user add `ROMLOADER_COPY += xxx` to Make.defs, you can use ROMLOADER_COPY in your Makefile directly.

fsutils/romloader/Makefile
xiaoxiang781216: I mean this patch just need create a general mechanism to gather file list and generate romfs/cromfs, or even better any filesystem supported by nuttx. And let user decide how to use filesystem bin: either convert to C array and embed to code or flash to the nan/nor/emmc storage.

Makefile
lucasssvaz: This was what I did in one of my first attempts. It works fine for static files. The problem was when the files are generated or downloaded during the compilation process. I wasn't able to call these files recipes from the romloader Makefile. Is there a way around this ?

fsutils/romloader/Makefile
lucasssvaz: Yeah, but I have no idea on how to embed that on the NSH 🤣 
If you can give me some overview I could try to do it myself.

Makefile
xiaoxiang781216: In this case, the user could do the generation in context target just like how we deal with the 3rd party libraries. I found that context phase is a best place to resolve the cross project dependence.

fsutils/romloader/Makefile
xiaoxiang781216: nsh give the user three choice to provide ROMFS/CROMFS content:
https://github.com/apache/nuttx-apps/blob/master/nshlib/Kconfig#L861-L883
The first option(NSH_DEFAULTROMFS) is a fixed and prebuilt content here:
https://github.com/apache/nuttx-apps/blob/master/nshlib/nsh_romfsimg.h
So, the simplest approach is overwrite this file dynamically with your general mechanism. Another approach is adding an new choice and do your work there.

lucasssvaz: That's a good idea. I'll try to implement this next week.

Makefile
lucasssvaz: Doing something like ```context:: $(ROMLOADER_COPY)``` in the romloader makefile ? I tried this but it couldn't find the target for the files as it is in a different makefile.
I'm probably missing something. Idk a lot of how the NuttX build system works. Could you point me to an example of where this is done ?

xiaoxiang781216: No, I mean the user could generate the romfs content in context phase:
```
I don't like it either but it was the only way I found to make sure that the romloader app is compiled after all the required files are generated (and without making it a kconfig dependency).
```

lucasssvaz:
Closing this as I'll re-implement this embedded in NSH.

PR #1702, opened 2023-04-04
nsh/nsh_parse: Fix handling of back-quotes

pussuw:
## Summary
The logic that handles back-quotes was faulty, i.e. example command
set FOO `ls -l` would be split into two tokens as follows:
- set FOO `ls
- -l`

This results in nsh: `: no matching ` error, this fixes that issue.
## Impact
Fix something I broke when shell aliases were introduced.
nsh> set FOO `ls -la`
nsh: `: no matching `

## Testing
icicle:nsh

### DIFF
--- a/nshlib/nsh_parse.c
+++ b/nshlib/nsh_parse.c
@@ -1769,7 +1769,13 @@ static FAR char *nsh_argument(FAR struct nsh_vtbl_s *vtbl,
 
               /* Is it a back-quote ? These are not removed here */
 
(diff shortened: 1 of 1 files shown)


pussuw:
Seems like github wants to re-format the back-quotes from the PR title. The commit message has the correct description.

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #1706, opened 2023-04-05
nsh/alias: Add support for alias arguments

pussuw:
## Summary
This adds support for more complex alias handling, such as:

$ alias ls='ls -l'

Previously such an alias was not split into the command verb and the argument correctly, instead the full alias string was handled as the verb, which obviously fails.

This commit fixes this by expanding the alias, checking whether it has arguments and if so, it merges the expanded alias + the old command line together, resulting in a completely new command line.

Example (assuming the alias above has been created):

$ ls /bin

Results in a new command line: "ls -l /bin" which is then parsed and executed.
## Impact
Adds support for aliases with arguments
## Testing
icicle:nsh

### DIFF
--- a/nshlib/nsh_parse.c
+++ b/nshlib/nsh_parse.c
@@ -59,11 +59,29 @@
 #  define HAVE_MEMLIST 1
 #endif
 
+/* If CONFIG_NSH_ALIAS is enabled, the alias strings might need dynamic
+ * memory, in case the alias has arguments and is set like:
+ *
+ * $ alias ls='ls -l'
+ *
+ * In this case the command verb and the arguments need to be separated, much
+ * like the argument separation is done with environment variable expansion.
+ *
+ * This needs a new working buffer in order to keep the original alias string
+ * intact.
+ */
+
+#ifdef CONFIG_NSH_ALIAS
+#  define ALIAS_ALLOCS           2
+#else
+#  define ALIAS_ALLOCS           0
+#endif
+
 #if defined(HAVE_MEMLIST) && !defined(CONFIG_NSH_MAXALLOCS)
 #  ifdef CONFIG_NSH_ARGCAT
-#    define CONFIG_NSH_MAXALLOCS (2*CONFIG_NSH_MAXARGUMENTS)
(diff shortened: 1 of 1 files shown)


nshlib/nsh_parse.c
```diff
@@ -59,11 +59,29 @@
...
+#ifdef CONFIG_NSH_ALIAS
+#  define ALIAS_ALLOCS           2
+#else
+#  define ALIAS_ALLOCS           0
+#endif
+
 #if defined(HAVE_MEMLIST) && !defined(CONFIG_NSH_MAXALLOCS)
 #  ifdef CONFIG_NSH_ARGCAT
-#    define CONFIG_NSH_MAXALLOCS (2*CONFIG_NSH_MAXARGUMENTS)
+#    define CONFIG_NSH_MAXALLOCS ((2*CONFIG_NSH_MAXARGUMENTS) + ALIAS_ALLOCS)
```
pkarashchenko, line 82:
```suggestion
#    define CONFIG_NSH_MAXALLOCS ((2 * CONFIG_NSH_MAXARGUMENTS) + ALIAS_ALLOCS)
```

pkarashchenko (approved):

nshlib/nsh_parse.c
pussuw: I like the suggestion but the reason I did not add the spaces originally was that it would exceed the 78 character limit. 

```
#    define CONFIG_NSH_MAXALLOCS ((2 * CONFIG_NSH_MAXARGUMENTS) + \ 
                                                                  ALIAS_ALLOCS)
```

So it is a trade-off, add spaces to improve readability but add extra white space, or keep it like it is.

The original code did not have the whitespaces so that is why I did not add them + the line change myself.

What do you think, which is better ? I'm fine either way.

pkarashchenko: I approved the PR, so I'm fine with current variant. If adding spaces makes macro to become multiline then I think better to keep as is.

pkarashchenko (approved):

masayuki2009 (approved):

PR #1729, opened 2023-04-28
apps/examples/pipe: test file_pipe opening and closing

tmedicci:
## Summary

Test `file_pipe()` opening and closing to make sure it doesn't block unexpectedly. 

Also, print messages to stderr to avoid messing with the duplicated stdout.

## Impact

Test https://github.com/apache/nuttx/pull/9130

## Testing

Internal CI + sim:citest

### DIFF
--- a/examples/pipe/pipe_main.c
+++ b/examples/pipe/pipe_main.c
@@ -45,7 +45,7 @@ static void *open_write_only(pthread_addr_t pvarg)
 {
   void *fd_addr = (void *)pvarg;
 
-  printf("open_write_only: Opening FIFO for write access\n");
+  fprintf(stderr, "open_write_only: Opening FIFO for write access\n");
   ((int *)fd_addr)[1] = open(FIFO_PATH1, O_WRONLY);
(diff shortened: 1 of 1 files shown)


examples/pipe/pipe_main.c
```diff
@@ -204,26 +206,59 @@ int main(int argc, FAR char *argv[])
...
+  fprintf(stderr, "pipe_main: pipe() redirection test PASSED\n");
+
+  /* Test file_pipe() command */
+
+  fprintf(stderr, "\npipe_main: Open with file_pipe()\n");
+
+  filepipep[0] = &filepipe[0];
+  filepipep[1] = &filepipe[1];
+
+  ret = file_pipe(filepipep, CONFIG_DEV_PIPE_SIZE, 0);
```
xiaoxiang781216, line 236: it isn't good to call kernel specific function from user space. let's test pipe or open_pty instead.

tmedicci: You are right!

We'd need to test using pty, but it doesn't seem to fit on this test. I'll drop it for now.

PR #1740, opened 2023-05-06
Added example for Bosch BMP280 pressure sensor

fdcavalcanti:
## Summary
Added BMP280 example to apps.
Currently, there is a BMP180/280 example which reads from _/dev/press0_. It is not compatible with BMP280 since the driver binds to _/dev/uorb/sensor_baro0_.

## Impact
New example add to apps/example.

## Testing
Tested on the BMP280 I2C sensor using a Tiva TM4C board. No issues noticed.

### DIFF
--- a/examples/bmp280/Kconfig
+++ b/examples/bmp280/Kconfig
@@ -1,0 +1,30 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config EXAMPLES_BMP280
+	tristate "BMP280 Barometer sensor example"
+	default n
+	depends on SENSORS_BMP280
+	---help---
+		Enable the BMP280 example
+
+if EXAMPLES_BMP280
+
+config EXAMPLES_BMP280_PROGNAME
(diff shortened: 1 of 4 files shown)


xiaoxiang781216 (approved):

xiaoxiang781216:
@fdcavalcanti you can try this tool directly:
https://github.com/apache/nuttx-apps/blob/master/system/uorb/listener.c

fdcavalcanti:
@xiaoxiang781216 thanks for the reply, I would be happy to improve the commit. However,  I'm afraid I did not understand your suggestion. Could you elaborate?

xiaoxiang781216:
> @xiaoxiang781216 thanks for the reply, I would be happy to improve the commit. However, I'm afraid I did not understand your suggestion. Could you elaborate?

This patch looks good. The comment isnot related to your change, just notice you there is a general tool to do the basic sensor testing and more.

jerpelea:
@fdcavalcanti  for the future please add a commit message describing the change

fdcavalcanti:
Thanks everyone.

PR #1756, opened 2023-05-15
ostest: sem need to be destroyed

hujun260:
## Summary
ostest: sem need to be destroyed

## Impact
none

## Testing
ostest


xiaoxiang781216:
LGTM.

PR #1759, opened 2023-05-17
add fdsan unit test

hujun260:
## Summary
Add fdsan unit test. Due to the current Nuttx not supporting death test, we removed death test from the original test cases of Android

## Impact
none

## Testing
none


btashton (approved):
preemptively merging since lint passed and there is nothing for the other cases to test as this is not enabled yet in a build.

xiaoxiang781216 (approved):

PR #1765, opened 2023-05-18
examples/foc: add real time data capture with nxscope lib

raiden00pl:
## Summary

- examples/foc/foc_thr: fix mutex logic
- examples/foc/foc_thr: get controller type from controller ID
- examples/foc: add real time data capture with the NxScope library

## Impact
It's now possible to capture controller data easily and then visualizing them. Useful for controller tuning and debugging.

## Testing
Tested with nxscope client https://github.com/railab/nxscli
Data stream on UART works fine.
Data stream on full-speed CDCACM device adds too much uC overload which breaks FOC control loops and is not recommended (usbdev full-speed packet limit is too small). High-speed CDCACM should be more practical, but I don't have proper hardware to test.

### DIFF
--- a/examples/foc/Kconfig
+++ b/examples/foc/Kconfig
@@ -421,5 +421,92 @@ config EXAMPLES_FOC_RUN_DISABLE
 	bool "FOC Disable FOC motor controller"
 	default n
 
+config EXAMPLES_FOC_NXSCOPE
+	bool "FOC nxscope support"
+	depends on LOGGING_NXSCOPE
+	select LOGGING_NXSCOPE_DISABLE_PUTLOCK
+	default n
+	---help---
+		This option enables a controller real-time data capture with
+		the NxScope library.
+
+if EXAMPLES_FOC_NXSCOPE
+
+config EXAMPLES_FOC_NXSCOPE_SERIAL
+	bool "FOC nxscope on serial port"
+	select LOGGING_NXSCOPE_INTF_SERIAL
+	default y
+
+if EXAMPLES_FOC_NXSCOPE_SERIAL
+
+config EXAMPLES_FOC_NXSCOPE_SERIAL_PATH
+	string "FOC nxscope serial path"
+	default "/dev/ttyS0"
+
+config EXAMPLES_FOC_NXSCOPE_SERIAL_BAUD
+	int "FOC nxscope serial baud"
+	default 115200
+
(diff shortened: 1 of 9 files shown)


cederom (approved):
looks cool! :-)

xiaoxiang781216 (approved):

cederom:
> @raiden00pl: Tested with nxscope client https://github.com/railab/nxscli Data stream on UART works fine. Data stream on full-speed CDCACM device adds too much uC overload which breaks FOC control loops and is not recommended (usbdev full-speed packet limit is too small). High-speed CDCACM should be more practical, but I don't have proper hardware to test.

@raiden00pl did you consider USB HID? It uses small fast interrupt transfers and provides quite good portability (+python modules on the host side). There is an interesting discussion at https://github.com/libusb/hidapi/issues/274#issuecomment-1542958282 on improving transfers at low level and hidapi vs hidraw implementation for various OS in favor of hidraw replacing hidapi :-)

I have just ordered cheap ESP32-S2 devkits maybe they will provide necessary USB Device to play with.. and the second device has OLED Display :-)
1. https://allegro.pl/oferta/esp32-s2-mini-s2fn4r2-wifi-wemos-micropython-4mb-13135527193
2. https://allegro.pl/oferta/esp32-s2-pico-oled-wifi-wemos-micropython-4mb-13135527582

raiden00pl:
@cederom that's interesting idea. I have no experience with usbdev as HID and NuttX doesn't seem to support HID device driver, so it's not so easy to verify :)

cederom:
There is a HIDKBD example app but I did not use it yet.. lets see when ESP32-S2 comes and some free time to play :-)

https://github.com/apache/nuttx-apps/tree/master/examples/hidkbd

If that chip works with both USB and WiFi then I could finally create virtual hid tester / fuzzer one day.. I have this idea to emulate various USB HID devices for host stack and libraries testing.. as for now people need to have whole bunch of devices while usually it is only the descriptor part that matters and some payload data all this could be provided over WiFi :-)

xiaoxiang781216 (approved):

mcuee:
@cederom 

For HIDAPI I have used the testing devices mentioned here.
https://github.com/libusb/hidapi/issues/478

ESP32 C3/S2/S3/C6 should have have USB HID support and BLE HID support.

PR #1775, opened 2023-05-25
system/audio: check if channels are valid

jinxiuxu:
check ac_channels upper four bits which means min channels.


xiaoxiang781216:
@jinxiuxu please fix "Remove Gerrit Change-ID's before submitting upstream".

jerpelea (approved):

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #1788, opened 2023-06-09
nshlib: disable nsh error print if NSH_DISABLE_ERROR_PRINT enabled

anchao:
## Summary

nshlib: disable nsh error print if NSH_DISABLE_ERROR_PRINT enabled

```
   text    data     bss     dec     hex filename
 398953   27088    4128  430169   69059 nuttx    /* before */
 389241   27072    4128  420441   66a59 nuttx    /* after  */
  -9712     -16
```

Signed-off-by: chao an <anchao@xiaomi.com>

## Impact

N/A

## Testing

sim/nsh


nshlib/nsh_console.c
```diff
@@ -63,8 +63,10 @@ static int nsh_consoleioctl(FAR struct nsh_vtbl_s *vtbl,
                             int cmd, unsigned long arg);
 static int nsh_consoleoutput(FAR struct nsh_vtbl_s *vtbl,
                              FAR const char *fmt, ...) printf_like(2, 3);
+#ifndef CONFIG_DEFAULT_SMALL
 static int nsh_erroroutput(FAR struct nsh_vtbl_s *vtbl,
```
patacongo, line 67: I would prefer that CONFIG_DEFAULT_SMALL not be used in this manner.  CONFIG_DEFAULT_SMALL is a property of the hardware (or should be), usually meaning is that the MCU has small memories.  Its function is to the choose a smaller default value for other settings and should not change for a given MCU.

My preference would be that you honor these semantics.  I suggest that add a new configuration like:

    config NSH_DISABLE_ERROR_PRINT
        bool "Disable NSH Error Printing"
       default y if DEFAULT_SMALL
       default n if !DEFAULT_SMALL

The option should be selected even if the MCU has CONFIG_DEFAULT_SMALL.

This comment applies to all of the size reduction changes you have made that uses CONFIG_DEFAULT_SMALL to directly control features.  It should not be used that way.

anchao: Done, Thanks for review.

nshlib/Kconfig
```diff
@@ -377,6 +377,11 @@ config NSH_DISABLE_HELP
 	bool "Disable help"
 	default DEFAULT_SMALL
 
+config NSH_DISABLE_ERROR_PRINT
+	bool "Disable NSH Error Printing"
+	default y if DEFAULT_SMALL
```
xiaoxiang781216, line 382: default DEFAULT_SMALL

anchao: Done

nshlib/nsh_console.c
xiaoxiang781216: @patacongo do you really mean CONFIG_SMALL_MEMORY not CONFIG_DEFAULT_SMALL?

nshlib/nsh_console.h
```diff
@@ -58,6 +58,15 @@
 #  define nsh_output            vtbl->output
 #endif
 
+#ifdef CONFIG_NSH_DISABLE_ERROR_PRINT
+# undef nsh_error
```
xiaoxiang781216, line 62: two space indent

nshlib/nsh_alias.c
```diff
@@ -47,22 +47,14 @@
...
-
-/****************************************************************************
- * Private Data
- ****************************************************************************/
-
 /* Alias message format */
 
-static const char g_aliasfmt[]    = "alias %s='%s'\n";
-static const char g_savefailfmt[] = "alias %s='%s' failed\n";
+#define alias_format            "alias %s='%s'\n"
```
xiaoxiang781216, line 52: should we keep g_ prefix like g_dd?

pkarashchenko: Let's change to `atic const char *g_aliasfmt = "alias %s='%s'\n";` that will have basically the same effect with optimisation turned on

nshlib/nsh_console.h
anchao: Done

nshlib/nsh_alias.c
anchao: Done

xiaoxiang781216 (approved):

nshlib/nsh_alias.c
pkarashchenko: So you still prefered to use define and not pointer to constant data? May I ask what is the difference in size/mem with define vs pointer to constant data?

xiaoxiang781216: The problem is that the compiler will complain static symbols isn't used.

PR #1789, opened 2023-06-10
wireless/bluetooth/btsak: fix help messages

pkarashchenko:
## Summary
Currently the help messages use "private" string while code expects "random".
Remove not needed `(CODE void *)` cast.
Use 'nitems' macro to calculate number of elements in array Align output in `btsak_cmd_scanget` command.
Fix style issues

## Impact
Should be none

## Testing
Pass CI
Tested with `btsak` application on ESP32

### DIFF
--- a/wireless/bluetooth/btsak/README.md
+++ b/wireless/bluetooth/btsak/README.md
@@ -54,7 +54,7 @@ Description:  Enable security (encryption) for a connection:
               This command may return error if required level of security
               is not possible to achieve due to local or remote device
               limitation (eg input output capabilities).
-Usage:        bt <ifname> security [-h] <addr> public|private <level>
+Usage:        bt <ifname> security [-h] <addr> public|random <level>
 Where:        <addr>  - The 6-byte address of the connected peer
               <level> - Security level, on of:
 
@@ -76,7 +76,7 @@ Where:        See "GATT Commands" below
 ```
 Command:      exchange-mtu
 Description:  Set MTU to out maximum and negotiate MTU with peer
-Usage:        bt <ifname> gatt exchange-mtu [-h] <addr> public|private
+Usage:        bt <ifname> gatt exchange-mtu [-h] <addr> public|random
 ```
 
 ```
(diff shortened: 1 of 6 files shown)


cederom:
Quick search [1] reveals this is true, confirmed by the Bluetooth Core Specification version 4.0 [2] and latest version 5.4 [3] :-)

Bluetooth Core Specification 4.0:
> 10.8 RANDOM DEVICE ADDRESS
> Random device addresses are defined in the Link Layer section in [Vol. 6], Part B Section 1.3. For the purposes of this profile, the random device address may be of either of the following two sub-types:
> • Static address
> • Private address.
> The term random device address refers to both static and private address types. The transmission of a random device address is optional. A device shall accept the reception of a random device address from a remote device.
> The private address may be of either of the following two sub-types:
> • Non-resolvable private address
> • Resolvable private address

Bluetooth Core Specification 5.4:

> 1.3 DEVICE ADDRESS
> Devices are identified using a device address and an address  type; the address type indicates either a public device address or a random device address.
> A public device address and a random device address are both 48 bits in length. A device shall use at least one type of device address and may use both.  The device may be addressed by any device address that it uses.
> A device's Identity Address is a Public Device Address or Random Static Device Address that it uses in packets it transmits. If a device is using Resolvable Private Addresses, it shall also have an Identity Address. Whenever two device addresses are compared, the  comparison shall include the device address type (i.e. if the two addresses have different types, they are different even if the two 48-bit addresses are the same).
> 1.3.1 Public device address
> The public device address shall be created in accordance with [Vol 2] Part B, Section 1.2, with the exception that the restriction on LAP values does not apply unless the public device address will also be used as a BD_ADDR for a BR/EDR Controller.
> 1.3.2 Random device address
> The random device address may be of either of the following:
> • Static address
> • Private address.


[1] https://novelbits.io/bluetooth-address-privacy-ble/
[2] https://www.bluetooth.com/specifications/specs/core-specification-4-0/
[3] https://www.bluetooth.com/specifications/specs/core-specification-5-4/

cederom (approved):
Looks good :-) Thank you @pkarashchenko :-)

xiaoxiang781216 (approved):

PR #1796, opened 2023-06-16
toywasm: disable tail call

yamt:
## Summary
toywasm: disable tail call

## Impact
fix occasional stack overflow on esp32

## Testing
Tested on esp32-devkitc


pkarashchenko:
Could you please update commit message of a second commit with few more words (it seems to be too laconic to me)?

yamt:
> Could you please update commit message of a second commit with few more words (it seems to be too laconic to me)?

done.

these files are generated by regen.sh.
after changing regen.sh, they need to be updated.

pkarashchenko:
Thank you so much.

pkarashchenko (approved):

PR #1810, opened 2023-06-28
import/mk: Update ELF flags to represent build configuration type.

g2gps:
## Summary
Allows board configuration to select the output ELF format for applications built using the `make import` target.

This PR should be paired with [#9635](https://github.com/apache/nuttx/pull/9635), from the NuttX main repository.

## Impact

The format of application ELFs built through `make import` can be set by board configuration.

## Testing

Copied from [#9635](https://github.com/apache/nuttx/pull/9635):

```
# Using a board configuration which has `CONFIG_BINFMT_ELF_EXECUTABLE=y`
$ cd nuttx
$ ./tools/configure.sh arty_a7:knsh
$ make
$ make export
$ cd ../apps
$ ./tools/mkimport.sh -z -x ../nuttx/nuttx-export-*.tar.gz
$ make import
$ riscv64-unknown-elf-readelf -a bin/init | grep Type:
  Type:                              EXEC (Executable file)

# Using a board configuration which does not define `CONFIG_BINFMT_ELF_EXECUTABLE`
$ cd nuttx
$ ./tools/configure.sh rv-virt:knsh32:knsh
$ make
$ make export
$ cd ../apps
$ ./tools/mkimport.sh -z -x ../nuttx/nuttx-export-*.tar.gz
$ make import
$ riscv64-unknown-elf-readelf -a bin/init | grep Type:
  Type:                              REL (Relocatable file)
```

### DIFF
--- a/import/Make.defs
+++ b/import/Make.defs
@@ -90,5 +90,8 @@ endif
 
 # ELF module definitions
 
-LDELFFLAGS += -r -e _start -Bstatic
+ifeq ($(CONFIG_BINFMT_ELF_RELOCATABLE),y)
+  LDELFFLAGS += -r
+endif
+LDELFFLAGS += -e _start -Bstatic
 LDELFFLAGS += $(addprefix -T,$(call CONVERT_PATH,$(TOPDIR)/scripts/gnu-elf.ld))


acassis:
Hi @g2gps thanks for fixing it.

Someone ping me few days ago saying that running external ELF binaries on NuttX as explained on my video: https://www.youtube.com/watch?v=oL6KAgkTb8M is failing.

Then I rested on versions: 12.1, 9.1 and 7.25 (that was used in the video). Only version 7.25 is working. Probably some modification after version 7.25 broke it and nobody paid attention.

I think we need to move the scripts from https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=139629542 to apps/ to let people to test it easily.

Something I noticed is that the addon/Makefile looks for nuttx-export/System.map that is not exported anymore.
So, since you working on it, could you please help me to fix it?

Neale is also fixing the issues on his ET_DYN Loader PR: https://github.com/apache/nuttx/pull/7202
So, soon we will have support to true libraries on NuttX

xiaoxiang781216 (approved):

g2gps:
Hi @acassis,

> Something I noticed is that the addon/Makefile looks for nuttx-export/System.map that is not exported anymore.
So, since you working on it, could you please help me to fix it?

I added #1816, which adds the System.map file back into the `import` directory. I can't guarantee it will fix the issue entirely, but at least the symbol table is there for external applications to build.

I had a look through your video, and it actually helped clarify a few things about the different ways Nuttx support building and launching applications.

As a brief summary (more for my understanding):

1. Nuttx can build applications using by resolving symbols against the addresses in the `System.map` file. This would only work in CONFIG_BUILD_FLAT (I believe). There are three variants of this method which you covered in your video.
2. Nuttx can load relocatable (partially linked) ELFs. This works in CONFIG_BUILD_KERNEL, however the kernel needs to process relocations for both library and application symbols. As a result, the application ELF files are much larger, take longer to load and are subject to unresolved symbols at load time.
3. Nuttx can load executable ELFs in `CONFIG_BUILD_KERNEL`.  This relies on the application being linked correctly to match the virtual memory configuration of the target, and kernel.

I've been chipping away at adding support for (3) over the last couple of months,  as the drawbacks of (2) were too much for our current application. As far as I know, the arty_a7 is the only platform which supports building applications in this manor. Loading fully linked applications should be supported generically.

> I think we need to move the scripts from https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=139629542 to apps/ to let people to test it easily.

I agree. The full extent of Nuttx's capability is not obvious, especially to a new-comer.  Perhaps the first step would be to add a brief outline of the currently supported options to the Documentation, with an example of a board configuration which supports it?

acassis:
@g2gps I think discussion to the mailing list to let more people to participate and then I can create a summary to include as Documentation here: https://nuttx.apache.org/docs/latest/components/binfmt.html#

PR #1813, opened 2023-06-28
Add small test of pthread_exit()/pthread_self() on main thread.

patacongo:
## Summary

Add a minimal test to assure that pthread_self() and pthread_exit() work correctly when called from the main thread.

## Impact

None expected.  Worst case would be an OS test failure

## Testing

sim/ostest

Test result:

1. Test has to run in a separate task because it will exit.

   pthread_exit_test:  Starting pthread_exit_main at PID=183

2. Test logic a) starts the pthread_exit_thread, b) waits 5 seconds and then c) calls pthread_exit()

    a) pthread_exit_main 183: Starting pthread_exit_thread
    b) pthread_exit_main 183: Sleeping for 5 seconds

3. Process does not terminate (correct) and pthread_exit_thread is started
4. The pthread_exit_thread d) sleeps for 5 seconds, then e) 5 more seconds, then f)  implicitly calls pthread_exit()

    d) pthread_exit_thread 189: Sleeping for 10 second
    c) pthread_exit_main 183: Calling pthread_exit()
    e) pthread_exit_thread 189: Still running...
    f) pthread_exit_thread 189: Exiting

5. Now the process terminates and the OS test continues

### DIFF
--- a/testing/ostest/Makefile
+++ b/testing/ostest/Makefile
@@ -66,6 +66,10 @@ CSRCS += cancel.c cond.c mutex.c timedmutex.c sem.c semtimed.c barrier.c
 CSRCS += timedwait.c
 CSRCS += pthread_rwlock.c pthread_rwlock_cancel.c
 
+ifeq ($(CONFIG_SCHED_WAITPID),y)
+CSRCS += pthread_exit.c
+endif
+
 ifneq ($(CONFIG_TLS_NELEM),0)
 CSRCS += specific.c
 endif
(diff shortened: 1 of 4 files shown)


patacongo:
Inspired by the discussion of #apache/nuttx/9632 and based on the discussion and examples at https://stackoverflow.com/questions/3559463/is-it-ok-to-call-pthread-exit-from-main

This is a minimal test that only verifies correct behavior in the conditions of the discussion there.  It could be extended to provide a more rigorous, complete test.

patacongo:
macOS CI tests are failing for unknown reasons.

xiaoxiang781216 (approved):

PR #1819, opened 2023-07-01
apps/libtest: add libtest demo

xiaoxiang781216:
## Summary

Redefine the name of BIN to support static library:
```
BIN = $(APPDIR)/libtest$(LIBEXT)
```

## Impact

new example

## Testing

ci

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -642,6 +642,21 @@ maintaining duplicate logic in the NuttX repository.
 This is a simple test of the board LED driver at
 `nuttx/drivers/leds/userled_*.c`.
 
+## `libtest` Static Library Test
+
+This example illustrates how you may create a static library. It does the following:
+
+It creates a static library called libtest.a that contains an object that provides the symbol library_test().
+
+At adds the library as an EXTRA_LIB in the build
+
+EXTRA_LIBS += -ltest
+E XTRA_LIBPATHS += -L$(APPDIR)/examples/libtest
+
+And optionally, it can be configured to:
+
+Generate a built-in command that can be executed by NSH. This command logic links with the symbol library_test() that will provided by the libtest.a static library.
+
 ## `luamod_hello` Hello World Lua module
 
 A Lua C module showing how to add built-in modules to the Lua interpreter.
--- a/examples/libtest/Kconfig
+++ b/examples/libtest/Kconfig
@@ -1,0 +1,37 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config EXAMPLES_LIBTEST
+	bool "Static library Example"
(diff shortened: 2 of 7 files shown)


patacongo:
If this is really a test and not  usage example, then it would really belong under apps/testing rather than apps/examples.

xiaoxiang781216:
> If this is really a test and not usage example, then it would really belong under apps/testing rather than apps/examples.

The name may confuse you, but it is really a demo how to generate a separated library from libapps.a.

cederom (approved):
cool :-)

cederom:
Very nice example @xiaoxiang781216 :-)

Right now I am writing a firmware that does similar, there is a main that call separate functions according to need, and they can be moved to a library that way in order to be easily upgradable :-)

Is dynamic library supported as well on NuttX?

cederom:
Maybe as @patacongo suggest the name could be "libexample" @xiaoxiang781216 ? :-)

patacongo:
> Is dynamic library supported as well on NuttX?

Yes, but no in kernel build.  Checkout apps/examples/sotest

Hmm... another case that is really an example.  The difference between an example and a test can be subtle.

cederom:
Okay, so there is `sotest` already and I guess `libtest` comes from that name :-)

If I had created that example, I would put both static and dynamic libraries under one example application and named it `libexample` to have all-in-one place.. but granularity of examples here may have its advantages too :-)

Anyways its good to see library support is that easy in NuttX! :-)

patacongo:
Should apps/examples/README.md be updated as well?  People have gotten in the habit of adding examples like this one without updating the README to mention what the example does.  That makes the examples less useful.

Similarly for apps/examples/testing/README.md.  It would be good to know generally what each test verifies.

xiaoxiang781216:
Done, @patacongo .

patacongo:
> If I had created that example, I would put both static and dynamic libraries under one example application and named it `libexample` to have all-in-one place.. but granularity of examples here may have its advantages too :-)

I think there is a misconception here.  The static libraries (archives) and shared libraries are very different:  Archives are only used by the build system on the host.  They are just containers for compiled .o files and are really only usable by the linker which draws out the .o files and links them into the binary image.  So unless you have a linker in your target the static library is useless on the target.

The shared libraries, on the the other hand, are created at build time but are ONLY used at run-time in the target when the target application calls dlopen() and gets the function addresses from the shared library.  So these shared libraries are not useful on the host; they are used only in the target.

cederom:
Yup, I know the difference between static and dynamic library, just considered "a generic example for both static and dynamic library in one place".. but it may be confusing I can see it now maybe it is better to keep them as separate examples, thanks @patacongo  :-)

When my project is ready and I am familiar enough with the internals I will try to help with the documentation :-)

PR #1823, opened 2023-07-05
wamr: add a kconfig to control wasi-threads

yamt:
## Summary
cf. https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/pthread_impls.md

### DIFF
--- a/interpreters/wamr/Kconfig
+++ b/interpreters/wamr/Kconfig
@@ -112,9 +112,17 @@ config INTERPRETERS_WAMR_THREAD_MGR
 	bool "Enable thread manager"
 	default n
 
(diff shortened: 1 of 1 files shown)


interpreters/wamr/Kconfig
```diff
@@ -112,9 +112,17 @@ config INTERPRETERS_WAMR_THREAD_MGR
 	bool "Enable thread manager"
 	default n
 
+config INTERPRETERS_WAMR_LIB_WASI_THREADS
```
xiaoxiang781216, line 115: should we use choice here?

yamt: no.
you can enable both of them if you want.

PR #1839, opened 2023-07-16
Redesigning cachespeed

crafcat7:
## Summary
Compared to the previous version, the following factors have been taken into account in the new version of cachespeed:
    1. the user needs to care about getting a piece of available memory (even if this piece of memory may memory stomp during use)
    2. High repetition of test code, poor readability, and overall high test coupling
    3. Not taking into account the alignment cache line situation
    4. When comparing values between different test tools, there is a little error because of the different ways to get the values (different fine reading)
    Therefore, in the new version according to the above problems have been improved:
    1. Provide a "test_skeleton", which contains the overall testing process
    2. The user only needs to execute, not to care about how to get the available memory address, and the memory allocated by the program also ensures the memory security and data accuracy
    3. The system to obtain the data required for the test reduces the difficulty of use and the possibility of inaccurate results due to data errors.
    4. Provide two kinds of precision data results, which can be configured through Kconfig
    5. Optimize the output log, now more intuitive and concise, to help the subsequent data organization and observation
    6. New test items for aligned/unaligned cache line
    7. Better readability and extensibility, making it easier to add/remove test items


system/cachespeed/Kconfig
```diff
@@ -4,7 +4,8 @@
 #
 
 config SYSTEM_CACHESPEED
-	tristate "CACHE Speed Test"
+	bool "CACHE Speed Test"
+	depends on (ARCH_ICACHE && ARCH_DCACHE) || (ARMV7M_ICACHE && ARMV7M_DCACHE) || (ARMV8M_ICACHE && ARMV8M_DCACHE)
```
xiaoxiang781216, line 8: remove " || (ARMV7M_ICACHE && ARMV7M_DCACHE) || (ARMV8M_ICACHE && ARMV8M_DCACHE)"

crafcat7: Done

xiaoxiang781216 (approved):

PR #1844, opened 2023-07-17
system/nxdiag: Fix race condition during build

lucasssvaz:
## Summary

This commit fixed a race condition that might happen by only unshallowing the Espressif HAL only after it was completely cloned.
This is done by checking the `.git/index` file of the HAL repository.

## Impact

Correct compilation behavior of NXdiag.

## Testing

ESP32-C3-DevKitC


xiaoxiang781216:
let's ignore the macOS ci broken.

xiaoxiang781216 (approved):

PR #1872, opened 2023-07-31
support openflashloader for nuttx

anjiahao1:
## Summary
support openflashloader for nuttx
wiki:https://wiki.segger.com/SEGGER_Flash_Loader

## Impact
If the nuttx driver supports it, we can flash any firmware through openflashloader

## Testing
stm32f429i-disco with internal flash

Details:[nuttx-9962](https://github.com/apache/nuttx/pull/9962)
Please ignore "error: Mixed case identifier found" since some symbols defined by Segger are mix case.

### DIFF
--- a/system/ofloader/segger.c
+++ b/system/ofloader/segger.c
@@ -1,0 +1,244 @@
+/****************************************************************************
+ * apps/system/ofloader/segger.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <mqueue.h>
+
+#include <nuttx/init.h>
+
+#include "ofloader.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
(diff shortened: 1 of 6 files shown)


xiaoxiang781216 (approved):

jerpelea (approved):

jerpelea (commented):
please squash the patches

system/ofloader/segger.c
```diff
@@ -0,0 +1,244 @@
+/****************************************************************************
+ * apps/system/ofloader/segger.c
```
raiden00pl, line 2: shouldn't this file be on the kernel side? direct call to `__start` doesn't seem appropriate in apps

xiaoxiang781216: Yes, it isn't good to call __start from app, but since ofloader is a very low-level tool which interact JLINK directly:

1. JLINK require flash programmer must start with SEGGER_FL_Prepare in PrgData section
2. ofloader  is designed to work only in FLAT mode, it can work __start without trouble

Do you have the better ideal?

raiden00pl: Maybe it's worth moving the whole `ofloader` to the kernel side ? Is it foreseen that the `ofloader` will be run from NSH? or only as an entry point?

If we want to keep it in apps then we should use `boardctl` interface to call `__start`.

xiaoxiang781216: > Maybe it's worth moving the whole `ofloader` to the kernel side ? Is it foreseen that the `ofloader` will be run from NSH? or only as an entry point?

Yes, it's better to run from NSH since we want to support multiple flash method(e.g. ymodem, fastboot and ofloader) at the same time.

> 
> If we want to keep it in apps then we should use `boardctl` interface to call `__start`.

Here is how it work under JLINK:

1. JLINK load ofloader to SRAM through JTAG
2. JLINK jump PC to SEGGER_FL_Prepare 
3. SEGGER_FL_Prepare initialize the chip and return
4. JLINK prepare the data and jump PC to SEGGER_FL_Program
5. SEGGER_FL_Program write a chunk of data to flash and return

Repeat step 4-5 until the whole image is written to flash. The whole flash is controlled by JLINK, and boardctl can't work here, since __start can only run once.

raiden00pl: I'm not going to block this PR but let's wait a little with the merge, maybe someone will have some valuable feedback. I'm not familiar with this tool from Segger and I'm unable to play with it at the moment.

xiaoxiang781216: Sure, ofloader is designed to work in bootloader, and most bootloader is built with flat mode to make it as simple as possible, so the limitation here isn't too rigid as the first glance.

xiaoxiang781216: Yes, we have tried our best to make the violation as minimal as possible. This is a best solution we got after @anjiahao1 try several possible approaches.

xiaoxiang781216 (approved):

PR #1889, opened 2023-08-03
mcuboot: update mcuboot to support swap using move

michallenc:
## Summary
Older versions of mcuboot did not support swap using move and therefore the build would have failed CONFIG_MCUBOOT_SWAP_USING_MOVE=y.

Newer mcuboot contains initial support of swap using move for NuttX. Also Makefile was updated to as some bootutil files were removed from mcuboot.

## Impact
Updates MCUboot and because of deletion of some files older versions will not compile.

## Testing
Tested on SAMv7 MCU.


xiaoxiang781216 (approved):

michallenc:
Also updated cmake format

jerpelea (approved):

xiaoxiang781216 (approved):

PR #1923, opened 2023-08-09
sensor: add force sensor

Jxudong:
A sensor of this type measures the force on it, and additionally compares the force with one or more specified thresholds. The sensor can output the force value directly. Moreover, it's usually applied as a press key. In that case, when it detects a force greater than some given threshold, a corresponding event is reported.


Jxudong:
abandonabandon

PR #1938, opened 2023-08-10
add test case for regulator

DannyPlusPlus:
## Summary
add test case for regulator


jerpelea (approved):
please add commit message describing the change

PR #1975, opened 2023-08-17
Added tools for NXP SE05x (secure element)

AndreHeinemans-NXP:
## Summary
Added tools for NXP SE05x (secure element)
Depends on the SE05x driver in nuttx (PR pending)

## Impact
Some configuration options added in MbedTLS KConfig. Some tools in this PR depend on those.


testing/setest/Make.defs
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+# Copyright 2023 NXP
```
xiaoxiang781216, line 21: remove

testing/setest/Makefile
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+# Copyright 2023 NXP
```
xiaoxiang781216, line 21: remove all NXP copyright from code base

PetervdPerk-NXP: Could you clarify this further?

xiaoxiang781216: The new written code is better to declare as Apache copyright, the proprietary license is hard to accept or at least user need get the explicit notice like what we have done for BSD/MIT here:
https://github.com/apache/nuttx/blob/master/Kconfig#L28-L90

PetervdPerk-NXP: But it's Apache 2.0 licensed, I don't see the problem here.

xiaoxiang781216: but as far as I know, the new code should avoid add the company copyright like "Copyright 2023 NXP".

PetervdPerk-NXP: It's indicating the original copyright holder, I don't see why this should be avoided. It's quite common practice in other projects to include this i.e. Linux, Zephyr. OpenCV and more
`Copyright <YEAR> <COPYRIGHT HOLDER>`

xiaoxiang781216: @acassis @jerpelea please take a look.

xiaoxiang781216 (approved):

testing/setest/Makefile
btashton: I don't think this is correct. If NXP is not donating the code I believe it must be treated as third party and handled via the NOTICE.

PR #1980, opened 2023-08-18
examples/bme680: Added test program for the BME680 sensor

simonatoaca:
## Summary
This is a usage example for the bme680 when all its sub-sensors are enabled and it can be easily modified to accomodate other configurations.

## Testing
Testing was done on the esp32-sparrow board.


jerpelea (approved):
please add a commit message

xiaoxiang781216 (approved):

PR #1981, opened 2023-08-18
net/ip: print ip addresses using ip4_addrN macro

pkarashchenko:
## Summary
Print ip addresses using ip4_addrN macro

## Impact
N/A

## Testing
Pass CI


examples/bridge/host_main.c
```diff
@@ -32,6 +32,7 @@
 #include <net/if.h>
 #include <arpa/inet.h>
 #include <netinet/in.h>
+#include <nuttx/net/ip.h>
```
xiaoxiang781216, line 35: host can't use nuttx specific heade file

pkarashchenko: Yes. Missed that. Will fix

pkarashchenko: Done

pkarashchenko: The UDP example needs the same change

xiaoxiang781216 (approved):

PR #1983, opened 2023-08-20
mq_open return status is set as ERROR

TaiJuWu:
## Summary
mq_open return status is set as ERROR instead of -1
## Impact
None
## Testing
Noe


pkarashchenko:
I do not see any benefit of this change. From application point of view `(mqd_t)-1` is more portable

TaiJuWu:
In my mind, error is more intuitive than -1.

pkarashchenko:
From POSIX perspective the `ERROR` is not defined, so from Linux/FreeBSD/OSX/NuttX portability point of view the -1 is better.

TaiJuWu:
@pkarashchenko Thank you for your reply.

PR #1987, opened 2023-08-22
nshlib/nsh_timcmds.c: Suppress warning about unused variable (errfmt)

pussuw:
## Summary
Fixes following warning:

CC:  grp/lib_getgrbufr.c nsh_timcmds.c: In function 'cmd_date': nsh_timcmds.c:370:19: warning: variable 'errfmt' set but not used [-Wunused-but-set-variable]
  370 |   FAR const char *errfmt;
      |
## Impact
Fixes build warning
## Testing
Code compiles without the warning

### DIFF
--- a/nshlib/nsh_timcmds.c
+++ b/nshlib/nsh_timcmds.c
@@ -367,7 +367,7 @@ int cmd_time(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
 int cmd_date(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
 {
   FAR char *newtime = NULL;
-  FAR const char *errfmt;
+  FAR const char *errfmt unused_data;
   bool utc = false;
   int option;
   int ret;


xiaoxiang781216 (approved):

xiaoxiang781216:
Let's ignore riscv ci temp fail and merge this simple change directly.

nshlib/nsh_timcmds.c
```diff
@@ -367,7 +367,7 @@ int cmd_time(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
 int cmd_date(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
 {
   FAR char *newtime = NULL;
-  FAR const char *errfmt;
+  FAR const char *errfmt unused_data;
```
pkarashchenko, line 370: 1. But I see that it is passed to `nsh_error(vtbl, errfmt, argv[0]);`
2. Don't we usually use `UNUSED` for such cases?

pussuw: The way the nsh_error macro expands this does produce an unused variable error:

https://github.com/apache/nuttx-apps/blob/77e6c39cbd4758e2b6c7994cf266a9326cbfecb7/nshlib/nsh_console.h#L61-L68

AFAIK there is no clean way to suppress the error as nsh_error is a variadic function.

xiaoxiang781216: @pussuw we can fix like this:
```
#ifdef CONFIG_NSH_DISABLE_ERROR_PRINT 
#  undef nsh_error 
#  ifdef CONFIG_CPP_HAVE_VARARGS 
#    define nsh_error(v, ...)   do { if (0) (v, ##__VA_ARGS__); } while (0)
#  else 
#    define nsh_error         (void)(vtbl) 
#  endif 
#endif
```
like _none in debug.h:
```
#ifdef CONFIG_CPP_HAVE_VARARGS
/* don't call syslog while performing the compiler's format check. */

#  define _none(format, ...) \
    do { if (0) syslog(LOG_ERR, format, ##__VA_ARGS__); } while (0)
#else
#  define _none       (void)
#endif
```

pussuw: Hmm, looks like a pretty elegant solution. I can try this. If it works I'll revert this change and do this instead.

PR #1992, opened 2023-08-24
nshlib/nsh_console.h: Add nsh_none, where empty output can be forwarded

pussuw:
## Summary
Adds way to forward any output into nothingness. Handles CONFIG_NSH_DISABLE_ERROR_PRINT case gracefully for any
amount of arguments.
## Impact
Only nsh_error output, when CONFIG_NSH_DISABLE_ERROR_PRINT is set
## Testing
Build passes, output works when enabled, output is silent when not

### DIFF
--- a/nshlib/nsh_console.h
+++ b/nshlib/nsh_console.h
@@ -51,20 +51,19 @@
 #define nsh_exit(v,s)          (v)->exit(v,s)
 
 #ifdef CONFIG_CPP_HAVE_VARARGS
-#  define nsh_error(v, ...)     (v)->error(v, ##__VA_ARGS__)
-#  define nsh_output(v, ...)    (v)->output(v, ##__VA_ARGS__)
+#  define nsh_error(v, ...)    (v)->error(v, ##__VA_ARGS__)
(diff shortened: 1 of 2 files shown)


pussuw:
@xiaoxiang781216 @pkarashchenko this is what I came up with regard to https://github.com/apache/nuttx-apps/pull/1987

The inline function is needed to eat the variadic argument list, even though `if (0) (v, ##__VA_ARGS__);` is perfectly valid C code, it does produce a warning, the compiler is sad that the comma expression is meaningless. Even casting the whole statement to void does not suppress the warning.

nsh_parse.c:2739:39: error: left-hand operand of comma expression has no effect [-Werror=unused-value]
 2739 |       nsh_error(vtbl, g_fmttoomanyargs, cmd);

xiaoxiang781216:
can you try this `(void)(1 || (v, ##__VA_ARGS__))`?

pussuw:
> can you try this `(void)(1 || (v, ##__VA_ARGS__))`?

Gives the same warning.

xiaoxiang781216 (approved):

PR #2003, opened 2023-08-27
add macro for ecc

xiaoxiang781216:
## Summary

application may use micro-ecc for some ecc operation, which may be faster than tinycrypt, due to assemble code, but both of them have same function definition, cause build error on there case.

## Impact

new option

## Testing

internal test


xiaoxiang781216:
@pkarashchenko could you review this simple change?

pkarashchenko (approved):

PR #2008, opened 2023-08-28
crypto: tinycrypt: Add macro for ecc

LingaoM:
application may use micro-ecc for some ecc operation, which may be faster than tinycrypt, due to assamble code, but both of them have same function defination, cause build error on those case.


LingaoM:
same with https://github.com/apache/nuttx-apps/pull/2003

PR #2017, opened 2023-08-30
Fix wrong parameter size

WORsjlt:
## Summary
Type int expects 4 bytes for function write(). Type char expects 1 byte which matches the real usage of function write().

## Testing

ci

### DIFF
--- a/system/cu/cu_main.c
+++ b/system/cu/cu_main.c
@@ -446,12 +446,15 @@ int main(int argc, FAR char *argv[])
 
   while (!cu->force_exit)
     {
-      int ch = getc(stdin);
+      char ch;
+      int c = getc(stdin);
 
-      if (ch < 0)
+      if (c < 0)
         {
           continue;
         }
+
+      ch = c;
 
       if (nobreak == 1)
         {


system/cu/cu_main.c
```diff
@@ -446,7 +446,7 @@ int main(int argc, FAR char *argv[])
 
   while (!cu->force_exit)
     {
-      int ch = getc(stdin);
+      signed char ch = getc(stdin);
```
pkarashchenko, line 449: This seems to be improper fix as `getc` return value is truncated.
We should stay with `int ch = getc(stdin);`, but replace `write(cu->outfd, &ch, 1);` with
```
char c = ch;
write(cu->outfd, &c, 1);
```
Or we can switch from `int outfd` to `FAR FILE *outfile` and use `fputc(ch, cu->outfile)` instead of `write(cu->outfd, ...);`

pkarashchenko (changes requested):

system/cu/cu_main.c
pkarashchenko: The issue with current code should only be a problem for BE architectures

WORsjlt:
Yes! Casting after non-negative check should be more appropriate.

system/cu/cu_main.c
```diff
@@ -446,13 +446,15 @@ int main(int argc, FAR char *argv[])
...
-      int ch = getc(stdin);
+      int c = getc(stdin);
 
-      if (ch < 0)
+      if (c < 0)
         {
           continue;
         }
 
+      char ch = c;
```
pkarashchenko, line 456: Looks good, but please move variable declaration to the beginning of the scope to respect C89 compatibility.
The other way is to eliminate `continue;` and put the whole code block under `if (c >= 0)` condition

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #2027, opened 2023-09-04
Fixed the error that when the driver scan result is empty, iwe_stream…

wangyingdong1:
## Summary
 Fixed the error that when the driver scan result is empty, iwe_stream->len is a random number because the user data buf is not initialized.
## Impact
minor

### DIFF
--- a/wireless/wapi/src/wireless.c
+++ b/wireless/wapi/src/wireless.c
@@ -267,14 +267,15 @@ static int wapi_event_stream_extract(FAR struct wapi_event_stream_s *stream,
   int ret = 1;
   FAR struct iw_event *iwe_stream;
 
-  if (stream->current + offsetof(struct iw_event, u) > stream->end)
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

jerpelea (approved):

pkarashchenko (approved):

wireless/wapi/src/wireless.c
```diff
@@ -1337,16 +1339,16 @@ int wapi_scan_coll(int sock, FAR const char *ifname,
       FAR char *tmp;
 
       buflen *= 2;
-      tmp = realloc(buf, buflen);
+      tmp = malloc(buflen);
+      free(buf);
```
pkarashchenko, line 1343: Do we need to free only if allocation succeededs or unconditionally?

xiaoxiang781216 (approved):

xiaoxiang781216:
let's ignore the ci false failure.

PR #2039, opened 2023-09-07
codecs: add md5_file API

0528Mike:
## Summary

Add the API for 'md5_file' to directly obtain the digest value by passing in the file path

## Impact
None
## Testing
None

### DIFF
--- a/netutils/codecs/md5.c
+++ b/netutils/codecs/md5.c
@@ -64,6 +64,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdint.h>
+#include <fcntl.h>
 
 #include "netutils/md5.h"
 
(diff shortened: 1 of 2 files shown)


netutils/codecs/md5.c
```diff
@@ -406,4 +411,59 @@ char *md5_hash(const uint8_t * addr, const size_t len)
...
+ * Description:
+ *   MD5 hash for a file
+ *
+ * Input Parameters:
+ *   path: File Path
+ *   mac : Buffer for the hash
+ *
+ ****************************************************************************/
+
+int md5_file(const char *path, uint8_t *mac)
```
pkarashchenko, line 426: General comment: Need to add `FAR` to pointers

pkarashchenko (approved):

jerpelea (approved):

netutils/codecs/md5.c
xiaoxiang781216: @pkarashchenko The whole code base doesn't add FAR, so it's better to put into the new patch.

pkarashchenko: Yes. Sure. It is more like a "note" comment

PR #2058, opened 2023-09-12
wamr: Generate libc binding from NuttX

no1wudi:
## Summary
Bypass libc and other system calls into NuttX directly, similar to WAMR's builtin libc but cover more interface, must works with --bounds-checks=0 now, so it's more like  an alternative to ELF (with addition XIP support from WAMR).

Compared to ELF, this solution has better footprint (don't need to load entire ELF into RAM), and can take safety feature from Wasm in future (support boundary checks).
## Impact
New feature
## Testing
Vela

### DIFF
--- a/interpreters/wamr/Kconfig
+++ b/interpreters/wamr/Kconfig
@@ -83,6 +83,15 @@ config INTERPRETERS_WAMR_LIBC_WASI
 		not implemented. (Mainly because of lack of openat family of
 		the API in NuttX.)
 
+config INTERPRETERS_WAMR_LIBC_NUTTX
+	bool "Bypass to NuttX libc"
+	default n
+
+config INTERPRETERS_WAMR_LIBC_NUTTX_POSIXMEMALIGN_MAP_SIZE
+	int "Map size for posix_memalign"
+	default 64
+	depends on INTERPRETERS_WAMR_LIBC_NUTTX
+
 config INTERPRETERS_WAMR_MULTI_MODULE
 	bool "Enable multi module support"
 	default n
--- a/interpreters/wamr/wamr_custom_init.c
+++ b/interpreters/wamr/wamr_custom_init.c
@@ -1,0 +1,657 @@
+/****************************************************************************
+ * apps/interpreters/wamr/wamr_custom_init.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <semaphore.h>
+#include <sys/param.h>
+#include <sys/types.h>
+#include <iconv.h>
+
+#include "wasm_export.h"
+#include "wasm_native.h"
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static sem_t g_aligned_memory_map_sem = SEM_INITIALIZER(1);
+static uintptr_t g_aligned_memory_map[CONFIG_INTERPRETERS_WAMR_LIBC_NUTTX_POSIXMEMALIGN_MAP_SIZE][2] = {{0, 0}};
+
+static bool
+add_to_aligned_map(uintptr_t mapped, uintptr_t raw)
+{
+  int i;
+  bool ret = false;
+
+  sem_wait(&g_aligned_memory_map_sem);
+
+  for (i = 0; i < CONFIG_INTERPRETERS_WAMR_LIBC_NUTTX_POSIXMEMALIGN_MAP_SIZE; i++)
+    {
+      DEBUGASSERT(g_aligned_memory_map[i][0] != mapped);
+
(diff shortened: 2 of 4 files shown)


yamt:
>  can take safety feature from Wasm in future (support boundary checks).

how?

no1wudi:
> > can take safety feature from Wasm in future (support boundary checks).
> 
> how?

Just as WAMR's builtin libc does, do mapping or copy on need (only possible for partial interfaces ).

yamt:
> > > can take safety feature from Wasm in future (support boundary checks).
> > 
> > 
> > how?
> 
> Just as WAMR's builtin libc does, do mapping or copy on need (only possible for partial interfaces ).

are you going to write wrappers on many of API functions manually?

interpreters/wamr/wamr_custom_init.c
```diff
@@ -0,0 +1,657 @@
...
+#define GLUE_FUNCTION_qsort
+void glue_qsort(wasm_exec_env_t env, uintptr_t parm1, uintptr_t parm2,
+                uintptr_t parm3, uintptr_t parm4)
+{
+  wasm_module_inst_t module_inst = get_module_inst(env);
+  pthread_mutex_lock(&g_compare_mutex);
+  g_compare_env = env;
+  g_compare_func = parm4;
+  qsort((FAR void *)parm1, (size_t)parm2,
+        (size_t)parm3, compare_proxy);
```
yamt, line 285: this seems broken as the user callback might not return. or it might even call qsort.

interpreters/wamr/wamr_custom_init.c
```diff
@@ -0,0 +1,657 @@
...
+    {
+      hdr->msg_iov[i].iov_base =
+        addr_native_to_app((uintptr_t)hdr->msg_iov[i].iov_base);
+    }
+
+  hdr->msg_iov = addr_native_to_app((uintptr_t)hdr->msg_iov);
+
+  if (hdr->msg_name != NULL && hdr->msg_namelen > 0)
+    {
+      hdr->msg_name = addr_native_to_app(hdr->msg_name);
```
yamt, line 351: are you assuming host pointer size?

tools/mkwamrglue.py
```diff
@@ -0,0 +1,365 @@
...
+HEADER_INDEX = 1
+COND_INDEX = 2
+RETTYPE_INDEX = 3
+PARM1_INDEX = 4
+
+POINTER_SIGNATURE = ['...', '*', '_sa_handler_t', ',mbstate_t']
+
+BLACK_LIST = ['sigqueue', 'pthread_create', 'pthread_detach',
+              'pthread_key_create', 'pthread_key_delete', 'pthread_getspecific',
+              'pthread_setspecific', 'mallinfo', 'inet_ntoa']
```
yamt, line 36: i guess it's simpler to use whitelist as only very simple things can work.

interpreters/wamr/wamr_custom_init.c
```diff
@@ -0,0 +1,657 @@
...
+#define GLUE_FUNCTION_posix_memalign
+uintptr_t glue_posix_memalign(wasm_exec_env_t env, uintptr_t parm1, uintptr_t parm2, uintptr_t parm3)
+{
+  uintptr_t rawptr;
+  wasm_module_inst_t module_inst = get_module_inst(env);
+
+  /* Extra size for align */
+
+  parm3 += parm2;
+  int ret = posix_memalign((FAR void **)parm1, (size_t)parm2, (size_t)parm3);
```
yamt, line 463: doesn't this return an aligned pointer?

interpreters/wamr/wamr_custom_init.c
no1wudi: It might be a issue since the host pointer maybe 8 byte size, the pointer size is 4 byte, that means on 64 bit platform, the offset (host pointer - start address of linear memory of wasm) may exceed the 32 bit len.

interpreters/wamr/wamr_custom_init.c
no1wudi: posix_memalign return an aligned pointer, but it may not an aligned value after convert to be offset of linear memory.

tools/mkwamrglue.py
no1wudi: Maybe ture, there are many hand writing wrappers now, but we are trying to do it by tools.

no1wudi:
> > > > can take safety feature from Wasm in future (support boundary checks).
> > > 
> > > 
> > > how?
> > 
> > 
> > Just as WAMR's builtin libc does, do mapping or copy on need (only possible for partial interfaces ).
> 
> are you going to write wrappers on many of API functions manually?

I prefer to do it by script except some corner case if possible.

interpreters/wamr/Kconfig
```diff
@@ -83,6 +83,15 @@ config INTERPRETERS_WAMR_LIBC_WASI
 		not implemented. (Mainly because of lack of openat family of
 		the API in NuttX.)
 
+config INTERPRETERS_WAMR_LIBC_NUTTX
+	bool "Bypass to NuttX libc"
+	default n
```
acassis, line 88: Please include --help with more information: 1) Why to bypass libc? 2) Scenarios where it could be useful, etc

interpreters/wamr/wamr_custom_init.c
```diff
@@ -0,0 +1,657 @@
...
+
+#include "wasm_export.h"
+#include "wasm_native.h"
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static sem_t g_aligned_memory_map_sem = SEM_INITIALIZER(1);
+static uintptr_t g_aligned_memory_map[CONFIG_INTERPRETERS_WAMR_LIBC_NUTTX_POSIXMEMALIGN_MAP_SIZE][2] = {{0, 0}};
```
acassis, line 42: Please fix the coding style issues

interpreters/wamr/wamr_custom_init.c
```diff
@@ -0,0 +1,657 @@
...
+  uintptr_t ret = 0;
+
+  if (mapped == 0)
+    {
+      return ret;
+    }
+
+  sem_wait(&g_aligned_memory_map_sem);
+
+  for (i = 0; i < CONFIG_INTERPRETERS_WAMR_LIBC_NUTTX_POSIXMEMALIGN_MAP_SIZE; i++)
```
acassis, line 82: Please try to reduce the symbol name, i.e.: just use MEMALIGN instead of POSIXMEMALIGN

no1wudi:
https://github.com/arjunr2/WALI

PR #2064, opened 2023-09-14
lte/alt1250: Fix bug that name resolution failed for a long time

SPRESENSE:
## Summary

Fix bug that name resolution failed for a long time under particular cases. If a DNS query for IPv6 succeeds and IPv4 fails, only the answer of the IPv6 query is saved in the cache. If only IPv4 is given from LTE, name resolution will fail while the answer is saved in the cache. Therefore, the ALT1250 daemon sets the given IP type to the DNS client, which can only send DNS queries for that IP type.

## Impact
Only alt1250 daemon.

## Testing


SPRESENSE:
This pull request has a dependency with the following pull request, please merge the nuttx side pull request first.
https://github.com/apache/nuttx/pull/10622

xiaoxiang781216 (approved):

pkarashchenko (approved):

pkarashchenko:
`dns_set_queryfamily` is missing. Most probably there is some dependency

xiaoxiang781216:
> `dns_set_queryfamily` is missing. Most probably there is some dependency

it's here: https://github.com/apache/nuttx/pull/10622

Issue #2046, opened 2023-09-08
tools/Wasm.mk problems


yamt:
* the idea to build wasm modules using nuttx headers is broken
* it produces very WAMR specific modules (it seems targeting WAMR's LIBC_BUILTIN). if we want to build wasm modules, it's better to target wasm32-wasi.
* WCC detection is broken. for example, github ubuntu-22.04 runners have /bin/clang, which confuses the detection.
* CONFIG_INTERPRETERS_WASM vs CONFIG_INTERPRETERS_WASM3 typo

my proposal: remove it

an alternative:
* target it to wasm32-wasi
* have an explicit config to enable it (disable it by default)

anchao:
@yamt 
We are not using the implementation of wasi/wasi-libc. We have a python script to parse nuttx csv files([syscall](https://github.com/apache/nuttx/blob/master/syscall/syscall.csv)/[libc](https://github.com/apache/nuttx/blob/master/libs/libc/libc.csv)/[libm](https://github.com/apache/nuttx/blob/master/libs/libm/libm.csv)) to automatically generate nuttx syscall/libc/libm glue codes. Currently, this feature works as expected without boundary checking in wamr, but it still needs some work if enable boundary check.

yamt:
> @yamt We are not using the implementation of wasi/wasi-libc. We have a python script to parse nuttx csv files([syscall](https://github.com/apache/nuttx/blob/master/syscall/syscall.csv)/[libc](https://github.com/apache/nuttx/blob/master/libs/libc/libc.csv)/[libm](https://github.com/apache/nuttx/blob/master/libs/libm/libm.csv)) to automatically generate nuttx syscall/libc/libm glue codes. Currently, this feature works as expected without boundary checking in wamr, but it still needs some work if enable boundary check.

where can i find the python script?

anchao:
> where can i find the python script?

I will upload a draft ASAP

yamt:
why do you prefer the "no sandbox" approach?
doesn't it spoil many of benefits of wasm?

no1wudi:
@yamt Please refer to this PR https://github.com/apache/nuttx-apps/pull/2058

PR #2065, opened 2023-09-15
Wasm.mk: disable by default

yamt:
## Summary
Fixes partly: https://github.com/apache/nuttx-apps/issues/2046

### DIFF
--- a/interpreters/wamr/Kconfig
+++ b/interpreters/wamr/Kconfig
@@ -75,6 +75,11 @@ config INTERPRETERS_WAMR_LIBC_BUILTIN
 	bool "Enable built-in libc"
 	default n
 
+config INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX
+	bool "Build WASM modules for NuttX system interface (Experimental)"
+	default n
+	depends on INTERPRETERS_WAMR_LIBC_BUILTIN
+
 config INTERPRETERS_WAMR_LIBC_WASI
 	bool "Enable WASI libc"
 	select PSEUDOFS_SOFTLINKS
--- a/tools/Wasm.mk
+++ b/tools/Wasm.mk
@@ -19,9 +19,8 @@
 #
 ############################################################################
 
-# Only build wasm if one of the following runtime is enabled
 
-ifneq ($(CONFIG_INTERPRETERS_WAMR)$(CONFIG_INTERPRETERS_WASM)$(CONFIG_INTERPRETERS_TOYWASM),)
+ifeq ($(CONFIG_INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX),y)
 include $(APPDIR)$(DELIM)tools$(DELIM)WASI-SDK.defs
 include $(APPDIR)$(DELIM)interpreters$(DELIM)wamr$(DELIM)Toolchain.defs
 
@@ -113,4 +112,4 @@ clean::
 
 endif # WASM_BUILD
 
-endif # CONFIG_INTERPRETERS_WAMR || CONFIG_INTERPRETERS_WASM || CONFIG_INTERPRETERS_TOYWASM
+endif


interpreters/wamr/Kconfig
```diff
@@ -70,6 +70,11 @@ config INTERPRETERS_WAMR_LIBC_BUILTIN
 	bool "Enable built-in libc"
 	default n
 
+config INTERPRETERS_WAMR_BUILD_WASM_MODULES_FOR_WAMR_LIBC_BUILTIN
```
xiaoxiang781216, line 73:
```suggestion
config INTERPRETERS_WAMR_BUILD_MODULES
```

yamt: i want to make it clear it's about a non-standard abi, rather than wasi.

acassis: @yamt what about: INTERPRETERS_WAMR_BUILD_MODULES_NO_STD_ABI ?

acassis: @yamt ping

yamt: well, what's wrong with the original name? too long?

tools/Wasm.mk
```diff
@@ -19,8 +19,8 @@
 #
 ############################################################################
 
-# Only build wasm if one of the following runtime is enabled
 
+ifeq ($(CONFIG_INTERPRETERS_WAMR_BUILD_WASM_MODULES_FOR_WAMR_LIBC_BUILTIN),y)
 ifneq ($(CONFIG_INTERPRETERS_WAMR)$(CONFIG_INTERPRETERS_WASM)$(CONFIG_INTERPRETERS_TOYWASM),)
```
xiaoxiang781216, line 24: should we remove this check

interpreters/wamr/Kconfig
xiaoxiang781216: WASM_BUILD is defined to control whether to build wasm module, why no change the default value of WASM_BUILD to n directly?

yamt: WASM_BUILD is supposed to be set by each apps, isn't it?
eg. https://github.com/apache/nuttx-apps/blob/ba075747be4bbd8b9a2a2f86a02593d4a5a40413/benchmarks/coremark/Makefile#L135
i'm not sure what you mean by "change the default value of WASM_BUILD to n directly".

tools/Wasm.mk
yamt: done

interpreters/wamr/Kconfig
xiaoxiang781216: > WASM_BUILD is supposed to be set by each apps, isn't it? eg.
> 

Yes.

> https://github.com/apache/nuttx-apps/blob/ba075747be4bbd8b9a2a2f86a02593d4a5a40413/benchmarks/coremark/Makefile#L135
> 
> 
> i'm not sure what you mean by "change the default value of WASM_BUILD to n directly".

The build system doesn't generate wasm module unless the user set WASM_BUILD=y in Makefile explicitly. So WASM_BUILD support the same functionality of INTERPRETERS_WAMR_BUILD_WASM_MODULES_FOR_WAMR_LIBC_BUILTIN.

yamt: > > https://github.com/apache/nuttx-apps/blob/ba075747be4bbd8b9a2a2f86a02593d4a5a40413/benchmarks/coremark/Makefile#L135
> > 
> > i'm not sure what you mean by "change the default value of WASM_BUILD to n directly".
> 
> The build system doesn't generate wasm module unless the user set WASM_BUILD=y in Makefile explicitly. So WASM_BUILD support the same functionality of INTERPRETERS_WAMR_BUILD_WASM_MODULES_FOR_WAMR_LIBC_BUILTIN.

do you mean to remove hardcoded WASM_BUILD from coremark etc?

xiaoxiang781216: Yes. It's unsuitable to set WASM_BUILD to both in the common config. It's better to add a new coremark specific option to build wasm version.

yamt: i agree it should be able to control wasm build on per-app basis.

on the other hand, my goal here is a bit different.
i want to disable the whole these wasm build stuff because it's causing problems.

xiaoxiang781216: but if WASM_BUILD is disabled by default, why do it make trouble to you? It's redundant to request people to enable both INTERPRETERS_WAMR_BUILD_WASM_MODULES_FOR_WAMR_LIBC_BUILTIN and xxx_WASM_MODULE for wasm module.

yamt: Well, if someone volunteer to do it in a way it fixes #2046, I probably don't object.
But I myself don't volunteer at this point. To me, it isn't obvious how the LINK_WASM part can be conditional on per app config.

Otherwise, please consider merging this PR for now to stop interfering people who is not even interested in this feature.
An extra step to enable it can be considered a good thing as this is a very experimental feature.

xiaoxiang781216: @no1wudi could you take a look?

no1wudi: I will do the further imporvement but as @yamt mentioned, a global configuration to control the behavior is OK, just like `up_enable_irq` and `up_irq_enable`.

yamt:
do anyone still have any concerns?
it was a bit surprising me that it was this controversial to disable a very experimental feature which is breaking unrelated workloads.

no1wudi:
@yamt I'm sorry to cause you trouble, and I think it's reasonable to add an option to control this behavior.
However, building Wasm applications with NuttX headers in the NuttX build system is a general design rather than just for WAMR. It's just that WAMR was the first to implement these interfaces, so calling the libc here as the NuttX libc would be more appropriate, rather than the built-in libc of WAMR.

Perhaps we could consider renaming the option to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC.

no1wudi:
@xiaoxiang781216 I think we should land this patch first because there are indeed users who use WAMR on NuttX, but they don't need to build applications from within NuttX. This way, the current behavior of compiling certain applications (such as coremark) into Wasm modules by default when the Wasm runtime is enabled will cause unnecessary trouble for them.

yamt:
> Perhaps we could consider renaming the option to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC.

i have no problem with the name.

xiaoxiang781216:
> > Perhaps we could consider renaming the option to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC.
> 
> i have no problem with the name.

it's a pure nuttx option, why need append _FOR_NUTTX_LIBC? INTERPRETERS_WAMR_BUILD_MODULES is enough to represent the intent.

no1wudi:
> it's a pure nuttx option, why need append _FOR_NUTTX_LIBC? INTERPRETERS_WAMR_BUILD_MODULES is enough to represent the intent.

In fact, these modules from NuttX's build system shared the same difinition with native NuttX (such as sys/types.h and fcntl.h), they can only works with NuttX now even some function signature is compatible with WAMR's builtin libc.

yamt:
> > > Perhaps we could consider renaming the option to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC.
> > 
> > 
> > i have no problem with the name.
> 
> it's a pure nuttx option, why need append _FOR_NUTTX_LIBC? INTERPRETERS_WAMR_BUILD_MODULES is enough to represent the intent.

wasm modules are often expected more "run anywhere" than eg. ELF objects.
there is even a standard (well, at least de-facto) system interface available, wasi.
just INTERPRETERS_WAMR_BUILD_MODULES gives me a false impression that it uses wasi.

xiaoxiang781216:
> > > > Perhaps we could consider renaming the option to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC.
> > > 
> > > 
> > > i have no problem with the name.
> > 
> > 
> > it's a pure nuttx option, why need append _FOR_NUTTX_LIBC? INTERPRETERS_WAMR_BUILD_MODULES is enough to represent the intent.
> 
> wasm modules are often expected more "run anywhere" than eg. ELF objects. there is even a standard (well, at least de-facto) system interface available, wasi. just INTERPRETERS_WAMR_BUILD_MODULES gives me a false impression that it uses wasi.

Ok, but it's better to remove _LIBC suffix from the option since many non libc api can expose too.

yamt:
> > > > > Perhaps we could consider renaming the option to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC.
> > > > 
> > > > 
> > > > i have no problem with the name.
> > > 
> > > 
> > > it's a pure nuttx option, why need append _FOR_NUTTX_LIBC? INTERPRETERS_WAMR_BUILD_MODULES is enough to represent the intent.
> > 
> > 
> > wasm modules are often expected more "run anywhere" than eg. ELF objects. there is even a standard (well, at least de-facto) system interface available, wasi. just INTERPRETERS_WAMR_BUILD_MODULES gives me a false impression that it uses wasi.
> 
> Ok, but it's better to remove _LIBC suffix from the option since many non libc api can expose too.

ok, it makes sense.

xiaoxiang781216:
@yamt please change INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX, thanks.

yamt:
> @yamt please change INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX, thanks.

done

yamt:
> > > @yamt please change INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX_LIBC to INTERPRETERS_WAMR_BUILD_MODULES_FOR_NUTTX, thanks.
> > 
> > 
> > done
> 
> but could you sqaush into one patch?

done

xiaoxiang781216 (approved):

PR #2072, opened 2023-09-18
ltp/Makefile: remove some blackwords to enable more testcases

extinguish:
## Summary
remove the blackwords related case that already verified in nuttx


pkarashchenko (approved):

extinguish:
after remove the blackwords, some internal ltp testcases have some build error and link error, need modified either

PR #2085, opened 2023-09-22
Fix osperf running error in SMP mode

Gary-Hobson:
## Summary
This issue (https://github.com/apache/nuttx/issues/10754) found that there are some problems using osperf on qemu-armv8a

In SMP mode, all threads must be thread synchronized and cannot rely on priority for synchronization.


TaiJuWu:
Thanks a lot!!

xiaoxiang781216 (approved):

PR #2118, opened 2023-10-15
Add stepper example

leducp:
## Summary
Add a stepper example.

### DIFF
--- a/examples/stepper/CMakeLists.txt
+++ b/examples/stepper/CMakeLists.txt
@@ -1,0 +1,31 @@
+# ##############################################################################
+# apps/examples/stepper/CMakeLists.txt
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  The ASF licenses this
+# file to you under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.  You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+# ##############################################################################
+
+if(CONFIG_EXAMPLES_STEPPER)
+  nuttx_add_application(
+    NAME
+    stepper
+    STACKSIZE
+    ${CONFIG_DEFAULT_TASK_STACKSIZE}
+    MODULE
+    ${CONFIG_EXAMPLES_STEPPER}
+    SRCS
+    stepper.c)
+endif()
--- a/examples/stepper/Kconfig
+++ b/examples/stepper/Kconfig
@@ -1,0 +1,10 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config EXAMPLES_STEPPER
+	tristate "Stepper controller test"
+	default n
+	---help---
+		Enable the stepper controller test
--- a/examples/stepper/stepper.c
+++ b/examples/stepper/stepper.c
@@ -1,0 +1,163 @@
+/****************************************************************************
+ * apps/examples/stepper/stepper.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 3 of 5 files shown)


examples/stepper/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+# ##############################################################################
+
+if(CONFIG_EXAMPLES_HELLOXX)
```
xiaoxiang781216, line 21:
```suggestion
if(CONFIG_EXAMPLES_STEPPER)
```

examples/stepper/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+  nuttx_add_application(
+    NAME
+    stepper
+    STACKSIZE
+    ${CONFIG_DEFAULT_TASK_STACKSIZE}
+    MODULE
+    ${CONFIG_EXAMPLES_STEPPER}
+    SRCS
+    stepper_main.c)
```
xiaoxiang781216, line 30:
```suggestion
    stepper.c)
```

examples/stepper/stepper.c
```diff
@@ -0,0 +1,144 @@
+//***************************************************************************
```
xiaoxiang781216, line 1: change ALL // to c comment

examples/stepper/Kconfig
```diff
@@ -0,0 +1,11 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config EXAMPLES_STEPPER
+	tristate "Stepper controller test"
+	default n
+	depends on HAVE_CXX
```
xiaoxiang781216, line 9: remove

acassis:
@leducp I think it is a good idea to move this example to apps/industrial as there is already the foc/ example. Categorizing the example makes things easier to users.

examples/stepper/stepper.c
```diff
@@ -0,0 +1,151 @@
...
+  if (argc < 3)
+    {
+      return -1;
+    }
+
+  args->path = argv[1];
+  args->microstep = -1;
+  args->speed = 200;
+
+  for (int i = 2; i < argc; ++i)
```
acassis, line 56: Please use C89, don't declare "i" in the for loop, declare it at begging of the function.

examples/stepper/stepper.c
```diff
@@ -0,0 +1,151 @@
...
+
+  /* Retrieve the absolute position since boot */
+
+  rc = read(fd, &status, sizeof(struct stepper_status_s));
+  if (rc != sizeof(struct stepper_status_s))
+    {
+      printf("read: %s\n", strerror(errno));
+    }
+
+  printf("Current position: %ld\n", status.position);
```
acassis, line 148: Maybe it could be interesting to print the: "Position before stepper motor moving: " and "Position after stepper motor moving: " for diagnostic purposes.

raiden00pl:
> @leducp I think it is a good idea to move this example to apps/industrial as there is already the foc/ example. Categorizing the example makes things easier to users.

that's not true :) `industrial/foc` is FOC library that can be used to build custom FOC applications. FOC example is still in `example/foc`

acassis:
> > @leducp I think it is a good idea to move this example to apps/industrial as there is already the foc/ example. Categorizing the example makes things easier to users.
> 
> that's not true :) `industrial/foc` is FOC library that can be used to build custom FOC applications. FOC example is still in `example/foc`

Thank you for correction @raiden00pl
In fact only the libs are there, the examples still in apps/examples. Maybe we could separate examples by type.

raiden00pl:
> Thank you for correction @raiden00pl
In fact only the libs are there, the examples still in apps/examples. Maybe we could separate examples by type.

good idea.  `examples/foc` is quite a large app now (and will be bigger) and it's more like the reference design of the FOC motor controller based on NuttX than just a simple example. So it should be moved somewhere else anyway. When I find a moment, I'll create and issue to discuss the problem.

examples/stepper/stepper.c
leducp: My bad, old habits

xiaoxiang781216:
@leducp please rebase your change, the ci break is fixed by: https://github.com/apache/nuttx-apps/pull/2131

xiaoxiang781216:
please rebase your patch, ci break is fixed with the recent patch.

leducp:
> please rebase your patch, ci break is fixed with the recent patch.

done :)

PR #2131, opened 2023-10-16
crypto/mbedtls: Unified configuration with mbedtls default configuration

ThePassionate:
(1)The configuration of Kconfig used by the nuttx before is not consistent with the mbedtls default configuration. This patch update configuration is consistent with the default configuration.
(2)And MBEDTLS_BLOWFISH_C/MBEDTLS_ARC4_C/MBEDTLS_SSL_DTLS_BADMAC_LIMIT/MBEDTLS_XTEA_C were removed in mbedtls-3.4.0

depends on patch https://github.com/apache/nuttx/pull/10930
Signed-off-by: makejian <makejian@xiaomi.com>


xiaoxiang781216 (approved):

crypto/mbedtls/Kconfig
```diff
@@ -17,7 +17,7 @@ config MBEDTLS_VERSION
 
 config MBEDTLS_DEBUG_C
 	bool "This module provides debugging functions."
-	default n
+	default y
```
acassis, line 20: I think the debugging functions should be enabled by default because it could increase the final firmware size if the user didn't see it is enabled. Maybe you could use "default y if DEBUG_FEATURES"

crypto/mbedtls/Kconfig
```diff
@@ -17,7 +17,7 @@ config MBEDTLS_VERSION
 
 config MBEDTLS_DEBUG_C
 	bool "This module provides debugging functions."
-	default n
+	default y if DEBUG_FEATURES
```
xiaoxiang781216, line 20:
```suggestion
	default DEBUG_FEATURES
```

xiaoxiang781216 (approved):

PR #2136, opened 2023-10-18
examples/foc: add phase angle observer support (sensorless mode)

raiden00pl:
## Summary

- examples/foc: add phase angle observer support (sensorless mode)
    This app can work now as sensorless ESC.
    Also introduce a cmd line option that force open-loop control
    which is useful when tuning angle/velocity observers

## Testing
ESC based on b-g431b-esc


raiden00pl:
needs https://github.com/apache/nuttx/pull/10972 to pass sim/foc

xiaoxiang781216 (approved):

PR #2138, opened 2023-10-18
Application.mk:resolve file conflicts with the same name when batching archives

xuxin930:
## Summary

This is a workaround to solve the problem of not being able to use `ar rcsP`

when there are too many target files in archive and need to be processed in batches,
in order to avoid file replacement due to duplicate names,
we dynamically rename the target files here.

```makefile
OBJS := SAME_NAME.foo.bar.o ...... SAME_NAME.foo.bar.o

# After call SPLIT splitting the variable, the .o file with the same name will be replaced in the archive
# therefore, new AROBJS rules are added and dynamically generated

AROBJS := SAME_NAME.foo.bar_1.o ...... SAME_NAME.foo.bar_2.o

#  For instance: 
#  $(AROBJS) : $(OBJS)
#        cp $< $@


.built : $(AROBJS)
``` 

## Impact
relate to (merge before):  https://github.com/apache/nuttx/pull/10973

## Testing
CI build

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -205,9 +205,10 @@ $(ZIGOBJS): %$(ZIGEXT)$(SUFFIX)$(OBJEXT): %$(ZIGEXT)
 	$(if $(and $(CONFIG_BUILD_LOADABLE), $(CELFFLAGS)), \
 		$(call ELFCOMPILEZIG, $<, $@), $(call COMPILEZIG, $<, $@))
 
-AROBJS := 
+AROBJS :=
 ifneq ($(OBJS),)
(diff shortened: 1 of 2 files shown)


Application.mk
```diff
@@ -298,6 +311,7 @@ depend:: .depend
 clean::
 	$(call DELFILE, .built)
 	$(call CLEAN)
+	$(Q) rm -f $(AROBJS)
```
anchao, line 314: $(call DELFILE, $(AROBJS))

xuxin930: > `$(call DELFILE, $(AROBJS))`

Done

anchao (approved):
LGTM

PR #2139, opened 2023-10-18
testing: add verbose output of correct/incorrect tests

michallenc:
## Summary
This adds counting of correct/incorrect tests to scanftest and fstest and thus allow to check whether the tests are really
successful. Current CI tests only check whether the test program does on end with hard fault but do not check whether
the results are correct (and they are not). This will allow to check if new patches break some OS functions.

## Impact
Better CI.

## Testing
CI test pass

### DIFF
--- a/testing/fstest/fstest_main.c
+++ b/testing/fstest/fstest_main.c
@@ -89,6 +89,10 @@ struct fstest_ctx_s
   struct mallinfo mmprevious;
   struct mallinfo mmafter;
 };
+
+static int tests_ok = 0;
+static int tests_err = 0;
+
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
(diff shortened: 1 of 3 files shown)


michallenc:
Should be merged before [ci: enhance scanf/fs/os tests to control their correctness](https://github.com/apache/nuttx/pull/10971)

testing/fstest/fstest_main.c
```diff
@@ -1183,6 +1191,11 @@ int main(int argc, FAR char *argv[])
   free(ctx->files);
   free(ctx);
 
+  printf("File system tests done... OK: %d, FAILED: %d\n", tests_ok,
+                                                           tests_err);
+
+  tests_ok = tests_err = 0;
```
xiaoxiang781216, line 1197: move to the start of main, since the program may exit in the middle of main

michallenc: Done

xiaoxiang781216 (approved):

PR #2141, opened 2023-10-19
towasm/libwasi: add config check for 'getrandom' in nuttx

ThePassionate:
check whether getrandom has '/dev/urandom' or '/dev/random' support during code linking
needed by patch https://github.com/apache/nuttx/pull/10925
Signed-off-by: makejian <makejian@xiaomi.com>


xiaoxiang781216:
please rebase your patch, ci break is fixed with the recent patch.

PR #2151, opened 2023-10-23
import/mk: Use libm from the export package when defined.

g2gps:
## Summary

Nuttx exports libm.a when `CONFIG_LIBM` is enabled. Adjust the application import makefile to pick this up when linking.

## Impact

Applications using `CONFIG_BUILD_KERNEL` and `CONIFG_LIBM` can link correctly when needed.

## Testing

Before this change, when trying to add  `CONFIG_EXAMPLES_CALIB_UDELAY`
```
nuttx/apps/examples/calib_udelay/calib_udelay_main.c:282: undefined reference to `ceil'
```

With this change, the `calib_udelay` application links as expected.

### DIFF
--- a/import/Make.defs
+++ b/import/Make.defs
@@ -70,12 +70,14 @@ ifeq ($(CONFIG_BUILD_KERNEL),y)
   ifeq ($(CONFIG_HAVE_CXX),y)
     LDLIBS += -lxx
   endif
-ifneq ($(CONFIG_LIBM),y)
+ifeq ($(CONFIG_LIBM_TOOLCHAIN),y)
   LIBM = ${shell "$(CC)" $(ARCHCPUFLAGS) --print-file-name=libm.a}
 ifneq ($(LIBM),".")
   LDLIBPATH += -L "${shell dirname $(LIBM)}"
   LDLIBS += -lm
 endif
+else ifneq($(CONFIG_LIBM_NONE),)
+  LDLIBS += -lm
 endif
 endif
 


import/Make.defs
```diff
@@ -76,6 +76,8 @@ ifneq ($(LIBM),".")
   LDLIBPATH += -L "${shell dirname $(LIBM)}"
   LDLIBS += -lm
 endif
+else
```
xiaoxiang781216, line 79:
```suggestion
else ifneq ($(CONFIG_LIBM_NONE,)
```
line 73 nee change to `ifeq ($(CONFIG_LIBM_TOOLCHAIN),y)`

xiaoxiang781216 (approved):

hitHuang:
@g2gps @xiaoxiang781216 
I came across this PR while debugging a ceil undefined reference in calib_udelay on a BUILD_KERNEL config with CONFIG_LIBM=y (rv-virt:knsh). It turns out the condition else ifneq ($(CONFIG_LIBM_NONE),) is inverted — it adds -lm only when
CONFIG_LIBM_NONE is set to y, but the intent is the opposite: add -lm for every math library choice except NONE. When CONFIG_LIBM=y, CONFIG_LIBM_NONE is empty so -lm never gets linked. The fix is simply:

-else ifneq ($(CONFIG_LIBM_NONE),)
+else ifneq ($(CONFIG_LIBM_NONE),y)

Am I missing something here?

PR #2171, opened 2023-11-01
Add missing variables and change to be more intuitive to use

GC-20-20:
## Summary
Add missing variables and change to be more intuitive to use
## Impact
Variables that do cause compilation to fail
## Testing
local test

### DIFF
--- a/examples/serialrx/Kconfig
+++ b/examples/serialrx/Kconfig
@@ -39,22 +39,22 @@ config EXAMPLES_SERIALRX_DEVPATH
 
 choice
 	prompt "Output method"
-	default EXAMPLES_SERIALRX_PRINTHYPHEN
+	default EXAMPLES_SERIALRX_PRINTSTR
 
-config EXAMPLES_SERIALRX_PRINTHYPHEN
-	bool "Hyphen"
+config EXAMPLES_SERIALRX_PRINTSTR
+	bool "String"
 	---help---
-		Print a hyphen after receiving a full buffer
+		Print the bytes received as a string
 
 config EXAMPLES_SERIALRX_PRINTHEX
 	bool "Hexadecimal"
 	---help---
 		Print the bytes received in hexadecimal
 
-config EXAMPLES_SERIALRX_PRINTSTR
-	bool "String"
+config EXAMPLES_SERIALRX_PRINTHYPHEN
+	bool "Hyphen"
 	---help---
-		Print the bytes received as a string
+		Print a hyphen after receiving a full buffer
 
 endchoice
 
--- a/examples/serialrx/Kconfig
+++ b/examples/serialrx/Kconfig
@@ -39,22 +39,22 @@ config EXAMPLES_SERIALRX_DEVPATH
 
 choice
 	prompt "Output method"
-	default EXAMPLES_SERIALRX_PRINTSTR
+	default EXAMPLES_SERIALRX_PRINTHYPHEN
 
-config EXAMPLES_SERIALRX_PRINTSTR
-	bool "String"
+config EXAMPLES_SERIALRX_PRINTHYPHEN
+	bool "Hyphen"
 	---help---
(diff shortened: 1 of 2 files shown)


examples/serialrx/Kconfig
```diff
@@ -39,22 +39,22 @@ config EXAMPLES_SERIALRX_DEVPATH
 
 choice
 	prompt "Output method"
-	default EXAMPLES_SERIALRX_PRINTHYPHEN
+	default EXAMPLES_SERIALRX_PRINTSTR
 
-config EXAMPLES_SERIALRX_PRINTHYPHEN
```
xiaoxiang781216, line 44: why change the order, let's keep as before

examples/serialrx/Kconfig
```diff
@@ -39,22 +39,22 @@ config EXAMPLES_SERIALRX_DEVPATH
 
 choice
 	prompt "Output method"
-	default EXAMPLES_SERIALRX_PRINTHYPHEN
+	default EXAMPLES_SERIALRX_PRINTSTR
```
xiaoxiang781216, line 42: why change the default setting

examples/serialrx/serialrx_main.c
```diff
@@ -118,8 +118,7 @@ int main(int argc, FAR char *argv[])
   while (cnt < bytecount)
     {
 #ifdef CONFIG_EXAMPLES_SERIALRX_BUFFERED
-      size_t n = fread(buf, 1, 26, f);
```
xiaoxiang781216, line 121: why make the follow change

examples/serialrx/Kconfig
GC-20-20: When I opened this application for the first time, it was very confusing for me to not have read the source code of the program and not have any display after using another serial port to send data, after all, there is no explanation in the HELP. So I would like to change the default value to EXAMPLES_SERIALRX_PRINTSTR so that later learners can be more intuitively clear that the previous operation is correct.

examples/serialrx/serialrx_main.c
GC-20-20: I'm very sorry, this change is really optional, this 26 I double-checked a few times without finding any clues, then I changed it to a simple number. Maybe it helps.

xiaoxiang781216: Ok, but after you finish the investigation is better to remove the temp change. If you think the temp change is worth to upstream, please split to the different patch. It's important to ensure one patch fix one problem, so the reviewer could understand your change quickly.

GC-20-20: Okay, I see what you mean.I'd still like to ask, but would these small changes be a waste of your time.After all, it doesn't change the logic of the code

xiaoxiang781216: You can have one PR which contain more than one patch

jerpelea (changes requested):
a commit message would be nice

examples/serialrx/serialrx_main.c
```diff
@@ -173,7 +172,6 @@ int main(int argc, FAR char *argv[])
 
       UNUSED(eof);
     }
-
```
xiaoxiang781216, line 176: revert this change

examples/serialrx/serialrx_main.c
```diff
@@ -118,8 +118,7 @@ int main(int argc, FAR char *argv[])
   while (cnt < bytecount)
     {
 #ifdef CONFIG_EXAMPLES_SERIALRX_BUFFERED
-      size_t n = fread(buf, 1, 26, f);
-      cnt++;
+      size_t n = fread(buf, 1, 1, f);
```
xiaoxiang781216, line 121: why not write 26 bytes.

xiaoxiang781216:
please squash your patch and update the commit message to reflect your change.

PR #2178, opened 2023-11-04
Add Settings Utility and Example App

TimJTi:
Last year @fjpanag offered to donate some code to the repo. This is my attempt at making his "settings" functions NuttX compatible along with an example app to check it out
Impact

None. This is new.
Testing

On my custom board (SAMA5D27C-D1G). It seems to work!

There are probably still some style and/or NuttX-convention errors/inconsistencies so I have left it as a draft for now.

### DIFF
--- a/examples/settings/Kconfig
+++ b/examples/settings/Kconfig
@@ -1,0 +1,53 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config EXAMPLES_SETTINGS
+	tristate "Settings example"
+	default n
+	---help---
+		Enable the "settings" example. This uses the functions provided in
+		utils/settings to test the storage and retrieval of various
+		configurable parameters
+
+if EXAMPLES_SETTINGS
+
+config EXAMPLES_SETTINGS_PROGNAME
+	string "Program name"
+	default "settings"
+	---help---
+		This is the name of the program that will be used when the NSH ELF
+		program is installed.
+
+config EXAMPLES_SETTINGS_PRIORITY
+	int "settings task priority"
+	default 100
+
+config EXAMPLES_SETTINGS_STACKSIZE
+	int "settings stack size"
+	default DEFAULT_TASK_STACKSIZE
+
+choice
+	prompt "Select settings storage location"
+	default EXAMPLES_SETTINGS_USE_TMPFS
+
+config EXAMPLES_SETTINGS_USE_TMPFS
+	bool "Use TMPFS"
+	select FS_TMPFS
+	---help---
+		TMPFS will be enabled and used
+
+config EXAMPLES_SETTINGS_USE_OTHER
+	bool "Use existing/other storage media"
+
+endchoice # Select settings storage location
+
+if EXAMPLES_SETTINGS_USE_OTHER
+
+config EXAMPLES_SETTINGS_EXISTING_STORAGE
+	string "Path to existing storage media"
+	default "/mnt/sdcard0"
+
+endif # EXAMPLES_SETTINGS_USE_OTHER
+endif # EXAMPLES_SETTINGS
--- a/examples/settings/settings_main.c
+++ b/examples/settings/settings_main.c
@@ -1,0 +1,275 @@
+/****************************************************************************
+ * apps/examples/settings/settings_main.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdio.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <sys/mount.h>
+
+#include "system/settings.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * settings_main
+ ****************************************************************************/
+
+int main(int argc, FAR char *argv[])
+{
+  int                  ret;
+  int                  fd;
+  enum settings_type_e stype;
+  char                 path[50];
+  char                 *teststr = "";
+  int                  testval = 0x5aa5;
+  char                 readstr[CONFIG_SETTINGS_VALUE_SIZE];
+  enum storage_type_e  storage_type = STORAGE_BINARY;
+  bool                 flag_present = false;
+  struct stat          sbuf;
+
+  if ((argc < 1) || *argv[1] == 0 || *(argv[1] + 1) == 0)
+    {
+      goto print_help;
+    }
+
+  if (*argv[1] == '-')
+    {
+      flag_present = true;
+      if (*(argv[1] + 1) == 'b')
+        {
+          storage_type = STORAGE_BINARY;
+        }
+      else if (*(argv[1] + 1) == 't')
+        {
+          storage_type = STORAGE_TEXT;
+        }
+      else
+        {
+          goto print_help;
+        }
+    }
+
+  if (flag_present && argc < 2)
+    {
+      goto print_help;
+    }
+
+#ifdef CONFIG_EXAMPLES_SETTINGS_USE_TMPFS
+#  ifndef CONFIG_FS_TMPFS
+#    error TMPFS must be enabled
+#  endif
+#  ifndef CONFIG_LIBC_TMPDIR
+#    error LIBC_TMPDIR must be defined
+#  endif
+  if (stat(CONFIG_LIBC_TMPDIR, &sbuf))
+    {
+      ret = nx_mount(NULL, CONFIG_LIBC_TMPDIR, "tmpfs", 0, NULL);
+      if (ret < 0)
+        {
+          printf("ERROR: Failed to mount tmpfs at %s: %d\n",
+                  CONFIG_LIBC_TMPDIR, ret);
+          goto end;
+        }
+    }
+
+  strcpy(path, CONFIG_LIBC_TMPDIR);
+  strcat(path, "/settings");
+#else
+  strcpy(path, CONFIG_EXAMPLES_SETTINGS_EXISTING_STORAGE);
+  if (path == NULL)
+    {
+      printf("Settings filepath is empty!");
+      goto end;
+    }
+
+  strcat(path, "/settings");
+#endif
+
+  printf("Example of settings usage: %s. Path: %s\n",
+         ((storage_type == STORAGE_TEXT) ? "text" : "binary"), path);
+  printf("-------------------------------------------------------\n");
+
+  ret = settings_init();
+  if (ret < 0)
+    {
+      printf("settings init failed: %d", ret);
+      goto end;
+    }
+
+  ret = settings_setstorage(path, storage_type);
+  if (ret == -ENOENT)
+    {
+      printf("INFO: no existing binary settings file found. Creating it.\n");
+      fd = open(path, O_CREAT);
+      if (fd < 0)
+        {
+          printf("failed to create settings file\n");
+          goto end;
+        }
+
+      close(fd);
+    }
+  else if (ret < 0)
+    {
+      printf("settings setstorage failed: %d\n", ret);
+      goto end;
+    }
+  else
+    {
+      printf("INFO: a settings file was found and loaded\n");
+      ret = settings_get("v1", SETTING_STRING, readstr,
+                         CONFIG_SETTINGS_VALUE_SIZE);
+      if (ret < 0)
+        {
+          printf("settings retrieve failed: %d\n", ret);
+          goto end;
+        }
+
+      printf("Retrieved settings value (v1) with value:%s\n",
+              readstr);
+    }
+
+  printf("Creating an integer settings value (v1) with value:0x%x\n",
+          testval);
+  ret = settings_create("v1", SETTING_INT, testval);
+  if (ret < 0)
+    {
+      printf("settings create failed: %d\n", ret);
+      goto end;
+    }
+
+  ret = settings_type("v1", &stype);
+  if (ret < 0)
+    {
+      printf("failed to read settings type: %d\n", ret);
+      goto end;
+    }
+
+  printf("Retrieved setting type is:%d\n", stype);
+
+  ret = settings_get("v1", SETTING_INT, &testval);
+  if (ret < 0)
+    {
+      printf("settings retrieve failed: %d\n", ret);
+      goto end;
+    }
+
+  printf("Retrieved integer settings value (v1) with value:0x%x\n",
+          testval);
+
+  ret = settings_get("v1", SETTING_BOOL, &testval);
+  if (ret < 0)
+    {
+      printf("Deliberate fail: BOOL type requested not INT. Error:%d\n",
+              ret);
+    }
+
+  ret = settings_get("v2", SETTING_INT, &testval);
+  if (ret < 0)
+    {
+      printf("Deliberate fail: non-existent setting requested. Error:%d\n",
+             ret);
+    }
+
+  teststr = "I'm a string!";
+  printf("Changing an integer settings value (v1) to string:%s\n",
+          teststr);
+  ret = settings_create("v1", SETTING_STRING, teststr);
+  if (ret < 0)
+    {
+      printf("settings changed failed: %d\n", ret);
+      goto end;
+    }
+
+  ret = settings_get("v1", SETTING_STRING, &readstr, sizeof(readstr));
+  if (ret < 0)
+    {
+      printf("settings retrieve failed: %d\n", ret);
+      goto end;
+    }
+
+  printf("Retrieved string settings value (v1) with value:%s\n",
+          readstr);
+
+  FAR struct in_addr save_ip;
+  FAR struct in_addr load_ip;
+  save_ip.s_addr = HTONL(0xc0a86401);
+
+  printf("Creating an IP settings value (IP0) with value:0x%08lx\n",
+          save_ip.s_addr);
+  ret = settings_create("IP0", SETTING_IP_ADDR, &save_ip);
+  if (ret < 0)
+    {
+      printf("IP address settings create failed: %d\n", ret);
+      goto end;
+    }
+
+  ret = settings_get("IP0", SETTING_IP_ADDR, &load_ip);
+  if (ret < 0)
+    {
+      printf("IP address settings retrieve failed: %d\n", ret);
+      goto end;
+    }
+
+  printf("Retrieved IP address settings value (IP0) with value:0x%08lx\n",
+          NTOHL(load_ip.s_addr));
+end:
+
+#ifdef CONFIG_SETTINGS_CACHED_SAVES
+  /* saves may not have been written out before we exit */
+
+  sleep(1);
+#endif
+  return ret;
+
+print_help:
+  printf("Usage...\n");
+  printf("settings [-b | -t] \n");
+  printf("    -i = use a binary storage file (default)\n");
+  printf("    -t = use a text   storage file\n");
+  printf(" Example:\n");
+  printf("   settings -b\n");
+  return -EINVAL;
+}
--- a/system/settings/Kconfig
+++ b/system/settings/Kconfig
@@ -1,0 +1,76 @@
+#
+# Settings Storage Configuration
+#
+
+config SYSTEM_SETTINGS
+	bool "Settings Functions"
+	default n
+		---help---
+		The settings storage can be used to store and retrieve various
+		configurable parameters. This storage is a RAM-based map (for fast access)
+		but one or more files can be used too (so values can persist across system
+		reboots).
+
+if SYSTEM_SETTINGS
+
+config SETTINGS_MAP_SIZE
+	int "Settings storage size"
+	default 20
+	---help---
+		The size of the settings storage, i.e. the
+		maximum number of settings that can be created.
+
+config SETTINGS_KEY_SIZE
+	int "Settings key size"
+	default 20
+	---help---
+		Maximum size of a setting key.
+
+config SETTINGS_VALUE_SIZE
+	int "Settings value size"
+	default 20
+	---help---
+		Maximum size of a string setting value.
+
+config SETTINGS_MAX_STORAGES
+	int "Settings files"
+	default 2
+	---help---
+		Maximum number of supported settings files.
+
+config SETTINGS_MAX_FILENAME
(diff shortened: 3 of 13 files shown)


TimJTi:
I have no idea why this is failing CI? HELP!

TimJTi:
OK - I assume it's because a created a new "top level" folder (utils) for the settings functions rather than placing it under some pre-existing folder?

If someone can propose a good folder for it to be re-homed to I can do it, or the new folder needs to be added to the scripts, I think?

xiaoxiang781216:
how about system/ folder? Many utility libraries put there.

TimJTi:
@xiaoxiang781216 It still fails CI - I HAVE NO IDEA!!!!!!!!!!!

raiden00pl:
@TimJTi when CI fails and you don't know what's going on, the most common problem is some error in Kconfig. For me configure this PR shows:

```
CMake Warning at CMakeLists.txt:348 (message):
  Kconfig Configuration Error: warning: the default selection
  EXAMPLES_SETTINGS_CREATE_RAMDISK (undefined) of <choice> (defined at
  /home/raiden00/git/RTOS/nuttx/apps/examples/settings/Kconfig:31) is not
  contained in the choice


```

TimJTi:
> @TimJTi when CI fails and you don't know what's going on, the most common problem is some error in Kconfig. For me configure with this PR shows:
> 
> ```
> CMake Warning at CMakeLists.txt:348 (message):
>   Kconfig Configuration Error: warning: the default selection
>   EXAMPLES_SETTINGS_CREATE_RAMDISK (undefined) of <choice> (defined at
>   /home/raiden00/git/RTOS/nuttx/apps/examples/settings/Kconfig:31) is not
>   contained in the choice
> ```
> 
> You're probably using `kconfig-frontends` which doesn't catch as many errors as `kconfiglib` used by CI

Thank you! I will check.

I do not use Cmake but saw the Cmake files begin to appear everywhere so thought I probably should make sure they were there - but since I don't use them I missed that error.

It's a long time since I set up my Linux build environment but I will certainly look to see whether I use kconfig-frontends or kconfiglib and changeover if I can.

TimJTi:
@raiden00pl Tests are running now - thank you.

I edit Kconfig files in VS Code, and have a syntax checker extension, and have kconfig-frontends as per NuttX install/setup guides for Ubuntu.

I can pip install kconfiglib, but am not sure how to make use of it. Can you share your Kconfig edit/create workflow with me so I can avoid these errors in the future?

I am actually thinking VS Code may not a hendrance not a help - I find a build will fail within VSCode for reasons I can never see, usually to do with apps, and end up having to "force save" a new .confif from make menuconfig and build from a terminal prompt before it all works OK again in VS Code.

May be time for me to jump ship to a different edit/build/debug workflow completely!!!!!!!!!!

raiden00pl:
In my case it was just installing `python-kconfiglib` package for my OS (I'm avoid pip installs when possible). Build system should auto-detect kconfiglib if available.
kconfiglib should make the `menuconfig` command available in your system, so you can check it with `command -v menuconfig` command which should return path to `menuconfig` command:

```
[raiden00:~/git/RTOS/nuttx/nuttx]$ command -v menuconfig                                                                               
/usr/bin/menuconfig
```
If this doesn't work, then probably the packages installed with pip (usually `~/.local/bin`) are not in system `PATH`

TimJTi:
@raiden00pl 
> In my case it was just installing `python-kconfiglib` package for my OS (I'm avoid pip installs when possible). Build system should auto-detect kconfiglib if available. kconfiglib should make the `menuconfig` command available in your system, so you can check it with `command -v menuconfig` command which should return path to `menuconfig` command:
> 

Thank you - installed and working!

system/settings/Kconfig
```diff
@@ -0,0 +1,76 @@
+#
+# Settings Storage Configuration
+#
+
+config SYSTEM_SETTINGS
+  bool "Settings Functions"
```
xiaoxiang781216, line 6: tab

system/settings/storage_bin.c
```diff
@@ -0,0 +1,246 @@
...
+
+  read(fd, &exp_crc, sizeof(uint32_t));
+
+  if (calc_crc != exp_crc)
+    {
+      ret = -EIO;
+      goto abort;
+    }
+
+  lseek(fd, (sizeof(uint16_t) *2), SEEK_SET);  /* Get after valid & size */
```
xiaoxiang781216, line 120:
```suggestion
  lseek(fd, (sizeof(uint16_t) * 2), SEEK_SET);  /* Get after valid & size */
```

system/settings/storage_bin.c
```diff
@@ -0,0 +1,246 @@
...
+  if (valid != VALID)
+    {
+      goto abort; /* Just exit - the settings aren't valid. What to do? */
+    }
+
+  count = 0;
+  read(fd, &count, sizeof(uint16_t));
+
+  uint32_t calc_crc = 0;
+  uint32_t exp_crc = 0;
```
xiaoxiang781216, line 102: move to the beginning of function

system/settings/storage_bin.c
```diff
@@ -0,0 +1,246 @@
...
+ *    file             - the filename of the storage to use
+ *
+ * Returned Value:
+ *   Success or negated failure code
+ *
+ ****************************************************************************/
+
+int load_bin(FAR char *file)
+{
+  static int fd;
```
xiaoxiang781216, line 78:
```suggestion
  int fd;
```

system/settings/storage_bin.c
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+  lseek(fd, (sizeof(uint16_t) *2), SEEK_SET);  /* Get after valid & size */
+
+  for (i = 0; i < count; i++)
+    {
+      setting_t setting;
+      read(fd, &setting, sizeof(setting_t));
+
+      FAR setting_t *slot = getsetting(setting.key);
```
xiaoxiang781216, line 127: move the declaration to the beginning of for loop

system/settings/storage_bin.c
```diff
@@ -0,0 +1,246 @@
...
+    {
+      ret = -ENODEV;
+      goto abort;
+    }
+
+  memset(buffer, 0xff, BUFFER_SIZE);
+  *((FAR uint16_t *)buffer) = VALID;
+  *(((FAR uint16_t *)buffer) + 1) = count;
+
+  off_t offset = sizeof(uint16_t) *2;  /* Valid & count */
```
xiaoxiang781216, line 190:
```suggestion
  off_t offset = sizeof(uint16_t) * 2;  /* Valid & count */
```
move the declaration to the beginning of function, same for data_size, rem_data, crc and rem_crc.

system/settings/storage_bin.c
```diff
@@ -0,0 +1,246 @@
...
+  size_t rem_crc = sizeof(crc);
+
+  while ((offset + rem_data + rem_crc) > 0)
+    {
+      size_t to_write = ((offset + rem_data) > BUFFER_SIZE) ?
+                         (size_t)(BUFFER_SIZE - offset) : rem_data;
+      memcpy((buffer + offset), (((FAR uint8_t *)g_settings.map) +
+             (data_size - rem_data)), to_write);
+
+      size_t j;
```
xiaoxiang781216, line 203: move to the beginning of while oop

system/settings/storage_text.c
```diff
@@ -0,0 +1,379 @@
...
+      free(backup_file);
+    }
+
+  f = fopen(file, "r");
+  if (f == NULL)
+    {
+      return -ENOENT;
+    }
+
+  FAR char *buffer = malloc(BUFFER_SIZE);
```
xiaoxiang781216, line 114: move to the beginning of function

system/settings/storage_text.c
```diff
@@ -0,0 +1,379 @@
...
+    {
+      ret = -ENOMEM;
+      goto abort;
+    }
+
+  while (fgets(buffer, BUFFER_SIZE, f))
+    {
+      /* Remove any line terminators */
+
+      for (int i = ((int)strlen(buffer) - 1); i > 0; i--)
```
xiaoxiang781216, line 125:
```suggestion
      int i;

      for (i = ((int)strlen(buffer) - 1); i > 0; i--)
```

system/settings/storage_text.c
```diff
@@ -0,0 +1,379 @@
...
+            }
+          else
+            {
+              break;
+            }
+        }
+
+      /* Separate the key / value pair */
+
+      FAR char *eq = strchr(buffer, '=');
```
xiaoxiang781216, line 139: move to the beginning of function

system/settings/storage_text.c
```diff
@@ -0,0 +1,379 @@
...
+      if (eq == NULL)
+        {
+          continue;
+        }
+
+      FAR char *key = buffer;
+      FAR char *val = eq + 1;
```
xiaoxiang781216, line 146: move to the beginning of function

system/settings/storage_text.c
```diff
@@ -0,0 +1,379 @@
...
+      /* Check if the key is valid */
+
+      if (!isalpha(key[0]))
+        {
+          continue;
+        }
+
+      /* Get the setting slot */
+
+      FAR setting_t *setting = getsetting(key);
```
xiaoxiang781216, line 158: move to the beginning of function

system/settings/storage_text.c
```diff
@@ -0,0 +1,379 @@
...
+
+  if (backup_file == NULL)
+    {
+      return -ENODEV;
+    }
+
+  strcpy(backup_file, file);
+  strcat(backup_file, "~");
+
+  FAR FILE *f;
```
xiaoxiang781216, line 279: move to the beginning of function

system/settings/settings.c
```diff
@@ -0,0 +1,1432 @@
...
+    {
+      return ret;
+    }
+
+  for (i = 0; i < CONFIG_SETTINGS_MAX_STORAGES; i++)
+    {
+      if ((g_settings.store[i].file[0] != '\0') &&
+           g_settings.store[i].save_fn)
+        {
+          ret = g_settings.store[i].save_fn(g_settings.store[i].file);
```
xiaoxiang781216, line 627: it isn't safe to call fs function in signal handler

system/settings/Kconfig
TimJTi: Every time a set VSCode up to not convert tabs to spaces for Kconfig files, something (VS Code updates maybe) comes along and that setting gets lost. Fixed now, ready for my next commit.

system/settings/storage_bin.c
TimJTi: As above

system/settings/storage_text.c
TimJTi: As above

system/settings/storage_text.c
TimJTi: As above

system/settings/storage_text.c
TimJTi: As above

system/settings/storage_text.c
TimJTi: As above

system/settings/settings.c
TimJTi: @xiaoxiang781216 - OK - will look tomorrow as need a clear mind for it :)

TimJTi: @xiaoxiang781216 - can you give me a little more information on what's not safe, so I can devise the best fix, please?

xiaoxiang781216: here: https://man7.org/linux/man-pages/man7/signal-safety.7.html.

TimJTi: Found that as soon as I replied a few minutes ago so deleted it. Thanks :)

system/settings/storage_bin.c
xiaoxiang781216: no change here.

TimJTi: i missed it, sorry, but what's the reasoning please? So I learn :)

fjpanag:
@TimJTi I would like to review this, if you don't mind, in case any change causes any issue already known to me.
I will try to check the app today, for proper usage of the API, and possibly tomorrow the actual settings code.

@xiaoxiang781216 How can I become a reviewer of the PR? Do I just leave comments inside of the commit page?

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+  enum settings_type_e stype;
+  char                 path[50];
+  char                 *teststr = "";
+  int                  testval = 0x5aa5;
+  char                 readstr[CONFIG_SETTINGS_VALUE_SIZE];
+  enum storage_type_e  storage_type = STORAGE_BINARY;
+  bool                 flag_present = false;
+  struct stat          sbuf;
+
+  ASSERT(0);
```
fjpanag, line 72:
```suggestion

```

Maybe left-over to be removed?

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+  if ((argc < 1) || *argv[1] == 0 || *(argv[1] + 1) == 0)
```
fjpanag, line 73:
```suggestion
  if ((argc > 1) && (*argv[1] == 0 || *(argv[1] + 1) == 0))
```
I think your check here is wrong, for `argc == 1`, you access `argv[1]` which is an overflow.

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+    {
+      goto print_help;
+    }
+
+  if (*argv[1] == '-')
```
fjpanag, line 78:
```suggestion
  if (argc > 1 && *argv[1] == '-')
```
Check first that the argument actually exists?

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+        {
+          storage_type = STORAGE_TEXT;
+        }
+      else
+        {
+          goto print_help;
+        }
+    }
+
+  if (flag_present && argc < 2)
```
fjpanag, line 95: I think that this is an impossible condition.

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+#    error LIBC_TMPDIR must be defined
+#  endif
+  if (stat(CONFIG_LIBC_TMPDIR, &sbuf))
+    {
+      ret = nx_mount(NULL, CONFIG_LIBC_TMPDIR, "tmpfs", 0, NULL);
+      if (ret < 0)
+        {
+          printf("ERROR: Failed to mount tmpfs at %s: %d\n",
+                  CONFIG_LIBC_TMPDIR, ret);
+          goto end;
```
fjpanag, line 114: This will eventually `return OK;`. Maybe it's better to return an error in all these cases?

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+
+  strcpy(path, CONFIG_LIBC_TMPDIR);
+  strcat(path, "/settings");
+#else
+  strcpy(path, CONFIG_EXAMPLES_SETTINGS_EXISTING_STORAGE);
```
fjpanag, line 121: I would greatly prefer to use `char *path` and a `strdup()` instead of hard-coding the size of the path to 50.

If you really want to do it like this, maybe use `strncpy()`? To ensure that a long path will not cause an overflow.

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+  if (path == NULL)
```
fjpanag, line 122: This check is invalid, as `path` will never be `NULL`.  
Either check with `strlen() != 0`, with the code as is, or change above to `strdup()`.

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+
+  ret = settings_init();
+  if (ret < 0)
+    {
+      printf("settings init failed: %d", ret);
+      goto end;
+    }
+
+  ret = settings_setstorage(path, storage_type);
+  if (ret == -ENOENT)
```
fjpanag, line 143: As far as I can tell, there is no `ENOENT` return from `settings_setstorage()`.

And I believe that there shouldn't be. If the file does not exist, settings should create it automatically.

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+  else if (ret < 0)
+    {
+      printf("settings setstorage failed: %d\n", ret);
+      goto end;
+    }
+  else
+    {
+      printf("INFO: a settings file was found and loaded\n");
+      ret = settings_get("v1", SETTING_STRING, readstr,
+                         CONFIG_SETTINGS_VALUE_SIZE);
```
fjpanag, line 164:
```suggestion
                         sizeof(readstr));
```

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+    }
+  else if (ret < 0)
+    {
+      printf("settings setstorage failed: %d\n", ret);
+      goto end;
+    }
+  else
+    {
+      printf("INFO: a settings file was found and loaded\n");
+      ret = settings_get("v1", SETTING_STRING, readstr,
```
fjpanag, line 163: This call should always fail.

You try to access a setting without creating it first.
It **must** be created, even if a value already exists in the file.

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+      goto end;
+    }
+
+  printf("Retrieved integer settings value (v1) with value:0x%x\n",
+          testval);
+
+  ret = settings_get("v1", SETTING_BOOL, &testval);
+  if (ret < 0)
+    {
+      printf("Deliberate fail: BOOL type requested not INT. Error:%d\n",
```
fjpanag, line 206: No, this is not a failure.

The integer setting should be automatically converted to a boolean.  
This should always succeed.

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+  if (ret < 0)
+    {
+      printf("Deliberate fail: non-existent setting requested. Error:%d\n",
+             ret);
+    }
+
+  teststr = "I'm a string!";
+  printf("Changing an integer settings value (v1) to string:%s\n",
+          teststr);
+  ret = settings_create("v1", SETTING_STRING, teststr);
```
fjpanag, line 220: This should always fail. You are trying to create a setting that already exists!

If you want to change its type, you should be able to just `settings_set()` it to the new type.

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+  if (ret < 0)
+    {
+      printf("settings changed failed: %d\n", ret);
+      goto end;
+    }
+
+  ret = settings_get("v1", SETTING_STRING, &readstr,
+                     CONFIG_SETTINGS_VALUE_SIZE);
```
fjpanag, line 228:
```suggestion
                     sizeof(readstr));
```

examples/settings/settings_main.c
```diff
@@ -0,0 +1,276 @@
...
+  save_ip.s_addr = HTONL(0xc0a86401);
+  ret = settings_create("IP0", SETTING_IP_ADDR, &save_ip);
+  if (ret < 0)
+    {
+      printf("IP address settings create failed: %d\n", ret);
+      goto end;
+    }
+
+  printf("Creating an IP settings value (IP0) with value:0x%08lx\n",
+          save_ip.s_addr);
```
fjpanag, line 249:
```suggestion

  printf("Creating an IP settings value (IP0) with value:0x%08lx\n",
          save_ip.s_addr);

  FAR struct in_addr save_ip;
  FAR struct in_addr load_ip;
  save_ip.s_addr = HTONL(0xc0a86401);
  ret = settings_create("IP0", SETTING_IP_ADDR, &save_ip);
  if (ret < 0)
    {
      printf("IP address settings create failed: %d\n", ret);
      goto end;
    }

```

Maybe  move the print before actually attempting the call?

fjpanag (changes requested):

examples/settings/settings_main.c
TimJTi: Absolutely!! Oops!!! Was yesterday evening when I was testing/trying stuff"!

system/settings/settings.c
TimJTi: @xiaoxiang781216 If I have understood this right:

1. Calls to the save function are obviously asynchronous
2. Each call to the save function sets a a new timer
3. That timer could trigger before the processing of a previous timer has completed. And that potentially violates the signal safety concern?

I have proven this to be an issue, to satisfy my curiosity by adding a sleep into the the save function and calling it a few times before the function awakens. It crashes, so - I hope - I've understood it?

@fjpanag Thoughts?

TimJTi:
> @TimJTi I would like to review this, if you don't mind, in case any change causes any issue already known to me. I will try to check the app today, for proper usage of the API, and possibly tomorrow the actual settings code.
> 
I will be pushing a new commit with minor changes, and will wait for you to finish your much needed review before working on them. It is inevitable I have misunderstood the purpose or reasoning of things along the way :)

examples/settings/settings_main.c
TimJTi: It is used in a variety of apps in the repo and seems to work though?

examples/settings/settings_main.c
TimJTi: It checks that there aren't too many arguments. If the app is called as "settings -b -rubbish" it traps for that; even though it may be unnecessary. But it is strict and therefore OK I think?

examples/settings/settings_main.c
TimJTi: So, it is the "return OK" at the end of the app that is wrong. That should be `return ret`

examples/settings/settings_main.c
TimJTi: But it has been loaded from a file that has been found. Why do you have to create it (again)?

examples/settings/settings_main.c
TimJTi: I was rather left to guess what the functions did, and reverse engineer it all, so I did my best as a first pass :)

To me "create" is the first time the app wants to make use of a setting and "set" is to set the value and "get" is to get the value. If you try and "set" or "get" a value of the wrong type, should it not fail? But, yes, if you do want to change the type what do we do? I chose "create" meaning, in this case, re-create.

It does work, as I have interpreted it, but if its not in line with what you intended I apologise, and please clarify/explain :)

examples/settings/settings_main.c
TimJTi: Agreed.

examples/settings/settings_main.c
TimJTi: Yep, got swapped around when I tweaked it I think.

system/settings/settings.c
TimJTi: @fjpanag @xiaoxiang781216 
Maybe the save() function should  only set a new timer if g_settings.wrpend == 0?

I am obviously struggling a little here, sorry :(

TimJTi: Hmmm...been reading more....the save and/or dump functions need to become reentrant. Will take another look tomorrow, but still open to suggestions :)

examples/settings/Kconfig
```diff
@@ -0,0 +1,53 @@
...
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config EXAMPLES_SETTINGS
+	tristate "Settings example"
+	default n
+	---help---
+        Enable the "settings" example. This uses the functions provided in
+        utils/settings to test the storage and retrieval od various
```
hartmannathan, line 11:
```suggestion
        utils/settings to test the storage and retrieval of various
```

TimJTi: Found some spaces not tabs in this too, so will correct that and this error, and push a new commit

system/settings/settings.c
TimJTi: @fjpanag @xiaoxiang781216 
The more I think about this the more I wondering whether the "dump_cache" function should be removed completely?

The code as-is presumes a write to a block device (e.g. SD card) but that may not be what's wanted. In my case, I actually need to write the data to EEPROM, and only bytes that have changed, so that needs a separate and unrelated thread to handle it.

This code currently has the ability to signal other threads that c hange has occured. Is that not enough? Off load any async file saving away from this settings function?

I am also inclined to move away from a signal and move to a poll method instead?

What do you both think?

Otherwise, I think I should change it back to a draft PR -  or even close it completely and forgot it? If it's not doing things "right" then I don't want to use it, nor will others, and it certainly isn't going to be merged of course.

TimJTi:
Due to lack of interest. Will work on my own variant and maybe try and resubmit in the future.

PR #2184, opened 2023-11-09
lte/lapi: Remove the PIN feature

SPRESENSE:
## Summary
Remove the PIN feature, because the SIM may accidentally become unusable.

## Impact
Only LTE feature.

## Testing

### DIFF
--- a/lte/lapi/src/lapi_pin.c
+++ b/lte/lapi/src/lapi_pin.c
@@ -36,106 +36,6 @@
 #include "lapi_dbg.h"
 
 /****************************************************************************
- * Pre-processor Definitions
- ****************************************************************************/
-
-#define SETPIN_TARGETPIN_MIN LTE_TARGET_PIN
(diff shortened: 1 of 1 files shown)


cederom:
* It this code to be removed works well in general?
* What will happen with SIM that requires PIN and the code gets now removed?
* Why not add option to disable that functionality and not delete working code?

raiden00pl:
Isn't it the user's responsibility to use the API correctly? If the API is dangerous for some reason, it is better to hide it behind an Kconfig option than to remove it completely.

hartmannathan:
I noticed the APIs do not have docstrings. Perhaps as @cederom and @raiden00pl suggest, don't delete the APIs? Instead, consider adding a comment block before the functions, consistent with most of NuttX, with text like, "WARNING: This API is dangerous because [reason]." If the caller needs to do something specific to mitigate the problems, maybe document that here also.

acassis:
ping @SPRESENSE

acassis:
Closing this PR, let's keep the PIN support

PR #2201, opened 2023-11-17
testing/sd_*: Restrict function and variable scope to current file.

g2gps:
## Summary

Add the static keyword to required function and variable delcaration to reduce their scope. Stops namespace pollution in `CONFIG_BUILD_FLAT`

## Impact

Stops symbol collision when `CONFIG_TESTING_SD_BENCH` and `CONFIG_TESTING_SD_STRESS` are enables simultaneously. 

## Testing

Enable `CONFIG_TESTING_SD_BENCH` and `CONFIG_TESTING_SD_STRESS` at the same time.

### DIFF
--- a/testing/sd_bench/sd_bench_main.c
+++ b/testing/sd_bench/sd_bench_main.c
@@ -80,22 +80,22 @@ typedef struct sdb_config
 static const char *BENCHMARK_FILE =
     CONFIG_TESTING_SD_BENCH_DEVICE "/sd_bench";
 
-const size_t max_block = 65536;
-const size_t min_block = 1;
-const size_t default_block = 512;
+static const size_t max_block = 65536;
+static const size_t min_block = 1;
+static const size_t default_block = 512;
 
-const size_t max_runs = 10000;
-const size_t min_runs = 1;
-const size_t default_runs = 5;
+static const size_t max_runs = 10000;
+static const size_t min_runs = 1;
+static const size_t default_runs = 5;
 
-const size_t max_duration = 60000;
-const size_t min_duration = 1;
-const size_t default_duration = 2000;
+static const size_t max_duration = 60000;
+static const size_t min_duration = 1;
+static const size_t default_duration = 2000;
 
-const bool default_keep_test = false;
-const bool default_fsync = false;
-const bool default_verify = true;
-const bool default_aligned = false;
+static const bool default_keep_test = false;
+static const bool default_fsync = false;
+static const bool default_verify = true;
+static const bool default_aligned = false;
 
 /****************************************************************************
  * Private Function Prototypes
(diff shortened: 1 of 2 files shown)


testing/sd_bench/sd_bench_main.c
```diff
@@ -107,26 +107,26 @@ static int read_test(int fd, sdb_config_t *cfg, uint8_t *block,
...
+static struct timespec get_abs_time(void)
 {
   struct timespec ts;
   clock_gettime(CLOCK_MONOTONIC, &ts);
   return ts;
 }
 
-uint64_t get_time_delta_us(const struct timespec *start,
+static uint64_t get_time_delta_us(const struct timespec *start,
                         const struct timespec *end)
```
xiaoxiang781216, line 130:
```suggestion
                                 const struct timespec *end)
```

testing/sd_stress/sd_stress_main.c
```diff
@@ -272,14 +272,14 @@ static bool rename_dir(const char *old_dir, const char *new_dir)
...
+static struct timespec get_abs_time(void)
 {
   struct timespec ts;
   clock_gettime(CLOCK_MONOTONIC, &ts);
   return ts;
 }
 
-uint64_t get_time_delta(const struct timespec *start,
+static uint64_t get_time_delta(const struct timespec *start,
                         const struct timespec *end)
```
xiaoxiang781216, line 283:
```suggestion
                              const struct timespec *end)
```

davids5 (changes requested):
Why not allocate these in a struct and reduce memory footprint that just including that app costs?

pkarashchenko:
> Why not allocate these in a struct and reduce memory footprint that just including that app costs?

Could you please expand more with maybe a kind of example. I'm not fully following your proposal. I think all those local `const`s will be optimized out by compiler anyway.

hartmannathan (approved):

davids5:
Yokogawa 700937 

> > Why not allocate these in a struct and reduce memory footprint that just including that app costs?
> 
> Could you please expand more with maybe a kind of example. I'm not fully following your proposal. I think all those local `static const`s will be optimized out by compiler anyway.

I did not think these were local. Also as globals they  do not conform to the coding standard. 

My point was placing all the variables in a struct allocated at run time will reduce the data size to a single pointer or 0 if it is passed as an argument.

acassis:
ping @g2gps

g2gps:
Sorry, I'd forgotten that this was pending.

These variables are marked static, so if I understand correctly, there scope is limited to the files the file they are declared in. 

There is an opportunity to reduce the scope of some of these variables, if required I can refactor.

xiaoxiang781216 (approved):

PR #2205, opened 2023-11-19
ostest: Add initial support for CONFIG_BUILD_KERNEL

fxysunshine:
## Summary

This is the continuation of the pull https://github.com/apache/nuttx-apps/pull/1500

task_* APIs are unavailable in kernel build mode, replacing with posix_spawn or pthread_* API to pass the case. But posix_spawn require some dependency, for example CONFIG_BUILTIN and CONFIG_LIBC_EXECFUNCS, so pthread_* APIs are used in most scenarios. Some tests should be re-visited because the intent is to user another task (in this case another process) to e.g. receive signals. That will require quite a bit of extra work.

Tests that had to be disabled:
- restart: task_restart() does not work at all with kernel mode so it is disabled entirely
- fpu: make sure the FPU test is not even attempted, because it will cause ASSERT() and stop the test
- vfork: vfork() does not work for some reason in CONFIG_BUILD_KERNEL, there is something missing on the kernel side, so just disable the test for now

Tests that should be re-visited:
- The signal tests, now they signal the process itself while before the signal was sent to another task. This will require building the part that receives the signal as a separate process
- waitpid: Like stated above, waitpid does not work for pthreads
- suspend: kill to send signal does not work for pthreads

## Impact
Adds basic support for ostest with CONFIG_BUILD_KERNEL=y

## Testing
sabre-6quad:knsh

### DIFF
--- a/testing/ostest/ostest.h
+++ b/testing/ostest/ostest.h
@@ -115,7 +115,9 @@ void aio_test(void);
 
 /* restart.c ****************************************************************/
 
+#ifndef CONFIG_BUILD_KERNEL
 void restart_test(void);
+#endif
 
 /* waitpid.c ****************************************************************/
 
--- a/testing/ostest/ostest_main.c
+++ b/testing/ostest/ostest_main.c
@@ -339,7 +339,8 @@ static int user_main(int argc, char *argv[])
       check_test_memory_usage();
 #endif
 
-#if defined(CONFIG_ARCH_FPU) && !defined(CONFIG_TESTING_OSTEST_FPUTESTDISABLE)
+#if defined(CONFIG_ARCH_FPU) && !defined(CONFIG_TESTING_OSTEST_FPUTESTDISABLE) && \
+    !defined(CONFIG_BUILD_KERNEL)
       /* Check that the FPU is properly supported during context switching */
 
       printf("\nuser_main: FPU test\n");
@@ -347,13 +348,15 @@ static int user_main(int argc, char *argv[])
       check_test_memory_usage();
 #endif
 
+#ifndef CONFIG_BUILD_KERNEL
       /* Checkout task_restart() */
 
       printf("\nuser_main: task_restart test\n");
       restart_test();
       check_test_memory_usage();
+#endif
 
-#ifdef CONFIG_SCHED_WAITPID
+#if defined(CONFIG_SCHED_WAITPID) && !defined(CONFIG_BUILD_KERNEL)
       /* Check waitpid() and friends */
 
       printf("\nuser_main: waitpid test\n");
@@ -508,7 +511,8 @@ static int user_main(int argc, char *argv[])
       signest_test();
       check_test_memory_usage();
 
-#if defined(CONFIG_SIG_SIGSTOP_ACTION) && defined(CONFIG_SIG_SIGKILL_ACTION)
+#if defined(CONFIG_SIG_SIGSTOP_ACTION) && defined(CONFIG_SIG_SIGKILL_ACTION) && \
+    !defined(CONFIG_BUILD_KERNEL)
       printf("\nuser_main: signal action test\n");
       suspend_test();
       check_test_memory_usage();
@@ -581,8 +585,16 @@ static int user_main(int argc, char *argv[])
 #endif
 
 #if defined(CONFIG_ARCH_HAVE_FORK) && defined(CONFIG_SCHED_WAITPID)
+#ifndef CONFIG_BUILD_KERNEL
       printf("\nuser_main: vfork() test\n");
       vfork_test();
+#else
+      /* REVISIT: The issue with vfork() is on the kernel side, fix the issue
+       * and re-enable this test with CONFIG_BUILD_KERNEL
+       */
+
+      printf("\nuser_main: vfork() test DISABLED (CONFIG_BUILD_KERNEL)\n");
+#endif
 #endif
 
 #ifdef CONFIG_SMP_CALL
@@ -635,11 +647,34 @@ static void stdio_test(void)
 
 int main(int argc, FAR char **argv)
 {
-  int result;
 #ifdef CONFIG_TESTING_OSTEST_WAITRESULT
   int ostest_result = ERROR;
 #else
   int ostest_result = OK;
+#endif
+
+#ifndef CONFIG_BUILD_KERNEL
+  int result;
+#else
+  struct sched_param param;
+  posix_spawnattr_t attr;
+  FAR const char * const arg[7] =
+  {
+    argv[0], "user_main", arg1, arg2, arg3, arg4, NULL
+  };
+
+  pid_t result;
+  int status;
+
+  /* Use posix_spawn API to spawn the new task for tests, should re-enter
+   * the main entry in new task, the second arg as to identify the real
+   * entry point of the test tasks.
+   */
+
+  if (argc >= 2 && strcmp(argv[1], "user_main") == 0)
+    {
+      return user_main(argc - 1 , &argv[1]);
+    }
 #endif
 
   /* Verify that stdio works first */
@@ -680,9 +715,19 @@ int main(int argc, FAR char **argv)
 
   /* Verify that we can spawn a new task */
 
+#ifdef CONFIG_BUILD_KERNEL
+  posix_spawnattr_init(&attr);
+  param.sched_priority = PRIORITY;
+  posix_spawnattr_setschedparam(&attr, &param);
+  posix_spawnattr_setstacksize(&attr, STACKSIZE);
+  status = posix_spawn(&result, arg[0], NULL, &attr,
+                       (char * const *)arg, NULL);
+  if (status != 0)
+#else
   result = task_create("ostest", PRIORITY, STACKSIZE, user_main,
                        (FAR char * const *)g_argv);
   if (result == ERROR)
+#endif
     {
       printf("ostest_main: ERROR Failed to start user_main\n");
       ASSERT(false);
--- a/testing/ostest/pthread_exit.c
+++ b/testing/ostest/pthread_exit.c
@@ -44,7 +44,7 @@
  * Private Functions
  ****************************************************************************/
 
-static void *pthread_exit_thread(FAR void *parameter)
+static FAR void *pthread_exit_thread(FAR void *parameter)
 {
   unsigned me = (unsigned)pthread_self();
 
@@ -56,7 +56,7 @@ static void *pthread_exit_thread(FAR void *parameter)
   return NULL;
 }
 
-static int pthread_exit_main(int argc, char **argv)
+static FAR void *pthread_exit_main(FAR void *arg)
 {
   pthread_t child;
 #ifdef SDCC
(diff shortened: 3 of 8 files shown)


testing/ostest/ostest_main.c
```diff
@@ -225,6 +225,7 @@ static int user_main(int argc, char *argv[])
   printf("\nuser_main: Begin argument test\n");
   printf("user_main: Started with argc=%d\n", argc);
 
+#ifndef CONFIG_BUILD_KERNEL
```
xiaoxiang781216, line 228: why skip

testing/ostest/ostest_main.c
```diff
@@ -347,13 +349,15 @@ static int user_main(int argc, char *argv[])
...
       /* Checkout task_restart() */
 
       printf("\nuser_main: task_restart test\n");
       restart_test();
       check_test_memory_usage();
+#endif
 
-#ifdef CONFIG_SCHED_WAITPID
+#if defined(CONFIG_SCHED_WAITPID) && !defined(CONFIG_BUILD_KERNEL)
       /* Check waitpid() and friends */
```
xiaoxiang781216, line 361: waittpid support in kernel mode, why skip

testing/ostest/ostest_main.c
```diff
@@ -581,8 +585,16 @@ static int user_main(int argc, char *argv[])
 #endif
 
 #if defined(CONFIG_ARCH_HAVE_FORK) && defined(CONFIG_SCHED_WAITPID)
+#ifndef CONFIG_BUILD_KERNEL
       printf("\nuser_main: vfork() test\n");
       vfork_test();
```
xiaoxiang781216, line 590: vfork should support too?

testing/ostest/ostest_main.c
```diff
@@ -706,6 +719,9 @@ int main(int argc, FAR char **argv)
     }
 
   printf("ostest_main: Exiting with status %d\n", ostest_result);
+#else
+  user_main(1, NULL);
```
xiaoxiang781216, line 723: why need #else

testing/ostest/signest.c
```diff
@@ -295,11 +298,12 @@ void signest_test(void)
 
   /* Start interfering thread  */
 
-  prio++;
-  printf("signest_test: Starting interfering task at priority %d\n", prio);
-  interferepid = task_create("interfere", prio, STACKSIZE,
```
xiaoxiang781216, line 300: let's change task_create to posix_spawn

testing/ostest/ostest_main.c
fxysunshine: vfork->arm_fork do not support in CONFIG_BUILD_KERNEL, and will hang for some reason, there will be much effect to support it.

xiaoxiang781216: Ok.

testing/ostest/ostest_main.c
fxysunshine: test task_create function in flat mode, we can skip in kernel mode
https://github.com/apache/nuttx-apps/blob/master/testing/ostest/ostest_main.c#L681

testing/ostest/signest.c
fxysunshine: OK. That requires quite a bit of extra work to change with posix_spawn for kernel mode. 
1. some case use many global variables to sync between tasks, but global variable is not the same copy between process for kernel mode
https://github.com/apache/nuttx-apps/blob/master/testing/ostest/signest.c#L56
2. waitpid to check WEXISTATUS is wrong in kernel mode.

testing/ostest/ostest.h
```diff
@@ -87,6 +87,8 @@
  * Public Function Prototypes
  ****************************************************************************/
 
+int user_main(int argc, char *argv[]);
```
xiaoxiang781216, line 90: add FAR for ALL pointers

testing/ostest/ostest_main.c
```diff
@@ -96,6 +96,32 @@ static const char g_bad_value2[]   = "BadValue2";
...
+POSIX_FUNC(waitpid_main),
+POSIX_FUNC(pthread_exit_main),
+#endif
+#if defined(CONFIG_SIG_SIGSTOP_ACTION) && defined(CONFIG_SIG_SIGKILL_ACTION)
+POSIX_FUNC(victim_main),
+#endif
+POSIX_FUNC(user_main)
+};
+
+static const int g_posix_spawn_table_count =
```
xiaoxiang781216, line 122: int->size_t

testing/ostest/ostest_main.c
```diff
@@ -295,7 +321,9 @@ static int user_main(int argc, char *argv[])
   show_environment(false, true, true);
   check_test_memory_usage();
 
+#ifndef CONFIG_BUILD_KERNEL
```
xiaoxiang781216, line 324: why skip

testing/ostest/ostest_main.c
```diff
@@ -347,13 +375,16 @@ static int user_main(int argc, char *argv[])
...
       /* Checkout task_restart() */
 
       printf("\nuser_main: task_restart test\n");
       restart_test();
       check_test_memory_usage();
+#endif
 
-#ifdef CONFIG_SCHED_WAITPID
+#if defined(CONFIG_SCHED_WAITPID) && \
+    (!defined(CONFIG_BUILD_KERNEL) || defined(CONFIG_SCHED_CHILD_STATUS))
```
xiaoxiang781216, line 387: why check CONFIG_BUILD_KERNEL

testing/ostest/ostest_main.c
```diff
@@ -641,6 +680,28 @@ int main(int argc, FAR char **argv)
...
+  {
+    argv[0], "user_main", arg1, arg2, arg3, arg4, NULL
+  };
+
+  /* posix_spawn API to spawn the new task for tests, need re-visit
+   * the main entry, Use the second arg string to identify the real
+   * entry of test tasks
+   */
+
+  for (i = 0; i < g_posix_spawn_table_count; i++)
```
xiaoxiang781216, line 697: nitems(g_posix_spawn_table)

testing/ostest/ostest_main.c
```diff
@@ -641,6 +680,28 @@ int main(int argc, FAR char **argv)
 #else
   int ostest_result = OK;
 #endif
+  int i;
```
xiaoxiang781216, line 683:
```
  struct sched_param param;
  posix_spawnattr_t attr;
  FAR const char * const arg[7] =
  {
    argv[0], "user_main", arg1, arg2, arg3, arg4, NULL
  };
#ifdef CONFIG_TESTING_OSTEST_WAITRESULT
  int ostest_result = ERROR;
#else
  int ostest_result = OK;
#endif
  int result;
  pid_t pid;
  size_t i;
```

testing/ostest/ostest_main.c
fxysunshine: waitpid function is not completed, and child exit status is wrong. Should enable SCHED_CHILD_STATUS in kernel mode.

testing/ostest/sighand.c
```diff
@@ -128,7 +128,7 @@ static void wakeup_action(int signo, siginfo_t *info, void *ucontext)
     }
 }
 
-static int waiter_main(int argc, char *argv[])
+static void *waiter_main(FAR void *parm)
```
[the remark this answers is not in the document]
fxysunshine: global variable to check status, but invalid in kernel mode
https://github.com/apache/nuttx-apps/blob/master/testing/ostest/sighand.c#L55

testing/ostest/ostest_main.c
fxysunshine: clearenv() will clear PATH, result in failure to find elf binary for sub-task spawn

testing/ostest/ostest_main.c
```diff
@@ -641,6 +680,28 @@ int main(int argc, FAR char **argv)
...
+  pid_t pid;
+  struct sched_param param;
+  posix_spawnattr_t attr;
+  FAR const char * const arg[7] =
+  {
+    argv[0], "user_main", arg1, arg2, arg3, arg4, NULL
```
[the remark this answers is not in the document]
fxysunshine: global variable argx

testing/ostest/signest.c
```diff
@@ -180,10 +180,10 @@ static int waiter_main(int argc, char *argv[])
   /* Just exit, the system should clean up the signal handlers */
 
   g_waiter_running = false;
-  return EXIT_SUCCESS;
+  return NULL;
 }
 
-static int interfere_main(int argc, char *argv[])
+static void *interfere_main(void *arg)
```
xiaoxiang781216, line 186:
```suggestion
static FAR void *interfere_main(FAR void *arg)
```

testing/ostest/signest.c
```diff
@@ -118,7 +118,7 @@ static void waiter_action(int signo)
   sem_post(&g_sem_signal_finish);
 }
 
-static int waiter_main(int argc, char *argv[])
+static void *waiter_main(void *arg)
```
xiaoxiang781216, line 121:
```suggestion
static FAR void *waiter_main(FAR void *arg)
```

testing/ostest/sighand.c
```diff
@@ -128,7 +128,7 @@ static void wakeup_action(int signo, siginfo_t *info, void *ucontext)
     }
 }
 
-static int waiter_main(int argc, char *argv[])
+static void *waiter_main(FAR void *arg)
```
xiaoxiang781216, line 131:
```suggestion
static FAR void *waiter_main(FAR void *arg)
```

testing/ostest/pthread_exit.c
```diff
@@ -56,7 +56,7 @@ static void *pthread_exit_thread(FAR void *parameter)
   return NULL;
 }
 
-static int pthread_exit_main(int argc, char **argv)
+static void *pthread_exit_main(FAR void *arg)
```
xiaoxiang781216, line 59:
```suggestion
static FAR void *pthread_exit_main(FAR void *arg)
```

testing/ostest/ostest_main.c
```diff
@@ -680,9 +715,19 @@ int main(int argc, FAR char **argv)
...
+#ifdef CONFIG_BUILD_KERNEL
+  posix_spawnattr_init(&attr);
+  param.sched_priority = PRIORITY;
+  posix_spawnattr_setschedparam(&attr, &param);
+  posix_spawnattr_setstacksize(&attr, STACKSIZE);
+  status = posix_spawn(&result, arg[0], NULL, &attr,
+                       (char * const *)arg, NULL);
+  if (status != 0)
+#else
   result = task_create("ostest", PRIORITY, STACKSIZE, user_main,
```
xiaoxiang781216, line 727: why not always use pthread_create

fxysunshine: Subtask user_main need check the argc and argv list, but no copy argv list in pthread_create for parameter. If need, should rework for user_main, but do not change anything if posix_spawn in kernel mode.
static int user_main(int argc, char *argv[])

xiaoxiang781216 (approved):

jerpelea (approved):

PR #2208, opened 2023-11-21
[Don't submit]: add test for SMP bug

GUIDINGLI:
## Summary

[Don't submit]: add test for SMP bug

## Impact

test only


hartmannathan:
This PR is labeled `[Don't submit]`. Do you want to mark it as Draft to avoid accidental merge?

PR #2210, opened 2023-11-24
PTPD improvements

PetteriAimonen:
## Summary

Implemented missing functionality in PTPD and fixed bugs.

Most importantly:

* Negative clock deltas were not correctly handled.
* Add `SO_TIMESTAMP` support to get accurate packet reception times
* Add network path delay measurement
* Add status and stop interfaces to control ptpd
* Improve clock adjustment algorithm, implement clock drift estimator

## Impact

Improves timing accuracy achievable.

## Testing

Tested on custom STM32F417 board, both against Linux `ptpd2 version 2.3.1` and between two STM32 boards.

My PC does not have PTP-compatible hardware timestamping, so it has some 50 µs of jitter.
Between two STM32 boards the clocks are synchronized within +- 5µs as confirmed by measuring GPIO PPS signal with oscilloscope.

Clock settling time depends primarily on `CONFIG_CLOCK_ADJTIME_PERIOD_MS` and `CONFIG_NETUTILS_PTPD_DRIFT_AVERAGE_S`. They present a tradeoff between sensitivity to measurement jitter and speed of reaction. The plot below is with `ADJTIME_PERIOD_MS=10000` and `DRIFT_AVERAGE_S=600`.

plot of test results

Current limit of timing resolution is the use of microseconds in `adjtime()` and `SO_TIMESTAMP` interfaces.
Implementing `adjtimex()` and `SO_TIMESTAMPNS` on NuttX side would permit nanosecond resolution.
However the achieved accuracy is more than sufficient for my own purposes so I'm not currently planning to do so.

### DIFF
--- a/netutils/ptpd/ptpd.c
+++ b/netutils/ptpd/ptpd.c
@@ -84,23 +84,45 @@ struct ptp_state_s
   uint16_t sync_seq;
   uint16_t delay_req_seq;
 
+  /* Previous measurement and estimated clock drift rate */
+
+  struct timespec last_delta_timestamp;
+  int64_t last_delta_ns;
+  int64_t last_adjtime_ns;
+  long drift_avg_total_ms;
+  long drift_ppb;
+
   /* Identity of currently selected clock source,
    * from the latest announcement message.
    *
-   * The timestamp is used for timeout when a source
-   * disappears, it is from the local monotonic clock.
+   * The timestamps are used for timeout when a source disappears.
+   * They are from the local CLOCK_MONOTONIC.
    */
 
-  struct ptp_announce_s selected_source;
-  struct timespec last_received_sync;
-  struct timespec last_received_multicast;
+  bool selected_source_valid;              /* True if operating as client */
+  struct ptp_announce_s selected_source;   /* Currently selected server */
+  struct timespec last_received_multicast; /* Any multicast packet */
+  struct timespec last_received_announce;  /* Announce from any server */
+  struct timespec last_received_sync;      /* Sync from selected source */
 
-  /* Last transmitted sync & announcement packets */
+  /* Last transmitted packet timestamps (CLOCK_MONOTONIC)
+   * Used to set transmission interval.
+   */
 
   struct timespec last_transmitted_sync;
   struct timespec last_transmitted_announce;
+  struct timespec last_transmitted_delayresp;
+  struct timespec last_transmitted_delayreq;
 
-  /* Latest received packet and its timestamp */
+  /* Timestamps related to path delay calculation (CLOCK_REALTIME) */
+
+  bool can_send_delayreq;
+  struct timespec delayreq_time;
+  int path_delay_avgcount;
+  long path_delay_ns;
+  long delayreq_interval;
+
+  /* Latest received packet and its timestamp (CLOCK_REALTIME) */
 
   struct timespec rxtime;
   union
@@ -109,15 +131,16 @@ struct ptp_state_s
     struct ptp_announce_s   announce;
     struct ptp_sync_s       sync;
     struct ptp_follow_up_s  follow_up;
+    struct ptp_delay_req_s  delay_req;
+    struct ptp_delay_resp_s delay_resp;
     uint8_t                 raw[128];
   } rxbuf;
 
-  union
-  {
-    uint8_t                 raw[64];
-  } rxcmsg;
+  uint8_t rxcmsg[CMSG_LEN(sizeof(struct timeval))];
 
-  /* Buffered sync packet for two-step clock setting */
+  /* Buffered sync packet for two-step clock setting where server sends
+   * the accurate timestamp in a separate follow-up message.
+   */
 
   struct ptp_sync_s twostep_packet;
   struct timespec twostep_rxtime;
@@ -223,6 +246,33 @@ static int64_t timespec_to_ms(struct timespec *ts)
   return ts->tv_sec * MSEC_PER_SEC + (ts->tv_nsec / NSEC_PER_MSEC);
 }
 
+/* Get positive or negative delta between two timespec values.
+ * If value would exceed int64 limit (292 years), return INT64_MAX/MIN.
+ */
+
+static int64_t timespec_delta_ns(const struct timespec *ts1,
+                                 const struct timespec *ts2)
+{
+  int64_t delta_s;
+
+  delta_s = ts1->tv_sec - ts2->tv_sec;
+
+#ifdef CONFIG_SYSTEM_TIME64
+  /* Conversion to nanoseconds could overflow if the system time is 64-bit */
+
+  if (delta_s >= INT64_MAX / NSEC_PER_SEC)
+    {
+      return INT64_MAX;
+    }
+  else if (delta_s <= INT64_MIN / NSEC_PER_SEC)
+    {
+      return INT64_MIN;
+    }
+#endif
+
+  return delta_s * NSEC_PER_SEC + (ts1->tv_nsec - ts2->tv_nsec);
+}
+
 /* Check if the currently selected source is still valid */
 
 static bool is_selected_source_valid(struct ptp_state_s *state)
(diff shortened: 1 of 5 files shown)


netutils/ptpd/ptpd.c
```diff
@@ -1262,6 +1275,110 @@ static int ptp_process_rx_packet(struct ptp_state_s *state, ssize_t length)
...
+
+  sigaction(SIGHUP, &act, NULL);
+  sigaction(SIGUSR1, &act, NULL);
+}
+
+/* Process status information request */
+
+static void ptp_process_statusreq(struct ptp_state_s *state)
+{
+  struct ptpd_status_s *status;
```
xiaoxiang781216, line 1313: FAR

netutils/ptpd/ptpd.c
```diff
@@ -1262,6 +1275,110 @@ static int ptp_process_rx_packet(struct ptp_state_s *state, ssize_t length)
...
+    }
+
+  status = state->status_req.dest;
+  status->clock_source_valid = state->selected_source_valid;
+
+  if (status->clock_source_valid)
+    {
+      /* Copy relevant parts of announce info to status struct */
+
+      struct ptp_announce_s *s = &state->selected_source;
```
xiaoxiang781216, line 1327: FAR

netutils/ptpd/ptpd.c
```diff
@@ -1262,6 +1275,110 @@ static int ptp_process_rx_packet(struct ptp_state_s *state, ssize_t length)
...
+             sizeof(status->clock_source_info.gm_id));
+
+      status->clock_source_info.stepsremoved =
+          ((uint16_t)s->stepsremoved[0] << 8) | s->stepsremoved[1];
+      status->clock_source_info.timesource = s->timesource;
+    }
+
+  /* Copy latest adjustment info */
+
+  status->last_clock_update   = state->last_delta_timestamp;
```
xiaoxiang781216, line 1353: remove extra spaces

netutils/ptpd/ptpd.c
```diff
@@ -1346,6 +1469,7 @@ static int ptp_daemon(int argc, FAR char** argv)
       ptp_periodic_send(state);
 
       state->selected_source_valid = is_selected_source_valid(state);
+      ptp_process_statusreq(state);
```
xiaoxiang781216, line 1472: why not process in the signal handler? it suffer the race condition process it here.

netutils/ptpd/ptpd.c
```diff
@@ -223,6 +246,33 @@ static int64_t timespec_to_ms(struct timespec *ts)
   return ts->tv_sec * MSEC_PER_SEC + (ts->tv_nsec / NSEC_PER_MSEC);
 }
 
+/* Get positive or negative delta between two timespec values.
+ * If value would exceed int64 limit (292 years), return INT64_MAX/MIN.
+ */
+
+static int64_t timespec_delta_ns(const struct timespec *ts1,
+                                 const struct timespec *ts2)
```
xiaoxiang781216, line 265: FAR

netutils/ptpd/ptpd.c
```diff
@@ -319,6 +390,7 @@ static int ptp_initialize_state(struct ptp_state_s *state,
                                 const char *interface)
 {
   int ret;
+  int arg;
```
xiaoxiang781216, line 404: add CONFIG_NET_TIMESTAMP guard

netutils/ptpd/ptpd.c
PetteriAimonen: If I understand correctly, signal handler could get invoked in middle of updating the fields in ptp_state_s. Would need masking & unmasking the signal then before write access to ptp_state_s. I also wanted to avoid any long processing in the signal handler, though that probably makes little difference.

It's true that processing it here could cause a race condition if two high-priority tasks call ptpd_status() simultaneously. Then one of the requests would timeout.

xiaoxiang781216: > If I understand correctly, signal handler could get invoked in middle of updating the fields in ptp_state_s. Would need masking & unmasking the signal then before write access to ptp_state_s.

This behavior is done by signal subsystem by default.

> I also wanted to avoid any long processing in the signal handler, though that probably makes little difference.
> 

The signal handler copy a small struct, so the process should be very quickly.

> It's true that processing it here could cause a race condition if two high-priority tasks call ptpd_status() simultaneously. Then one of the requests would timeout.

Yes, it's also doesn't work in protected and kernel mode. So, it's better to use mqueue/socket, or creete a named share memory.

PetteriAimonen: I'm sorry, but I don't understand what kind of approach you are suggesting. `ptp_process_statusreq()` takes some time, and it also accesses multiple fields in `ptp_state_s`. If signal handler would occur in middle of e.g. `ptp_process_rx_packet()` and would then call `ptp_process_statusreq()` inside the signal handler, the fields could be in inconsistent state.

I think mqueue/socket is excessive for such simple status requests. Shared memory makes sense as future expansion, but for flat builds it is unnecessary.

If the task requesting status is lower priority than PTPD, simultaneous requests won't happen.

xiaoxiang781216: > I'm sorry, but I don't understand what kind of approach you are suggesting. `ptp_process_statusreq()` takes some time, and it also accesses multiple fields in `ptp_state_s`.

ptp_process_statusreq just copy a bunch of small structure, plus sem_post, so the timing isn't a big issue.

> If signal handler would occur in middle of e.g. `ptp_process_rx_packet()` and would then call `ptp_process_statusreq()` inside the signal handler, the fields could be in inconsistent state.
> 

Yes, this is a big problem, it's a bad idea to call ptp_process_statusreq in signal handler.

> I think mqueue/socket is excessive for such simple status requests. Shared memory makes sense as future expansion, but for flat builds it is unnecessary.
> 
> If the task requesting status is lower priority than PTPD, simultaneous requests won't happen.

Ok, let's keep the current desgin.

xiaoxiang781216 (approved):

netutils/ptpd/ptpd.c
```diff
@@ -476,9 +583,48 @@ static int ptp_destroy_state(struct ptp_state_s *state)
...
+  if (timespec_to_ms(&delta) > CONFIG_NETUTILS_PTPD_MULTICAST_TIMEOUT_MS)
+    {
+      /* Remove and re-add the multicast group */
+
+      state->last_received_multicast = time_now;
+
+      mcast_addr.s_addr = HTONL(PTP_MULTICAST_ADDR);
+      ipmsfilter(&state->interface_addr.sin_addr,
+                  &mcast_addr,
+                  MCAST_EXCLUDE);
```
pkarashchenko, line 611: remove 1 space

netutils/ptpd/ptpd.c
```diff
@@ -272,55 +336,83 @@ static uint16_t ptp_get_sequence(struct ptp_header_s *hdr)
...
+  struct cmsghdr *cmsg;
 
-  return OK;
+  for_each_cmsghdr(cmsg, rxhdr)
+    {
+      if (cmsg->cmsg_level == SOL_SOCKET &&
+          cmsg->cmsg_type == SO_TIMESTAMP &&
+          cmsg->cmsg_len == CMSG_LEN(sizeof(struct timeval)))
+        {
+          TIMEVAL_TO_TIMESPEC((struct timeval *)CMSG_DATA(cmsg), ts);
```
pkarashchenko, line 384: Add `FAR` here

netutils/ptpd/ptpd.c
```diff
@@ -582,20 +732,68 @@ static int ptp_send_sync(struct ptp_state_s *state)
...
+  req.header = state->own_identity.header;
+  req.header.messagetype = PTP_MSGTYPE_DELAY_REQ;
+  req.header.messagelength[1] = sizeof(req);
+  ptp_increment_sequence(&state->delay_req_seq, &req.header);
+
+  ptp_gettime(state, &state->delayreq_time);
+  timespec_to_ptp_format(&state->delayreq_time, req.origintimestamp);
+
+  ret = sendto(state->tx_socket, &req, sizeof(req), 0,
+    (struct sockaddr *)&addr, sizeof(addr));
```
pkarashchenko, line 765: Add `FAR` here and align with upper `(`

netutils/ptpd/ptpd.c
```diff
@@ -618,53 +816,101 @@ static int ptp_periodic_send(struct ptp_state_s *state)
...
+
+#ifdef CONFIG_NETUTILS_PTPD_SEND_DELAYREQ
+  if (state->selected_source_valid && state->can_send_delayreq)
+    {
+      struct timespec time_now;
+      struct timespec delta;
+
+      clock_gettime(CLOCK_MONOTONIC, &time_now);
+      clock_timespec_subtract(&time_now,
+        &state->last_transmitted_delayreq, &delta);
```
pkarashchenko, line 829: align with upper `(`

netutils/ptpd/ptpd.c
```diff
@@ -618,53 +816,101 @@ static int ptp_periodic_send(struct ptp_state_s *state)
...
+      state->last_adjtime_ns = 0;
+      state->drift_avg_total_ms = 0;
+      state->drift_ppb = 0;
 
       if (ret == OK)
         {
-          ptpinfo("Jumped to timestamp %ld.%09ld s\n",
-            (long)local_time.tv_sec, (long)local_time.tv_nsec);
+          ptpinfo("Jumped to timestamp %lld.%09ld s\n",
+            (long long)new_time.tv_sec, (long)new_time.tv_nsec);
```
pkarashchenko, line 913: Align with upper `(`

netutils/ptpd/ptpd.c
```diff
@@ -673,30 +919,127 @@ static int ptp_update_local_clock(struct ptp_state_s *state,
...
+      /* Apply adjustment and store information for next time */
+
+      state->last_delta_ns = delta_ns;
+      state->last_delta_timestamp = *local_timestamp;
+      state->last_adjtime_ns = adjustment_ns;
+
+      ptpinfo("Delta: %+lld ns, adjustment %+lld ns, drift rate %+lld ppb\n",
+        (long long)delta_ns,
+        (long long)state->last_adjtime_ns,
+        (long long)state->drift_ppb);
```
pkarashchenko, line 1017: Align with upper `(`

netutils/ptpd/ptpd.c
```diff
@@ -753,18 +1092,144 @@ static int ptp_process_followup(struct ptp_state_s *state,
...
+  memcpy(resp.reqidentity, msg->header.sourceidentity,
+         sizeof(resp.reqidentity));
+  memcpy(resp.reqportindex, msg->header.sourceportindex,
+         sizeof(resp.reqportindex));
+  memcpy(resp.header.sequenceid, msg->header.sequenceid,
+         sizeof(resp.header.sequenceid));
+  resp.header.logmessageinterval = CONFIG_NETUTILS_PTPD_DELAYRESP_INTERVAL;
+
+  ret = sendto(state->tx_socket, &resp, sizeof(resp), 0,
+    (struct sockaddr *)&addr, sizeof(addr));
```
pkarashchenko, line 1135: Add `FAR` here and align with upper `(`

netutils/ptpd/ptpd.c
```diff
@@ -805,10 +1284,114 @@ static int ptp_process_rx_packet(struct ptp_state_s *state, ssize_t length)
...
+    }
+
+  status = state->status_req.dest;
+  status->clock_source_valid = state->selected_source_valid;
+
+  if (status->clock_source_valid)
+    {
+      /* Copy relevant parts of announce info to status struct */
+
+      struct ptp_announce_s *s = &state->selected_source;
```
pkarashchenko, line 1336: Add `FAR` here

netutils/ptpd/ptpd.c
```diff
@@ -805,10 +1284,114 @@ static int ptp_process_rx_packet(struct ptp_state_s *state, ssize_t length)
...
+
+  state->status_req.done = NULL;
+  state->status_req.dest = NULL;
+}
+
+/* Main PTPD task */
+
 static int ptp_daemon(int argc, FAR char** argv)
 {
   const char *interface = "eth0";
```
pkarashchenko, line 1393: Add `FAR` here

system/ptpd/ptpd_main.c
```diff
@@ -29,6 +29,118 @@
 
 #include "netutils/ptpd.h"
 
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+static int do_ptpd_start(const char *interface)
```
pkarashchenko, line 36: Add `FAR` here

PetteriAimonen:
@pkarashchenko Thanks for the review, I have added the fixes in separate pull request #2212.

PR #2217, opened 2023-12-03
examples/modbus: Fix modbus example to keep running

acassis:
## Summary
examples/modbus: Fix modbus example to keep running
More info: https://github.com/apache/nuttx/issues/11305
## Impact
Users will be able to use modbus example
## Testing
STM32F4Discovery


examples/modbus/modbus_main.c
```diff
@@ -432,20 +432,22 @@ int main(int argc, FAR char *argv[])
 
   /* Handle command line arguments */
 
-  g_modbus.quit = false;
+  g_modbus.quit = true;
```
pkarashchenko, line 435: why `quit` is defaulted to `true`?

acassis: Because now I use it to decide if the program needs to finishes normally or keep waiting. The only option where it keeps waiting is when user passes "modbus -e". Initially I was "g_modbus.quit = true" inside each switch/case option except the case 'e'. So I realized it was better to invert the value to simplify the code.

acassis: Ah ok, I forgot to update the comment

acassis: wait, it was a mistake! Probably I inverted in the top, removed the others quit = true and inverted that from false to true...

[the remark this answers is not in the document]
acassis: Yes, I think we don't need it to be shared between threads, I will remove it.

xiaoxiang781216 (approved):

PR #2239, opened 2023-12-21
apps: Select PIPES when enabling system popen

acassis:
## Summary
Select PIPES when enabling system popen we explained here: https://github.com/apache/nuttx/issues/11435
## Impact
Users will be able to get popen working correctly
## Testing
Build

### DIFF
--- a/system/popen/Kconfig
+++ b/system/popen/Kconfig
@@ -7,7 +7,8 @@ config SYSTEM_POPEN
 	bool "popen()/pclose() Functions"
 	default n
 	select SCHED_WAITPID
-	depends on NSH_LIBRARY && PIPES
+	select PIPES
+	depends on NSH_LIBRARY
 	---help---
 		Enable support for the popen() and pclose() interfaces.
 		This will support execution of NSH commands from C code with


xiaoxiang781216:
here is some recommendation for `select` from zephyr:
https://docs.zephyrproject.org/latest/build/kconfig/tips.html#select-recommendations

acassis:
> here is some recommendation for `select` from zephyr: https://docs.zephyrproject.org/latest/build/kconfig/tips.html#select-recommendations

@xiaoxiang781216 in this case CONFIG_SYSTEM_POPEN is selecting a feature that doesn't "slip" or get "out of sync". This is a clear case where the feature popen() requires SCHED_WAITPID and PIPES to work correctly:

```
        select SCHED_WAITPID
        select PIPES
``` 

Also PIPES doesn't have external dependencies that could create confusion or circular dependence!

I think what people from zephyr is trying to avoid are those other cases where "select" could create problems cause by symbols that many dependencies.

In this case here we don't have this risk, because CONFIG_SYSTEM_POPEN is not enabled by default and PIPES doesn't have other dependencies.

xiaoxiang781216:
We found that select expose some bad effect, for example:

1. Feature A depends on Feature C
2. Feature B select Feature C

The initial defconfig set both A and C to y to get feature A. Next, he wants feature B and enable it, defconfig contain just contains A and B. In the later, B feature doesn't need and then remove B from defconfig, but defconfig fail since C is removed from step 2.

acassis:
@xiaoxiang781216 right, but in this case here it is a direct dependence:
Let's say assume Features:
A = popen example app
B = system popen
C = pipes

-> A depends on B (so user needs to example B first to see A)
-> When user enable B, it automatically will select C.

In this case defconfig with end up with A, B and C.

The only way it could fail is if the user manually disable PIPES, but are want to improve beginner users experience and reduce their friction. So users that follow the common path will not face this issue.

xiaoxiang781216:
But, there are other path which directly use pipes, if someday the project decides to avoid use popen to save the code size, the end user will surprise that the linker can't find pipes after he disable popen a related feature. So, it's prefer to use `depends on` over  `select`:

1. Both `depends on` and `select` can describe the relationship correctly
2. And `depends on` doesn't hide the relationship (user still need enable in defconfig explicitly)
3. But `select` hide the relationship which introduce many problems in the engineering process

acassis:
@xiaoxiang781216 but popen is not used by default by the project, so there is not possibility to it be disabled and break the relation.

I think "depends on" doesn't protect the feature as we are thinking, because the user could there and enable manually the feature in the .config. And even worst, "depdends on" hide the feature in menuconfig.

The user will search for it, will find it, when when he goes there the option is not there. I think this is a case where "select" is better than "depends on"

xiaoxiang781216:
> @xiaoxiang781216 but popen is not used by default by the project, so there is not possibility to it be disabled and break the relation.
> 

user may call both pipes and popen initially, and remove popen call later, and will surprise that linker complain pipes function can't be found after he remove CONFIG_SYSTEM_POPEN from defconfig.

> I think "depends on" doesn't protect the feature as we are thinking, because the user could go there and enable manually the feature in the .config.

The user will find that the feature doesn't enable actually because the dependence doesn't match, instead linker error. Why do you think `depends on` doesn't protect the feature?

> And even worst, "depends on" hides the feature in menuconfig.
>
> The user will search for it, will find it, but when he goes there the option is not there. I think this is a case where "select" is better than "depends on"

The user need understand the basic Kconfig concept, I don't believe that they can work with NuttX or Zephyr/Linux efficiently if they don't understand `depends on` and `select`.

acassis:
> > @xiaoxiang781216 but popen is not used by default by the project, so there is not possibility to it be disabled and break the relation.
> 
> user may call both pipes and popen initially, and remove popen call later, and will surprise that linker complain pipes function can't be found after he remove CONFIG_SYSTEM_POPEN from defconfig.
> 

Xiang, if the user enable both features (not matter which order) and if later he/she disables popen, pipes still enabled. If his/her defconfig had popen enabled and later on popen feature was removed from menuconfig, I understand that pipes will not be found, unless other feature also selects it.

I think in general "select" and "depends on" are similar, but although "depends on" is safer, it has a drawback because it "hides" that feature from newcomer users. But I agree with you, users should learn how Kconfig works and how to search and fix symbols dependence.

> > I think "depends on" doesn't protect the feature as we are thinking, because the user could go there and enable manually the feature in the .config.
> 
> The user will find that the feature doesn't enable actually because the dependence doesn't match, instead linker error. Why do you think `depends on` doesn't protect the feature?
>

Because a beginner user could bypass the menuconfig enabling the feature directly editing the .config and then complain in the mailing list that NuttX functions are broken :-)

> > And even worst, "depends on" hides the feature in menuconfig.
> > The user will search for it, will find it, but when he goes there the option is not there. I think this is a case where "select" is better than "depends on"
> 
> The user need understand the basic Kconfig concept, I don't believe that they can work with NuttX or Zephyr/Linux efficiently if they don't understand `depends on` and `select`.
>

Agree!

> BTW, user still encounter the similar problem unless you change ALL `depends on` to `select`.

Not in this case, because the user just need to use popen, so if he enable the popen examble it could automatically select the system open and the pipe. In the other hand I think "depends on" is safe if user doesn't see the example popen in the menuconfig. If we don't have a documentation saying that examples exists, it will become a hidden feature.

If you agree I can change the popen example to also select system popen, so the problem will not exist for end users

xiaoxiang781216:
> > > @xiaoxiang781216 but popen is not used by default by the project, so there is not possibility to it be disabled and break the relation.
> > 
> > 
> > user may call both pipes and popen initially, and remove popen call later, and will surprise that linker complain pipes function can't be found after he remove CONFIG_SYSTEM_POPEN from defconfig.
> 
> Xiang, if the user enable both features (not matter which order) and if later he/she disables popen, pipes still enabled. If his/her defconfig had popen enabled and later on popen feature was removed from menuconfig, I understand that pipes will not be found, unless other feature also selects it.

But, if we change popen to select pipes, CONFIG_PIPES will be removed from defconfig after normalization. You see the bad side effect with select.

> 
> I think in general "select" and "depends on" are similar, but although "depends on" is safer, it has a drawback because it "hides" that feature from newcomer users. But I agree with you, users should learn how Kconfig works and how to search and fix symbols dependence.
> 
> > > I think "depends on" doesn't protect the feature as we are thinking, because the user could go there and enable manually the feature in the .config.
> > 
> > 
> > The user will find that the feature doesn't enable actually because the dependence doesn't match, instead linker error. Why do you think `depends on` doesn't protect the feature?
> 
> Because a beginner user could bypass the menuconfig enabling the feature directly editing the .config and then complain in the mailing list that NuttX functions are broken :-)
> 

But I think it isn't a linker error.

> > > And even worst, "depends on" hides the feature in menuconfig.
> > > The user will search for it, will find it, but when he goes there the option is not there. I think this is a case where "select" is better than "depends on"
> > 
> > 
> > The user need understand the basic Kconfig concept, I don't believe that they can work with NuttX or Zephyr/Linux efficiently if they don't understand `depends on` and `select`.
> 
> Agree!
> 
> > BTW, user still encounter the similar problem unless you change ALL `depends on` to `select`.
> 
> Not in this case, because the user just need to use popen, so if he enable the popen examble it could automatically select the system open and the pipe. In the other hand I think "depends on" is safe if user doesn't see the example popen in the menuconfig. If we don't have a documentation saying that examples exists, it will become a hidden feature.
> 

I mean there are many features which describe the dependence through `depends on` not `select`. We need fix them too, otherwise the user will complain again and again.

> If you agree I can change the popen example to also select system popen, so the problem will not exist for end users

so, basically all dependence will describe through `depends on`. Before you do that, I would suggest you search the relative discussion on the web. Both Linux/Zephyr already learn from the practice and suggest to avoid select as much as possible.

PR #2241, opened 2023-12-25
wamr: bump the default stack guard size

yamt:
## Summary
using 0 by default seems a bit too aggressive to me.

1024 here is what would have been used before this kconfig knob was introduced by https://github.com/apache/nuttx-apps/pull/1259. It also matches some expectations in the recent wamrc. https://github.com/bytecodealliance/wasm-micro-runtime/blob/40b430fd240ad45f5c7a7dd0239fa2425e7c294e/core/iwasm/compilation/aot_emit_aot_file.c#L2672-L2678


yamt:
@no1wudi how do you think?

no1wudi:
> @no1wudi how do you think?

LGTM, and maybe we should notify user that using thread, since default thread on nuttx is 2048 for many platforms.

interpreters/wamr/Kconfig
```diff
@@ -103,10 +103,9 @@ config INTERPRETERS_WAMR_ENABLE_SPEC_TEST
 
 config INTERPRETERS_WAMR_STACK_GUARD_SIZE
 	int "Custom stack guard size"
-	default 0
+	default 1024
```
xiaoxiang781216, line 106: Is it good to resolve 1KB for stack guard for the default setting?

yamt: what do you mean by "resolve" in this context?

xiaoxiang781216: typo I  mean `reserve`

yamt: 1KB might be a bit too large for typical nuttx targets.
but it's definitely safer than 0.
also, 1KB is what wamr uses by default for other platforms.

xiaoxiang781216: I assume that the stack size consumed by wamr function is fixed, why wasm engine can't catch the stack overflow before jump to the new function?

yamt: because your assumption is wrong?
at least in the current wamr aot implementation, a caller doesn't have precise knowledge about how much stack the callee consumes. the callees need to check the overflow by itself and the check itself might consume stack. (thus the small amount of stack is reserved here.)

yamt:
> > @no1wudi how do you think?
> 
> LGTM, and maybe we should notify user that using thread, since default thread on nuttx is 2048 for many platforms.

i added some description in the help text.

no1wudi (approved):

interpreters/wamr/Kconfig
```diff
@@ -103,10 +103,18 @@ config INTERPRETERS_WAMR_ENABLE_SPEC_TEST
...
+		Reserve some space in stack as guard to detect stack overflow.
+
+		Please consider this extra stack consumption when deciding the stack
+		size of your NuttX task on which you run WAMR.
+
+		The default value here (1024) was chosen to match the default value on
+		the majority of other platforms, including Linux. It also matches
+		the expections in the WAMR AoT compiler.
+		On the other hand, it might be a bit larger than what's strictly
+		necessary, especially for typical NuttX targets.
```
xiaoxiang781216, line 117: so, why not give a small value match the nuttx normal case?

interpreters/wamr/Kconfig
xiaoxiang781216: But 1KB isn't a small amount of stack.

interpreters/wamr/Kconfig
yamt: just because it's difficult to fine tune this kind of things.

PR #2252, opened 2024-01-04
wamr: Add new option to enable tail call support

no1wudi:
## Summary

See also:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2975
## Impact
No
## Testing
WAMR Community

### DIFF
--- a/interpreters/wamr/Kconfig
+++ b/interpreters/wamr/Kconfig
@@ -106,6 +106,12 @@ config INTERPRETERS_WAMR_ENABLE_SPEC_TEST
 	bool "Enable spectest support"
 	default n
 
+config INTERPRETERS_WAMR_TAIL_CALL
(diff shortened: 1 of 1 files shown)


interpreters/wamr/Kconfig
```diff
@@ -106,6 +106,12 @@ config INTERPRETERS_WAMR_ENABLE_SPEC_TEST
 	bool "Enable spectest support"
 	default n
 
+config INTERPRETERS_WAMR_TAIL_CALL
+	bool "Enable tail call support"
+	default y
```
yamt, line 111: why default to y?

no1wudi: CHanged to n now

xiaoxiang781216 (approved):

PR #2262, opened 2024-01-17
wamr: Add new option to enable quick aot entry

no1wudi:
## Summary
wamr: Add new option to enable quick aot entry

WAMR side: https://github.com/bytecodealliance/wasm-micro-runtime/pull/3040
## Impact
No.
## Testing
CI

### DIFF
--- a/interpreters/wamr/Kconfig
+++ b/interpreters/wamr/Kconfig
@@ -36,6 +36,17 @@ config INTERPRETERS_WAMR_AOT
 	select ARCH_USE_TEXT_HEAP if ARCH_HAVE_TEXT_HEAP
 	default n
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

no1wudi:
> please add a commit message

I guess this change is simple enough, the commit "title" can represent it clearly, and help info in change is also complete, should we dup it in commit message body?

jerpelea:
> > please add a commit message
> 
> I guess this change is simple enough, the commit "title" can represent it clearly, and help info in change is also complete, should we dup it in commit message body?

it is a good practice to add it even on small commits

no1wudi:
> it is a good practice to add it even on small commits
OK

PR #2270, opened 2024-01-26
nsh_syscmds: update rptun_ping to rpmsg_ping, add cmd_rpmsg.

wyr8899:
## Summary

To support rpmsg ioctl, add cmd_rpmsg function, and update rptun ping to rpmsg ping.
depends on https://github.com/apache/nuttx/pull/11618

## Impact

Able to use rpmsg ioctl cmd.

## Testing

Tested in sim vela and 86 panel.

### DIFF
--- a/nshlib/Kconfig
+++ b/nshlib/Kconfig
@@ -569,6 +569,11 @@ config NSH_DISABLE_ROUTE
 	depends on FS_PROCFS && NET_ROUTE && !FS_PROCFS_EXCLUDE_NET && !FS_PROCFS_EXCLUDE_ROUTE
 	default DEFAULT_SMALL
 
+config NSH_DISABLE_RPMSG
+	bool "Disable rpmsg"
+	default DEFAULT_SMALL
+	depends on RPMSG
+
 config NSH_DISABLE_RPTUN
 	bool "Disable rptun"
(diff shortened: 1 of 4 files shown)


xiaoxiang781216 (approved):

jerpelea (commented):
please add a commit message

acassis:
@xiaoxiang781216 why rptun and rpmsg commands are implemented as NSH commands? I think it should be just a regular command like "pppd" or other. Also although it is used for test under NSH, keeping it outside will be easier to use as standalone application.

xiaoxiang781216:
> @xiaoxiang781216 why rptun and rpmsg commands are implemented as NSH commands?

Since rptun/rpmsg is a simple mapping to kernel rptun/rpmsg driver, just like cmd_losetup operate on /dev/loop.

> I think it should be just a regular command like "pppd" or other.

Yes, it could be move to apps/system, but it belong another task.

> Also although it is used for test under NSH, keeping it outside will be easier to use as standalone application.

Since the action of rptun/rpmsg command is very simple:

1. Parse command line argument
2. Open the node under /dev/ and call ioctl

If the user disables nsh, it's simpler to call ioctl directly, instead launch the standalone application.

jerpelea (approved):

PR #2279, opened 2024-02-01
examples/dac: Add test mode

PilnyTomas:
Implements simple function generating triangular pattern / saw tooth.

## Summary

cmd_dac_test - create simple configurable triangle wave to test the DAC
-c count  can create this wave multiple times
-d delay  introduces delay between each value - this can be handy when
measuring only with multimeter
-b bit  resolution determines max value - this depends on the
underlying architecture. By default this is set to 8 bits and
therefore max value 255.
-s step  is the increment added to the previous value. If you need only few
measuring points you can set high value such as 64, if you need finer
steps , set to lower value. The default is 8.
Example use for 8-bit DAC using multimeter
nsh> dac -d 5000 -b 8 -s 16 test
Example use for 12-bit DAC using osciloscope
nsh> dac -d 10 -b 12 -s 4 test

## Impact

Easier verification of the entire scope of values

### DIFF
--- a/examples/dac/dac_main.c
+++ b/examples/dac/dac_main.c
@@ -48,6 +48,13 @@
 #  define CONFIG_EXAMPLES_DAC_DEVPATH "/dev/dac0"
 #endif
 
+#define DEFAULT_COUNT          1
+#define DEFAULT_DELAY          0
+#define DEFAULT_CHANNEL        0
+#define DEFAULT_TEST_BIT_RES   8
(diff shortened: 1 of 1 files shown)


acassis (approved):

acassis:
@PilnyTomas please submit a patch to improve the Documentation about dac app: https://nuttx.apache.org/docs/latest/applications/examples/dac/index.html

Please use the examples adc documentation as reference (although even adc is missing usage examples).

PR #2291, opened 2024-02-19
nsh/uname: alloc dynamic heap on non-GNU compilers

anchao:
## Summary

nsh/uname: alloc dynamic heap on non-GNU compilers

The alloca() function is machine- and compiler-dependent.  For certain applications, its use can improve efficiency compared to the use of malloc(3) plus free(3).  In certain cases, it can also simplify memory deallocation in applications that use longjmp(3) or siglongjmp(3). Otherwise, its use is discouraged.

## Impact

N/A

## Testing

Tasking

### DIFF
--- a/nshlib/nsh_syscmds.c
+++ b/nshlib/nsh_syscmds.c
@@ -690,6 +690,7 @@ int cmd_uname(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
   FAR const char *str;
   struct lib_memoutstream_s stream;
   struct utsname info;
+  struct utsname output;
   unsigned int set;
   int option;
   bool badarg;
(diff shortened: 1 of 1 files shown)


nshlib/nsh_syscmds.c
```diff
@@ -780,8 +781,12 @@ int cmd_uname(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
   /* Process each option */
 
   first = true;
-  lib_memoutstream(&stream, alloca(sizeof(struct utsname)),
-                   sizeof(struct utsname));
+#ifdef __GNUC__
+  putsname = alloca(sizeof(struct utsname));
```
xiaoxiang781216, line 785: let's declare "struct utsname xxx" directly

anchao: ustname structure is quite large, how about maintain the current style to avoid stack overflow?

anchao: Done

xiaoxiang781216 (approved):

PR #2302, opened 2024-02-22
fix variable set but not used

Gary-Hobson:
## Summary
These variables will trigger variable 'ret' set but not used warnings due to different configurations.

For using variables only in assert, added UNUSED to declare
Warnings in external repositories are ignored in the makefile

## Impact
Let's ignore the following format check, which seems to have nothing to do with this change
```
apps/modbus/nuttx/porttimer.c:51:9: error: Mixed case identifier found
  return xMBPortSerialSetTimeout(ulTimeOut);
```


acassis (approved):

acassis:
Please verify the issues with sim/zipfs

xiaoxiang781216:
I think it doesn't need with the new approach. Let's close it now.

PR #2304, opened 2024-02-22
cmake: add lvgl cmake support

XuNeo:
## Summary
Add LVGL cmake support.

## Testing

cmake -Bbuild -DBOARD_CONFIG=boards/sim/sim/sim/configs/lvgl_fb -GNinja

### DIFF
--- a/examples/lvgldemo/CMakeLists.txt
+++ b/examples/lvgldemo/CMakeLists.txt
@@ -28,6 +28,8 @@ if(CONFIG_EXAMPLES_LVGLDEMO)
     ${CONFIG_EXAMPLES_LVGLDEMO_STACKSIZE}
     MODULE
     ${CONFIG_EXAMPLES_LVGLDEMO}
+    DEPENDS
+    lvgl_nuttx
     SRCS
     lvgldemo.c)
 endif()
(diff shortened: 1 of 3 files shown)


examples/lvgldemo/CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ if(CONFIG_EXAMPLES_LVGLDEMO)
     ${CONFIG_EXAMPLES_LVGLDEMO_STACKSIZE}
     MODULE
     ${CONFIG_EXAMPLES_LVGLDEMO}
+    DEPENDS
+    lvgl_nuttx
```
xiaoxiang781216, line 32: let's add cmake support when upgrading lvgl to v9?

XuNeo: The idea is to make lvgl v8 working firslty and migrate to v9.0.1 since v9.0.0 is buggy. It should be easier than v8 because lvgl has builtin nuttx support.

XuNeo:
There's still one issue for cmake is the CONFIG_INIT_ARGS must be set to empty string for config `lvgl_fb`. The generated config.h is wrong `#define CONFIG_INIT_ARGS \widgets\`

examples/lvgldemo/CMakeLists.txt
xiaoxiang781216: Ok.

xiaoxiang781216 (approved):

PR #2306, opened 2024-02-24
apps: Make /dev/temp0 as default to system/lm75

acassis:
## Summary
The stm32 common logic board uses /dev/temp0 as
default, but system/lm75/lm75.c is trying to
open /dev/temp instead
## Impact
Make things work out of box
## Testing
stm32


cederom (approved):
thanks @acassis :-)

cederom:
CI failed (cancelled status) but this change seems trivial and already verified :-)

xiaoxiang781216 (approved):

PR #2334, opened 2024-03-19
dhcpc: reset the state of the "cancel" variable when restarting the async dhcpc thread

slorquet:
## Summary
dhcpc can run in async mode with a background thread. this can be cancelled (when the link goes down)

but when the link goes up again the variable dhcpc_state->cancel is not reinitialized, so the async mode works only once

## Impact

dhcpc async thread can be started multiple times

## Testing

verified to work experimentally on a stm32f4 custom board

### DIFF
--- a/netutils/dhcpc/dhcpc.c
+++ b/netutils/dhcpc/dhcpc.c
@@ -959,6 +959,7 @@ int dhcpc_request_async(FAR void *handle, dhcpc_callback_t callback)
     }
 
   pdhcpc->callback = callback;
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
> Hi, can you apply this as soon as checks complete? I have more changes to send, I would like to avoid parallel pull requests, thanks.

It's better to create pull request parallel If you change is independent each other. All committers come from the different time zone and have daily job to be done in the working time.

PR #2337, opened 2024-03-22
nsh: Include option to disable uname timestamp

acassis:
## Summary
Currently uname command will print the timestamp
when the binary was built, and it generates an issue because two identical built binaries will have differents hashes/CRC.

## Impact
It could be possible to disable this feature and generate identical binaries

## Testing
SIM


acassis:
I'll close it can move the configuration to the kernel side

PR #2351, opened 2024-04-08
examples/pipe:removes useless data from test cases

GC-20-20:
## Summary
A file was created during the test, and it was not deleted when it was used up
image

### DIFF
--- a/examples/pipe/pipe_main.c
+++ b/examples/pipe/pipe_main.c
@@ -157,6 +157,12 @@ int main(int argc, FAR char *argv[])
       return 6;
     }
 
+  ret = remove(FIFO_PATH1);
+  if (ret != 0)
+    {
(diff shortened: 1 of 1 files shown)


examples/pipe/pipe_main.c
```diff
@@ -157,6 +157,11 @@ int main(int argc, FAR char *argv[])
       return 6;
     }
 
+  ret = remove(FIFO_PATH1);
+  if (ret != 0)
+    {
+      fprintf(stderr, "pipe_main: remove failed with errno=%d\n", errno);
+    }
```
xiaoxiang781216, line 164: fix the warning

xiaoxiang781216 (approved):

examples/pipe/pipe_main.c
GC-20-20: ok

PR #2355, opened 2024-04-12
apps/nsh_commands: Use quit to poweroff

acassis:
## Summary
Currently new users when try to run NuttX using the SIM get stuck into simulator, because they have no idea that poweroff command is used to leave it. Let use KISS approach and use quit as alias to poweroff command.
## Impact
Make it easier to user to leave SIM
## Testing
sim


cederom:
Unix uses `exit` rather than `quit` can we change that to `exit` ?

Also it would be nice to configure `exit` behavior to be `poweroff` or `reboot`..?

acassis:
I tried to use "exit", but exit is already used internally to exit from nsh or telnet, where poweroff is used to turn-off the board

xiaoxiang781216 (approved):

PR #2381, opened 2024-05-10
system/adb: Ignore warnings for using variables

Gary-Hobson:
## Summary
Will cause compilation warning if NDEBUG is defined We can't modify the code of the external library, so let's ignore it

microADB/hal/hal_uv_client_usb.c:90:13: warning: unused variable 'ret' [-Wunused-variable]
   90 |         int ret = uv_read_start((uv_stream_t*)&client->pipe,


xiaoxiang781216 (approved):

xiaoxiang781216:
Let's provide a patch to fix the upstream project too, @Gary-Hobson.

PR #2385, opened 2024-05-13
wamr: Add a few options for esp32s3

yamt:
## Summary
The corresponding WAMR change:
https://github.com/bytecodealliance/wasm-micro-runtime/pull/2348

### DIFF
--- a/interpreters/wamr/Kconfig
+++ b/interpreters/wamr/Kconfig
@@ -47,6 +47,16 @@ config INTERPRETERS_WAMR_AOT_QUICK_ENTRY
 		signature.
 		Enable this option will increase the size of runtime, ~8KB.
 
+config INTERPRETERS_WAMR_AOT_WORD_ALIGN_READ
+	bool "Make WAMR AOT loader use word-aligned read"
+	default y
+	depends on INTERPRETERS_WAMR_AOT && ARCH_CHIP_ESP32S3
+
+config INTERPRETERS_WAMR_MEM_DUAL_BUS_MIRROR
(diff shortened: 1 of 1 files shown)


interpreters/wamr/Kconfig
```diff
@@ -47,6 +47,16 @@ config INTERPRETERS_WAMR_AOT_QUICK_ENTRY
 		signature.
 		Enable this option will increase the size of runtime, ~8KB.
 
+config INTERPRETERS_WAMR_AOT_WORD_ALIGN_READ
+	bool "Make WAMR AOT loader use word-aligned read"
+	default y
+	depends on INTERPRETERS_WAMR_AOT && ARCH_CHIP_ESP32S3
```
xiaoxiang781216, line 53: why check ARCH_CHIP_ESP32S3 in common code?

yamt: because these things in wamr are specific to esp32s3 at least for now.

xiaoxiang781216: please remove esp32s3 specific code from wamr and remove ARCH_CHIP_ESP32S3 from here.

yamt: i'm not happy with the current structure either.
but i want to make it work first.

xiaoxiang781216: can we remove ARCH_CHIP_ESP32S3 from patch? espressif is working on to make the change in wamr more general for other arch/chip too.

yamt: ok. done.

xiaoxiang781216 (approved):

PR #2393, opened 2024-05-15
benchmarks/ramspeed: Don't expose interrupt control with kernel configuration.

g2gps:
## Summary
Don't expose interrupt control with kernel configuration.

enter_critical_section and leave_critical_section aren't reliable interfaces to expose in usermode, as they aren't available if CONFIG_IRQCOUNT is enabled.

## Impact

Allows the ramspeed benchmark to be built when both `CONFIG_BUILD_KERNEL` and `CONFIG_IRQCOUNT` are enabled. 

## Testing

Litex based board with the above configuration enabled.

### DIFF
--- a/benchmarks/ramspeed/ramspeed_main.c
+++ b/benchmarks/ramspeed/ramspeed_main.c
@@ -64,6 +64,17 @@
       } \
   } while (0)
 
+  #define HAS_IRQ_CONTROL !defined(CONFIG_BUILD_KERNEL) && \
+                          !defined(CONFIG_BUILD_PROTECTED)
+
+  #if HAS_IRQ_CONTROL
+  #  define ENABLE_IRQ(flags) leave_critical_section(flags);
+  #  define DISABLE_IRQ(flags) flags=enter_critical_section();
+  #else
+  #  define ENABLE_IRQ(flags) (void)flags;
+  #  define DISABLE_IRQ(flags) (void)flags;
+  #endif
+
 /****************************************************************************
  * Private Types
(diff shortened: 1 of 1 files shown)


benchmarks/ramspeed/ramspeed_main.c
```diff
@@ -400,7 +404,9 @@ static void memcpy_speed_test(FAR void *dest, FAR const void *src,
 
       if (irq_disable)
         {
+          #ifndef CONFIG_BUILD_KERNEL
```
xiaoxiang781216, line 407: CONFIG_BUILD_PROTECTED need the same special handle too

benchmarks/ramspeed/ramspeed_main.c
```diff
@@ -423,7 +429,9 @@ static void memcpy_speed_test(FAR void *dest, FAR const void *src,
...
```
xiaoxiang781216, line 432: how about define leave_critical_section to empty macro in case of protected/kernel build

benchmarks/ramspeed/ramspeed_main.c
```diff
@@ -465,7 +473,9 @@ static void memset_speed_test(FAR void *dest, uint8_t value,
...
```
xiaoxiang781216, line 476: should remove all intent

g2gps:
Should be a bit clearer, and covers CONFIG_BUILD_PROTECTED.

xiaoxiang781216 (approved):

PR #2397, opened 2024-05-17
Revert "testing/ostest: Fix config name"

Donny9:
## Summary
 this patch https://github.com/apache/nuttx-apps/pull/2394 enable vfork test, but vfork test is failed in sim:citest so, let's recover this patch after fix crash.

## Impact
fix ci break
## Testing
local test

### DIFF
--- a/testing/ostest/ostest_main.c
+++ b/testing/ostest/ostest_main.c
@@ -590,7 +590,7 @@ static int user_main(int argc, char *argv[])
       check_test_memory_usage();
 #endif
 
-#if defined(CONFIG_ARCH_HAVE_FORK) && defined(CONFIG_SCHED_WAITPID)
+#if defined(CONFIG_ARCH_HAVE_VFORK) && defined(CONFIG_SCHED_WAITPID)
 #ifndef CONFIG_BUILD_KERNEL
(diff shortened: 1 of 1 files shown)


Donny9:
@xiaoxiang781216 @ttnie @SPRESENSE

SPRESENSE:
@Donny9 
`CONFIG_ARCH_HAVE_VFORK` is not defined anywhere.
I think it would be better to remove the `select` for `CONFIG_ARCH_HAVE_FORK` from `CONFIG_ARCH_SIM` to avoid `sim:citest` failure so that other architectures are not affected.

xiaoxiang781216:
> @Donny9 `CONFIG_ARCH_HAVE_VFORK` is not defined anywhere. I think it would be better to remove the `select` for `CONFIG_ARCH_HAVE_FORK` from `CONFIG_ARCH_SIM` to avoid `sim:citest` failure so that other architectures are not affected.

good suggestion. @Donny9 let's disable fork test on sim util the problem gets fixed.

Donny9:
@SPRESENSE @xiaoxiang781216 okay，new patch:https://github.com/apache/nuttx/pull/12375 on nuttx/

PR #2434, opened 2024-07-05
dd_main.c: Add oseek optional argument.

dry-75:
Add oseek optional parameter to seek/skip on output file.

## Summary
If oseek=N passed, will seek N*bs on output file before witing.
Optional,  keeping existing seek and bs as-is.

## Testing
dd if=/dev/zero of=/dev/mtd0 bs=1 count=32 oseek=15728640

### DIFF
--- a/system/dd/dd_main.c
+++ b/system/dd/dd_main.c
@@ -65,6 +65,7 @@ struct dd_s
   int          outfd;      /* File descriptor of the output device */
   uint32_t     nsectors;   /* Number of sectors to transfer */
   uint32_t     skip;       /* The number of sectors skipped on input */
+  uint32_t     seek;       /* The number of sectors seeked on output */
   bool         eof;        /* true: The end of the input or output file has been hit */
   uint16_t     sectsize;   /* Size of one sector */
   uint16_t     nbytes;     /* Number of valid bytes in the buffer */
@@ -176,7 +177,7 @@ static void print_usage(void)
 {
   fprintf(stderr, "usage:\n");
   fprintf(stderr, "  %s if=<infile> of=<outfile> [bs=<sectsize>] "
-    "[count=<sectors>] [skip=<sectors>]\n", g_dd);
+    "[count=<sectors>] [skip=<sectors>] [seek=<sectors>]\n", g_dd);
 }
 
 /****************************************************************************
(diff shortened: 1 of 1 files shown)


yamt:
i feel it a bit inconsistent to have skip and oseek while you don't have iseek or seek.

xiaoxiang781216 (approved):

system/dd/dd_main.c
```diff
@@ -65,6 +65,7 @@ struct dd_s
   int          outfd;      /* File descriptor of the output device */
   uint32_t     nsectors;   /* Number of sectors to transfer */
   uint32_t     skip;       /* The number of sectors skipped on input */
+  uint32_t     oseek;      /* The number of sectors seeked on output */
```
xiaoxiang781216, line 68: let's remove o from ALL place

xiaoxiang781216:
> i feel it a bit inconsistent to have skip and oseek while you don't have iseek or seek.

Yes, dd from host name it `skip`:
```
dd --help
Usage: dd [OPERAND]...
  or:  dd OPTION
Copy a file, converting and formatting according to the operands.
```
  seek=N          skip N obs-sized blocks at start of output
  skip=N          skip N ibs-sized blocks at start of input
```
@dry-75 please use the same option as host version.

File: system/dd/dd_main.c (hunk 1 above)
Comment (dry-75) [reply system/dd/dd_main.c:68]: okey

**Comment by dry-75** [comment]
updated.

**Review by xiaoxiang781216 — APPROVED** [review]

File: system/dd/dd_main.c [hunk 2]
```diff
@@ -271,6 +276,18 @@ int main(int argc, FAR char **argv)
         }
     }
 
+  if (dd.seek)
+    {
+      ret = lseek(dd.outfd, dd.seek * dd.sectsize, SEEK_SET);
+      if (ret < -1)
```
yamt, line 282: although i guess this is a copy-and-paste from the above dd.skip case,
i don't understand this "ret < -1".
i suspect it should be "ret < 0" or "ret == -1".

xiaoxiang781216: @dry-75 please update skip case too?

dry-75: Yes c&p  ( on 'minimal change' principle...)
Yes looks like bug.

Yes to update.

dry-75: Update pushed with the above poitned out issue.

Would be good now to focus on actual change, not trying fix other issues in this (very small) feature update.

It is - this change - is usefull however, as I'm exercising it now.

xiaoxiang781216 (approved):

PR #2447, opened 2024-07-18
`i2ctool dev` command now uses an empty write header to scan the I2C bus

### LINKED ISSUE #994: i2c-tools is not able to detect pressure sensor MS5611
I decide to report it to prevent forgetting this issue. Although the device is not detected writing to it working correctly:
```
nsh> i2c set -a 77 1e
WROTE Bus: 0 Addr: 77 Subaddr: 00 Value: 1e
nsh> i2c set -a 77 a0
WROTE Bus: 0 Addr: 77 Subaddr: 00 Value: a0
nsh> i2c dump -a 77 2
DUMP Bus: 0 Addr: 77 Subaddr: --
00000000 08 04
```

linguini1:
## Summary

The `i2ctool dev` command now supports an additional command-line flag, `-z`, to the `dev` command. This flag will use a zero-byte write request over I2C as the scan instead of the default 1 byte read. This sometimes allows for discovery of more devices, such as the SHT41 temperature and humidity sensor, and the MS5611 barometric pressure sensor.

This PR closes #994.

## Impact

This should save some developers time debugging or wondering why their devices didn't show up (and trying to wrongly diagnose a hardware issue like myself :laughing:).

## Testing

Using the previous version of this command on an I2C bus with the following sensors:

- MS5611 barometric pressure sensor (2, at 0x76 and 0x77)
- SHT41 temperature and humidity sensor (0x44)
- MAXM10S GPS (0x42)
- M24C02 EEPROM (0x50)
- LSM6DSO32 IMU (0x6b)

The output of `i2c dev 00 77` was:

```console
nsh> i2c dev 00 77
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00: 00 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- 42 -- -- -- -- -- -- -- -- -- -- -- -- -- 
50: 50 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
60: -- -- -- -- -- -- -- -- -- -- -- 6b -- -- -- -- 
70: -- -- -- -- -- -- -- --                         
```

Even providing an argument to the register flag (`i2c dev -r 0 00 77`) resulted in the saem output.

With these proposed changes, the result when using the `-z` flag is:

```console
nsh> i2c dev -z 00 77
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- 42 -- 44 -- -- -- -- -- -- -- -- -- -- -- 
50: 50 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
60: -- -- -- -- -- -- -- -- -- -- -- 6b -- -- -- -- 
70: -- -- -- -- -- -- 76 77
```

Now all devices show up in the scan. Additionally, the general call address of `00` does not appear anymore, which is good as it is not a real device.

When the `-z` flag is not passed, the user is presented with a message indicating that some devices may not appear in the scan, and they can try a zero-byte write scan instead:

```console
nsh> i2c dev 00 77
NOTE: Some devices may not appear with this scan.
You may also try a scan with the -z flag to discover more devices using a zero-byte write request.
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00: 00 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- 42 -- -- -- -- -- -- -- -- -- -- -- -- -- 
50: 50 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
60: -- -- -- -- -- -- -- -- -- -- -- 6b -- -- -- -- 
70: -- -- -- -- -- -- -- --                         
```

### DIFF
--- a/system/i2c/i2c_common.c
+++ b/system/i2c/i2c_common.c
@@ -173,6 +173,10 @@ int i2ctool_common_args(FAR struct i2ctool_s *i2ctool, FAR char **arg)
         i2ctool->width = (uint8_t)value;
         return ret;
 
+      case 'z':
+        i2ctool->zerowrite = true;
+        return 1;
+
       default:
         goto invalid_argument;
     }
--- a/system/i2c/i2c_dev.c
+++ b/system/i2c/i2c_dev.c
@@ -64,6 +64,14 @@ int i2ccmd_dev(FAR struct i2ctool_s *i2ctool, int argc, FAR char **argv)
   saveaddr         = i2ctool->regaddr;
   i2ctool->regaddr = 0;
 
+  /* For backwards compatibility, the default behaviour while scanning will
+   * be to send a read request on the I2C bus. It is also possible to specify
+   * the use of a zero-byte write request instead, but this option will not
+   * be sticky.
+   */
+
+  i2ctool->zerowrite = false;
+
   /* Parse any command line arguments */
 
   for (argndx = 1; argndx < argc; )
@@ -126,6 +134,19 @@ int i2ccmd_dev(FAR struct i2ctool_s *i2ctool, int argc, FAR char **argv)
       goto errout;
     }
 
+  /* Display message warning user about some devices which don't appear
+   * unless using zero-byte write header.
+   */
+
+  if (!i2ctool->zerowrite)
+    {
+      i2ctool_printf(
+          i2ctool,
+          "NOTE: Some devices may not appear with this "
+          "scan.\nYou may also try a scan with the -z flag to "
+          "discover more devices using a zero-byte write request.\n");
+    }
+
   /* Probe each address */
 
   i2ctool_printf(i2ctool,
@@ -146,13 +167,22 @@ int i2ccmd_dev(FAR struct i2ctool_s *i2ctool, int argc, FAR char **argv)
 
           /* Set up data structures */
 
-          regaddr       = i2ctool->regaddr;
-
           msg.frequency = i2ctool->freq;
-          msg.addr      = addr;
-          msg.flags     = I2C_M_READ;
-          msg.buffer    = &regaddr;
-          msg.length    = 1;
+          msg.addr = addr;
+
+          if (i2ctool->zerowrite)
+            {
+              msg.flags = 0;
+              msg.buffer = NULL;
+              msg.length = 0;
+            }
+          else
+            {
+              regaddr = i2ctool->regaddr;
+              msg.flags = I2C_M_READ;
+              msg.buffer = &regaddr;
+              msg.length = 1;
+            }
 
           ret = i2cdev_transfer(fd, &msg, 1);
 
--- a/system/i2c/i2c_main.c
+++ b/system/i2c/i2c_main.c
@@ -149,6 +149,10 @@ static int i2ccmd_help(FAR struct i2ctool_s *i2ctool, int argc,
   i2ctool_printf(i2ctool,
                  "  [-r regaddr] is the I2C device register index (hex)."
                  "  Default: not used/sent\n");
+  i2ctool_printf(
+      i2ctool,
+      "  [-z] instructs the 'dev' command to scan the I2C bus by sending "
+      "zero-byte write headers (if the architecture supports it)\n");
 
   i2ctool_printf(i2ctool, "\nNOTES:\n");
 #ifndef CONFIG_DISABLE_ENVIRON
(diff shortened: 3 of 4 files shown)


xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

davids5 (changes requested):
Is this safe or will it over write values.
Do all the archs support 0 length writes and actually make it to the bus?

acassis (approved):

linguini1:
> @linguini1 did you test i2ctools on Linux and look at logic analyzer? Let's confirm Linux also do this way.

I won't have a change to access a logic analyzer for a little while. I had a look at the description for the Linux I2C tools here:

https://github.com/MLAB-project/i2c-tools/blob/master/tools%2Fi2cdetect.8

It seems they either use a single byte read (as was the previous implementation) or a write command, depending on the address by default. Although they do warn that scanning the bus can affect the I2C devices since there is no actual standard for scanning in the I2C protocol.

acassis:
> > @linguini1 did you test i2ctools on Linux and look at logic analyzer? Let's confirm Linux also do this way.
> 
> I won't have a change to access a logic analyzer for a little while. I had a look at the description for the Linux I2C tools here:
> 
> https://github.com/MLAB-project/i2c-tools/blob/master/tools%2Fi2cdetect.8
> 
> It seems they either use a single byte read (as was the previous implementation) or a write command, depending on the address by default. Although they do warn that scanning the bus can affect the I2C devices since there is no actual standard for scanning in the I2C protocol.
> 
> Perhaps a good compromise would be trying to read one byte from the address, and if that fails, attempt to send a 0 byte write?

I'm curious to know if Linux i2ctool is able to detect these sensors.

@linguini1 maybe to avoid David's worryings (in fact it could break I2C for arch that don't work when msgc is 0 or msg is NULL), I suggest you implementing your suggestion as special search case (like -z for zero byte reading).

linguini1:
> @linguini1I suggest you implementing your suggestion as special search case (like -z for zero byte reading).

Do you think it would be worth while for the i2ctool to have some `#ifdef` to check the architecture and disable the -z functionality for architectures which do not support the 0 byte write?

cederom:
> > @linguini1I suggest you implementing your suggestion as special search case (like -z for zero byte reading).
> 
> Do you think it would be worth while for the i2ctool to have some `#ifdef` to check the architecture and disable the -z functionality for architectures which do not support the 0 byte write?

+ verbose message in either situation could be helpful..?

acassis:
> > @linguini1I suggest you implementing your suggestion as special search case (like -z for zero byte reading).
> 
> Do you think it would be worth while for the i2ctool to have some `#ifdef` to check the architecture and disable the -z functionality for architectures which do not support the 0 byte write?

I don't think we can check the architecture now, because we don't know which arch will fail and which will work. My suggestion for now is adding the "-z" and maybe include a suggestion just before the scan start:

If your device didn't appear in the bus, try to use the -z option

linguini1:
@davids5 As far as I can tell, this should be safe. A 0 byte write means that any sensors which allow a write with no register address first specified should not perform the write regardless, since there is no data sent to be written. A device should see the stop condition and discard the write request, but I suppose it is possible for a poorly designed device to not handle this case.

Architectures which do not support a 0 byte write through their I2C driver will return some error, which will cause any device to just not appear. This is fine since with the addition of the `-z` option, the default behaviour is still a single byte read. The `-z` is a fallback for the user to check in case they do not see their expected device appearing in the scan.

Please let me know if there's anything I've overlooked!

xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216:
@davids5 the patch look fine since zero write is guarded by option in the new change.

davids5 (approved):
The option mitigates my concern. Thank you

PR #2451, opened 2024-07-27
examples/camera: Fix wrong device file path

SPRESENSE:
## Summary

In camera_main.c, the dev paths of capture_initialize() and open() are different. So open could always failed.

## Impact

Only camera example.

## Testing

Test the camera example works well with sprsense:camera config.


xiaoxiang781216 (approved):

xiaoxiang781216:
let's ignore msys2 ci error which is unrelated to this change.

PR #2490, opened 2024-08-18
greenhills support: fix the variables set but never used build warning

extinguish:
## Summary
apps/canutils: fix the build warning on greenhills compiler

### DIFF
--- a/canutils/candump/candump.c
+++ b/canutils/candump/candump.c
@@ -99,7 +99,6 @@
 const char col_on [MAXCOL][19] = {BLUE, RED, GREEN, BOLD, MAGENTA, CYAN};
 const char col_off [] = ATTRESET;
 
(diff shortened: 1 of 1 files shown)


extinguish:
@xuxin930 please help check, the origin "canutils/candump/candump.c" file itself cannot pass the nxstyle format check. If we want to submit modifications to this file, do we need to first modify the file to ensure it passes the format check, or do we need to bypass the format check for this file?

xiaoxiang781216:
> @xuxin930 please help check, the origin "canutils/candump/candump.c" file itself cannot pass the nxstyle format check. If we want to submit modifications to this file, do we need to first modify the file to ensure it passes the format check, or do we need to bypass the format check for this file?

@extinguish candump is designed to keep the original coding style, let's keep it as before.

PR #2501, opened 2024-08-19
netutils/webclient: Fix bug that the socket is not close

SPRESENSE:
## Summary
Fix to close socket when it fails to resolve hostname.

## Impact
webclient. 

## Testing

### DIFF
--- a/netutils/webclient/webclient.c
+++ b/netutils/webclient/webclient.c
@@ -428,7 +428,9 @@ static inline int wget_parsestatus(struct webclient_context *ctx,
               DEBUGASSERT(strlen(g_http10) == 8);
               DEBUGASSERT(strlen(g_http11) == 8);
 
-              if (ws->line[8] != ' ')  /* SP before the status-code */
+              /* SP before the status-code */
+
+              if (ws->line[8] != ' ')
                 {
                   return -EINVAL;
                 }
@@ -443,7 +445,9 @@ static inline int wget_parsestatus(struct webclient_context *ctx,
                   return -EINVAL;
                 }
 
-              if (*ep != ' ')  /* SP before reason-phrase */
+              /* SP before reason-phrase */
+
+              if (*ep != ' ')
                 {
                   return -EINVAL;
                 }
@@ -598,7 +602,10 @@ static inline int wget_parseheaders(struct webclient_context *ctx,
            */
 
           found = false;
-          if (ndx > 0) /* Should always be true */
+
+          /* Should always be true */
+
+          if (ndx > 0)
             {
               ninfo("Got HTTP header line%s: %.*s\n",
                     got_nl ? "" : " (truncated)",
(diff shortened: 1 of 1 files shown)


jerpelea (approved):

jerpelea (changes requested):
please fix the commit

xiaoxiang781216 (approved):

xiaoxiang781216:
Please fix the style issue, @SPRESENSE

SPRESENSE:
@jerpelea @xiaoxiang781216 
I have fixed the style issue. Please check.

xiaoxiang781216:
@acassis and @LuchianMihai does https://github.com/apache/nuttx/pull/12794 change good?

LuchianMihai:
Strictly regarding the [C Coding Standard](https://nuttx.apache.org/docs/latest/contributing/coding_style.html), is stated that the:

> Comments to the right of statements in C source files are discouraged.

> Indentation Comments should, typically, be placed before the code section to which they apply. The comment indentation should be the same as the follow indentation rules as the following code (if applicable).

Now there is nothing explicitly stating that there should not be comments to the right of __if statements__, but I saw no comment at the right of the statement throughout the examples given there (in C Coding Standard page).

My change will report an error if there is anything to the right of the statement, because checking that the comment is correctly formatted (as the comment should be indented accordingly) would be harder, and maybe not worth the time penalty for checking that corner case. Maybe we can report this as an warning?

xiaoxiang781216:
> Hi @xiaoxiang781216 , @acassis
> 
> Strictly regarding the [C Coding Standard](https://nuttx.apache.org/docs/latest/contributing/coding_style.html), is stated that the:
> 
> > Comments to the right of statements in C source files are discouraged.
> 
> > Indentation Comments should, typically, be placed before the code section to which they apply. The comment indentation should be the same as the follow indentation rules as the following code (if applicable).
> 
> Now there is nothing explicitly stating that there should not be comments to the right of **if statements**, but I saw no comment at the right of the statement throughout the examples given there (in C Coding Standard page).
> 

@patacongo @acassis what's your opinion?

> My change will report an error if there is anything to the right of the statement, because checking that the comment is correctly formatted (as the comment should be indented accordingly) would be harder, and maybe not worth the time penalty for checking that corner case. Maybe we can report this as an warning?

The code base is large, new check will break many codes which pass the ci before. So, we need be more conservative here.

xiaoxiang781216 (approved):

LuchianMihai:
Just run a quick search through the codebase:
> grep -r "if (.*)\ */\*" | wc -l         # This reports the total number of occurrences: 78

> grep -r "if (.*)\ */\*" -l | wc -l      # This reports the total number of independent files affected: 32

I'm only counting correctly formatted if statements which are followed by spaces until the beginning of a comment is found. We add a few more changes if I include incorrectly formatted if statements.

But nevertheless, it's quite a big surface area as it is, maybe my changes are not adding more value than adding issues.

PR #2511, opened 2024-08-20
util.c:Add mode when creating files in the wapi_save_config

wangchen61698:
## Summary
 no permission bit was provided when creating the file, resulting in the loss of permission to create the file and ultimately unable to read it
## Impact
wapi save config
## Testing
image


xiaoxiang781216:
let's ignore the stye violation from cJSON_XxxxYxxx.

xiaoxiang781216 (approved):

PR #2522, opened 2024-08-22
nshlib: use sh -c replace pthread detach when nsh background.

jasonbu:
## Summary
nsh_parse using nsh_arg_clone and pthread detach.
when pthread detached, and parent task exit, the sub-thread will quit directly.
lead to **memleak**. we should prefer use sub-process to do nsh background tasks.
this case can re-produce memory leak.
```C
{
  printf("Hello, World!!\n");
  system("sleep 1 &");
  return 0;
}
```

use 
```
sh -c
```
to do backtground command, will also compatible with kernel build.

## Impact
before fix, spawn feature used many times. and try run buildin_cmds & quit quickly, will cause a memory leak.
after fix, spawn required nsh all from nsh_spawn.c

## Testing
No more memleak in previouse case.
basic features still works as expect.
enable tmpfs and work with qemu-armv8a & qemu-armv7a.
```C
{
  char buffer[100];
  fgets(buffer, 100, stdin);

  printf("Hello, World!!\n");
  printf("stdin %s\n", buffer);
  return 0;
}
```
```bash
cd tmp
hello
hello > a
hello < a
hello >> a
hello >> a &
hello < a >> a &
```

### DIFF
--- a/nshlib/nsh_parse.c
+++ b/nshlib/nsh_parse.c
@@ -30,7 +30,6 @@
 #include <string.h>
 #include <errno.h>
 #include <debug.h>
-#include <pthread.h>
 #include <sched.h>
 #include <unistd.h>
 
@@ -126,19 +125,6 @@
  * Private Types
  ****************************************************************************/
 
-/* These structure describes the parsed command line */
-
-#ifndef CONFIG_NSH_DISABLEBG
-struct cmdarg_s
-{
-  FAR struct nsh_vtbl_s *vtbl;      /* For front-end interaction */
-  int fd_in;                        /* FD for output redirection */
-  int fd_out;                       /* FD for output redirection */
-  int argc;                         /* Number of arguments in argv */
-  FAR char *argv[MAX_ARGV_ENTRIES]; /* Argument list */
-};
-#endif
-
 /* This structure describes the allocation list */
 
 #ifdef HAVE_MEMLIST
@@ -172,13 +158,6 @@ static void nsh_alist_add(FAR struct nsh_alist_s *alist,
                           FAR struct nsh_alias_s *alias);
 static void nsh_alist_free(FAR struct nsh_vtbl_s *vtbl,
                            FAR struct nsh_alist_s *alist);
-#endif
-
-#ifndef CONFIG_NSH_DISABLEBG
-static void nsh_releaseargs(struct cmdarg_s *arg);
-static pthread_addr_t nsh_child(pthread_addr_t arg);
-static struct cmdarg_s *nsh_cloneargs(FAR struct nsh_vtbl_s *vtbl,
-               int fd_in, int fd_out, int argc, FAR char *argv[]);
 #endif
 
 static int nsh_saveresult(FAR struct nsh_vtbl_s *vtbl, bool result);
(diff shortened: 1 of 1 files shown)


jasonbu:
looks like the nuttx/boards/arm64/qemu/qemu-armv8a/configs/knsh/defconfig
did not export the **task_spawn** API ?

xiaoxiang781216:
> looks like the nuttx/boards/arm64/qemu/qemu-armv8a/configs/knsh/defconfig did not export the **task_spawn** API ?

task_spawn doesn't exist in kernel build.

nshlib/nsh_parse.c
```diff
@@ -441,97 +421,25 @@ static void nsh_alist_free(FAR struct nsh_vtbl_s *vtbl,
...
-}
-#endif
-
 /****************************************************************************
  * Name: nsh_child
  ****************************************************************************/
 
 #ifndef CONFIG_NSH_DISABLEBG
-static pthread_addr_t nsh_child(pthread_addr_t arg)
+static CODE int nsh_child(int argc, FAR char *argv[])
```
pkarashchenko, line 429:
```suggestion
static int nsh_child(int argc, FAR char *argv[])
```

jasonbu: just reverted the nsh_parse change. here only need thread to do bg stuff , no spawn required.

builtin/exec_builtin.c
```diff
@@ -24,11 +24,11 @@
 
 #include <nuttx/config.h>
 
-#include <spawn.h>
 #include <fcntl.h>
 #include <errno.h>
 #include <debug.h>
 
+#include "nshlib/nshlib.h"
```
anchao, line 31: nshlib is just a library for the NuttX shell. I think it is not a good idea to reverse-use nshlib's APIs in builtin/system. They are not part of the shell.

anchao: For example, if the application does not enable nshlib, then it cannot use the system() call? This does not make sense.

xiaoxiang781216: builtin apps couldn't depend on nshlib, but system already depends on shell since it launches nsh internally, this behavior also explicitly specified in spec:
https://pubs.opengroup.org/onlinepubs/009695399/functions/system.html

jasonbu:
there is a memleak case can reproduce in current code.

```C
{
  printf("Hello, World!!\n");
  system("111 &");
  // system("sleep 5 &");
  return 0;
}
```
put these code in hello_main.c

and check free before & after call hello,
can see obviously memleak.

direct cause is nsh_parse.c when do bg thread np, the foreground task quit quickly, cause memleak, 

if disable https://github.com/apache/nuttx-apps/blob/89dd0c29e2e3c5dba42240b08ff4ca407b5848c6/nshlib/nsh_parse.c#L752-L853 can see no memleak any more.

already fixed in private fork. but look like not fixed by this commit, still doing root cause find.

jasonbu:
```bash
Configuration/Tool: rv-virt/knsh_paging
2024-08-29 14:26:28
/github/workspace/sources/apps /github/workspace/sources/nuttx
Warning: fpu.c:55:8: warning: #warning "FPU test not built; Only available in the flat build (CONFIG_BUILD_FLAT)" [-Wcpp]
   55 | #      warning "FPU test not built; Only available in the flat build (CONFIG_BUILD_FLAT)"
      |        ^~~~~~~
```
should not because of my commit?

builtin/exec_builtin.c
jasonbu: exec_builtin.c already revert.

system and nsh all use spawn to do background execute.

nshlib/nsh_parse.c
```diff
@@ -822,43 +657,56 @@ static int nsh_execute(FAR struct nsh_vtbl_s *vtbl,
...
-      pthread_attr_setschedpolicy(&attr, SCHED_NSH);
-      pthread_attr_setschedparam(&attr, &param);
-
-      /* Execute the command as a separate thread at the appropriate
-       * priority.
-       */
-
-      ret = pthread_create(&thread, &attr, nsh_child, (pthread_addr_t)args);
-      if (ret != 0)
+      spawn_argv[0] = "nsh_child";
```
xiaoxiang781216, line 660: can't find nsh_child in kernel mode

nshlib/nsh_parse.c
```diff
@@ -174,13 +161,6 @@ static void nsh_alist_free(FAR struct nsh_vtbl_s *vtbl,
                            FAR struct nsh_alist_s *alist);
 #endif
 
-#ifndef CONFIG_NSH_DISABLEBG
-static void nsh_releaseargs(struct cmdarg_s *arg);
```
xiaoxiang781216, line 178: remove np_bg process from nsh_builtin.c, nsh_fileapps. and nsh_timcmds.c too.

nshlib/nsh_parse.c
```diff
@@ -607,6 +498,171 @@ static int nsh_saveresult(FAR struct nsh_vtbl_s *vtbl, bool result)
...
+ *
+ * Returned Value:
+ *   >0          The bg spawn created task pid.
+ *   <0          The bg spawn failed.
+ *
+ ****************************************************************************/
+
+#ifndef CONFIG_NSH_DISABLEBG
+static
+int nsh_bg_spawn(FAR const char *appname, main_t main,
```
xiaoxiang781216, line 515: let's call nsh_builtin/nsh_fileapp directly

nshlib/nsh_parse.c
```diff
@@ -443,105 +423,16 @@ static void nsh_alist_free(FAR struct nsh_vtbl_s *vtbl,
...
-      for (i = 0; i < argc; i++)
-        {
-          ret->argv[i] = strdup(argv[i]);
-        }
-    }
-
-  return ret;
-}
+#  ifndef CONFIG_BUILD_KERNEL
+#    define nsh_child_task nsh_system
```
xiaoxiang781216, line 432: add nsh_system to apps/builtin/builtin_list.c

nshlib/nsh_parse.c
jasonbu: nsh_builtin & nsh_fileapps will handle bg independently. to minum the influence,did not change the nsh_builtin & nsh_fileapps code.

nshlib/nsh_parse.c
jasonbu: no more required.

xiaoxiang781216 (approved):

PR #2571, opened 2024-09-13
ppp: Fix wrong lcp_state handling on peer reconnection

lzungri:
If a PPP peer disconnects and then tries to reconnect it will send an 'LCP configure request' packet. The code that handles that scenario seems to be clearing the wrong `lcp_state` flag (`LCP_RX_UP` instead of `LCP_TX_UP`) and thus the nuttx ppp client will keep sending IPCP packets which are rightfully dropped by the new peer since it is still in the LCP negotiation phase.

### DIFF
--- a/netutils/pppd/lcp.c
+++ b/netutils/pppd/lcp.c
@@ -135,10 +135,11 @@ void lcp_rx(struct ppp_context_s *ctx, uint8_t * buffer, uint16_t count)
       /* In case of new peer connection */
 
       ipcp_init(ctx);
-      
-      /* Clear LCP state to keep it in negotiation phase. LCP_TX_UP will be re-set
-       * once a CONF_ACK is received. */
-      
+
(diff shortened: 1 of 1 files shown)


cederom:
* Thank you @lzungri very nice catch :-)
* You have tested the patch and it fixes the issue in practice with various different setups? Please add short testing / verification description / confirmation.
* Comment needs reformat to pass CI, see https://nuttx.apache.org/docs/latest/contributing/coding_style.html.

lzungri:
> You have tested the patch and it fixes the issue in practice with various different setups? Please add short testing / verification description / confirmation.

@cederom I have a custom version of the pppd code from nuttx so I have only tested it in my specific setup, which doesn't involve running it under nuttx OS.
My testing involved connecting a linux `pppd` daemon to my custom nuttx code and confirming that the connection is re-established smoothly after restarting the linux `pppd` process.

xiaoxiang781216 (approved):

PR #2572, opened 2024-09-14
apps/mbedtls-alt: Provide system alternative implementation for mbedtls

ThePassionate:
## Summary
1. provide aes-cmac alternative implemetation.
The computing process and software algorithm of AESCMAC are already provided in the crypto module of NuttX, and the mbedtls-alt interface is provided for the convenience of the upper layer, which can be used to directly replace the implementation of the mbedtls interface. Drive capability is used imperceptibly, and that replacement has gone through the MBEDTLS internal selftest case.
2. provide hardware random for PSA.
There is a conflict between using the PSA random number interface provided by MBEDTLS and trying to use AES to drive the acceleration function. This is due to the fact that the software random number source of the PSA is a global variable, and the driving ability of AES is exposed through fd, and there will be cross-process call errors. So there is a need to provide the ability for PSA to also use random number driven.
## Impact
1. The Nuttx encrypted driver capability can be used to independently implement the MBEDTLS interface, which can provide a general interface without using the driver capability imperceptibly
2. Provides the ability for mbedTLS to use hardware random number drivers, but does not directly verify driver performance. In order to solve the conflict between the PSA global software random number entropy source and the MBEDTLS AES using cryptographic drivers
## Testing
ci & mbedselftest

`ap> mbedselftest cmac

  AES 128 CMAC subkey #1: passed
  AES 128 CMAC subkey #2: passed
  AES 128 CMAC subkey #3: passed
  AES 128 CMAC subkey #4: passed
  AES 128 CMAC #1: passed
  AES 128 CMAC #2: passed
  AES 128 CMAC #3: passed
  AES 128 CMAC #4: passed`


cederom (changes requested):
Thank you @ThePassionate :-)
1. Please update PR Summary and git commit messages with more detailed description why the change is needed, what it does and how.
2. Please update PR Impact section - there are two important impacts one is provision of new encryption method (why it is called alternative if it also comes from mbedtls?) and second is providing entropy from `/dev/random` that may be considered insecure (inadequate entropy). What hardware random number generator did you use? Did you verify its entropy?
3. Please update PR Testing section with test run logs.
4. There are build errors, I have restarted CI lets see if that helps.

ThePassionate:
> * Please update PR Testing section with test run logs.

thanks, 3 and 4 i will verify in local and update after.

cederom:
Thank you @ThePassionate :-)

xiaoxiang781216 (approved):

PR #2577, opened 2024-09-14
module build:replace BUILD_LOADABLE with MODULES

xuxin930:
## Summary

this patch is relate to: https://github.com/apache/nuttx/pull/13452

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -209,35 +209,35 @@ define RENAMEMAIN
 endef
 
 $(RAOBJS): %.s$(SUFFIX)$(OBJEXT): %.s
-	$(if $(and $(CONFIG_BUILD_LOADABLE),$(AELFFLAGS)), \
+	$(if $(and $(CONFIG_MODULES),$(AELFFLAGS)), \
 		$(call ELFASSEMBLE, $<, $@), $(call ASSEMBLE, $<, $@))
 
 $(CAOBJS): %.S$(SUFFIX)$(OBJEXT): %.S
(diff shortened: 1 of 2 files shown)


cederom (commented):
@patacongo please advise on nomenclature change :-)

patacongo:
> @patacongo please advise on nomenclature change :-)

I am not so much involved these days but in the past, configuration names began with CONFIG_ (of course) then what was being modified ( for example, NET_ for anyting affecting the network, and then the remainder describing what is being configured.

So by those old rules, CONFIG_MODULES would be wrong.  Since it affects the build system, it should be CONFIG_BUILD_MODULES.

What is configured?  BUILD_ system
What is controlled in the build system?  MODULE support.

cederom (changes requested):
Lets finish the nomenclature change discussion resolution first.

masayuki2009:
@xuxin930  @xiaoxiang781216 
I noticed that the PR does not work with kconfig-frontends but works with kconfiglib.
Do you mean that kconfig-frontends is no longer supported?

xuxin930:
> @xuxin930 @xiaoxiang781216 I noticed that the PR does not work with kconfig-frontends but works with kconfiglib. Do you mean that kconfig-frontends is no longer supported?

hi @masayuki2009  i add both MODULE config and modules option in this patch https://github.com/apache/nuttx/pull/13532. 
which both work in two Kconfig implementation library.

PR #2593, opened 2024-09-20
examples/userfs: Add initial value for d_ino

no1wudi:
## Summary

Currently implementations of dirent.h do not have a field for the inode number, but it is required by POSIX.

For better compatibility with POSIX based applications, it is should be added even if it is not actually implemented.

Refer to https://github.com/apache/nuttx/pull/13556, should be merged together.

## Impact

* `include/dirent.h`.
* `include/nuttx/fs/hostfs.h`.
* Makes implementation compatible with POSIX.

## Testing

CI

### DIFF
--- a/examples/userfs/userfs_main.c
+++ b/examples/userfs/userfs_main.c
@@ -140,19 +140,19 @@ static char g_file3_data[UFSTEST_FILE3_MXSIZE] = "This is file 3";
 static struct ufstest_file_s g_rootdir[UFSTEST_NFILES] =
 {
     {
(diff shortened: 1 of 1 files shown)


examples/userfs/userfs_main.c
```diff
@@ -140,19 +140,19 @@ static char g_file3_data[UFSTEST_FILE3_MXSIZE] = "This is file 3";
 static struct ufstest_file_s g_rootdir[UFSTEST_NFILES] =
 {
```
cederom, line 141: compiler warnings :-(

PR #2594, opened 2024-09-21
nshlib:enable O_CLOEXEC expliciti to avoid potential fd leak

gneworld:
## Summary
  leaking here means fork/vfork will duplicate fd without O_CLOEXEC flag to the child process.
## Impact
  nsh
## Testing
  cortex M55


xiaoxiang781216 (approved):

cederom (approved):
Thank you @gneworld :-)

PR #2596, opened 2024-09-22
resetcause:restore to factory, factory to factory_inquiry

gneworld:
## Summary
 add restore to indicate "restore the factory settings" and  the previous "factory" represents the confirmation of restoring to factory settings.
## Impact
 resetcause
## Testing
 CI test

### DIFF
--- a/nshlib/nsh_syscmds.c
+++ b/nshlib/nsh_syscmds.c
@@ -118,6 +118,7 @@ static FAR const char * const g_resetflag[] =
   "panic",
   "bootloader",
   "recovery",
+  "restore",
   "factory",
   NULL
 };


cederom (changes requested):
* I like the idea but this changes existing logic and may impact existing applications. 
* After introducing `restore` existing applications may not catch the `factory` event and skip the factory defaults reset.
* `factory` is already used to trigger the "perform reset to factory defaults".
* After successful factory defaults reset one more reset can be triggered with `NULL`  and device should land in a state that should be handled by a firmware accordingly based on existing configuration.
* `restore` name is a bit ambiguous and may be confusing, its not self explanatory, maybe another keyword can be used?
* Maybe something like `defaults` would fit here better? `factory` as before would cause factory defaults reset, and then call reset with `defaults` mark to note everything is default or needs to be set default (this state should be also unambiguous)?
* Documentation needs an update in this area. Will try to work more with documentation in a free moment.
* Long story short if new resetflag is to create new and unique state that would be fine, but if it changes existing behavior this will break compatibility and we do not want that.
* I would like to hear what other devs think about this idea :-)

xiaoxiang781216 (approved):

xiaoxiang781216:
> * I like the idea but this changes existing logic and may impact existing applications.
> * After introducing `restore` existing applications may not catch the `factory` event and skip the factory defaults reset.
> * `factory` is already used to trigger the "perform reset to factory defaults".
> * After successful factory defaults reset one more reset can be triggered with `NULL`  and device should land in a state that should be handled by a firmware accordingly based on existing configuration.
> * `restore` name is a bit ambiguous and may be confusing, its not self explanatory, maybe another keyword can be used?
> * Maybe something like `defaults` would fit here better? `factory` as before would cause factory defaults reset, and then call reset with `defaults` mark to note everything is default or needs to be set default (this state should be also unambiguous)?
> * Documentation needs an update in this area. Will try to work more with documentation in a free moment.
> * Long story short if new resetflag is to create new and unique state that would be fine, but if it changes existing behavior this will break compatibility and we do not want that.
> * I would like to hear what other devs think about this idea :-)

@cederom this change is to align with nuttx side definition:
https://github.com/apache/nuttx/blob/master/include/sys/boardctl.h#L461-L471

acassis (approved):

PR #2597, opened 2024-09-22
nshlib: enable O_CLOEXEC explicit to avoid potential fd leak

gneworld:
## Summary
 leaking here means fork/vfork will duplicate fd without O_CLOEXEC flag to the child process.
## Impact
 nshlib
## Testing
 ci test


cederom (approved):
Thank you @gneworld :-)

PR #2599, opened 2024-09-22
github/workflows: add PR size labeler

raiden00pl:
## Summary
Add workflow that assigns labels based on the PR's size.
The same as on kernel side: https://github.com/apache/nuttx/pull/13542

change from https://github.com/apache/nuttx/pull/13558 is not yet ported, because the apps repo requires different labels.

## Impact
Each PR will have a label assigned depending on its size:
xs_label: 'Size: XS', max size = 10
s_label: 'Size: S', max size = 100
m_label: 'Size: M', max size = 500
l_label: 'Size: L', max size = 1000
xl_label: 'Size: XL', > 1000

## Testing
not tested, should work the same as on NuttX repo

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,39 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


cederom (approved):
Thank you @raiden00pl :-)

PR #2608, opened 2024-09-27
build(enhancement):enhance apps to support independent out-of-tree builds

xuxin930:
## Summary

Add a `PREFIX` extension to your application compilation.

by defining the `PREFIX` variable and setting it to an external path, 
`make -C /some/app` can be called independently to build a separate application out_of_tree.

modify Make.defs and Application.mk so that their dependent variables can be overwritten.

Usage:
```makefile
# in some app's Makefile
PREFIX = out/of/tree/dir/

``` 
compiled products and intermediate products can be separated from the source tree

## Impact

forward compatibility

## Testing

CI build

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -61,6 +61,13 @@ ifeq ($(BUILD_MODULE),y)
   LDLIBS += $(COMPILER_RT_LIB)
 endif
 
+# Apps compilation can achieve out-of-tree intermediate products
+# by specifying "PREFIX" to a directory in its own Makefile.
+# Default value is NULL,
+# Make sure the out-of-tree directory exists and ends with $(DELIM) when setting it.
+
+PREFIX ?=
+
 SUFFIX ?= $(subst $(DELIM),.,$(CWD))
 
 PROGNAME := $(subst ",,$(PROGNAME))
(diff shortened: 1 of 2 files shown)


nuttxpr:



The PR Summary is missing some details:

* **What functional part of the code is being changed?**  Specify which files (Make.defs, Application.mk) are modified.
* **How does the change exactly work?** You mention overwriting dependent variables. Briefly explain how this is achieved.

**Impact:**

* **Is new feature added? Is existing feature changed?**  This seems like a new feature. Specify that.
* **Impact on user:**  While you say "forward compatibility", be explicit: Will existing users be affected? Will they need to change anything in their workflow? 
* **Impact on build:**  You mention changes to the build process. Briefly explain what changes users might see.
* **Impact on documentation:** A new feature like this likely requires documentation updates. Mention if you've included those.

**Testing:**

* **Build Host(s):** Be specific. List the OS, CPU architecture, and compiler used for testing.
* **Target(s):** List the specific architectures and boards you tested on.
* **Testing logs:**  "CI build" is not sufficient. Provide snippets of logs showing the build process working as expected before and after the change, specifically highlighting the new `PREFIX` functionality. 

**In short:** The PR description is a good start but needs more details, especially in the Impact and Testing sections, to meet the NuttX requirements.

cederom (changes requested):
* Very cool @xuxin930 :-)
* Please provide documentation update with detailed description of this feature :-)

xiaoxiang781216 (approved):

cederom:
* Does that allow full out of tree app build with no source tree modification?
* Can similar approach be applied (in separate PR) for nuttx repo and out of tree boards - that is build for board:config that is out of tree without touching the source tree? This seems highly demanded feature :-)
* Instead Makefile modification maybe env variable can be used too? `PREFIX` has some standard use, so maybe `NXAPP_PREFIX` would be better here? This part needs extra explanation in the docs as it may be widely used :-)

xuxin930:
> * Does that allow full out of tree app build with no source tree modification?
> * Can similar approach be applied (in separate PR) for nuttx repo and out of tree boards - that is build for board:config that is out of tree without touching the source tree? This seems highly demanded feature :-)
> * Instead Makefile modification maybe env variable can be used too? `PREFIX` has some standard use, so maybe `NXAPP_PREFIX` would be better here? This part needs extra explanation in the docs as it may be widely used :-)

thanks @cederom 
I think our `CMake` base can achieve a complete out-of-tree build.

This patch serves the `Makefile` base, and is used for independent apps to be compiled into **separate** .a static libraries for external use.

For example, directly calling like:
```shell
# in my applications Makefile i can set BIN:=custom.a

make -C /my/app/path context
make -C /my/app/path depend
make -C /my/app/path all
```

cederom (approved):
Thank you @xuxin930 :-)

PR #2613, opened 2024-09-27
fb_main: fix draw_rect24

jianglianfang:
## Summary
it should be written into act_fbmem  rather than fbmem in draw_rect24
## Impact
fb exmaple with 24 bpp
## Testing
run fb


cederom (approved):
Thank you @jianglianfang :-)

xiaoxiang781216 (approved):

PR #2615, opened 2024-09-27
greenhills support: fix green hills toolchain build warnings

extinguish:
## Summary
1. fix the variable set but never used warning
2. fix pointless comparison of unsigned integer with zero warning

## Impact
has no impact on current implementation

## Testing
1. has pass the ostest
2. has tested on gcc and ghs compiler

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 *.a
+*.dbo
+*.dba
 *.adb
 *.asm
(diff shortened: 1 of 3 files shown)


cederom (approved):
Thank you @extinguish :-)

xiaoxiang781216 (approved):

PR #2660, opened 2024-10-09
Application.mk:no need to add DEPPATH by default

xuxin930:
## Summary

After:https://github.com/apache/nuttx/pull/13958

current directory is always added in mkdeps.c for dep-path there is no need to manually add multiple times

## Impact

Generation of Make.dep

## Testing
CI build

### DIFF
--- a/Application.mk
+++ b/Application.mk
@@ -147,9 +147,6 @@ RUSTELFFLAGS ?= $(RUSTFLAGS)
 DELFFLAGS ?= $(DFLAGS)
 SWIFTELFFLAGS ?= $(SWIFTFLAGS)
 
-DEPPATH += --dep-path .
-DEPPATH += --obj-path .
-
 VPATH += :.
 
 # Targets follow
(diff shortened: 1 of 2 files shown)


nuttxpr:



**PR Requirements Met?** 

Partially.

**What's Good:**

* **Testing:**  Referencing CI build is sufficient.
* **Impact:** Correctly identifies the impacted area (Make.dep generation).

**What Needs Improvement:**

* **Summary:**
    * **Why?**  The summary lacks a clear explanation of *why* this change is necessary. What problem does it solve? Was there a bug, inefficiency, or redundancy?
    * **What/How?**  The summary needs more detail on *what* part of `mkdeps.c` is changed and *how* the change addresses the issue. Simply stating "no need to manually add multiple times" isn't specific enough. 
* **Impact:**
    * **User Impact:**  Will this change how users interact with NuttX or its build system? Be explicit.
    * **Other Impacts:**  The PR should address all other impact categories, even if the answer is "NO." This demonstrates thoroughness. 

**Example of a More Complete Summary:**

"This PR addresses an inefficiency in `mkdeps.c` where the current directory was being added multiple times to the dependency path. This was redundant and could potentially lead to slower build times. The change modifies [specific function/logic in `mkdeps.c`] to ensure the current directory is added only once, streamlining the dependency generation process."

cederom:
Does CI build error reveal that this is in fact required? :-)

xuxin930:
> Does CI build error reveal that this is in fact required? :-)

@cederom thanks, yes I think you are right.I'm reconsidering this change

PR #2684, opened 2024-10-11
coremark/Makefile: avoid crc function multiple definition

CV-Bowen:
## Summary
/data/project/oh2/rel-4.0/prebuilts/gcc/linux/arm/bin/../lib/gcc/arm-none-eabi/11.3.1/../../../../arm-none-eabi/bin/ld: crc16_sw.c.data.project.oh2.rel-4.0.external.zblue.zblue_1.o (symbol from plugin): in function `crc16':
(.text+0x0): multiple definition of `crc16'; core_util.c.data.project.oh2.rel-4.0.apps.benchmarks.coremark_1.o (symbol from plugin):(.text+0x0): first defined here

## Impact
coremark

## Testing
compile with coremark

### DIFF
--- a/benchmarks/coremark/Makefile
+++ b/benchmarks/coremark/Makefile
@@ -85,6 +85,11 @@ else
 CFLAGS += ${DEFINE_PREFIX}ITERATIONS=0
 endif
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
let's ignore macOS ci temp break.

PR #2689, opened 2024-10-11
wapi_json_load: the buf need add null character

lipengfei28:
wapi_json_load: the buf need add null character

If the buf no null character, the strlen acces buf maybe out of bounds


GUIDINGLI (approved):

xiaoxiang781216 (approved):

xiaoxiang781216:
let's ignore macOS ci temp break and format check.

xiaoxiang781216:
let's ignore macOS ci temp break.

PR #2697, opened 2024-10-12
nsh/fsutils: fix build break if CONFIG_NSH_DISABLE_PKILL not disabled

anchao:
## Summary

nsh/fsutils: fix build break if CONFIG_NSH_DISABLE_PKILL not disabled

```
arm-none-eabi/bin/ld: apps/libapps.a(nsh_proccmds.c.obj): in function `cmd_pkill':
apps/nshlib/nsh_proccmds.c:861:(.text.cmd_pkill+0x38): undefined reference to `nsh_getpid'
```

Signed-off-by: chao an <anchao@lixiang.com>

## Impact

N/A

## Testing

ci-check

### DIFF
--- a/nshlib/nsh.h
+++ b/nshlib/nsh.h
@@ -1394,7 +1394,7 @@ int nsh_foreach_direntry(FAR struct nsh_vtbl_s *vtbl, FAR const char *cmd,
  *
  ****************************************************************************/
 
-#if defined(CONFIG_FS_PROCFS) && !defined(CONFIG_NSH_DISABLE_PIDOF)
+#ifdef CONFIG_FS_PROCFS
 ssize_t nsh_getpid(FAR struct nsh_vtbl_s *vtbl, FAR const char *name,
                    FAR pid_t *pids, size_t count);
 #endif
--- a/nshlib/nsh_fsutils.c
+++ b/nshlib/nsh_fsutils.c
@@ -63,7 +63,7 @@ struct getpid_arg_s
  *
  ****************************************************************************/
 
-#if defined(CONFIG_FS_PROCFS) && !defined(CONFIG_NSH_DISABLE_PIDOF)
+#ifdef CONFIG_FS_PROCFS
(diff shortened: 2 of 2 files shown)


xiaoxiang781216 (approved):

nshlib/nsh.h
```diff
@@ -1394,7 +1394,8 @@ int nsh_foreach_direntry(FAR struct nsh_vtbl_s *vtbl, FAR const char *cmd,
  *
  ****************************************************************************/
 
-#if defined(CONFIG_FS_PROCFS) && !defined(CONFIG_NSH_DISABLE_PIDOF)
+#if defined(CONFIG_FS_PROCFS) && \
+  (!defined(CONFIG_NSH_DISABLE_PIDOF) || !defined(CONFIG_NSH_DISABLE_PKILL))
```
xiaoxiang781216, line 1398: let's remove `(!defined(CONFIG_NSH_DISABLE_PIDOF) || !defined(CONFIG_NSH_DISABLE_PKILL))`? to avoid the new break in the future.

anchao: Done

xiaoxiang781216 (approved):

cederom (approved):
Thank you @anchao :-)

We have similar issue with `free` command not enabled when `PROCFS` is not enabled. I mean why not use positive selection only (i.e. `ENABLE_FREE` and `ENABLE_PROCFS`) instead one thing is disabled and other thing is enabled and that makes things more complicated than should be :-)

xiaoxiang781216:
> Thank you @anchao :-)
> 
> We have similar issue with `free` command not enabled when `PROCFS` is not enabled. I mean why not use positive selection only (i.e. `ENABLE_FREE` and `ENABLE_PROCFS`) instead one thing is disabled and other thing is enabled and that makes things more complicated than should be :-)

I guess that the original design wants to enable the command by default for newbie.

xiaoxiang781216 (approved):

PR #2718, opened 2024-10-14
apps/benchmarks/dhrystone: Fix compilation errors

lipengfei28:
Error details:
dhrystone/v2.2/dry.c:714:3: error: 'Int_2_Loc' may be used uninitialized [-Werror=maybe-uninitialized]
  714 |   fprintf (stderr, "Int_2_Loc:           %d\n", Int_2_Loc);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
dhrystone/v2.2/dry.c:543:25: note: 'Int_2_Loc' was declared here
  543 |   REG   One_Fifty       Int_2_Loc;
      |                         ^~~~~~~~~

### DIFF
--- a/benchmarks/dhrystone/CMakeLists.txt
+++ b/benchmarks/dhrystone/CMakeLists.txt
@@ -44,7 +44,10 @@ if(CONFIG_BENCHMARK_DHRYSTONE)
     endif()
   endif()
 
-  set(CFLAGS -DMSC_CLOCK)
+  set(CFLAGS
+      -DMSC_CLOCK -Wno-maybe-uninitialized -Wno-implicit-int
+      -Wno-strict-prototypes -Wno-implicit-function-declaration
+      -Wno-return-type)
   set(SRCS dhrystone/v2.2/dry.c dhrystone/v2.1/dhry_2.c)
 
   nuttx_add_application(
--- a/benchmarks/dhrystone/Makefile
+++ b/benchmarks/dhrystone/Makefile
@@ -29,6 +29,7 @@ MODULE    = $(CONFIG_BENCHMARK_DHRYSTONE)
 
 CFLAGS   += -DMSC_CLOCK -Wno-implicit-int -Wno-strict-prototypes
 CFLAGS   += -Wno-implicit-function-declaration -Wno-return-type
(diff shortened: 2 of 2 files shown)


benchmarks/dhrystone/CMakeLists.txt
```diff
@@ -44,7 +44,10 @@ if(CONFIG_BENCHMARK_DHRYSTONE)
     endif()
   endif()
 
-  set(CFLAGS -DMSC_CLOCK)
+  set(CFLAGS
```
cederom, line 47: Wouldn't it be better to fix the code than disable warnings? :-)

xiaoxiang781216: dhrystone is the 3rd party library, no critical change should upstream to the official project, so I would suggest you push the change to them, and we can update the download to new version.

xiaoxiang781216 (approved):

benchmarks/dhrystone/CMakeLists.txt
xiaoxiang781216: Since the different project has different warning setting. It's useless to fix the warning message or even wrong, before the fix gets accepted by the project maintainer.

cederom: Not really, we already patch the release code `malloc` and boolean statements, so we modify the code anyway, we can also patch the uninitialized variables. Instead we want to disable build warnings that may hide potential problems in future.

This kind of patching is common also in FreeBSD Ports. Then patches are supposed to land in the upstream so they are not needed in future releases :-)

I can see some flags were present before, so okay, we may include them :-) I would go for patching the upstream too :-)

cederom (approved):
Thank you @lipengfei28 :-)

benchmarks/dhrystone/CMakeLists.txt
xiaoxiang781216: we made this modification because there is no way to disable it from -Wno-xxx option or it's too dangerous to disable it. For the minor warning, it's better to avoid patching the code before the change upstreamed.

PR #2735, opened 2024-10-15
nshlib:Add macro restrictions to code that uses floating point numbers

anjiahao1:
## Summary
nshlib:Add macro restrictions to code that uses floating point numbers
## Impact
nsh fs
## Testing
qmue

some paltform can't use flaot


cederom (approved):
Thank you @anjiahao1 :-)

xiaoxiang781216 (approved):

PR #2791, opened 2024-10-28
resmonitor: add showinfo/filldisk/fillcpu/fillmem

jasonbu:
## Summary
add tools from 
after enable TESTING_RESMONITOR, use showinfo -i 2 can show cpuload to syslog every two seconds.

```text
showinfo [-i <interval>] [-p <pid>] [-a] [-d <dir>] [-l <logpath>] [-k <logpath>]
     -i: Set the refresh interval (in seconds), i.e., how often to print the result. Default is 1 second.
     -p: Set the pid of the process to monitor. If not set, only system information is shown.
     -a: Effective only if -p is set. If not set, the displayed CPU and memory usage reflect only the specified pid
         (excluding child threads or child tasks). If enabled, the displayed CPU and memory usage include the
         total for the process and all its child threads (i.e., the same task group).
     -d: View file system information. Total size represents the total size of the file system, free size shows
         the remaining size of the file system (including superblock, inode blocks, etc.), avail size is the 
         available size for users (data blocks only), and block size is the block size.
     -l: Output CPU and memory results to the file specified by logpath.
     -k: Output file system results to the file specified by logpath.

Examples:
     - To view system CPU and memory usage: `showinfo -i 1 &`
     - To monitor the process with pid 1000: `showinfo -p 1000 &`
     - To check the size of the /data file system: `showinfo -d /data &`
```

## Impact
default no impact. 
optional enable to add tool command

## Testing
Ci-test, local cortex-a7 smp board.

### DIFF
--- a/testing/resmonitor/CMakeLists.txt
+++ b/testing/resmonitor/CMakeLists.txt
@@ -1,0 +1,63 @@
+# ##############################################################################
+# apps/testing/resmonitor/CMakeLists.txt
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  The ASF licenses this
+# file to you under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.  You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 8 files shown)


cederom (approved):
Thank you @jasonbu :-)

Would it be possible to provide several words of functionalities description and use case examples in the Documentation in a free moment in a separate PR? :-)

jasonbu:
> Thank you @jasonbu :-)
> 
> Would it be possible to provide several words of functionalities description and use case examples in the Documentation in a free moment in a separate PR? :-)

hi @cederom 

updated the content of the pr, and add three most use case command sample in our side. maybe will update more document/samples later.

PR #2796, opened 2024-10-29
ymodem:fix bug,When sending a retransmission, HEAD is overwritten.

anjiahao1:
## Summary
ymodem bug fix
## Impact
ymodem
## Testing

Create some scenarios that require retransmission, and trigger retransmission to enable normal transmission


cederom (approved):
Thank you @anjiahao1 :-)

PR #2809, opened 2024-10-31
boot: add NuttX bootloader with update and recovery support

michallenc:
## Summary
This commit adds NuttX based bootloader with the support for image update and recovery if not confirmed. The algorithm utilizes three flash partitions: primary (image runs from this area), secondary and tertiary. Secondary and tertiary areas are used for update upload and recovery.

The update is performed by simple copy from update area to primary area with recovery being created in recovery area if not already present. Once image is confirmed by the user, the image in update area is confirmed as well, update area becomes recovery area and vice versa. This means the recovery is always present (except for the first update) and subsequent updates just copy image from update to primary. This makes the update significantly faster and more considerable to flash wear while keeping the recovery/revert possibility.

A header (aligned to flash's erase size) must be added to the beginning of the image. Python script nximage.py can be used to prepend this header to built binary. The algorithm also uses one erase page at the end of a partition (partition, not image!) to store flags used to indicate image confirm status and to detect update/recovery partitions. Any program uploading update image to the update partition has to erase this page for the boot to work correctly!

The algorithm implementation is based on a patch initially developed for MCUboot project but rejected by the project's maintainers

https://github.com/mcu-tools/mcuboot/pull/1902

## Impact
New bootloader that can be used with NuttX. Implemented as a standalone application, does not affect current upstream (both core and apps) if not configured.

## Testing
Tested on SAMv7 MCU.

### DIFF
--- a/boot/nxboot/CMakeLists.txt
+++ b/boot/nxboot/CMakeLists.txt
@@ -1,0 +1,32 @@
+# ##############################################################################
+# apps/boot/nxboot/CMakeLists.txt
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  The ASF licenses this
+# file to you under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.  You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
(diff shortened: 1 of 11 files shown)


michallenc:
The complete documentation is in a separate pull request to main repository, see [docs: add nxboot documentation entry](https://github.com/apache/nuttx/pull/14584)

xiaoxiang781216:
please correct the style issue

michallenc:
> please correct the style issue

I am not sure if this is a style issue.

```
 e45cec88c boot: add NuttX bootloader with update and recovery support
../nuttx/tools/checkpatch.sh -m -g 34ca696b869366de87ff2ab06a21bd1858b527a0..HEAD
../nuttx/tools/checkpatch.sh: line 87: black: command not found
../nuttx/tools/checkpatch.sh: line 88: flake8: command not found
../nuttx/tools/checkpatch.sh: line 89: isort: command not found
../nuttx/tools/checkpatch.sh: line 92: isort: command not found
```

This seems more like an incorrect configuration of checkpatch that is missing python modules. Local checkpatch passes fine on my PC.

PR #2823, opened 2024-11-05
nsh_vars: Remove rogue sched_unlock()

pussuw:
## Summary
There is no corresponding sched_lock() for this and therefore this causes random crashes.

Found when running a long init script which utilizes shell variables in SMP mode.

## Impact
Remove sched_unlock() from a place that does not use sched_lock()
## Testing
Downstream MPFS target which utilizes nsh_vars.

### DIFF
--- a/nshlib/nsh_vars.c
+++ b/nshlib/nsh_vars.c
@@ -346,7 +346,6 @@ int nsh_unsetvar(FAR struct nsh_vtbl_s *vtbl, FAR const char *name)
         }
     }
 
-  sched_unlock();
   return ret;
 }
 #endif


cederom (commented):
Thank you @pussuw :-)

cederom (approved):
Thank you @pussuw :-)

cederom:
@patacongo is it safe to remove this abandoned `shed_unlock()` or the `sched_lock()` is missing? From a history of file I can see it is here from the beginning of time :-)

pussuw:
There should be no reason to do mutual exclusion (sched_lock) here as pstate is allocated per shell and only the shell itself can add / remove vars.

patacongo:
> @patacongo is it safe to remove this abandoned `shed_unlock()` or the `sched_lock()` is missing?

Ville has it right.  It might have had meaning at one time.  Originally, NSH was part of the nuttx repository and was moved out, losing the history that far back.  The apps were separated at NuttX-5.19 since they are not architecturally a part of the OS.

NOTE also:  sched_lock() does not work does not provide mutual exclusion in the case of SMP; the locked code section can still be re-entered from other CPUs.  So it is wrong as is on many levels.

cederom:
Thank you @patacongo and @pussuw this PR can be safely merged :-)

xiaoxiang781216 (approved):

PR #2824, opened 2024-11-05
nxboot: fix incorrect confirm state for directly flashed image

michallenc:
## Summary
API function `nxboot_get_confirm` was returning incorrect value if primary image was flashed directly into the embedded flash (this image does not have a tail, but is automatically considered as valid and stable based on the header magic value).

## Impact
Fixes return value for `nxboot_get_confirm` function.

## Testing
SAMv7 MCU custom board.


cederom (approved):
Thank you @michallenc :-)

xiaoxiang781216 (approved):

PR #2835, opened 2024-11-08
build: increase FetchContent timeout to 120 seconds for LVGL

Jasinsky:
## Summary

LVGL 9.2.1 archive is ~70MB. Over poor connections, it may take longer to download it. This renders CMake build unusable in such environments.

## Impact

CMake build for LVGL configurations is now usable over poor network connections.

## Testing

Local CMake configure step for 
```
cmake .. -G Ninja -DBOARD_CONFIG=sim:lvgl_fb
```


cederom (approved):
Thank you @Jasinsky :-)

PR #2858, opened 2024-11-18
tools: Optimize size by wasm-opt

no1wudi:
## Summary
wasm-opt is a tool from binaryen that can do
further optimization for Wasm module, to reduce
the size of bytecode.
## Impact
New experimental feature, for wasm build
## Testing
Local machine

### DIFF
--- a/tools/Wasm.mk
+++ b/tools/Wasm.mk
@@ -42,7 +42,12 @@ define LINK_WASM
 	    $(if $(RETVAL), \
 	        $(error wasm build failed for $(PROGNAME).wasm) \
 	    ) \
-		$(call WAMR_AOT_COMPILE) \
+	    $(eval RETVAL=$(shell $(WASI_SDK_PATH)/wasm-opt -Oz -o $(BINDIR)$(DELIM)wasm$(DELIM)$(PROGNAME).wasm \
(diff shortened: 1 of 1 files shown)


cederom (commented):
Thank you @no1wudi :-)

Do we want to have it as option? You say this tool is experimental so maybe its good to make it optional just in case it produces broken binary? :-)

Does CMake build is in sync with Make build both will have this optimization enabled?

tools/Wasm.mk
```diff
@@ -42,7 +42,12 @@ define LINK_WASM
 	    $(if $(RETVAL), \
 	        $(error wasm build failed for $(PROGNAME).wasm) \
 	    ) \
-		$(call WAMR_AOT_COMPILE) \
+	    $(eval RETVAL=$(shell $(WASI_SDK_PATH)/wasm-opt -Oz -o $(BINDIR)$(DELIM)wasm$(DELIM)$(PROGNAME).wasm \
```
yamt, line 45: * as far as i know, wasi-sdk doesn't have wasm-opt.
* see also: https://github.com/llvm/llvm-project/pull/95208

no1wudi: Yes, it's from binaryen, I'll update the tool path

cederom:
ping :-)

PR #2861, opened 2024-11-20
bin:Copy debug info to a separate folder

anjiahao1:
## Summary
bin:Copy debug info to a separate folder
## Impact
Reduce bin size
## Testing

sim:sotest with debug symbol

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@
 .kconfig
 *.lock
 /bin
+/bin_debug
 /boot_romfsimg.h
 /external
 /Kconfig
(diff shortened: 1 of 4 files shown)


cederom (approved):
Thank you @anjiahao1 good idea :-)

.gitignore
```diff
@@ -28,6 +28,7 @@
 .kconfig
 *.lock
 /bin
+/bin_debug
```
xiaoxiang781216, line 31: should we name it symbols?

cederom: Initially I wrote about two possibilities but then deleted because `bin_debug` is quite self-explanatory :-P

1. `bin/release`, `bin/devel`, `bin/debug` so all is under `bin/` and then subdiretory (more elegant).
2. `bin/release-$TIMESTAMP`, `bin/debug-$TIMESTAMP` so all is just under `bin/` (less elegant).

xiaoxiang781216: OK, let's keep it.

xiaoxiang781216 (approved):

PR #2863, opened 2024-11-22
testing/ostest/signest.c: fix the range of signals

yamt:
## Summary

Note that MAX_SIGNO is a valid signal number.

## Testing

esp32s3-devkit:smp

### DIFF
--- a/testing/ostest/signest.c
+++ b/testing/ostest/signest.c
@@ -144,7 +144,7 @@ static FAR void *waiter_main(FAR void *arg)
   act.sa_flags   = 0;
 
   sigemptyset(&act.sa_mask);
-  for (i = 1; i < MAX_SIGNO; i += 2)
+  for (i = 1; i <= MAX_SIGNO; i += 2)
     {
       if (signest_catchable(i))
         {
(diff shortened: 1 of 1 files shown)


nuttxpr:



This PR appears to be **missing crucial information** to meet the NuttX requirements.  While it provides a skeletal structure, it lacks the necessary details in almost every section.

**Missing Information:**

* **Summary:**  The summary only mentions `MAX_SIGNO` being a valid signal number.  This doesn't explain *why* this change is needed, *what* code is changed, or *how* the change works. Is this a bug fix? A new feature?  What problem does it solve?
* **Impact:**  All impact sections are empty.  This makes it impossible to assess the ramifications of the change.  Even if the impact is minimal for some categories, explicitly stating "NO" with a brief justification is better than leaving it blank.
* **Testing:** While a target is listed (`esp32s3-devkit:smp`),  the critical "before" and "after" testing logs are empty.  Without these logs, there's no evidence that the change works as intended or that the problem being addressed is actually fixed.  The build host details are also missing.

**To make this PR compliant, the author needs to provide:**

* A complete summary explaining the context, the change, and its purpose.
* Fill out *all* impact sections, even if the answer is "NO."  Justify "NO" answers briefly.  If the answer is "YES," provide a detailed explanation.
* Include actual testing logs demonstrating the behavior before and after the change. Specify the build host details (OS, CPU, compiler).


Without this information, reviewers cannot effectively evaluate the PR.  It's likely to be rejected in its current state.

xiaoxiang781216 (approved):

PR #2929, opened 2025-01-07
mtetest: add thread to control mte test separately

W-M-R:
## Summary

After thread A holds the semaphore, it first accesses it safely. After releasing it, thread B closes MTE for unsafe access. After accessing, it switches back to thread A for unsafe access. At this time, an error should be reported. log:
```
Process 1 holding lock
Process 2 holding lock
Process 1 holding lock again
default_fatal_handler: (IFSC/DFSC) for Data/Instruction aborts: synchronous tag check fault arm64_exception_handler: CurrentEL: MODE_EL1
arm64_exception_handler: ESR_ELn: 0x96000011
arm64_exception_handler: FAR_ELn: 0xf00000040441700 arm64_exception_handler: ELR_ELn: 0x402ee5f4
print_ec_cause: DABT (current EL)
print_ec_cause: Data Abort taken without a change in Exception level
```


nuttxpr:



This PR description does **not** fully meet the NuttX requirements. Here's why and how to fix it:

**Missing Information in Summary:**

* **Why is this change necessary (fix, update, new feature)?**  It's implied this is a bug fix, but should be stated explicitly.  e.g., "Fixes a race condition leading to memory corruption when accessing shared resources after disabling Memory Tagging Extension (MTE)."
* **What functional part of the code is being changed?**  Which semaphore implementation?  Which part of the MTE handling? Be specific (e.g., "the `sem_wait` and `sem_post` functions in `semaphore/sem_wait.c` and the MTE configuration in `arch/<architecture>/src/<path_to_mte_code>`").
* **How does the change exactly work (what will change and how)?** This is the most crucial missing piece.  Explain the code changes made to prevent the race condition.  For example, "Added checks to ensure MTE is not disabled while other threads hold the semaphore.  This is accomplished by..." (and then describe the mechanism, e.g., a reference count, a new lock, etc.).

**Incomplete Impact Assessment:**

The entire Impact section is missing.  You *must* address each point:

* **Is new feature added? Is existing feature changed?**  This is a bug fix, so "Existing feature changed."
* **Impact on user:**  Potentially YES if the user was relying on the faulty behavior (unlikely, but must be considered).  Otherwise NO. Explain if YES.
* **Impact on build:**  Likely NO, but state explicitly.
* **Impact on hardware:**  Possibly YES if changes were architecture-specific. Specify the affected architecture(s). Otherwise NO.
* **Impact on documentation:**  Maybe YES if documentation needs to be updated to reflect the fix.  Specify if YES.
* **Impact on security:** YES, this fixes a security issue related to MTE. Explain the security implications of the original bug and how the fix mitigates it.
* **Impact on compatibility:**  Likely NO, but state it clearly.
* **Anything else to consider:**  Mention any potential performance implications of the changes.


**Incomplete Testing Information:**

* **Build Host(s):** Provide details about your build environment.
* **Target(s):** Specify the target architecture and board configuration used for testing.
* **Testing logs before change:** The provided log snippet shows the error, but you need more context. What steps reproduce the error reliably? Include the complete log from a test case demonstrating the issue.
* **Testing logs after change:**  Show a log from the same test case demonstrating that the fix prevents the error. The logs should ideally show the expected, correct behavior after the change.


**Example of Improved Summary and Impact:**

```
## Summary

Fixes a race condition leading to memory corruption when accessing shared resources after disabling Memory Tagging Extension (MTE). This bug occurs when one thread releases a semaphore and another thread subsequently disables MTE, leading to potential unsafe accesses to shared memory protected by the semaphore. The change modifies the `sem_wait` and `sem_post` functions in `semaphore/sem_wait.c` and the MTE configuration in `arch/arm64/src/kernel/arm64_mte.c`.  A reference counter is added to track the number of threads actively using the semaphore and prevent MTE from being disabled while the counter is greater than zero.  Related [NuttX Issue](put issue link here).

## Impact

* Is new feature added? Is existing feature changed? Existing feature changed.
* Impact on user: NO. The fix addresses an internal bug and should not impact user applications directly.
* Impact on build: NO.
* Impact on hardware: YES.  This fix is specific to the ARM64 architecture with MTE enabled.
* Impact on documentation: YES.  The documentation will be updated to explain the fix and its implications for MTE usage.
* Impact on security: YES.  The original bug could lead to memory corruption and potential security vulnerabilities. This fix prevents such scenarios by ensuring MTE remains active while shared resources are being accessed.
* Impact on compatibility: NO.
* Anything else to consider: The addition of the reference counter might have a negligible performance impact on semaphore operations.
```


By providing this complete information, reviewers can understand the changes, assess the impact, and verify the fix effectively, leading to a smoother and faster review process. Remember to replace placeholders like "put issue link here" with actual values.

xiaoxiang781216 (approved):

PR #2931, opened 2025-01-07
The placement of cases under testing is a bit messy, and cases of the same type need to be merged

txy-21:
## Summary
Merge the test cases under testing into the same folder.
## Impact
The cases under apps/testing will be separated and placed in different folders.
## Testing
Local self-test can be done through.

PR #2961, opened 2025-01-21
apps/testing:move {drivertest fftest irtest monkey nand_sim pcitest sd_bench sd_stress sensortest} folders to the new driver folder

txy-21:
## Summary

This is one step that merge test case as discussed in https://github.com/apache/nuttx-apps/pull/2931.
Create driver folder and move {drivertest fftest irtest monkey nand_sim pcitest sd_bench sd_stress sensortest}  folders to the new driver folder.

## Impact

This PR only affects the apps/testing folder, changing the location of some subfolders under this folder.
new folder : driver
move : drivertest fftest irtest monkey nand_sim pcitest sd_bench sd_stress sensortest

## Testing

CI test.


nuttxpr:



No, this PR does not fully meet the NuttX requirements.  While it provides a summary and describes the impact, it lacks crucial information.

Here's what's missing:

* **Summary:**  While the summary explains *what* is being done, it lacks the *why*.  What problem does moving these test folders solve?  How does this relate to the linked PR?  Is this a cleanup effort? Is it preparing for future work?  The summary needs to justify the change. It also needs the corresponding NuttX issue reference if one exists.
* **Impact:** While the PR mentions affected files, it doesn't explicitly answer all the required impact questions with YES/NO followed by details if applicable. Even if the answer is NO, stating it explicitly ensures reviewers don't have to guess.  For example:
    * Impact on user: NO
    * Impact on build: Potentially YES (if any build scripts relied on the old paths.  Even if unlikely, this needs to be considered).
    * Impact on hardware: NO
    * Impact on documentation:  YES (Any documentation referencing these test applications will need updating).
    * Impact on security: NO
    * Impact on compatibility: Potentially YES (if external scripts or tools rely on the old paths).
* **Testing:** "CI test" is insufficient.  The requirements explicitly request testing logs *before* and *after* the change. Even if the CI passes, providing snippets of relevant output demonstrating the tests still function as expected would greatly improve the PR.  Also, details of the build host and target are missing.


In short, the PR needs more detail and explicit answers to the checklist items to be considered complete.  While the changes themselves might be simple, demonstrating due diligence in the PR description is essential for proper review and acceptance.

cederom (approved):
Thank you @txy-21 :-)

cederom:
Didn't we want `drivers/`? :-P

txy-21:
@cederom ok, change driver to dirvers

testing/driver/monkey/Make.defs
```diff
@@ -1,5 +1,5 @@
 ############################################################################
-# apps/testing/monkey/Make.defs
+# apps/testing/driver/monkey/Make.defs
```
xiaoxiang781216, line 2: monkey is more like a command line, let's move to testing/tools/ or testing/utils

testing/driver/sd_bench/Makefile
```diff
@@ -1,5 +1,5 @@
 ############################################################################
-# apps/testing/sd_bench/Makefile
+# apps/testing/driver/sd_bench/Makefile
```
xiaoxiang781216, line 2: benchmark shold move to apps/benchmark

txy-21:
@cederom Thanks for your suggestion.About irtest,sd_bench,monkey and sensortest folders, @xiaoxiang781216 has a different suggestion.  His opinion is somewhat different from what we discussed in https://github.com/apache/nuttx-apps/pull/2931. What do you think of his suggestion?

testing/driver/irtest/cmd.cxx
```diff
@@ -1,5 +1,5 @@
 /****************************************************************************
- * apps/testing/irtest/cmd.cxx
+ * apps/testing/driver/irtest/cmd.cxx
```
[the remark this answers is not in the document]
cederom: Hmm, we need some documentation on this test [1], but isn't this IRDA (infrared comms) driver test? If so drivers looks good location. If not then we need some documentation :D 

[1] https://nuttx.apache.org/docs/latest/applications/testing/irtest/index.html

testing/driver/sensortest/sensortest.c
```diff
@@ -1,5 +1,5 @@
 /****************************************************************************
- * apps/testing/sensortest/sensortest.c
+ * apps/testing/driver/sensortest/sensortest.c
```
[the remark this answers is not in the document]
cederom: For me it looks like various sensors drivers test so also good location? :-)

testing/driver/monkey/Make.defs
cederom: This also looks like input device drivers test so drivers location looks good? We talked about this case in https://github.com/apache/nuttx-apps/pull/2931 :-)

testing/driver/irtest/cmd.cxx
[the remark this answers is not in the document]
GUIDINGLI: We should clear what we test, not the test method.
For example the mediatool should move to media folder, not tools folder.
The bttool should move to bt folder, not tools folder either.

testing/driver/sd_bench/Makefile
cederom: We talked about this these sd tests in https://github.com/apache/nuttx-apps/pull/2931.. looks like good test case for sd card drivers in terms of benchmark and stress-test? But no strong opinion on this one maybe benchmark is better location for this sd_bench :-)

cederom:
> @txy-21: @cederom Thanks for your suggestion.About irtest,sd_bench,monkey and sensortest folders, @xiaoxiang781216 has a different suggestion. His opinion is somewhat different from what we discussed in #2931. What do you think of his suggestion?

Always good to discuss :-) Maybe better solution can be found that way :-)

testing/driver/monkey/Make.defs
xiaoxiang781216: please reference this wiki: https://en.wikipedia.org/wiki/Monkey_testing
to understand why we call it monkey, not input.

lupyuen (approved):

testing/driver/sd_bench/Makefile
xiaoxiang781216: test should report pass/fail clearly, so we can run it and check the result automatically. sd_bench doesn't have this key characteristic.

testing/driver/sensortest/sensortest.c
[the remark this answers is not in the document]
cederom: Ah, okay, this is something @raiden00pl mentioned - we don't have clear split of testing utilities and test scenarios and I understand the point even better now. For now we have everything mixed in the `apps/testing/` location. Should we keep testing utilities in `apps/testing` and separate test scenarios to `apps/tests` next?

testing/driver/monkey/Make.defs
cederom: As above, looks like need to split testing utilities and tests scenarios?

testing/driver/sd_bench/Makefile
cederom: Yup, again need of testing utils and tests cases separation?

testing/driver/irtest/cmd.cxx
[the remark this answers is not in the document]
xiaoxiang781216: Whether the program belong to a tool or test? The key point is that whether this program can report pass/fail at the end. All programs I point out don't report pass/fail and impossible to add.

cederom:
Okay it seems @raiden00pl remarks on separating `apps/testing` into testing utilities (something to perform the tests) and `apps/tests` test scenarios (list of specific tests that returns PASS/FAIL) makes more sense now :-)

What would be better / possible approach for now @txy-21 @xiaoxiang781216 ?
1. `apps/testing` cleanup by organizing existing stuff by function first then identify test cases and move them next to `apps/tests` based on categories created in `apps/testing`?
2. Use current time/work and split `apps/testing/function` and `apps/tests/function` right away?

testing/driver/irtest/cmd.cxx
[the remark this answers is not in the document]
cederom: I think the solution here is to keep testing utilities in `apps/testing` and move out test scenarios that can be batched and expects PASS/FAIL output to `apps/tests` ? Whatever tool can be used in more general way can be moved to a function specific location? :-)

Question is whether we want / can only re-organize existing `apps/testing` now or do some bigger cleanup straight away? :-)

xiaoxiang781216:
@GUIDINGLI suggest we move the tools/utils to apps/system and benchmark to apps/benchmark.

testing/driver/sensortest/sensortest.c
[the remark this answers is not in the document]
raiden00pl: > From my point of view, test should verify the output and report pass/fail clearly, but sensortest don't do any verification.

I think this definition from @xiaoxiang781216 is good. If program returns clear pass/fail output then we can treat it as a test. Otherwise its something else

raiden00pl:
@xiaoxiang781216 I thought the same - we can move some tools back to `apps/system` and some to `apps/benchmark`. But what about `cmoka` , `unity` and `fff`? These fits quite well in `apps/testing`. Maybe something like this:
- programs that return pass/fail output - `apps/tests`
- testing frameworks to be used by user- `apps/testing`
- performance tests  - `apps/benchmark`
- small testing apps that doesn't fit above - `apps/system`

cederom:
I liked the @txy-21 idea that some testing utils for instance targeted for drivers tests could be placed in the `apps/testing/drivers` (or similar).. but looking at `apps/system` most programs there are testing utilities already.. the same with benchmarks :-) So `apps/testing` should be reserved for testing frameworks only according to original concept?

For sure @xiaoxiang781216 initial remarks here on what are "tests" in terms to return PASS/FAIL are in align with previous discussion @raiden00pl suggestions and I fully agree these should be separated and `apps/tests` seems best / standard / acclaimed location :-)

xiaoxiang781216:
> @xiaoxiang781216 I thought the same - we can move some tools back to `apps/system` and some to `apps/benchmark`. But what about `cmoka` , `unity` and `fff`? These fits quite well in `apps/testing`. Maybe something like this:
> 
> * programs that return pass/fail output - `apps/tests`
> * testing frameworks to be used by user- `apps/testing`

I prefer that testing frameworks and test case under one top level folder(either tests or testing is fine for me), we can move the testing framework to utils or frameworks etc.

> * performance tests  - `apps/benchmark`
> * small testing apps that doesn't fit above - `apps/system`

LGTM.

PR #2970, opened 2025-01-23
Add vela alsa lib interface

yangyalei:
## Summary
Add vela alsa lib interface

## Impact
Introducing ALSA compatible audio subsystem on NuttX.

## Testing
Use alsa lib interface play music in allwinnertech&BES platform

### DIFF
--- a/audioutils/alsa-lib/bits_convert.c
+++ b/audioutils/alsa-lib/bits_convert.c
@@ -1,0 +1,172 @@
+/****************************************************************************
+ * apps/audioutils/alsa-lib/bits_convert.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <stdio.h>
+
+#include "bits_convert.h"
+#include "pcm_util.h"
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+int bitsconv_init(FAR struct bitsconv_data **bc, int channels,
+                  snd_pcm_format_t src_format, snd_pcm_format_t dst_format,
+                  int buf_size)
+{
+  if (*bc != NULL)
+    {
+      return 0;
+    }
+
+  /* support s16->s32, s32->s16 only */
+
+  if ((src_format != SND_PCM_FORMAT_S16_LE &&
+       src_format != SND_PCM_FORMAT_S32_LE) ||
+      (dst_format != SND_PCM_FORMAT_S16_LE &&
+       dst_format != SND_PCM_FORMAT_S32_LE))
+    {
+      SND_ERR("format not supported! src: %d, dst: %d", src_format,
+              dst_format);
+      return -ENOTSUP;
+    }
+
+  *bc = malloc(sizeof(struct bitsconv_data));
+  if (*bc == NULL)
+    {
+      SND_ERR("bitsconv_init failed");
+      return -ENOMEM;
+    }
+
+  (*bc)->bitsconv_buf = malloc(buf_size);
+  if ((*bc)->bitsconv_buf == NULL)
+    {
+      free(*bc);
+      *bc = NULL;
+      SND_ERR("malloc bitsconv_buf failed");
+      return -ENOMEM;
+    }
+
+  (*bc)->channels = channels;
+  (*bc)->src_format = src_format;
+  (*bc)->dst_format = dst_format;
+  (*bc)->buf_size = buf_size;
+
+  return 0;
+}
+
+int bitsconv_process(FAR struct bitsconv_data *bc, FAR const void *in_data,
+                     int in_size, FAR void **out_data, FAR int *out_size)
+{
+  int out_len;
+  int frames;
+  int ch;
+
+  if (!bc)
+    {
+      SND_ERR("bitsconv_init not run");
+      return -EPERM;
+    }
+
+  if (bc->src_format == SND_PCM_FORMAT_S16_LE &&
+      bc->dst_format == SND_PCM_FORMAT_S32_LE)
+    {
+      FAR const int16_t *src;
+      FAR int32_t *dst;
+
+      out_len = in_size * bc->channels * sizeof(int32_t);
+      if (out_len > bc->buf_size)
+        {
+          SND_ERR("bitsconv out_buf too small");
+          return -EPERM;
+        }
+
+      for (ch = 0; ch < bc->channels; ch++)
+        {
+          frames = in_size;
+          src = (FAR const int16_t *)in_data;
+          dst = (FAR int32_t *)bc->bitsconv_buf;
+          while (frames-- > 0)
+            {
+              *(dst + ch) = (int32_t)((*(src + ch)) << 16);
+              src += bc->channels;
+              dst += bc->channels;
+            }
+        }
+    }
+  else if (bc->src_format == SND_PCM_FORMAT_S32_LE &&
+           bc->dst_format == SND_PCM_FORMAT_S16_LE)
+    {
+      FAR int32_t *src;
+      FAR int16_t *dst;
+
+      out_len = in_size * bc->channels * sizeof(int16_t);
+      if (out_len > bc->buf_size)
+        {
+          SND_ERR("bitsconv out_buf too small");
+          return -EPERM;
+        }
+
+      for (ch = 0; ch < bc->channels; ch++)
+        {
+          frames = in_size;
+          src = (FAR int32_t *)in_data;
+          dst = (FAR int16_t *)bc->bitsconv_buf;
+          while (frames-- > 0)
+            {
+              *(dst + ch) = (int16_t)((*(src + ch)) >> 16);
+              src += bc->channels;
+              dst += bc->channels;
+            }
+        }
+    }
+  else
+    {
+      SND_ERR("unknown format");
+      return -ENOTSUP;
+    }
+
+  *out_data = bc->bitsconv_buf;
+  *out_size = in_size;
+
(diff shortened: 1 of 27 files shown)


nuttxpr:



No, this PR does not fully meet the NuttX requirements. While it provides a summary and mentions testing, it lacks crucial details.  Here's a breakdown of what's missing:

* **Summary:** While it states "Add vela alsa lib interface," it doesn't explain *why* this change is necessary.  Is it a new feature?  Does it fix a bug?  What part of the code is affected?  How does the "vela alsa lib interface" work and what changes does it introduce?  Any related NuttX issues should also be linked.
* **Impact:** Simply stating "nop" is insufficient.  Each impact item needs to be addressed explicitly with "NO" or "YES," and if "YES," a description is required. Even if there's no impact, stating "NO" for each item clarifies that these aspects were considered.
* **Testing:**  The testing description is vague.  It mentions playing music on specific platforms, but doesn't provide details about the build host or the specific boards/configurations used.  Critically, it's missing the "before" and "after" testing logs, which are essential for demonstrating the change's effect.  What constitutes "works as intended"?  What specific tests were run?

**To meet the requirements, the PR needs to be revised to include the missing information.** For example, the impact section should look like this (even if there's no impact):

```
* Is new feature added? YES (Adds the vela alsa lib interface)
* Impact on user (will user need to adapt to change)? NO
* Impact on build (will build process change)? NO
* Impact on hardware (will arch(s) / board(s) / driver(s) change)? YES (Affects Allwinnertech and BES platforms - please specify which boards and how)
* Impact on documentation (is update required / provided)? YES (Documentation should be added explaining the vela alsa lib interface and how to use it)
* Impact on security (any sort of implications)? NO
* Impact on compatibility (backward/forward/interoperability)? NO
* Anything else to consider? NO
```

The testing section needs actual logs and specific build/target details.

audioutils/alsa-lib/pcm_dmix.h
```diff
@@ -0,0 +1,95 @@
...
+  int period_bytes;
+  int buffer_bytes;
+
+  snd_pcm_uframes_t period_frames;
+  snd_pcm_uframes_t buffer_frames;
+
+  FAR struct bitsconv_data *bc;
+  FAR struct bitsconv_data *bc_s16;
+  FAR struct chmap_data *cm;
+  FAR SpeexResamplerState *resampler;
```
cederom, line 82: This comes from ALSA and needs to stay that way @yangyalei ?

cederom:
Thank you @yangyalei amazing work!! :-)

audioutils/alsa-lib/pcm_dmix.h
yangyalei: > This comes from ALSA and needs to stay that way @yangyalei ?

This is just a simple way to support format convert and resampling... This is not the standard implementation of ALSA; it's just a basic implementation.

cederom: Our format check complains here because of mixed case.. but I guess this name is enforced by some standard API and we should not change that to maintain compatibility? :-)

yangyalei: > Our format check complains here because of mixed case.. but I guess this name is enforced by some standard API and we should not change that to maintain compatibility? :-)

I used libspeexdsp (BSD license, tested performance is quite good) for resampling, but the interface definitions of libspeexdsp follow camel case naming convention. I have discussed this with @xiaoxiang781216,  it's OK.

raiden00pl (changes requested):
what is the license of this code? is it somehow based on `alsa-lib` ? Is it GPL? 

If this code is GPL then it can't be merged. GPL licensed code cannot be hosted in ASF projects.

jerpelea:
> > what is the license of this code? is it somehow based on `alsa-lib` ? Is it GPL?
> > If this code is GPL then it can't be merged. GPL licensed code cannot be hosted in ASF projects.
> 
> it's rewrite from scratch, but follow alsa-lib public API, just like nuttx v4l2 driver framework: https://github.com/apache/nuttx/blob/master/include/sys/videoio.h The implementation is totally different from https://git.alsa-project.org/?p=alsa-lib.git;a=tree.

I will run a scan after merge

audioutils/alsa-lib/bits_convert.c
```diff
@@ -0,0 +1,172 @@
...
+        {
+          SND_ERR("bitsconv out_buf too small");
+          return -EPERM;
+        }
+
+      for (ch = 0; ch < bc->channels; ch++)
+        {
+          frames = in_size;
+          src = (const int16_t *)in_data;
+          dst = (int32_t *)bc->bitsconv_buf;
```
raiden00pl, line 109: FAR

audioutils/alsa-lib/bits_convert.c
```diff
@@ -0,0 +1,172 @@
...
+              src += bc->channels;
+              dst += bc->channels;
+            }
+        }
+    }
+  else if (bc->src_format == SND_PCM_FORMAT_S32_LE &&
+           bc->dst_format == SND_PCM_FORMAT_S16_LE)
+    {
+      int32_t *src;
+      int16_t *dst;
```
raiden00pl, line 122: FAR

audioutils/alsa-lib/bits_convert.c
```diff
@@ -0,0 +1,172 @@
...
+        {
+          SND_ERR("bitsconv out_buf too small");
+          return -EPERM;
+        }
+
+      for (ch = 0; ch < bc->channels; ch++)
+        {
+          frames = in_size;
+          src = (int32_t *)in_data;
+          dst = (int16_t *)bc->bitsconv_buf;
```
raiden00pl, line 135: FAR

audioutils/alsa-lib/channels_map.c
```diff
@@ -0,0 +1,223 @@
...
+
+  out_len = in_size * dst_ch * sizeof(int16_t);
+  if (out_len > cm->buf_size)
+    {
+      SND_ERR("cm out_buf too small");
+      return -EPERM;
+    }
+
+  in = (const int16_t *)in_data;
+  out = (int16_t *)cm->chmap_buf;
```
raiden00pl, line 94: FAR

audioutils/alsa-lib/channels_map.c
```diff
@@ -0,0 +1,223 @@
...
+
+  out_len = in_size * dst_ch * sizeof(int32_t);
+  if (out_len > cm->buf_size)
+    {
+      SND_ERR("cm out_buf too small");
+      return -EPERM;
+    }
+
+  in = (const int32_t *)in_data;
+  out = (int32_t *)cm->chmap_buf;
```
raiden00pl, line 147: FAR

audioutils/alsa-lib/pcm.c
```diff
@@ -0,0 +1,808 @@
...
+  snd_pcm_dump_sw_setup(pcm);
+
+  return 0;
+}
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+const char *snd_pcm_name(snd_pcm_t *pcm)
```
raiden00pl, line 107: FAR and in other places

audioutils/alsa-lib/pcm.c
```diff
@@ -0,0 +1,808 @@
...
+}
+
+FAR const char *snd_pcm_format_description(snd_pcm_format_t format)
+{
+  if (format > SND_PCM_FORMAT_LAST)
+    {
+      return NULL;
+    }
+
+  const char *_snd_pcm_format_descriptions[] =
```
raiden00pl, line 455: this file looks like highly inspired by https://github.com/alsa-project/alsa-lib/blob/master/src/pcm/pcm.c#L2001

raiden00pl:
I'm definitely not licences expert so I don't know if licensing this code to Apache is OK.

But for sure caution is required here. GPL is like virus. 
Where does "interface inspiration" ends and "derivative work" begin? I see a lot of identical fragments here as in https://github.com/alsa-project/alsa-lib/tree/master/src not only in headers, so I would wounder if we aren't dealing here with  "derivative GPL work".

audioutils/alsa-lib/pcm.c
cederom: in order to maintain compatibility?

raiden00pl: I doubt that compatibility is a sufficient reason not to name this implementation as "derivative GPL work".
Even internal static functions are called the same like in alsa-lib.

raiden00pl:
> it's rewrite from scratch, but follow alsa-lib public API, just like nuttx v4l2 driver framework:

If this is **rewriten** based on alsa-lib then this is derivative work. If this is derivative work of GPL, then this code is GPL. If this code is GPL then it can't be relicensed to Apache.

That's how I understand copyleft licences. I would be happy if there is someone more competent than me and says otherwise.

cederom:
maybe we should ask ALSA project directly to have no doubts?

if the interface is compatible but its a rewrite from scratch then should be no problem? how otherwise create alternatives?

raiden00pl:
The code structure is very similar to alsa-lib, some code fragments are the same, many private functions has the same names. This is not just a compatible interface, but highly inspired on alsa-lib work. Asking the alsa-lib team won't change anything here if it's a derivative work of GPL. I doubt they relicence it to permisive license, I doubt if its even possible.

If you want to avoid such problems you have to write EVERYTHING from scratch,  the only thing you can take from GPL project are PUBLIC header interfaces.

hartmannathan (commented):
Please add git commit log messages.

yangyalei:
> The code structure is very similar to alsa-lib, some code fragments are the same, many private functions has the same names. This is not just a compatible interface, but highly inspired on alsa-lib work. Asking the alsa-lib team won't change anything here if it's a derivative work of GPL. I doubt they relicence it to permisive license, I doubt if its even possible.
> 
> If you want to avoid such problems you have to write EVERYTHING from scratch, the only thing you can take from GPL project are PUBLIC header interfaces.



> The code structure is very similar to alsa-lib, some code fragments are the same, many private functions has the same names. This is not just a compatible interface, but highly inspired on alsa-lib work. Asking the alsa-lib team won't change anything here if it's a derivative work of GPL. I doubt they relicence it to permisive license, I doubt if its even possible.
> 
> If you want to avoid such problems you have to write EVERYTHING from scratch, the only thing you can take from GPL project are PUBLIC header interfaces.

I have updated to modify file names, function names, function implementations, and other similar code fragments to be more different with the standard ALSA library. Please help to review it again. @xiaoxiang781216

raiden00pl:
I doubt that changing the names of the functions changes anything here. It's still a derivative of GPL code.

cederom:
Does it mean no one can implement ALSA? :-(

There is much work done already, would be shame to waste it.. maybe we should just mark it as GPL tainted and it still should be available to open-source only projects?

raiden00pl:
GPL code can't be hosted on Apache Fundation repositories if I remember correctly. Workaround for this is to host GPL code outside Apache Fundation repositories and keep it as external resource. Or write it completely from scratch rather than rewrite from GPL code.

cederom:
ACK! Thanks @raiden00pl :-)

PR #2984, opened 2025-02-03
examples/amg88xx: fix kconfig

LuchianMihai:
## Summary
Small fix for my previous sloppy work on amg88xx example.
I forgot to add correct dependency for kconfig options.
Also, I've followed DTH driver as example, you can see that in the kconfig stack size option :D

## Impact
In non-interactive build, some defines previously polluted the header file with no impact on build. Now is fixed.
In interactive build (running make menuconfig ~or equivalent~) the amg88xx example was visible without enabling the needed driver

## Testing
Manually.
Checked that the define is no longer included in the output header.
Check that the amg88xx example is not longer present in the menuconfig if the amg88xx driver is disabled.

### DIFF
--- a/examples/amg88xx/Kconfig
+++ b/examples/amg88xx/Kconfig
@@ -1,6 +1,7 @@
 config EXAMPLES_AMG88XX
 	tristate "AMG88xx sensor example"
 	default n
+	depends on SENSORS_AMG88XX
 	---help---
 		Enable the AMG88xx sensor example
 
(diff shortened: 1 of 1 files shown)


jerpelea (approved):

cederom (approved):
Thank you @LuchianMihai :-)

LuchianMihai:
> Thank you @LuchianMihai :-)

No worries

simbit18 (approved):

anchao (approved):

PR #3007, opened 2025-02-24
system/debugpoint: fix bug the length option is parsed error

chirping78:
## Summary

The `debugpoint` program has a option `-l` which requires an number argument,

which means the `optstring` to  `int getopt(int argc, char * const argv[], const char *optstring)` should be `l:`.
## Impact
`debugpoint` program

## Testing
Use `-l` to test `debugpoint`


jerpelea (changes requested):
please add a commit message

simbit18 (approved):

chirping78:
> please add a commit message

Updated the commit message. @jerpelea

raiden00pl (approved):

acassis (approved):

PR #3024, opened 2025-03-12
[BREAKING] boot/nxboot: enhance bootloader capabilities and decision logic

michallenc:
## Summary
This commit enhances the bootloader capabilities. The image's header is extended with header version, size, platform identifier and pointer to optional next header. CRC32 now includes part of the header in its calculation as well.

The change also avoids having two different magics for image uploaded over programmer and update image. Both these images have the same magic and this magic is changed internally by the bootloader's logic. The change is needed because image with standard magic is automatically considered as a confirmed image (uploaded with programmer).

The current implementation avoids tails at all, therefore the user application uploading the image does not have to erase the tail before new upload. The image is considered as confirmed if it has standard magic or its recovery is present. This means the bootloader has to erase the header of the update image after the update is done (to avoid update loop and to mark the image as unstable). This page is written back during the confirmation.

This is a breaking change, but necessary for the future development of the bootloader. The added header version field will allow to add minor/major updates while keeping the backwards compatibility.

## Impact
Breaking change for nxboot bootloader. I think no one is using it so far (except for us at Elektroline), so it should not be a big issue. This also adds version field in the header so future breaking changes could be tracked easily. It is also a better and more user friendly design, so I think these advantages are big enough to break the old images that are not used by anyone so far.

Documentation updated at https://github.com/apache/nuttx/pull/15981

## Testing
Tested on SAMv7 custom board with various scenarios:

- revert after update
- confirm after update
- multiple updates in a row
- board power restarts during recovery creation, update and revert processes

All these scenarios lead to the valid image in primary slot.

### DIFF
--- a/boot/nxboot/Kconfig
+++ b/boot/nxboot/Kconfig
@@ -43,6 +43,15 @@ config NXBOOT_HEADER_SIZE
 		Note that this size should be aligned with the program memory write
 		page size!
 
+config NXBOOT_PLATFORM_IDENTIFIER
+	hex "A unique platform identifier"
+	default 0x0
+	---help---
+		This is a unique platform identifier used by the bootloader to
+		verify whether the image should be run on a given platform. An update
+		(or even a firmware uploaded via a programmer) is rejected if the
+		value in image's header doesn't match this option.
+
 config NXBOOT_BOOTLOADER
 	bool "Build nxboot bootloader application"
 	default n
@@ -80,7 +89,7 @@ config NXBOOT_PREVENT_DOWNGRADE
 		performed only for newer versions  (according to Semantic Version
 		preference rules).
 
-		WARNING: NXboot currently implementes preferences only for
+		WARNING: NXboot currently implements preferences only for
 		MAJOR.MINOR.PATCH and ignores prerelease.
 
 endif # NXBOOT_BOOTLOADER
--- a/boot/nxboot/include/nxboot.h
+++ b/boot/nxboot/include/nxboot.h
@@ -39,26 +39,32 @@
 #define NXBOOT_SECONDARY_SLOT_NUM (1)
 #define NXBOOT_TERTIARY_SLOT_NUM  (2)
 
-/* Offsets to write pages containing confirmed and updated flags. These
- * pages are located at the end of the partition, therefore index 0 means
- * the first page from the end.
- */
-
-#define NXBOOT_CONFIRMED_PAGE_INDEX (0)
-#define NXBOOT_UPDATED_PAGE_INDEX   (1)
-
-#define NXBOOT_HEADER_MAGIC     0x534f584e /* NXOS. */
-#define NXBOOT_HEADER_MAGIC_INV 0xaca0abb1 /* NXOS inverted. This is used
-                                            * for images uploaded directly
-                                            * to the primary flash with
-                                            * the debugger. These images
-                                            * does not have precalculated
-                                            * CRC and flags at the
-                                            * end of the partition, but
-                                            * are considered to be valid.
+#define NXBOOT_HEADER_MAGIC     0x534f584e /* NXOS. The NX images, both
+                                            * uploaded directly to primary
+                                            * partition via debugger and to
+                                            * update via some application
+                                            * are used with this magic. If
+                                            * this image is uploaded to
+                                            * primary flash, it is considered
+                                            * valid.
+                                            */
+#define NXBOOT_HEADER_MAGIC_INT 0xaca0abb0 /* NXOS internal. This is used
+                                            * for internal bootloader
+                                            * handling and operations. It is
+                                            * switch internally to distinguish
+                                            * between images uploaded via
+                                            * debugger or the ones updated
+                                            * after the bootloader performed
+                                            * its operation. The first two
+                                            * bits are reserved to point
+                                            * what partition is a recovery
+                                            * for this image.
                                             */
 
-#define NXBOOT_HEADER_PRERELEASE_MAXLEN 110
+#define NXBOOT_HEADER_MAGIC_INT_MASK 0xfffffff0
+#define NXBOOT_RECOVERY_PTR_MASK 0x3
+
+#define NXBOOT_HEADER_PRERELEASE_MAXLEN 94
 
 /****************************************************************************
  * Public Types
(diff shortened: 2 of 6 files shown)


nuttxpr:



Yes, this PR appears to meet the NuttX requirements, although some clarifications could strengthen it.

**Strengths:**

* **Clear Summary:** The summary explains the "why," "what," and "how" of the changes.  The reference to the related documentation PR is excellent.
* **Impact Acknowledged:** The impact section addresses most points, particularly highlighting the breaking change and justifying it.
* **Testing Description:** The testing description covers various scenarios and provides context (platform used).

**Areas for Improvement (to make it even stronger):**

* **Summary - Issue Reference:** If there's a related NuttX issue, include a link. Even if the change is internally driven, creating an issue to track it can be beneficial.
* **Impact - User Impact (Elaborate):**  While it states "no one is using it," providing more detail about *how* users would adapt (e.g., regenerate images with the new format) is helpful, even if the user base is small.
* **Impact - Build Impact:**  Clarify if the build process itself changes (e.g., new build options, changes to existing build scripts).  Even if the answer is "NO," explicitly stating it is good practice.
* **Impact - Hardware Impact:** Specify the affected architecture(s) (SAMv7 is mentioned in testing, but reiterate it in the impact section).
* **Impact - Security Impact:** While you might not expect any security implications, explicitly stating "NO" and briefly explaining why (e.g., "No new security vulnerabilities introduced") is a good defensive practice.
* **Impact - Compatibility:** Elaborate on the "breaking change."  How will existing images be handled? Are there migration steps? This ties in with user impact.
* **Testing - Be More Specific:** "Various scenarios" is good, but adding concise bullet points for each scenario and the *expected result* would improve the testing section significantly. For example:
    * **Revert after update:**  The bootloader successfully boots the original primary slot image.
    * **Confirm after update:** The bootloader boots the updated image and marks it as permanent.
* **Testing - Logs (Consider):** While the template requests logs, they might not be strictly necessary for all changes.  However, if readily available, including snippets of relevant log output (especially for error conditions or unusual behavior) can be very helpful.  If logs are omitted, briefly explain why (e.g., "Logs are not included as they are voluminous and don't provide additional insight beyond the described test cases").


By addressing these points, the PR would be even more robust and easier for reviewers to assess.

cederom:
Thank you @michallenc very interesting! As this is breaking change I will start a discussion on the dev@ mailing list :-)

jerpelea:
I would prefer that this change should be scheduled for NuttX 13.0.0. we have already branched out the current release so I have no concerns

jerpelea (approved):

TimJTi:
Just a question if I may as this seems to be making NXboot much more like MCUboot? What is the reasoning for improving NXboot rather than using MCUboot?

michallenc:
> Just a question if I may as this seems to be making NXboot much more like MCUboot? What is the reasoning for improving NXboot rather than using MCUboot?

I would not say this commit makes NXboot more like MCUboot, it just simplifies the previous, few months old, design and I don't think any of these changes move it closer to MCUboot, but rather opposite (no need for image's tail, one image for both primary and update).

Regarding MCUboot, there are several significant problems with the project. Swapping algorithms are not usable if you have secondary partition at external NOR flash as the update process takes several minutes (!!!), swap with scratch also causes a significant flash wear. Move algorithm is not usable for external flash at all because it requires the same size of erase page (or at least required the last time I checked) for internal and external flash.

I initially planned to merge the algorithm used in NXboot to MCUboot project but this was rejected by the maintainers (mostly because they didn't want to maintain another algorithm, which is a fair point). Therefore the implementation of NXboot. This commit does not change the used algorithm, but simplifies some user interface and API.

Also MCUboot image is padded to partition size, so you have to upload 2 MB image (if you have 2 MB slot) even if the actual code size is much less than this. But you still want to keep the entire primary flash for the image because of future extensions.

lupyuen (approved):

ppisa:
Hello everybody, we have had lean to not propose another small systems boot solution and we have discussed that in Elektroline team a lot. @michallenc  has even downloaded whole MCU boot with Zephyr dependencies (which pulls all possible HAL invention from many different companies) to test framework  etc. But reaction of MCUboot maintainers was far from promising that they ever take our arguments, we have plenty of external Flash but writes are slow and Flash can have (in NAND case) even 2 MB erase pages, i.e. bigger than whole internal Flash etc... So the end is solution (backup plan) which has no dependencies outside of NuttX. And I like result much more than MCU boot after this refinement boot.

We have lot of discussions still and yes, it is disadvantage that there is change of the header format at this moment. But I think that it worth to be provided in this newer version which is optimized to be easier for updates, only the 1:1 copy of the image to the right of two OTAs is required to update image and then single confirm command in the test run...

I would propose to push update even into new prepared release. May it be whole NXboot should be under experimental option for now, and as it is tested between releases it can be declared as stable for next one. Image header has mechanism of major and minor version now so I think that this updated format can server for some years and can be extended with backward compatibility.

raiden00pl:
A custom bootloader solution for NuttX is a great idea. Personally, I don't really like MCUBoot and if there is an option to get rid of it from my projects, I will gladly take it.

TimJTi:
> A custom bootloader solution for NuttX is a great idea. Personally, I don't really like MCUBoot and if there is an option to get rid of it from my projects, I will gladly take it.

I (recently) implemented MCUboot to replace Uboot...but not impressed with it TBH.

I will 100% be trying NXboot as it is not too late to jump ship for my project, and it should be a better platform for enhancements from a community that cares and is responsive (Y)

xiaoxiang781216:
It's strange to use uboot or mcuboot in NuttX. NuttX is small enough as base for any type of hardware envrionment. Actually, we use the same NuttX code base but different defconfig for bootloader, ota, tee, mcu, and dsp. it's very convenient to use the same developement environment for all kind of task in the same chip.

ppisa:
Our idea with MCUboot has been to use it as application only in the NuttX build or boot loader and main firmware. That mode has been supported by MCUboot and @michallenc  has worked on it. But MCUboot mainline did require testing on the full Zephyr based project, which is OK and has been handled, but then did not listen to our arguments to update MCUboot to be usable for applications with slow secondary Flashes and original algorithm has high scratchpad area wear level so it is another disadvantage.

acassis (approved):

PR #3051, opened 2025-04-07
tools/bitmap_converter.py: Fix print to allow use of Python3

TimJTi:
Amends print statement to Python3 standard

I gladly declare that I have never coded in Python, but I believe python3 is the current standard and I was trying to use this converter in relation to a PR I'm working in and it wouldn't run for this reason.

If I've missed anything that means this PR is not necessary let me know and I can close it.

I haven't run checkpatch...I have assumed it doesn't work for Python anyway, and the change is trivial.

## Testing
I have used to it successfully to convert a PNG file to a c file using Python3

### DIFF
--- a/tools/bitmap_converter.py
+++ b/tools/bitmap_converter.py
@@ -120,7 +120,7 @@ if __name__ == '__main__':
   import os.path
 
   if len(sys.argv) != 3:
-    print "Usage: bitmap_converter.py source.png output.cxx"
+    print("Usage: bitmap_converter.py source.png output.cxx")
     sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


TimJTi:
OK - lack of Python knowledge ringing out like a bell. I'm assuming the original script is not up to standard but OI am not about to start to learn Python to fix this.
Happy to be the typist if anyone wants to advise...will leave for a few days as I am away, but then probably close it and raise it as an issue rather than try and fix it myself!

TimJTi:
OK - this tool creates cpp code and is not suitable for the kernel. Looks like I'm learning Python after all!

PR #3059, opened 2025-04-15
Make slcan brin the interface up/down with the appropiate commands

csanchezdll:
## Summary

A recent change (https://github.com/apache/nuttx/pull/16199) has made the bitrate setting no longer bring the interface up. Moreover, it is now no longer possible to change bitrate of a CAN interface if it is up. Therefore, slcan needs to bring the interface down to change it. Fortunately, it already had commands to open and close the interface which map nicely to this.

## Impact

slcan would be unable to change the interface bitrate after the aforementioned (already merged) Nuttx patch, without this change.

## Testing

Tested using a second serial line (/dev/gsmtty2) and CAN interface (named can_6_14) on a custom board. The CAN bus is attacked using an adapter talking socketcand protocol:
```
< hi >< can_6_14 B 1000000 -1 -1 -1 -1 -1 -1 -1 >
< ok >< open can_6_14 >
< ok >< send 123 0 >
```
Before this patch (and with Nuttx updated so the patch there is applied) changing bitrate no longer brings the interface up, so no frames can be seen. After this patch, I can explicitly bring the interface up with the "O" command and see the traffic:
```
nsh> slcan can_6_14 /dev/gsmtty2
[01/01/70 00:00:28.620152] Starting slcan on NuttX
[01/01/70 00:00:28.620274] slcanBus
[01/01/70 00:00:28.620427] CAN socket open
[01/01/70 00:00:28.620518] Serial interface open /dev/gsmtty2
[01/01/70 00:00:34.440030] set speed 1000000
[01/01/70 00:00:53.280061] Open interface
[01/01/70 00:01:03.243936] R1, Id:0x123
```
I also checked bringing the interface down with "C" stops receiving frames.

I had to make another small fix to slcan.c (in a separate commit also in this patchset) because it originally only supported CAN interface names of up to 4 characters.

### DIFF
--- a/canutils/slcan/slcan.c
+++ b/canutils/slcan/slcan.c
@@ -137,8 +137,7 @@ static int caninit(char *candev, int *s, struct sockaddr_can *addr,
       syslog(LOG_ERR, "Error opening CAN socket\n");
       return -1;
     }
-  strncpy(ifr.ifr_name, candev, 4);
-  ifr.ifr_name[4] = '\0';
+  strlcpy(ifr.ifr_name, candev, IFNAMSIZ);
   ifr.ifr_ifindex = if_nametoindex(ifr.ifr_name);
   if (!ifr.ifr_ifindex)
     {
--- a/canutils/slcan/slcan.c
+++ b/canutils/slcan/slcan.c
@@ -316,9 +316,22 @@ int main(int argc, char *argv[])
                         {
                           /* open CAN interface */
 
-                          mode = 1;
-                          debug_print("Open interface\n");
-                          ok_return(fd);
+                          struct ifreq ifr;
+
+                          strlcpy(ifr.ifr_name, argv[1], IFNAMSIZ);
+
+                          ifr.ifr_flags = IFF_UP;
+                          if (ioctl(s, SIOCSIFFLAGS, &ifr) < 0)
+                            {
+                              syslog(LOG_ERR, "Open interface failed\n");
+                              fail_return(fd);
+                            }
+                          else
+                            {
+                              mode = 1;
+                              debug_print("Open interface\n");
+                              ok_return(fd);
+                            }
                         }
                       else if (buf[0] == 'S')
                         {
@@ -394,9 +407,22 @@ int main(int argc, char *argv[])
                         {
                           /* close interface */
 
-                          mode = 0;
-                          debug_print("Close interface\n");
-                          ok_return(fd);
+                          struct ifreq ifr;
+
+                          strlcpy(ifr.ifr_name, argv[1], IFNAMSIZ);
+
+                          ifr.ifr_flags = 0;
+                          if (ioctl(s, SIOCSIFFLAGS, &ifr) < 0)
+                            {
+                              syslog(LOG_ERR, "Close interface failed\n");
+                              fail_return(fd);
+                            }
+                          else
+                            {
+                              mode = 0;
+                              debug_print("Close interface\n");
+                              ok_return(fd);
+                            }
                         }
                       else if (buf[0] == 'T')
                         {


nuttxpr:



Yes, this PR appears to meet the NuttX requirements, although some sections could be more explicit.

**Strengths:**

* **Clear Summary:** The summary explains the "why," "what," and "how" of the change, and links it to a prior NuttX PR.
* **Impact on User:**  Clearly explains the user impact (inability to change bitrate without the fix).
* **Testing:** Provides a reasonable test setup and logs demonstrating the issue before and after the fix.  The additional fix for interface names is also mentioned.

**Areas for Improvement:**

* **Summary:** While the related NuttX PR is linked, a related issue (if one exists) should also be linked. If not, state that there isn't a related issue.
* **Impact:**  While the user impact is clear, other impact sections are marked "NO" without explanation. Even if there's no impact, explicitly stating "NO - No change to the build process" is better for clarity.  Be very explicit.
* **Impact - Compatibility:**  The summary mentions a backward compatibility issue with a prior NuttX change.  This should be explicitly called out in the "Impact on compatibility" section.  Is this change backward compatible with *previous* versions of slcan?
* **Testing:**  While the test setup is described, the host environment isn't fully specified (OS, compiler, etc.). While `can_6_14` is mentioned as a CAN interface on a custom board, more information on the board and the *target* architecture would be helpful. For instance: "Target: ARM, custom board (STM32F4xx based) with a CAN controller".  Also specify the NuttX version being tested against.

**Example of Improved Impact Section:**

```
## Impact

* Is new feature added? NO
* Is existing feature changed? YES, the bitrate setting logic is modified to accommodate a recent change in the NuttX CAN driver.
* Impact on user (will user need to adapt to change)? NO.  Users who were previously using the bitrate setting feature of slcan will find it working correctly again.
* Impact on build (will build process change)? NO - No changes to the build process.
* Impact on hardware (will arch(s) / board(s) / driver(s) change)? NO - No changes to hardware or drivers.  This change affects the slcan application's interaction with the CAN driver.
* Impact on documentation (is update required / provided)? NO -  No documentation updates required.
* Impact on security (any sort of implications)? NO - No security implications.
* Impact on compatibility (backward/forward/interoperability)? YES -  This change restores compatibility with a recent change in NuttX (https://github.com/apache/nuttx/pull/16199) that broke the bitrate setting functionality in slcan. This change is backward compatible with previous versions of slcan that were working correctly before the NuttX change.
* Anything else to consider?  The maximum length of the CAN interface name supported by slcan has been increased (addressed in a separate commit).
```

By being more thorough and explicit in the Impact and Testing sections, the PR will be easier to review and merge.

cederom (approved):
Thank you @csanchezdll :-)

canutils/slcan/slcan.c
```diff
@@ -137,8 +137,8 @@ static int caninit(char *candev, int *s, struct sockaddr_can *addr,
       syslog(LOG_ERR, "Error opening CAN socket\n");
       return -1;
     }
-  strncpy(ifr.ifr_name, candev, 4);
-  ifr.ifr_name[4] = '\0';
+  strncpy(ifr.ifr_name, candev, IFNAMSIZ);
```
xiaoxiang781216, line 140: strlcpy

canutils/slcan/slcan.c
```diff
@@ -137,8 +137,8 @@ static int caninit(char *candev, int *s, struct sockaddr_can *addr,
...
+  ifr.ifr_name[IFNAMSIZ] = '\0';
```
xiaoxiang781216, line 141: remove

csanchezdll: Makes sense. I was trying to make minimal changes, but this is better with strlcpy. Changed.

csanchezdll:
I fixed the format errors and used checkpatch from the nuttx repo.

I found the initialization of ifr = {0} that I had fails because format rules want the braces on separate lines. Too much code for little meaning, so I remove the initialization altogether and used = instead of |= when assigning flags. The result is the same.

canutils/slcan/slcan.c
```diff
@@ -137,8 +137,7 @@ static int caninit(char *candev, int *s, struct sockaddr_can *addr,
       syslog(LOG_ERR, "Error opening CAN socket\n");
       return -1;
     }
-  strncpy(ifr.ifr_name, candev, 4);
-  ifr.ifr_name[4] = '\0';
+  strlcpy(ifr.ifr_name, candev, IFNAMSIZ);
```
acassis, line 140: @csanchezdll suggestion: include the ifr.irf_name[IFNAMSIZ] = '\0'; for safety reason, please also include it to line 365

canutils/slcan/slcan.c
```diff
@@ -316,9 +316,22 @@ int main(int argc, char *argv[])
                         {
                           /* open CAN interface */
 
-                          mode = 1;
-                          debug_print("Open interface\n");
-                          ok_return(fd);
+                          struct ifreq ifr;
+
+                          strlcpy(ifr.ifr_name, argv[1], IFNAMSIZ);
+
```
acassis, line 322: Please include the null termination here too. It will protect in cases where users pass a longer name than IFNAMSIZ.

acassis (approved):

canutils/slcan/slcan.c
csanchezdll: This was here on my first version of the PR, with the original strncpy. Changing to strlcpy and removing the explicit \0 addition was suggested by @xiaoxiang781216 just above https://github.com/apache/nuttx-apps/pull/3059#discussion_r2045139275.
I agree with him, strlcpy is safer, as it is guaranteed to add a \0 (strncpy will only *copy* the original \0 if it fits in the buffer). Therefore, the explicit \0 is no longer needed.

acassis: @csanchezdll my fault, I ran "man strlcpy" but only take a look at function description. In fact it seems a good solution.

acassis: @xiaoxiang781216 do you think these BUGS reported in the man pages aren't a concern: 

```
BUGS
       All  catenation  functions  share the same performance problem: Shlemiel the painter.
       As a mitigation, compilers are able to transform some calls to  catenation  functions
       into  normal copy functions, since strlen(dst) is usually a byproduct of the previous
       copy.

       strlcpy(3) and strlcat(3) need to read the entire src string, even if the destination
       buffer is small.  This makes them vulnerable to Denial of Service (DoS) attacks if an
       attacker can control the length of the src string.  And if not, they're still  unnec‐
       essarily slow.
```

xiaoxiang781216: strlcpy always add \0 to the destination buffer, so you don't need add '\0' again. DoS is another problem, if you want to address this problem, you need more safer api(strscpy):
https://staticthinking.wordpress.com/2023/10/30/strcpy-strncpy-strlcpy-and-strscpy/

xiaoxiang781216: we can port safec library from:
https://sourceforge.net/p/safeclib/code/ci/master/tree/
https://github.com/intel/safestringlib

csanchezdll: Agreed `strscpy` is the way to go. However, it is IMHO out of scope if this PR, which is required to keep slcan working after recent Nuttx changes. Options:
1) Leave the code totally untouched with the 4 limit length.
2) Leave `srtncpy` + explicit `\0` as it was originally (just change the max length) [this was my original patch]
3) Use `strlcpy` [suggested by @xiaoxiang781216]
4) Port `strscpy` as part of this PR
5) Block this PR until `strscpy` is ported in another PR

I'd vote for 3. Otherwise 1 and 2 are fine. I think 4 and 5 are bad ideas.

canutils/slcan/slcan.c
```diff
@@ -395,9 +407,22 @@ int main(int argc, char *argv[])
...
                           /* close interface */
 
-                          mode = 0;
-                          debug_print("Close interface\n");
-                          ok_return(fd);
+                          struct ifreq ifr;
+
+                          strlcpy(ifr.ifr_name, argv[1], IFNAMSIZ);
+
+                          ifr.ifr_flags = IFF_DOWN;
```
csanchezdll, line 414: `IFF_DOWN` was removed in https://github.com/apache/nuttx/pull/13842 so build checks failed. Fixing.

canutils/slcan/slcan.c
xiaoxiang781216: yes, strlcpy is used in many places, it's better to create a bunch of change to fix it.

xiaoxiang781216 (approved):

PR #3067, opened 2025-05-02
testing/ostest/signest: Replace kill with pthread_kill to fix running in SMP

jlaitine:
## Summary

"kill" sends the signal to any of the the threads in the group. The intention of the test is to send signals only to the "waiter" thread.

I believe this test has been broken since 2023 when the test changed from using tasks to using pthreads. This has gone unnoticed since the test doesn't really fail in single CPU systems (since the threads have different priorities), even if the signals were executed in other thread's context than the "waiter".

In SMP systems, the signest test threads actually run in parallel, so scheduling signal actions to different threads in the same group causes also signal actions to run in parallel (for different signals of course).

Running signal actions in parallel is not compatible behaviour with the signest test, which assumes that signals are being run one after another. For example running signals 38 and 40 in parallel on two threads/two cpus would cause the test incorrectly fail on "even signals nested". Or, sending two signals with the same number sequentially may be scheduled to two threads/two cpus, and end up running in paralle..

## Impact

Fixes ostest:
1. Removes false negatives in SMP caused by the explanation in Summary
2. Removes false positives in SMP caused by signals being run by other threads than intended - before this they may get executed by "ostest" or "interferer" threads as well.

## Testing

Tested in qemu rv-virt:smp w. 8 cores and mpfs with 4 cores.

### DIFF
--- a/testing/ostest/signest.c
+++ b/testing/ostest/signest.c
@@ -328,13 +328,13 @@ void signest_test(void)
         {
           if (signest_catchable(j))
             {
-              kill(waiterpid, j);
+              pthread_kill(waiterpid, j);
               odd_signals++;
             }
 
           if (signest_catchable(j + 1))
             {
-              kill(waiterpid, j + 1);
+              pthread_kill(waiterpid, j + 1);
               even_signals++;
             }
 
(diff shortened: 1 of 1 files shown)


nuttxpr:



Yes, this PR appears to meet the NuttX requirements.  The summary clearly explains the bug and its root cause, the impact section details the effects of the fix, and the testing section provides relevant platform information and confirms testing on appropriate multi-core targets. While it lacks "before" and "after" logs, the description of the issue and the targeted testing on SMP systems suggests a good understanding of the problem and its solution.  Providing those logs would further strengthen the PR.

jlaitine:
Related issue: https://github.com/apache/nuttx/issues/15059

jerpelea (approved):

xiaoxiang781216 (approved):

PR #3070, opened 2025-05-07
testing/ostest: add checking of signal deliver thread

jlaitine:
## Summary

This adds checking that the signal is delivered to a correct thread, if sent with pthread_kill. The check is added to signest test, as it fits there with minimal addition.

There is currently an issue in nuttx, that if the signal is masked at the time of dispatch, it is added to the task groups pending queue. From there it is later delivered, but always to any thread in the group (whichever thread happens to remove the pending signal).

## Impact

This adds functionality to ostest's signest testing case.

## Testing

Tested that after this patch,  the signest test eventually fails (at least on an SMP system), for the test.

Note: Integrating this will cause failures before the actual issue is resolved in NuttX. I will provide a suggestion for fixing it as well for the NuttX repo.

Maintainers have to decide whether this addition is 1) wanted for ostest and 2)  whether it should be merged before or after the issue is fixed in nuttx.

### DIFF
--- a/testing/ostest/signest.c
+++ b/testing/ostest/signest.c
@@ -52,6 +52,8 @@
  * Private Data
  ****************************************************************************/
 
+static pthread_t g_waiterpid;
+static pthread_t g_interferepid;
 static sem_t g_waiter_sem;
 static sem_t g_interferer_sem;
 static sem_t g_sem_thread_started;
@@ -90,9 +92,28 @@ static bool signest_catchable(int signo)
   return true;
 }
 
+static void interfere_action(int signo)
+{
+  pthread_t pid = pthread_self();
+  if (pid != g_interferepid)
+    {
+      printf("interfere_action: Delivered to wrong TID %d, expected %d\n",
+             pid, g_interferepid);
+      ASSERT(false);
+    }
+}
+
 static void waiter_action(int signo)
 {
   int nest_level;
+
+  pthread_t pid = pthread_self();
+  if (pid != g_waiterpid)
+    {
+      printf("waiter_action: Delivered to wrong TID %d, expected %d\n",
+             pid, g_waiterpid);
+      ASSERT(false);
+    }
 
   sched_lock();
   nest_level = g_nest_level++;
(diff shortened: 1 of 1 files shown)


testing/ostest/signest.c
```diff
@@ -52,6 +53,8 @@
  * Private Data
  ****************************************************************************/
 
+static pid_t waiterpid;
```
xiaoxiang781216, line 56: should we change to pthread_t

testing/ostest/signest.c
```diff
@@ -90,10 +93,29 @@ static bool signest_catchable(int signo)
   return true;
 }
 
+static void interfere_action(int signo)
+{
+  pid_t pid = _SCHED_GETTID();
```
xiaoxiang781216, line 98: should we use pthread_self

testing/ostest/signest.c
jlaitine: yes, I believe it is more appropriate. I was just  copying the old code...

testing/ostest/signest.c
jlaitine: sure

jerpelea (approved):

jerpelea (changes requested):
please add a commit message

jlaitine:
> please add a commit message

Oops, thanks for reminding, done!

xiaoxiang781216 (approved):

jlaitine:
This fails on ostest, which is expected  due to the bug that this change is supposed to expose.

Need to re-trigger this after https://github.com/apache/nuttx/pull/16334 is proper and integrated.

jerpelea (approved):

cederom (approved):
Thank you @jlaitine :-)

jlaitine:
I just realized that this may cause unwanted side effects since there was an existing bug in the test; it doesn't unregister the signal action handlers after the test is done.

This causes later tests which fire signals to call these handlers still, as it is all in "ostest" group.

Added unregistering the handlers to this same, so that the TID checking ASSERT doesn't fire later during other tests.

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #3073, opened 2025-05-10
[BREAKING] uORB/sensor: Add timestamp for GNSS data

JianyuWang0623:
## Summary
Add timestamp for GNSS measurements and clock, for matching.

When recovering the combination of `GnssMeasurement` and `GnssClock` into [GnssData](https://android.googlesource.com/platform/hardware/libhardware/+/refs/heads/android14-release/include/hardware/gps.h#1748), the reason for splitting is that the uORB buffer is not large enough to accommodate `GnssData`, since different topics cannot guarantee the same timing when publishing and subscribing data, we need an index to match.

**Depends on https://github.com/apache/nuttx/pull/16355**
## Impact
Developers using the structures `struct sensor_gnss_measurement` and `struct sensor_gnss_clock` need to be aware of the changes to the structure members (a new member of type `uint64_t` has been added at the beginning).

## Testing
CI

### DIFF
--- a/system/uorb/sensor/gnss.c
+++ b/system/uorb/sensor/gnss.c
@@ -57,6 +57,7 @@ static const char sensor_gnss_format[] =
   UORB_DEBUG_FORMAT_SENSOR_GNSS ",firmware_version:%" PRIu32 "";
 
 static const char sensor_gnss_clock_format[] =
+  "timestamp:%" PRIu64 ","
   "flags:%" PRIx32 ",leap_second:%" PRId32 ",time_ns:%" PRId64 ","
   "time_uncertainty_ns:%hf,hw_clock_discontinuity_count:%" PRIu32 ","
   "full_bias_ns:%" PRId64 ",bias_ns:%hf,bias_uncertainty_ns:%hf,"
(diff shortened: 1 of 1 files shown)


cederom:
Thank you @JianyuWang0623 :-)

Maybe it would be helpful for uorb users to also put that warning in the commit message and tag topic with [BREAKING]? Its not a big impact just this application but users would have clear information on the change :-)

WARNING: Developers using the structures struct sensor_gnss_measurement and struct sensor_gnss_clock need to be aware of the changes to the structure members (a new member of type uint64_t has been added at the beginning).

JianyuWang0623:
> Thank you @JianyuWang0623 :-)
> 
> Maybe it would be helpful for uorb users to also put that warning in the commit message and tag topic with [BREAKING]? Its not a big impact just this application but users would have clear information on the change :-)
> 
> WARNING: Developers using the structures struct sensor_gnss_measurement and struct sensor_gnss_clock need to be aware of the changes to the structure members (a new member of type uint64_t has been added at the beginning).

Done, thank you!

cederom (approved):
Thank you @JianyuWang0623 :-)

jerpelea (approved):

PR #3090, opened 2025-06-08
system/syslogd: Initial implementation

linguini1:
## Summary

Initial implementation of syslogd including version information, simple usage help and UDP transmission. The current implementation transmits RFC 5424 formatted syslog entries over UDP for consumption by a syslog collector. Only some options from the manpage are implemented since NuttX doesn't currently have the ability for some of the more complex features (-l, etc.).

## Impact

Users can use this application alongside the new RFC 5424 support (see other PR here: https://github.com/apache/nuttx/pull/16493) to send syslog entries over the network in a format that is understood by various logging servers and network utilities (i.e. WireShark, rsyslog, etc).

## Testing

Testing was performed by completing a local build on the `sim` and `raspberrpi-pico-w` platforms with all new `syslogd` options enabled. Builds succeeded.

On the simulator (`tcpblaster` config used as a basis), `syslogd` was used with the host routing mode to verify that logs were being sent over UDP to the configured server address and port. This was confirmed with WireShark, which also successfully parsed the messages using the syslog protocol.

On the Pico W, the `telnetd` configuration was used as a basis. `syslogd` was run using my host computer as the server address and I verified that all logs were continuously received (as they occurred) by watching traffic using WireShark.

image

Debug output was confirmed to work by running with the `-d` flag and verifying that all logs being transmitted were also being logged to `stdout`, along with information about socket setup.

I also tried recording with `tshark` (the command line WireShark) and a Python implementation of a syslog server which can be found here: https://github.com/easypost/syslog-rfc5424-parser

Both of those programs were able to parse the syslogd messages:

[picow-syslogd-parser.json](https://github.com/user-attachments/files/20646212/picow-syslogd-parser.json)
[picow-syslogd-tshark.json](https://github.com/user-attachments/files/20646214/picow-syslogd-tshark.json)

I was not able to make the `syslogd` process run in the background by default using `fork()` (when run with no `-d` or `-n` command). If anyone has a suggestion on how to make that happen, please let me know!

### DIFF
--- a/system/syslogd/CMakeLists.txt
+++ b/system/syslogd/CMakeLists.txt
@@ -1,0 +1,36 @@
+# ##############################################################################
+# apps/system/syslogd/CMakeLists.txt
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  The ASF licenses this
+# file to you under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.  You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+# ##############################################################################
+
+if(CONFIG_SYSTEM_SYSLOGD)
+  nuttx_add_application(
+    MODULE
+    ${CONFIG_SYSTEM_SYSLOGD}
+    NAME
+    ${CONFIG_SYSTEM_SYSLOGD_PROGNAME}
+    STACKSIZE
+    ${CONFIG_SYSTEM_SYSLOGD_STACKSIZE}
+    PRIORITY
+    ${CONFIG_SYSTEM_SYSLOGD_PRIORITY}
+    SRCS
+    syslogd_main.c)
+
+endif()
--- a/system/syslogd/syslogd_main.c
+++ b/system/syslogd/syslogd_main.c
@@ -1,0 +1,376 @@
+/****************************************************************************
+ * apps/system/syslogd/syslogd_main.c
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for asyslogditional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#ifdef CONFIG_LIBC_EXECFUNCS
+#include <spawn.h>
+#endif
+
+#include <getopt.h>
+#include <syslog.h>
+
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* syslogd program version */
+
+#define SYSLOGD_VERSION "0.0.0"
+
+/* Minimum buffer size check */
+
+#if CONFIG_SYSTEM_SYSLOGD_ENTRYSIZE < 480
+#error "SYSTEM_SYSLOGD_ENTRYSIZE must be more than 480 to satisfy RFC 5424"
+#endif
+
+/* Maximum number of arguments that can be passed to syslogd */
+
+#define MAX_ARGS 8
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: print_usage
+ ****************************************************************************/
+
+static void print_usage(void)
+{
+  fprintf(stderr, "usage:\n");
+  fprintf(stderr, "  %s [-vdn]\n", CONFIG_SYSTEM_SYSLOGD_PROGNAME);
+}
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+int main(int argc, FAR char **argv)
+{
+  int fd;
+  int sock;
+  int c;
+  ssize_t bread;
+  ssize_t bsent;
+  ssize_t endpos;
+  char *end;
+  size_t bufpos = 0;
+  struct sockaddr_in server;
+  char buffer[CONFIG_SYSTEM_SYSLOGD_ENTRYSIZE];
+  bool debugmode = false;
+  bool skiplog = false;
+#ifdef CONFIG_LIBC_EXECFUNCS
+  pid_t pid;
+  bool background = true;
+  char *new_argv[MAX_ARGS + 1];
+#endif
+
+  /* Parse command line options */
+
+  while ((c = getopt(argc, argv, ":vdn")) != -1)
+    {
+      switch (c)
+        {
+        case 'v':
+
+          /* Print version and exit */
+
+          printf("%s " SYSLOGD_VERSION " (NuttX)\n", argv[0]);
+          return EXIT_SUCCESS;
+
+        case 'd':
+
+          /* Enable debug mode and stay in foreground */
+
+          debugmode = true;
+          printf("Enabling debug mode.\n");
+#ifdef CONFIG_LIBC_EXECFUNCS
+          background = false;
+#endif
+          break;
+
+        case 'n':
+
+          /* Stay in foreground */
+
+#ifdef CONFIG_LIBC_EXECFUNCS
+          background = false;
+#endif
+          break;
+
+        case '?':
+          print_usage();
+          exit(EXIT_FAILURE);
+          break;
+        }
+    }
+
+    /* Run this program in the background as a spawned task if the background
+     * option was selected.
+     */
+
+#ifdef CONFIG_LIBC_EXECFUNCS
+  if (background)
+    {
+      /* Set up the arguments, which is identical to the original except with
+       * an added `-n` flag to ensure that the new process does not 'respawn'
+       */
+
+      if (argc > MAX_ARGS)
+        {
+          fprintf(stderr,
+                  "Cannot spawn syslogd daemon: arg count %d exceeds %d",
+                  argc, MAX_ARGS);
+          return EXIT_FAILURE;
+        }
+
+      new_argv[0] = argv[0]; /* Same program name */
+      new_argv[1] = "-n";    /* Prevent daemon from spawning another child */
+      memcpy(&new_argv[2], &argv[1], sizeof(char *) * (argc - 1));
+
+      /* Spawn the child for backgrounding now */
+
+      if (posix_spawn(&pid, argv[0], NULL, NULL, new_argv, NULL) != 0)
+        {
+          fprintf(stderr, "Failed to fork() to background process: %d\n",
+                  errno);
+          return EXIT_FAILURE;
+        }
+      else
+        {
+          /* We succeeded in spawning, exit now. */
+
+          return EXIT_SUCCESS;
+        }
+    }
+#endif /* CONFIG_LIBC_EXECFUNCS */
+
+  /* Set up client connection information */
+
+  server.sin_family = AF_INET;
+  server.sin_port = htons(CONFIG_SYSTEM_SYSLOGD_PORT);
+  server.sin_addr.s_addr = inet_addr(CONFIG_SYSTEM_SYSLOGD_ADDR);
+
+  if (server.sin_addr.s_addr == INADDR_NONE)
+    {
+      fprintf(stderr, "Invalid address '%s'\n", CONFIG_SYSTEM_SYSLOGD_ADDR);
+      return EXIT_FAILURE;
+    }
+
+  /* Create a UDP socket */
+
+  if (debugmode)
+    {
+      printf("Creating UDP socket %s:%u\n", CONFIG_SYSTEM_SYSLOGD_ADDR,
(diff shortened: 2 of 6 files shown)


system/syslogd/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+# ##############################################################################
+# apps/system/dd/CMakeLists.txt
```
xiaoxiang781216: dd->syslogd

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }
+
+  /* Once all arguments are parsed, fork here to run in the background.
+   * TODO: what's a good way to tell if NuttX platform has fork()?
+   */
+
+  if (g_background)
+    {
+      err = fork();
```
xiaoxiang781216: fork  doesn't work well on nuttx, so it's better to put into background from nsh by &

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,291 @@
...
+
+  while ((c = getopt(argc, argv, ":vdn")) != -1)
+    {
+      switch (c)
+        {
+        case 'v':
+
+          /* Print version and exit */
+
+          printf("%s " SYSLOGD_VERSION " (NuttX)\n", g_syslogd);
```
xiaoxiang781216: use argv[0] and remove g_syslogd

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,291 @@
...
+ ****************************************************************************/
+
+static ssize_t read_log(int fd, char *buf, size_t n)
+{
+  ssize_t count = 0;
+  ssize_t bread;
+
+  while (count < n)
+    {
+      bread = read(fd, buf, 1);
```
xiaoxiang781216: the performance is very poor if we read one byte one time

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,291 @@
...
+
+  /* Set up client connection information */
+
+  client.sin_family = AF_INET;
+  client.sin_port = htons(CONFIG_SYSTEM_SYSLOGD_PORT);
+  client.sin_addr.s_addr = inet_addr(CONFIG_SYSTEM_SYSLOGD_ADDR);
+
+  /* Create a UDP socket */
+
+  if (g_debugmode)
```
xiaoxiang781216: change to local variable

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,291 @@
...
+#error "SYSTEM_SYSLOGD_ENTRYSIZE must be more than 480 to satisfy RFC 5424"
+#endif
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+bool g_debugmode = false;
+bool g_background = true;
+char g_buffer[CONFIG_SYSTEM_SYSLOGD_ENTRYSIZE];
```
xiaoxiang781216: why not use local buffer

system/syslogd/syslogd_main.c
linguini1: How does NSH achieve backgrounding? I am trying to make this as similar as possible to the Linux/BSD version of the program, so I would like to try and get it to run in the background unless called with `-d` or `-n`.

I know `fork()` can't be used everywhere on NuttX though. Is it possible to check the `ARCH_HAVE_FORK` definition from a user program? I can surround this code in a `#ifdef` so it's not used to forkless systems.

system/syslogd/syslogd_main.c
linguini1: I agree, this was the simplest way I could think of scanning for a newline without relying on `fgets`. Do you think I should just read the whole buffer and mark where the newline starts by looping through after the read?

system/syslogd/syslogd_main.c
linguini1: Saves more stack-size, but local is fine! I can change it. These should all really have been marked `static` anyways.

system/syslogd/syslogd_main.c
xiaoxiang781216: yes, it's more simple to read as much as possible and search '\n' by self.

system/syslogd/syslogd_main.c
xiaoxiang781216: > How does NSH achieve backgrounding?

same as Linux/BSD, append `&` to the end of command.

> I am trying to make this as similar as possible to the Linux/BSD version of the program, so I would like to try and get it to run in the background unless called with `-d` or `-n`.
> 
> I know `fork()` can't be used everywhere on NuttX though. Is it possible to check the `ARCH_HAVE_FORK` definition from a user program? I can surround this code in a `#ifdef` so it's not used to forkless systems.

if you really want to support '-n', it's better to simulate through posix_spawn by self.

linguini1: Sorry, I mean how does NuttX internally handle backgrounding? Does it use `posix_spawn` to start the program in the background? I can try doing it that way.

xiaoxiang781216: > Sorry, I mean how does NuttX internally handle backgrounding?

All tasks are started by posix_spawn or task_create, the main difference between foreground and background is that whether nsh wait the new task finish or not:
https://github.com/apache/nuttx-apps/blob/master/nshlib/nsh_fileapps.c#L248-L349

> Does it use `posix_spawn` to start the program in the background? I can try doing it that way.

if you append '&', nsh continue to accept the new console input without waiting your command finish. If you want your program automatically move to the background by self, you need posix_spawn your program again and exit the parent instance immediately to make nsh exit wait_pid.

linguini1: I've resolved to just use fork (in this latest version), protected by `CONFIG_ARCH_HAVE_FORK`. Systems that can't support it will have to run it in the foreground or use `nsh`'s backgrounding `&`.

I resolved against `posix_spawn` because it would result in an infinite loop of the program re-spawning itself unless I ran the spawned child process with the `-n` flag, requiring me to modify the original `argv` table. `fork()` seemed like a more elegant solution, and it works on a device even as small as the RP2040 seemingly quite well.

xiaoxiang781216: > I've resolved to just use fork (in this latest version), protected by `CONFIG_ARCH_HAVE_FORK`. Systems that can't support it will have to run it in the foreground or use `nsh`'s backgrounding `&`.
> 

but even CONFIG_ARCH_HAVE_FORK enable, fork doesn't work as your expect, so please don't use this feature in the production if you have other method.

> I resolved against `posix_spawn` because it would result in an infinite loop of the program re-spawning itself unless I ran the spawned child process with the `-n` flag, requiring me to modify the original `argv` table. 

yes, you need call memmove to erase `-n` from argv before posix_spawn self.

> `fork()` seemed like a more elegant solution, and it works on a device even as small as the RP2040 seemingly quite well.

to make fork work on NO-MMU device, the implementation on NuttX is very simple, and has many limitation.

linguini1: > yes, you need call memmove to erase `-n` from argv before posix_spawn self.

I actually needed to *add* the `-n` argument. Can you take a look at the new version and see if there's a better way of doing it? This method does work in my testing, but I'm unsure of the semantics of `posix_spawn` on NuttX. Does the `argv` array that I pass get copied to the child program (so that it is safe for me to `free()` it)?
 
> to make fork work on NO-MMU device, the implementation on NuttX is very simple, and has many limitation.

Is this in the NuttX docs? I could not find this mentioned and it's important for other users to be aware of too. It is likely obvious that `fork()` isn't meant for embedded environments, but a description of the limitations/differences between the architecture implementations would be good information to make available.

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,376 @@
...
+/****************************************************************************
+ * Name: find_end_pos
+ *
+ * Description:
+ *   Find the position of the newline character separating the syslog entry.
+ *
+ * Return: Index of the newline character, -1 if not found.
+ ****************************************************************************/
+
+static ssize_t find_endpos(char *buf, size_t n)
```
xiaoxiang781216: let's call `strchrnul(buf, n, '\n')` which is more fast.

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,376 @@
...
+    {
+      switch (c)
+        {
+        case 'v':
+
+          /* Print version and exit */
+
+          printf("%s " SYSLOGD_VERSION " (NuttX)\n", argv[0]);
+          return EXIT_SUCCESS;
+          break;
```
xiaoxiang781216: remove

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+  /* Transmit syslog messages forever */
+
+  if (debugmode)
+    {
+      printf("Beginning to continuously transmit syslog entries.\n");
+    }
+
+  while (true)
```
xiaoxiang781216: for (; ;)

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,376 @@
...
+       */
+
+      bread = read(fd, &buffer[bufpos], sizeof(buffer) - bufpos);
+      if (bread < 0)
+        {
+          fprintf(stderr, "Failed to read from syslog: %d", errno);
+          continue; /* Just try again */
+        }
+
+      if (bread == 0 && bufpos == 0)
```
xiaoxiang781216, line 249: we should output the buffer log and exit loop if `bread <= 0`:
```suggestion
      if (bread <= 0)
```

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,376 @@
...
+        {
+          /* Stream is over, terminate the program. */
+
+          if (debugmode)
+            {
+              printf("Syslog stream depleted, exiting...\n");
+            }
+
+          close(fd);
```
xiaoxiang781216: break to avoid the code dup

system/syslogd/syslogd_main.c
linguini1: This is not possible here, all `strchr` variants rely on a null terminated string, and take only two arguments: the buffer and the character to look for.

system/syslogd/syslogd_main.c
linguini1: No, this keeps `syslogd` more compatible with the man page definitions: https://linux.die.net/man/8/syslogd

The `-v` flag must print the version and exit.

system/syslogd/syslogd_main.c
linguini1: Not sure what you mean here? There's already a check to see if `bread` indicates an error directly above. I can end the program if `bread` indicates an error instead of continuing though.

xiaoxiang781216: since read return == 0 once, it will always 0 immediately in future, if the last message doesn't end with `\n`, syslogd will busy loop forever.

system/syslogd/syslogd_main.c
xiaoxiang781216: we can simply append '\0` before invoking strchrnul, which is simple and fast than cloning strchrnul locally.

system/syslogd/syslogd_main.c
xiaoxiang781216: but `break` after `return` is dead code.

system/syslogd/syslogd_main.c
xiaoxiang781216: > > yes, you need call memmove to erase `-n` from argv before posix_spawn self.
> 
> I actually needed to _add_ the `-n` argument. Can you take a look at the new version and see if there's a better way of doing it? This method does work in my testing, but I'm unsure of the semantics of `posix_spawn` on NuttX. Does the `argv` array that I pass get copied to the child program (so that it is safe for me to `free()` it)?
> 

yes, all string will copy to posix_spawn, you can man posix_spawn, NuttX implementation follow the spec.

> > to make fork work on NO-MMU device, the implementation on NuttX is very simple, and has many limitation.
> 
> Is this in the NuttX docs? I could not find this mentioned and it's important for other users to be aware of too. It is likely obvious that `fork()` isn't meant for embedded environments, but a description of the limitations/differences between the architecture implementations would be good information to make available.

not sure, I learn from the implementation of fork/vfork.

system/syslogd/syslogd_main.c
linguini1: Ah, I thought you meant to remove the whole option. I understand now.

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,376 @@
...
+  ssize_t endpos;
+  size_t bufpos = 0;
+  struct sockaddr_in server;
+  char buffer[CONFIG_SYSTEM_SYSLOGD_ENTRYSIZE];
+  bool debugmode = false;
+  bool skiplog = false;
+#ifdef CONFIG_LIBC_EXECFUNCS
+  pid_t pid;
+  bool background = true;
+  char **new_argv;
```
xiaoxiang781216:
```suggestion
  char *new_argv[argc + 1];
```
or use alloca to avoid the heapp allocation

linguini1: I'm worried about doing this since it's machine dependent and could overflow the stack on small devices. Why is it bad to use malloc for the brief heap allocation?

xiaoxiang781216: syslogd just accept a few argument, we can define a fix array(e.g. `char *new_argv[8]`).

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,377 @@
...
+
+      /* Take whatever bytes were leftover from our bulk read, and move them
+       * to the front of the buffer. Mark the new starting point for the next
+       * read so we don't overwrite them.
+       *
+       * endpos + 1 to overwrite vestigial newline character.
+       */
+
+      bufpos = (bufpos + bread) - (endpos + 1);
+      assert(bufstart >= 0);
```
cederom: Is `bufstart` undeclared?

```
% gmake -j24 CROSSDEV=riscv32-esp-elf-
In file included from /XXX/nuttx.git/include/errno.h:30,
                 from syslogd_main.c:29:
syslogd_main.c: In function 'syslogd_main':
syslogd_main.c:352:14: error: 'bufstart' undeclared (first use in this function); did you mean 'va_start'?
       assert(bufstart >= 0);
              ^~~~~~~~
syslogd_main.c:352:14: note: each undeclared identifier is reported only once for each function it appears in
```

cederom (changes requested):
Thank you @linguini1 amazing work! :-)

* I tired a local build and it failed at `bufstart` undeclared? Strange that CI did not catch this one?
* Would it be possible to add some documentation on this new functionality?
* Would it be possible to add dedicated `sim:syslogd` configuration to have quick test?

system/syslogd/syslogd_main.c
linguini1: Weird, my compiler didn't catch this either. Removed now!

system/syslogd/syslogd_main.c
```diff
@@ -0,0 +1,376 @@
...
+#include <netinet/in.h>
+#include <sys/socket.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* syslogd program version */
+
+#define SYSLOGD_VERSION "0.0.0"
```
acassis, line 55: Please define a valid version, maybe 0.0.1

linguini1: How come it's not valid? It's the initial implementation, all 0s

cederom: Yeah 0.0.1 looks better even for the initial implementation, 0.0.0 looks like version read error or some similar special case.. but we don't have any recommendation / guide for versioning right? :-P

PR #3097, opened 2025-06-16
boot/nxboot: fix compile errors caused by nxboot_progress

michallenc:
## Summary
Function `nxboot_progress` is located in `nxboot_main.c`, but this file is not compiled when NXboot is used as a library to an existing application and not as a bootloader. Therefore, the declaration of `nxboot_progress` function should be dependent on `NXBOOT_PROGRESS` config option.

## Impact

Fixes NXboot compilation if used as a library without `nxboot_main.c` file.

## Testing
Tested on samv7 custom board. Both bootloader and library now compiles and work as expected.

### DIFF
--- a/boot/nxboot/Kconfig
+++ b/boot/nxboot/Kconfig
@@ -122,9 +122,14 @@ config NXBOOT_PREVENT_DOWNGRADE
 		WARNING: NXboot currently implements preferences only for
 		MAJOR.MINOR.PATCH and ignores prerelease.
 
+config NXBOOT_PROGRESS
+	bool
+	default n
+
 config NXBOOT_PRINTF_PROGRESS
 	bool "Enable progress messages to be sent to STDOUT"
 	default y
+	select NXBOOT_PROGRESS
 	---help---
 		This will display progress during typically lengthy operations:
 			- Calculating checksums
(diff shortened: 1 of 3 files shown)


TimJTi:
> @TimJTi could you test this with your configuration to check if I hadn't break anything for your use case?

Of course...but I'm away at the moment, so can't until next week. The changes look sensible though.

boot/nxboot/Kconfig
```diff
@@ -122,9 +122,14 @@ config NXBOOT_PREVENT_DOWNGRADE
 		WARNING: NXboot currently implements preferences only for
 		MAJOR.MINOR.PATCH and ignores prerelease.
 
+config NXBOOT_PROGRESS
```
xiaoxiang781216, line 125: but why add new option

michallenc: To be ready for possible other progress outputs. This is a hidden option automatically selected by `NXBOOT_PRINTF_PROGRESS`. If someone adds for example `NXBOOT_SYSLOG_PROGRESS` or whatever in the future, that new option will just select `NXBOOT_PROGRESS` and there won't be need to edit ifdefs in the code.

xiaoxiang781216 (approved):

cederom (approved):
Thank you @michallenc :-)

TimJTi:
Hopefully I won't find any issues when, next week as said, I test the default use-case of using the boot app rather than as a library....

Was there an urgency to merge this now?

PR #3110, opened 2025-06-25
system/settings: general code cleaning

jeanthom:
## Summary

* Remove duplicate checks, use switch/case whenever possible
* Remove assertions just before if () condition checking the exact same thing
* Replace if (condition) assert(0) with assert(condition)

## Impact

No impact on users.

## Testing

Ran the example "settings" app using sim:nsh NuttX 12.9.0 configuration.

Before:
```
NuttShell (NSH) NuttX-12.9.0
nsh> settings
Example of settings usage to file: /tmp/settings.bin and /tmp/settings.txt:--------------------------------------------------------------
No existing binary storage file found. Creating it.
No existing text storage file found. Creating it.
Retrieved settings value (v1) with value:default value
Trying to (re)create a setting that already exists (v1)
Retrieved setting type is: 4
Trying to change setting (v1) to integer type
Deliberate fail: settings change invalid: -13
Deliberate fail: non-existent setting requested. Error:-2
Trying to change setting (v1) from int to string: I'm a string
Creating a string settings value (s1):I'm a string
Retrieved string settings value (s1) with value:I'm a string
Changing setting to an IP value (s1) with value:192.168.100.1
Retrieved IP address settings value (s1) with value:0xc0a86401
syncing storages
exiting settings example app
```

After:

```
NuttShell (NSH) NuttX-12.9.0
nsh> settings
Example of settings usage to file: /tmp/settings.bin and /tmp/settings.txt:--------------------------------------------------------------
No existing binary storage file found. Creating it.
No existing text storage file found. Creating it.
Retrieved settings value (v1) with value:default value
Trying to (re)create a setting that already exists (v1)
Retrieved setting type is: 4
Trying to change setting (v1) to integer type
Deliberate fail: settings change invalid: -13
Deliberate fail: non-existent setting requested. Error:-2
Trying to change setting (v1) from int to string: I'm a string
Creating a string settings value (s1):I'm a string
Retrieved string settings value (s1) with value:I'm a string
Changing setting to an IP value (s1) with value:192.168.100.1
Retrieved IP address settings value (s1) with value:0xc0a86401
syncing storages
exiting settings example app
```

### DIFF
--- a/system/settings/settings.c
+++ b/system/settings/settings.c
@@ -291,16 +291,19 @@ static int get_int(FAR setting_t *setting, FAR int *i)
           *i = setting->val.i;
         }
         break;
+
       case SETTING_BOOL:
         {
           *i = !!setting->val.i;
         }
         break;
+
       case SETTING_FLOAT:
         {
           *i = (int)setting->val.f;
         }
         break;
+
       default:
         {
           return -EACCES;
@@ -368,6 +371,7 @@ static int get_bool(FAR setting_t *setting, FAR int *i)
           *i = !!setting->val.i;
         }
         break;
+
       default:
         {
           return -EACCES;
@@ -432,11 +436,13 @@ static int get_float(FAR setting_t *setting, FAR double *f)
           *f = setting->val.f;
         }
         break;
+
       case SETTING_INT:
         {
           *f = (double)setting->val.i;
         }
         break;
+
       default:
         {
           return -EACCES;
@@ -505,11 +511,13 @@ static int get_ip(FAR setting_t *setting, FAR struct in_addr *ip)
           ret = OK;
         }
         break;
+
       case SETTING_STRING:
         {
           ret = inet_pton(AF_INET, setting->val.s, ip);
         }
         break;
+
       default:
         {
           return -EACCES;
--- a/system/settings/settings.c
+++ b/system/settings/settings.c
@@ -249,7 +249,6 @@ static int set_string(FAR setting_t *setting, FAR char *str)
       return -EACCES;
     }
 
-  ASSERT(strlen(str) < CONFIG_SYSTEM_SETTINGS_VALUE_SIZE);
   if (strlen(str) >= CONFIG_SYSTEM_SETTINGS_VALUE_SIZE)
     {
       return -EINVAL;
@@ -284,28 +283,29 @@ static int set_string(FAR setting_t *setting, FAR char *str)
 static int get_int(FAR setting_t *setting, FAR int *i)
 {
   assert(setting);
-  if ((setting->type != SETTING_INT) &&
-      (setting->type != SETTING_BOOL) &&
-      (setting->type != SETTING_FLOAT))
-    {
-      return -EACCES;
-    }
 
-  if (setting->type == SETTING_INT)
+  switch (setting->type)
     {
-      *i = setting->val.i;
-    }
-  else if (setting->type == SETTING_BOOL)
-    {
-      *i = !!setting->val.i;
-    }
-  else if (setting->type == SETTING_FLOAT)
-    {
-      *i = (int)setting->val.f;
-    }
-  else
-    {
-      return -EINVAL;
+      case SETTING_INT:
+        {
+          *i = setting->val.i;
+        }
+        break;
+      case SETTING_BOOL:
+        {
+          *i = !!setting->val.i;
+        }
+        break;
+      case SETTING_FLOAT:
+        {
+          *i = (int)setting->val.f;
+        }
+        break;
+      default:
+        {
+          return -EACCES;
+        }
+        break;
     }
 
   return OK;
@@ -359,19 +359,20 @@ static int set_int(FAR setting_t *setting, int i)
 static int get_bool(FAR setting_t *setting, FAR int *i)
 {
   assert(setting);
-  if ((setting->type != SETTING_BOOL) &&
-      (setting->type != SETTING_INT))
-    {
-      return -EACCES;
-    }
 
-  if ((setting->type == SETTING_INT) || (setting->type == SETTING_BOOL))
+  switch (setting->type)
     {
-      *i = !!setting->val.i;
-    }
-  else
-    {
-      return -EINVAL;
+      case SETTING_INT:
+      case SETTING_BOOL:
+        {
+          *i = !!setting->val.i;
+        }
+        break;
+      default:
+        {
+          return -EACCES;
+        }
+        break;
     }
 
   return OK;
@@ -423,23 +424,24 @@ static int set_bool(FAR setting_t *setting, int i)
 static int get_float(FAR setting_t *setting, FAR double *f)
 {
   assert(setting);
-  if ((setting->type != SETTING_FLOAT) &&
-      (setting->type != SETTING_INT))
-    {
-      return -EACCES;
-    }
 
-  if (setting->type == SETTING_FLOAT)
+  switch (setting->type)
     {
-      *f = setting->val.f;
-    }
-  else if (setting->type == SETTING_INT)
-    {
-      *f = (double)setting->val.i;
-    }
-  else
-    {
-      return -EINVAL;
+      case SETTING_FLOAT:
+        {
+          *f = setting->val.f;
+        }
+        break;
+      case SETTING_INT:
+        {
+          *f = (double)setting->val.i;
+        }
+        break;
+      default:
+        {
+          return -EACCES;
+        }
+        break;
     }
 
   return OK;
@@ -494,24 +496,24 @@ static int get_ip(FAR setting_t *setting, FAR struct in_addr *ip)
 {
   int ret;
   assert(setting);
-  if ((setting->type != SETTING_IP_ADDR) &&
-      (setting->type != SETTING_STRING))
-    {
-      return -EACCES;
-    }
 
-  if (setting->type == SETTING_IP_ADDR)
+  switch (setting->type)
     {
-      memcpy(ip, &setting->val.ip, sizeof(struct in_addr));
-      ret = OK;
-    }
-  else if (setting->type == SETTING_STRING)
-    {
-      ret = inet_pton(AF_INET, setting->val.s, ip);
-    }
-  else
-    {
-      ret = -EINVAL;
+      case SETTING_IP_ADDR:
+        {
+          memcpy(ip, &setting->val.ip, sizeof(struct in_addr));
+          ret = OK;
+        }
+        break;
+      case SETTING_STRING:
+        {
+          ret = inet_pton(AF_INET, setting->val.s, ip);
+        }
+        break;
+      default:
+        {
+          return -EACCES;
+        }
     }
 
   return ret;
@@ -673,10 +675,7 @@ static void dump_cache(union sigval ptr)
   int i;
 
   ret = pthread_mutex_lock(&g_settings.mtx);
-  if (ret < 0)
-    {
-      assert(0);
-    }
+  assert(ret >= 0);
 
   for (i = 0; i < CONFIG_SYSTEM_SETTINGS_MAX_STORAGES; i++)
     {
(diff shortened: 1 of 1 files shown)


system/settings/settings.c
```diff
@@ -284,28 +283,29 @@ static int set_string(FAR setting_t *setting, FAR char *str)
...
-    }
-  else if (setting->type == SETTING_FLOAT)
-    {
-      *i = (int)setting->val.f;
-    }
-  else
-    {
-      return -EINVAL;
+      case SETTING_INT:
+        {
```
cederom, line 290: brackets not necessary here :-)

system/settings/settings.c
```diff
@@ -359,19 +359,20 @@ static int set_int(FAR setting_t *setting, int i)
...
+  switch (setting->type)
     {
-      *i = !!setting->val.i;
-    }
-  else
-    {
-      return -EINVAL;
+      case SETTING_INT:
+      case SETTING_BOOL:
+        {
```
cederom, line 370: no brackets :-)

system/settings/settings.c
```diff
@@ -284,28 +283,29 @@ static int set_string(FAR setting_t *setting, FAR char *str)
...
+      case SETTING_FLOAT:
+        {
+          *i = (int)setting->val.f;
+        }
+        break;
+      default:
+        {
+          return -EACCES;
+        }
+        break;
```
cederom, line 311: break not necessary here :-)

system/settings/settings.c
```diff
@@ -359,19 +359,20 @@ static int set_int(FAR setting_t *setting, int i)
...
+      case SETTING_BOOL:
+        {
+          *i = !!setting->val.i;
+        }
+        break;
+      default:
+        {
+          return -EACCES;
+        }
+        break;
```
cederom, line 379: break not necessary here :-)

system/settings/settings.c
```diff
@@ -423,23 +424,24 @@ static int set_bool(FAR setting_t *setting, int i)
...
-  else if (setting->type == SETTING_INT)
-    {
-      *f = (double)setting->val.i;
-    }
-  else
+  switch (setting->type)
     {
-      return -EINVAL;
+      case SETTING_FLOAT:
+        {
```
cederom, line 435: no brackets

system/settings/settings.c
```diff
@@ -423,23 +424,24 @@ static int set_bool(FAR setting_t *setting, int i)
...
+      case SETTING_INT:
+        {
+          *f = (double)setting->val.i;
+        }
+        break;
+      default:
+        {
+          return -EACCES;
+        }
+        break;
```
cederom, line 450: no break

system/settings/settings.c
```diff
@@ -494,24 +496,24 @@ static int get_ip(FAR setting_t *setting, FAR struct in_addr *ip)
...
-    }
-  else if (setting->type == SETTING_STRING)
-    {
-      ret = inet_pton(AF_INET, setting->val.s, ip);
-    }
-  else
-    {
-      ret = -EINVAL;
+      case SETTING_IP_ADDR:
+        {
```
cederom, line 509: no brackets

system/settings/settings.c
```diff
@@ -1114,26 +1098,18 @@ int settings_create(FAR char *key, enum settings_type_e type, ...)
...
-  assert(strlen(key));
-  if (strlen(key) == 0)
+  if (type == SETTING_EMPTY)
     {
       return -EINVAL;
     }
 
-  assert(strlen(key) < CONFIG_SYSTEM_SETTINGS_KEY_SIZE);
-  if (strlen(key) >= CONFIG_SYSTEM_SETTINGS_KEY_SIZE)
+  if (strlen(key) == 0 || strlen(key) >= CONFIG_SYSTEM_SETTINGS_KEY_SIZE)
```
cederom, line 1116: I guess this condition before was invalid as it would fail assertion when shorter and return error when valid size or larger?

I suppose it was supposed to check if `strlen(key)` is `CONFIG_SYSTEM_SETTINGS_KEY_SIZE` right?

 Should the condition be `strlen(key) != CONFIG_SYSTEM_SETTINGS_KEY_SIZE` ?

system/settings/settings.c
```diff
@@ -249,7 +249,6 @@ static int set_string(FAR setting_t *setting, FAR char *str)
       return -EACCES;
     }
 
-  ASSERT(strlen(str) < CONFIG_SYSTEM_SETTINGS_VALUE_SIZE);
   if (strlen(str) >= CONFIG_SYSTEM_SETTINGS_VALUE_SIZE)
```
cederom, line 252: `strlen(str) != CONFIG_SYSTEM_SETTINGS_VALUE_SIZE` ? :-)

system/settings/settings.c
jeanthom: I think the `strlen(key) < CONFIG_SYSTEM_SETTINGS_KEY_SIZE` is correct: the keys can have mixed lengths.

system/settings/settings.c
jeanthom: Similarly to my comment about key lengths: values can have mixed lengths.

acassis:
@TimJTi since you are the original author, please take a look too!

system/settings/settings.c
acassis: @jeanthom so KEY_SIZE mean max size in this case, right? Maybe we can rename it to CONFIG_SYSTEM_SET_MAX_KEY_SIZE. @TimJTi ?

TimJTi: > @jeanthom so KEY_SIZE mean max size in this case, right? Maybe we can rename it to CONFIG_SYSTEM_SET_MAX_KEY_SIZE. @TimJTi ?

Yes it's a maximum, which is stated in Kconfig, so a change to the name does make sense.

jeanthom: What's the usual way of dealing with config name changes in NuttX?

TimJTi: > What's the usual way of dealing with config name changes in NuttX?

Not sure if there's an official answer...but I have now learnt to use make oldconfig when things aren't going right after NuttX updates. That finds inconsistencies and allows you to sort things out. But I imagine some folk might really prefer not to see config name changes as it creates work!!!

@acassis ??

acassis: @TimJTi in this case we will not have (AFAIK) any side effect because there is no board modifying the default value of CONFIG_SYSTEM_SETTINGS_VALUE_SIZE. So this change will not impact the current status of mainline. Just change in nuttx-apps: examples/settings/settings_main.c, include/system/settings.h, system/settings/Kconfig, system/settings/readme.md, system/settings/settings.c and system/settings/storage_text.c.

system/settings/settings.c
cederom: thanks for the feedback and all updates @jeanthom :-)

system/settings/settings.c
cederom: Looking at Kconfig descriptions says its the "maximum size" of key or value so it is okay to leave name untouched and avoid possible configuration breaks :-)

https://github.com/apache/nuttx-apps/blob/c4ff19a7e447dec2cdb188415650c5b0e2df9d9c/system/settings/Kconfig#L23-L27

https://github.com/apache/nuttx-apps/blob/c4ff19a7e447dec2cdb188415650c5b0e2df9d9c/system/settings/Kconfig#L29-L33

Looking again at the code, first assert triggered when `len(key)` was larger than `KEY_SIZE`, and later check returns error when len is larger equal than `KEY_SIZE`. Because `KEY_SIZE` may be shorter all is okay, thanks for the info, lets keep it that way! :-)

```
assert(strlen(key) < CONFIG_SYSTEM_SETTINGS_KEY_SIZE);
if (strlen(key) >= CONFIG_SYSTEM_SETTINGS_KEY_SIZE)
    {
      return -EINVAL;
    }
```

system/settings/settings.c
cederom: brackets still here inside case, just this one and we are ready to go, thanks @jeanthom  :-)

cederom (changes requested):
Thanks @jeanthom for all the updates and feedback, just one bracket inside case and we are ready to go :-)

system/settings/settings.c
TimJTi: Are you saying no brackets for the case? NuttX style guide says there should be?

jeanthom: > Case logic in braces. It is preferable that all case logic (except for the break) be enclosed in braces. If you need to instantiate local variables in case logic, then that logic must be surrounded with braces.

Case logic in braces is indeed preferred. Should I revert my previous commit or removing the last bracket in existence?

TimJTi: > > Case logic in braces. It is preferable that all case logic (except for the break) be enclosed in braces. If you need to instantiate local variables in case logic, then that logic must be surrounded with braces.
> 
> Case logic in braces is indeed preferred. Should I revert my previous commit or removing the last bracket in existence?

My opinion? Yes. I would like consistent code everywhere and don't like the "preferable" qualification in the guide. But I am not an official reviewer so suggest the views of others wins the day....and maybe we need to update the guide here?

cederom: Argh, you are right @TimJTi, braces are preferred [1], its a bit inconsistent in existing code, my apologies @jeanthom, please revert the braces, my mistake sorry :-( :-(

[1] https://nuttx.apache.org/docs/latest/contributing/coding_style.html#switch-statement

cederom: I hope around 13 we can move to some standard format to auto format and format verification tools can be applied easily :-)

TimJTi: > I hope around 13 we can move to some standard format to auto format and format verification tools can be applied easily :-)

I guess this is something nxstyle could trap...but when it's "optional...unless" it's difficult. So if this is really the preferred style, we should change it to be a "must" and update nxstyle?

I believe "our" style is GNU, or, at least, derived from GNU....looking for a reference right now!

TimJTi: OK...so not GNU. Must be a Gregory-ism lol.
https://www.gnu.org/software/c-intro-and-ref/manual/c-intro-and-ref.html#switch-Example

I am not brave enough to suggest we drop this nxstyle guideline!

jeanthom: Hehe no worries 😄  I'll fix a minor code style issue (there is currently no line return after the case 'break' statement) and revert the commit tomorrow morning.

cederom: Thank you @jeanthom and sorry again for inconvenience :-)

jerpelea (approved):

TimJTi:
> > @TimJTi: @acassis @jerpelea I have been asked to review this, which I would be delighted to do!
> > Can either of you - or anyone else - suggest an efficient way to do this with some tips, so I don't waste tons of time doing things the hard way?
> 
> Tim you can clone the source repo switch to branch that constitutes this PR build and see if that works for you :-) If you have existing application that uses settings you may want to build that too and see if anything is broken and works as expected - all should work fine with your existing application. Thank you :-)
> 
> I have dedicated nuttx/pr/nuttx.git and nuttx/pr/nuttx-apps.git clones where I add remotes to be tested, here is an example:
> 
> 1. `mkdir nuttx-pr` - create pr test workbench.
> 2. `cd nuttx-pr` - go to pr test workbench.
> 3. `git clone https://github.com/apache/nuttx.git nuttx` - clone nuttx repo.
> 4. `git clone https://github.com/apache/nuttx-apps.git` - clone nuttx-apps repo.
> 5. `cd nuttx-apps` - go to cloned apps repo.
> 6. `git remote add lambdaconcept https://github.com/lambdaconcept/nuttx-apps.git` - add remote where pr source resides.
> 7. `git fetch --all` - get changes from all remotes.
> 8. `git checkout lambdaconcept/system-settings-cleanup` - switch to branch that sources this PR.
> 9. you are ready to test :-)

Thanks @cederom 

And I see nicely "quoted"  added/deleted lines that are then commented on (and in the past I have seen suggested changes that can be accepted here). Is that done manually from the git change log or is there a "trick" to that too?

acassis:
Thank you @jeanthom !

PR #3112, opened 2025-06-26
apps/can: Replace fprintf with dprint

acassis:
## Summary

This modification will remove the dependency on
CONFIG_FILE_STREAM, useful for small devices.

## Impact

Fix build error when CONFIG_FILE_STREAM is not enabled

## Testing

stm32f103-minimum

### DIFF
--- a/examples/can/can_main.c
+++ b/examples/can/can_main.c
@@ -105,7 +105,7 @@
 
 static void show_usage(FAR const char *progname)
 {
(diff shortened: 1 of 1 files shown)


cederom (approved):
Thank you @acassis :-) I can see you also fixed stderr -> stdout :-)

pkarashchenko (approved):

PR #3122, opened 2025-07-07
system/coredump: Fix missing loglevel to logmask conversion

nmaggioni:
Companion PR to apache/nuttx#16690, since the same logic is also duplicated in the `coredump` app here.

### DIFF
--- a/system/coredump/coredump.c
+++ b/system/coredump/coredump.c
@@ -418,7 +418,7 @@ static int coredump_now(int pid, FAR char *filename)
 #endif
 
   printf("Start coredump:\n");
-  logmask = setlogmask(LOG_ALERT);
+  logmask = setlogmask(LOG_UPTO(LOG_ALERT));
 
   /* Initialize hex output stream */
 


cederom (changes requested):
Thank you @nmaggioni good catch!! :-)

Could you please update commit topic and message to something like this below? Just as in nuttx#16690 :-)

> coredump: Fix missing loglevel to logmask conversion.
> 
> The current setlogmask call used in coredump_dump_syslog specifies a raw log level
> instead of a bitmask, this causes erroneous evaluations later on when that value
> is checked (i.e. lib_syslog.c:57), therefore LOG_UPTO macri is used for conversion.
>
> Signed...

This is because git logs should contain brief description of the change for people that does not use github (cannot see the pr details / description) and we are fixing missing loglevel to logmask conversion and not adding any new macro thus also topic update :-)

xiaoxiang781216 (approved):

jerpelea (approved):
please update commit title to 
system/coredump: Fix missing loglevel to logmask conversion.

jerpelea (approved):

PR #3133, opened 2025-07-14
system/zlib: expose zlib interface to other applications

pkarashchenko:
## Summary
Expose zlib interface to other applications

## Impact
Other applications may include `zlib.h` and use zlib APIs

## Testing
Tested by using `zlib.h` with other NuttX based application

### DIFF
--- a/system/zlib/Make.defs
+++ b/system/zlib/Make.defs
@@ -23,4 +23,8 @@
 
 ifneq ($(CONFIG_LIB_ZLIB),)
 CONFIGURED_APPS += $(APPDIR)/system/zlib
+
+CFLAGS +=  ${INCDIR_PREFIX}$(APPDIR)/system/zlib/zlib
+CXXFLAGS +=  ${INCDIR_PREFIX}$(APPDIR)/system/zlib/zlib
+
 endif


jerpelea (approved):
please add a commit message to explain the change

system/zlib/Make.defs
```diff
@@ -23,4 +23,8 @@
 
 ifneq ($(CONFIG_LIB_ZLIB),)
 CONFIGURED_APPS += $(APPDIR)/system/zlib
+
+CFLAGS +=  ${INCDIR_PREFIX}$(APPDIR)/system/zlib/zlib
```
xiaoxiang781216, line 27: but many intetrnal header files will expose to the global search path, which may introduce the conflict.

pkarashchenko: Let me check if `${INCDIR_PREFIX}$(APPDIR)/system/zlib/zlib/zlib.h` would work

pkarashchenko: Nope, it does not work. @xiaoxiang781216 what would be your suggestions? Abandon this change?

pkarashchenko: Let me try creating `include/zlib.h` wrapper at application layer. Maybe that would work.

pkarashchenko: Didn't work well, but I just added
```
CFLAGS +=  ${INCDIR_PREFIX}$(APPDIR)/system/zlib/zlib
CXXFLAGS +=  ${INCDIR_PREFIX}$(APPDIR)/system/zlib/zlib
```
to my application Makefile and put `ZLIB` dependency in `Kconfig`, so seems like that resolved an issue.

PR #3136, opened 2025-07-17
Nxboot direct access

michallenc:
## Summary

**619ea01 boot/nxboot/loader/flash.c: open partition with O_DIRECT flag**

There is no need to use buffers in a bootloader. We expect a sequential
access, therefore we just need to erase the erase page before writing
to it for the first time, which is something FTL layer already takes
care of.

**95e4b0c boot/nxboot/loader/boot.c: copy partition with blocksize large writes**

The previous logic `MAX(info_from.blocksize, info_where.blocksize)` was
incorrect. The most effective access with by writing the entire
size of the block, therefore just decide the size based on the
target page size.

## Impact

Nxboot now utilizes changes presented in https://github.com/apache/nuttx/pull/16642. Should not have impact on current implementation as the additional flags are ignored if changes in https://github.com/apache/nuttx/pull/16642 are not applied.

## Testing

Tested with https://github.com/apache/nuttx/pull/16642 patch applied. Nxboot successfully updates and reverts new image. All works as before, but additional buffering is skipped.

### DIFF
--- a/boot/nxboot/loader/boot.c
+++ b/boot/nxboot/loader/boot.c
@@ -164,7 +164,7 @@ static int copy_partition(int from, int where, struct nxboot_state *state,
 #ifdef CONFIG_NXBOOT_PRINTF_PROGRESS_PERCENT
   total_size = remain * 100;
 #endif
-  blocksize = MAX(info_from.blocksize, info_where.blocksize);
+  blocksize = info_where.blocksize;
 
   buf = malloc(blocksize);
   if (!buf)
--- a/boot/nxboot/loader/flash.c
+++ b/boot/nxboot/loader/flash.c
@@ -64,7 +64,7 @@ int flash_partition_open(const char *path)
 {
   int fd;
 
-  fd = open(path, O_RDWR);
+  fd = open(path, O_RDWR | O_DIRECT);
   if (fd < 0)
     {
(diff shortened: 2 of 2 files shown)


michallenc:
@TimJTi adding this as a draft now, could you test it with your environment when you have time? Thanks!

jerpelea (approved):

michallenc:
@TimJTi Hi, did you find time to test the patch? Thanks.

TimJTi:
Today's task (one of). Sorry for the delay !

TimJTi:
@michallenc @xiaoxiang781216 
I can confirm this work nicely on my board. I rebuilt my bootloader with this patch, uploaded a new secondary image and it correctly copied primary to backup and secondary to primary. I did not validate the new image, and it correctly reverted primary to the backup image.

And it seems very much faster too, unless it's my imagination!?

PS: I found an error in my recently added progress percentage display for NXBoot, and will shortly submit a small PR for that.

TimJTi:
> Thanks! Yes, it should in theory be faster with the direct access, we avoid some unnecessary buffering and erases.

Excellent 👍

cederom (approved):
Thank you @michallenc great work! Also thank you @TimJTi for verification! :-)

PR #3137, opened 2025-07-18
kasantest: -fno-builtin is enabled by default

W-M-R:
The strlen function called by kasantest will be optimized by the compiler, and the strlen function implemented in NX will not be called. Adding -fno-builtin can solve the problem.

## Summary

The strlen function called by kasantest will be optimized by the compiler, and the strlen function implemented in NX will not be called. Adding -fno-builtin can solve the problem.

## Impact

No

## Testing

ci


jerpelea (approved):
please update the commit title to 
testing/kasantest: -fno-builtin is enabled by default

PR #3154, opened 2025-08-09
[EXPERIMENTAL] Makefile: EXTRA_APPS_LIBS and EXTRA_APPS_INCPATHS vars

zdebanos:
The variables are to be used to mainly by the export rule to export external libraries (.a files) and respective include paths.

The exported external libraries are packed together in the libapps.a library (basically, every .o files are extracted and then packed together).

Signed-off-by: Stepan Pressl <pressl.stepan@gmail.com>

## Summary
Improving the build system with these variables so all the symbols from the external libraries can be packed together within the libapps.a library. This is particularly useful, when we'd like to use this external library and all of its symbols in the nuttx export.

I'm not expecting to be this merged, yet. But I'd like to start a debate over possible solutions and discuss other solutions and improvements. I'm thinking about generalizing the variables over multiple Makefiles, so any directory in the NuttX project would have some kind of a variable. Or possibly only one variable.

A different way of doing this is to just all of these libraries separately, and not withing libapps.a, etc.

Also, external libraries are, in the current state, linked using the EXTRA_LIBS variable. But that is done only in the last stage of the build and the linker only picks symbols that are needed. In the end, the exported library only has a subset of all the symbols.

Also cmake support must be added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,28 @@ preconfig: Kconfig
 export:
 ifneq ($(EXPORTDIR),)
 	$(Q) mkdir -p "${EXPORTDIR}"$(DELIM)registry || exit 1;
+	if [ -n "${EXTRA_APPS_LIBS}" ]; then \
+		mkdir -p $(APPDIR)$(DELIM)libcomb; \
+		cp libapps.a $(APPDIR)$(DELIM)libcomb; \
+		for l in ${EXTRA_APPS_LIBS}; do \
+			echo "AR: $$(basename $${l})"; \
+			cp $${l} $(APPDIR)$(DELIM)libcomb; \
+		done; \
+		cd $(APPDIR)$(DELIM)libcomb; \
+		for l in *.a; do \
+			$(AR_EXTRACT) $${l}; \
+		done; \
+		$(AR) libmerged.a *.o; \
+		mv libmerged.a ..$(DELIM)libapps.a; \
+		cd ..; \
+		rm -rf $(APPDIR)$(DELIM)libcomb; \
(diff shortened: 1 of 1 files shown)


acassis:
@zdebanos I suggest converting it to Draft until we have an equivalent support to CMakefile to avoid build system asymmetry.

cederom:
Thanks @zdebanos, converted PR to Draft as suggested by @acassis, when all is set we will mark it ready for review :-)

zdebanos:
Does Nuttx actually support the `export` target as of today?

acassis:
> Does Nuttx actually support the `export` target in cmake as of today?

I think it is not supported yet, cmake is relatively new on NuttX and it is not the official building system (maybe it could become in the future or maybe not) :-)

ppisa:
For documentation, the `EXPORTED_INCLUDES` define from `Make.defs` allows export of include files provided by third party downloaded packages, see commit

https://github.com/apache/nuttx-apps/commit/564cabe9b1a9d37b058471b81480f7cb27632c8d build: introduce EXPORTED_INCLUDES machanism

and for `EXPORTED_INCLUDES` use

https://github.com/apache/nuttx-apps/pull/3689/changes/f5e91f7f4e2b3810303186386b5f69f7fa8b60fb example/microwindows: exports microwindows/src/include

SHV headers are exported as well

https://github.com/apache/nuttx-apps/commit/94012303daa58baf5e956a8e234c07bbe22182e8 netutils/libshvc: implement headers export and update shv-libs4c

CMake support is not implemented yet

PR #3159, opened 2025-08-12
crypto/libsodium: Update libsodium to 1.0.20

nmaggioni:
## Summary

This PR updates libsodium to the latest 1.0.20 tagged release.

## Impact

N/A

## Testing

These changes were tested on a Linux host, targeting a custom Cortex-M0+ (RP2040) board using NuttX v12.8.0 as the baseline. They were then manually ported to the current master branch.


jerpelea (changes requested):
please add a commit message which describes the chane

cederom:
Thanks @nmaggioni :-) Is 1.0.20 compatible with 1.0.18? No other application will be broken after this update?

nmaggioni:
@cederom Not as far as I know: the [changelogs](https://github.com/jedisct1/libsodium/releases) report no breaking changes and my HMAC usages in NuttX still work.

On the other hand, IIRC, the tests included in libsodium were already failing compilation before the upgrade. It's been a while since I poked at them so don't take my word for it, and I didn't dive much into that anyway: the line between a simple patch like the one that's already included here and something heavier that would need upstreaming is quite thin.

PR #3161, opened 2025-08-12
tools/mksymtab: Replace deprecated fgrep usages

nmaggioni:
## Summary

This PR replaces the usages of `fgrep`, deprecated in most distros in favor of `grep -F`.

## Impact

This removes noisy deprecation warnings emitted at runtime.

## Testing

These changes were tested on a Linux host.

### DIFF
--- a/tools/mksymtab.sh
+++ b/tools/mksymtab.sh
@@ -72,10 +72,10 @@ if [ -z "$varlist" ]; then
   if [ ! -z "$execlist" ]; then
 
 # Get all undefined symbol names
-    varlist=`nm $execlist 2>/dev/null | fgrep ' U ' | sed -e "s/^[ ]*//g" | cut -d' ' -f2 | sort | uniq`
+    varlist=`nm $execlist 2>/dev/null | grep -F ' U ' | sed -e "s/^[ ]*//g" | cut -d' ' -f2 | sort | uniq`
 
 # Get all defined symbol names
-    deflist=`nm $execlist 2>/dev/null | fgrep -v -e ' U ' -e ':' | sed -e "s/^[0-9a-z]* //g" | cut -d' ' -f2 | sort | uniq`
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

cederom (approved):
Thanks @nmaggioni :-) Tested with stm32l432kc, `grep -F` works on FreeBSD :-)

Ekhm, these "modern distros" ale the Linux distros ;-) If you look at `man fgrep` [1] "grep command first appeared in Version 6 AT&T UNIX" and that dates back to May 1975 [2] wow :-)

```
GREP(1)                 FreeBSD General Commands Manual                GREP(1)

NAME
     grep, egrep, fgrep, rgrep – file pattern searcher

SYNOPSIS
     grep [-abcdDEFGHhIiLlmnOopqRSsUVvwxz] [-A num] [-B num] [-C num]
          [-e pattern] [-f file] [--binary-files=value] [--color[=when]]
          [--colour[=when]] [--context=num] [--label] [--line-buffered]
          [--null] [pattern] [file ...]
(..)
     -F, --fixed-strings
             Interpret pattern as a set of fixed strings (i.e., force grep to
             behave as fgrep).
(..)
STANDARDS
     The grep utility is compliant with the IEEE Std 1003.1-2008 (“POSIX.1”)
     specification.

     The flags [-AaBbCDdGHhILmopRSUVw] are extensions to that specification,
     and the behaviour of the -f flag when used with an empty pattern file is
     left undefined.

     All long options are provided for compatibility with GNU versions of this
     utility.

     Historic versions of the grep utility also supported the flags [-ruy].
     This implementation supports those options; however, their use is
     strongly discouraged.

HISTORY
     The grep command first appeared in Version 6 AT&T UNIX.

FreeBSD 14.2-RELEASE-p1        December 18, 2022       FreeBSD 14.2-RELEASE-p1
```

[1] https://man.freebsd.org/cgi/man.cgi?query=fgrep
[2] https://en.wikipedia.org/wiki/Version_6_Unix

PR #3169, opened 2025-08-27
boot/nxboot/loader/boot.c: avoid unwanted confirm on double update

michallenc:
## Summary
Following steps led to incorrect behavior:
 - upload update image
 - restart device and perform the update
 - upload the same update image again without confirming

This led to the unwanted confirm of the update image. This change ensures the update image is not confirmed by uploading the same image to the update slot. The correct behavior is to perform revert to the last stable image if this occurs.

## Impact
None on current update process, tested with both standard update and revert, works as before.

ping @TimJTi if you want to verify

## Testing
Tested on samv7 custom board for standard update and revert processes + double update described above.

### DIFF
--- a/boot/nxboot/loader/boot.c
+++ b/boot/nxboot/loader/boot.c
@@ -313,6 +313,8 @@ static enum nxboot_update_type
           nxboot_progress(nxboot_progress_end);
           return NXBOOT_UPDATE_TYPE_UPDATE;
         }
+
+        flash_partition_erase_first_sector(update);
     }
 
   nxboot_progress(nxboot_progress_end);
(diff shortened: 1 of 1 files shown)


TimJTi:
@michallenc I have seen this behaviour (thanks to progress reporting lol) and did wonder.

I presume, after this change, that if the first upload of an image is confirmed and valid, that a second upload of the same image won't "revert", as such, but retain the already confirmed, previous upload, as the primary?

I won't get an opportunity to check this change for a week or two, but I have no problem with it and happy if it is committed before I do get a chance to try it 👍

PR #3181, opened 2025-09-14
netutils: Add plcatool

matiamic:
## Summary

This patch adds `plcatool` app into netutils. It is a simple tool that allows configuration of the Physical Layer Collision Avoidance (PLCA) functionality in 10BASE-T1S PHYs. The syntax and behavior are very similar to `ethtool`'s PLCA options. The tool is compatible with PHYs that follow the [OA-TC14 specification](https://opensig.org/wp-content/uploads/2025/03/2024-12-10BASE-T1S-PLCA-Management-Registers-v1.4.pdf) defining the mapping of PLCA registers into MMD 31.

## Impact

A new directory `plcatool` in `netutils` with application files is created.

## Testing
The app was tested with Microchip LAN865x and Onsemi NCN7410 (NCN26010) MAC-PHYs on ESP32C6-devkitC, ESP32C6-devkitM and SAMV71-Xult boards. The behavior of coordinator (node-id 0) was verified on an oscilloscope by looking for transmitted beacon signal in expected intervals. Follower (node-id > 0) was verified by checking for '1' in PLCA status register when periodically receiving beacon and '0' otherwise.


netutils/plcatool/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+# ##############################################################################
+# apps/netutils/plcatool/CMakeLists.txt
```
xiaoxiang781216, line 2: should we name it ethtool and make plca as one of component, so we can add more functionality to ethtool step by step.

cederom: +1 :-)

ppisa: I am not sure because there is some remains of `ethtool` mentioning in the NuttX code and may it be, separate tool for specific purpose is better for start when it is complete for its basic task. The use it as template for other more generic tool or even combining functionality into one application could be good evolution. But I think that providing application which would take the name `ethtool` and people would find it and then be disappointed that it does not have basic equivalent functionality of Linux `ethtool` is not ideal.

May it be @michallenc or @Cynerd could provide their input there as well. They are in need of PLCA functionality now.

matiamic: I agree with @ppisa. Support for PLCA in `ethtool` is relatively new (2023) and I think it is not the feature most people use it for. Given it would be the only feature supported in the NuttX version for some time I can imagine this could lead to confusion.

Also I think that the eventual `ethtool` implementation should be part of more focused effort in order to achieve correct argument parsing and extensibility.

cederom: sounds good :-) lets start with what we have and works :-)

cederom (approved):
Thank you @matiamic :-) Please remember to add documentation! :-)

netutils/plcatool/plcatool.c
```diff
@@ -0,0 +1,717 @@
...
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+static int get_num(FAR const char *str, FAR int *result)
+{
+  FAR char *endptr;
+  *result = (int) strtol(str, &endptr, 0);
+  if (errno)
```
xiaoxiang781216: need zero errno first

netutils/plcatool/plcatool.c
```diff
@@ -0,0 +1,717 @@
...
+static void plcatool_usage(bool err);
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+static int get_num(FAR const char *str, FAR int *result)
+{
+  FAR char *endptr;
+  *result = (int) strtol(str, &endptr, 0);
```
xiaoxiang781216:
```suggestion
  *result = (int)strtol(str, &endptr, 0);
```

netutils/plcatool/plcatool.c
```diff
@@ -0,0 +1,717 @@
...
+  if (errno)
+    {
+      return ERROR;
```
xiaoxiang781216: replace ALL ERROR with the correct EXXX

netutils/plcatool/plcatool.c
```diff
@@ -0,0 +1,717 @@
...
+                  PLCATOOL_CFG_SET(cfg, enable, 0);
+                }
+              else
+                {
+                  return ERROR;
+                }
+            }
+          else if (strcmp(param, "node-id") == 0)
+            {
+              int N;
```
xiaoxiang781216:
```suggestion
              int n;
```

netutils/plcatool/plcatool.c
```diff
@@ -0,0 +1,717 @@
...
+                }
+              else
+                {
+                  fprintf(stderr, "node-id out of range\n");
+                  return ERROR;
+                }
+            }
+          else if (strcmp(param, "node-cnt") == 0)
+            {
+              int N;
```
xiaoxiang781216:
```suggestion
              int n;
```

netutils/plcatool/plcatool.c
```diff
@@ -0,0 +1,717 @@
...
+                }
+              else
+                {
+                  fprintf(stderr, "node-cnt out of range\n");
+                  return ERROR;
+                }
+            }
+          else if (strcmp(param, "to-tmr") == 0)
+            {
+              int N;
```
xiaoxiang781216:
```suggestion
              int n;
```

netutils/plcatool/plcatool.c
```diff
@@ -0,0 +1,717 @@
...
+                }
+              else
+                {
+                  fprintf(stderr, "to_tmr out of range\n");
+                  return ERROR;
+                }
+            }
+          else if (strcmp(param, "burst-cnt") == 0)
+            {
+              int N;
```
xiaoxiang781216:
```suggestion
              int n;
```

netutils/plcatool/plcatool.c
```diff
@@ -0,0 +1,717 @@
...
+                }
+              else
+                {
+                  fprintf(stderr, "burst-cnt out of range\n");
+                  return ERROR;
+                }
+            }
+          else if (strcmp(param, "burst-tmr") == 0)
+            {
+              int N;
```
xiaoxiang781216:
```suggestion
              int n;
```

netutils/plcatool/plcatool.c
ppisa: I would not use `errno` for the check of `strtol` at all. There two way how to check that conversion succeeded
```
if (str == endptr)
   {
      report conversion error, this one does not detect if the start of the number is correct but it ends by garbage
  }
```
or
```
if (*endptr)
   {
      error
  }
```
For the regular NULL character terminated string the, whole argument should be processed by `strtol` if there is something left then it is error. If there could be more values in the single argument then check can be
```
if (*endptr && !isblank(*endptr) && (*endptr !=','))
   {
      error
  }
```

netutils/plcatool/CMakeLists.txt
Cynerd: I just want to add my agreement. Let's generalize when we have the need. For now, I think that having the PLCA tool is a good step. If additional tools are created for MDIO, then we can think about merging the functionality, but at that point, it should be much clearer what that would look like.

netutils/plcatool/plcatool.c
matiamic: I think (and based on tests) the endptr condition does not catch when the number string is out of range of the output type. In that case `strtol` returns either `LONG_MAX` or `LONG_MIN`. This changes the number the user probably thinks was entered.
But I agree in this case it is redundant because the number is later checked for much smaller range. Also casting to `(int)` can have the same effect and is not cared for in any way.

netutils/plcatool/plcatool.c
matiamic: Here the `ERROR` return value is part of the internal logic of the program. Every "shell-facing" return value is the correct EXXX.

jerpelea (approved):

PR #3185, opened 2025-09-18
oot ci build test

trns1997:
Complement with https://github.com/apache/nuttx/pull/17017

### DIFF
--- a/testing/oot-build/.gitignore
+++ b/testing/oot-build/.gitignore
@@ -1,0 +1,2 @@
+build/
+nuttx-export*
--- a/testing/oot-build/CMakeLists.txt
+++ b/testing/oot-build/CMakeLists.txt
@@ -1,0 +1,35 @@
+cmake_minimum_required(VERSION 3.12...3.31)
+
+project(OOTCpp
+    VERSION 1.0
+    DESCRIPTION "Out Of Tree Build C++ NuttX"
+)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# Use CMAKE_SOURCE_DIR for source files
+set(SOURCE_FILES
+    ${CMAKE_SOURCE_DIR}/src/HelloWorld.cpp
+    ${CMAKE_SOURCE_DIR}/src/main.cpp
+)
+
+set(EXE_NAME oot)
+
+add_executable(${EXE_NAME} ${SOURCE_FILES})
+
+# Add include directory so you can #include "..." from include/
+target_include_directories(${EXE_NAME}
+        PRIVATE
+        ${CMAKE_SOURCE_DIR}/include
+)
+
+# Generate a .bin file from the ELF after build
+add_custom_command(
+    TARGET ${EXE_NAME}
+    POST_BUILD
+    COMMAND ${CMAKE_OBJCOPY} -S -O binary
+            ${CMAKE_BINARY_DIR}/${EXE_NAME}
+            ${CMAKE_BINARY_DIR}/${EXE_NAME}.bin
+    COMMENT "Generating binary image ${EXE_NAME}.bin"
+)
--- a/testing/oot-build/include/HelloWorld.hpp
+++ b/testing/oot-build/include/HelloWorld.hpp
@@ -1,0 +1,13 @@
+#pragma once
+
+class CHelloWorld
+{
+public:
+    CHelloWorld();
+    ~CHelloWorld() = default;
+
+    bool HelloWorld();
+
+private:
+    int mSecret;
+};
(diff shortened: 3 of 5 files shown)


testing/cxx-oot-build/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+cmake_minimum_required(VERSION 3.12...3.31)
```
xiaoxiang781216, line 23: do we need add Kconfig?

trns1997: no this cannot be built using using classical apporaches. The aim to to build only out of tree using the export

testing/oot-build/.gitignore
```diff
@@ -0,0 +1,2 @@
+build/
+nuttx-export*
```
trns1997: add new line

cederom (dismissed):
Thank you @trns1997 :-) Please take a look at [1] [2] [3] and update git commit message and PR description to match requirements :-)

[1] https://github.com/apache/nuttx/blob/master/CONTRIBUTING.md
[2] https://nuttx.apache.org/docs/latest/contributing/index.html
[3] https://nuttx.apache.org/docs/latest/contributing/coding_style.html

testing/cxx-oot-build/CMakeLists.txt
cederom: What about people using Make and Kconfig? Would it build for them? Please provide steps how to reproduce in Testing section. Thanks :-)

trns1997: the aim was just to have add a out of tree ci test, and not necessarily expose this as a testing app, but if you guys want i can look into it, but keep in mind the build via Kconfig and make is not a representative test for what we intended to do. It's a simple oot of tree build test to prevent regressions when packaging nuttx export, the app itself doesn't do much :).

trns1997: @cederom where is the testing section. I'll add a link in the section orienting users to the doc in nuttx, Once i address this comment https://github.com/apache/nuttx/pull/17017#issuecomment-3293446463 from @acassis

trns1997: ping @cederom

trns1997:
> @trns1997 you need fix ci error first https://github.com/apache/nuttx-apps/actions/runs/17955938892/job/51067275467?pr=3185
> 
> ```
> CMake Error at /github/workspace/sources/apps/testing/cxx-oot-build/CMakeLists.txt:19 (add_executable):
>   Cannot find source file:
> 
>     /github/workspace/sources/nuttx/src/HelloWorld.cpp
> ```

is there a mechanism to add this example folder in an ignore list?

testing/cxx-oot-build/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
```
xiaoxiang781216, line 23: add copyright

testing/cxx-oot-build/src/main.cpp
```diff
@@ -0,0 +1,14 @@
+#include <memory>
```
xiaoxiang781216, line 23: add copyright

testing/cxx-oot-build/include/HelloWorld.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
xiaoxiang781216, line 23: copyright too

testing/cxx-oot-build/src/HelloWorld.cpp
```diff
@@ -0,0 +1,30 @@
+#include <cstdio>
```
xiaoxiang781216, line 23: copyright too

trns1997:
@xiaoxiang781216 do we want to add the stage from https://github.com/apache/nuttx/pull/17017/files to the `build.yml` workflow file in the app as well? If yes, i suggest we merge this first, that way we can validate the pipeline succeed on nuttx and then once validated i'll create another PR to add the stage to nuttx-apps. What do you think?

raiden00pl:
> Or i update nuttx_add_subdirectory function to accommodate an exclusion logic?

This solution seems to be the best: add an optional argument to `nuttx_add_subdirectory()` (like `IGNORE_DIRS`) where you can pass ignored directories. This can be a useful feature for later.

trns1997:
> > Or i update nuttx_add_subdirectory function to accommodate an exclusion logic?
> 
> This solution seems to be the best: add an optional argument to `nuttx_add_subdirectory()` (like `IGNORE_DIRS`) where you can pass ignored directories. This can be a useful feature for later.

ah @raiden00pl, i already implemented the guard approach. But i do not mind creating this feature in another PR. I do not want to mix this test with this feature. If you do not mind can you create the issue and assign it to me?

PR #3192, opened 2025-10-17
system/nxinit: Add NuttX Init ("NxInit")

JianyuWang0623:
## Summary

Related PR:
* Documentation: https://github.com/apache/nuttx/pull/17215.
* qemu/fastboot config update: https://github.com/apache/nuttx/pull/17216

### Add NuttX Init ("NxInit")
This component (abbreviated as "NxInit") is specifically developed for NuttX and intended for system initialization. While we have adopted the Android Init Language among various syntax options, this is a brand-new implementation—it is not a port or variant of Android Init.

Refer to the implementation of Android Init Language, consists of five broad classes of statements: Actions, Commands, Services, Options, and Imports.

Actions support two types of triggers: event and action. Action triggers also support runtime triggering. Services support lifecycle management, including automatic restart (at specified intervals), and starting/stopping individually or by class. Import supports files or directories, and we may add a static method in the future. The following are some differences:
  1. The Android Init Language treats lines starting with `#` as comments, while we use a preprocessor to handle comments.
  2. For action commands, we can omit "exec" and directly execute built-in apps or nsh builtins.
  3. Regarding the property service, users can either adapt it by self or directly use the preset NVS-based properties.
  4. Only part of standard action commands and service options are implemented currlently.

To enable system/init:
  ```diff
  -CONFIG_INIT_ENTRYPOINT="nsh_main"
  +CONFIG_INIT_ENTRYPOINT="init_main"
  +CONFIG_SYSTEM_NXINIT=y
  ```

For format and additional details, refer to: https://android.googlesource.com/platform/system/core/+/master/init/README.md

### Add class start/stop for service
Usage:
```
      class_start <classname>
      class_stop <classname>
```

### Add NSH builtin support for action
Enable `CONFIG_SYSTEM_SYSTEM` to support nsh builtins, which depends on nsh.

### Warning for long commands
If the command of an action takes too long (greater than `CONFIG_SYSTEM_NXINIT_ACTION_WARN_SLOW` milliseconds, 50 ms by default), a warning log will be output for analysis and debugging.

For example:

  1. sleep 1
  ```
       [    0.340000] [ 3] [ 0] init_main: executing NSH command 'sleep 1'
       [    1.360000] [ 3] [ 0] init_main: NSH command 'sleep 1' exited 0
     > [    1.360000] [ 3] [ 0] init_main: command 'sleep' took 1020 ms
  ```
  2. hello (add sleep(1) to examples/hello)
  ```
       [    1.390000] [ 3] [ 0] init_main: executed command 'hello' pid 14
       [    1.390000] [ 3] [ 0] init_main: waiting 'hello' pid 14
       Hello, World!!
     > [    2.400000] [ 3] [ 0] init_main: command 'hello' pid 14 took 1010 ms
       [    2.400000] [ 3] [ 0] init_main: command 'hello' pid 14 exited status 0
  ```

### Add oneshot support for service
Add support for the oneshot option to the service.

Test
  - RC
  ```
      service telnet telnetd
          class test
    >     oneshot
          restart_period 3000
  ```
  - Runtime
   ```
      [    0.150000] [ 3] [ 0] init_main: == Dump Services ==
      ...
      [    0.160000] [ 3] [ 0] init_main: Service 0x40486aa8 name 'telnet' path 'telnetd'
      [    0.160000] [ 3] [ 0] init_main:   pid: 0
      [    0.160000] [ 3] [ 0] init_main:   arguments:
      [    0.160000] [ 3] [ 0] init_main:       [0] 'service'
      [    0.160000] [ 3] [ 0] init_main:       [1] 'telnet'
      [    0.160000] [ 3] [ 0] init_main:       [2] 'telnetd'
      [    0.160000] [ 3] [ 0] init_main:   classes:
      [    0.160000] [ 3] [ 0] init_main:     'test'
      [    0.170000] [ 3] [ 0] init_main:   restart_period: 3000
      [    0.170000] [ 3] [ 0] init_main:   reboot_on_failure: -1
      [    0.170000] [ 3] [ 0] init_main:   flags:
    > [    0.170000] [ 3] [ 0] init_main:     'oneshot'
      ...
      [    0.370000] [ 3] [ 0] init_main: starting service 'telnet' ...
      [    0.380000] [ 3] [ 0] init_main: service 'telnet' flag 0x2 add 0x4
      [    0.380000] [ 3] [ 0] init_main:   +flag 'running'
      [    0.380000] [ 3] [ 0] init_main: service 'telnet' flag 0x6 add 0x8
      [    0.380000] [ 3] [ 0] init_main:   -flag 'restarting'
      [    0.380000] [ 3] [ 0] init_main: service 'telnet' flag 0x6 add 0x1
      [    0.380000] [ 3] [ 0] init_main:   -flag 'disabled'
      [    0.380000] [ 3] [ 0] init_main: started service 'telnet' pid 9
      ...
      nsh> kill -9 9
      nsh> [    7.350000] [ 3] [ 0] init_main: service 'telnet' flag 0x6 add 0x4
      [    7.350000] [ 3] [ 0] init_main:   -flag 'running'
      [    7.350000] [ 3] [ 0] init_main: service 'telnet' flag 0x2 add 0x80000001
      [    7.350000] [ 3] [ 0] init_main:   +flag 'disabled'
      [    7.350000] [ 3] [ 0] init_main:   +flag 'remove'
      [    7.350000] [ 3] [ 0] init_main: service 'telnet' pid 9 exited status 1
    > [    7.360000] [ 3] [ 0] init_main: removing service 'telnet' ...
   ```

### Add exec_start support for action
Format: `exec_start <service>`

Start the specified service and pause the processing of any additional initialization commands until the service completes its execution. This command operates similarly to the `exec` command; the key difference is that it utilizes an existing service definition rather than requiring the `exec` argument vector.

This feature is particularly intended for use with the `reboot_on_failure` built-in command to perform all types of essential checks during system boot.

### Add reboot_on_failure for service
Add support for the reboot_on_failure option to the service.

When the execution of a command within a certain action fails (returning a non-zero status code), NxInit will continue to execute subsequent commands or actions and will not proactively terminate the startup process. To implement the functionality of "terminating the startup process after a command execution fails", there are two methods:
1. Execute conditional statements (if/then/else/fi) via exec command, but this depends on sh:
   ```
   on init
       exec -- set +e; \
               mount -t fatfs /dev/data /data ; \
               if [ $? -ne 0 ] ; \
               then \
                 echo "failed" ; \
                 reboot ; \
               else \
                 echo "succeed" ; \
               fi;
   ```
2. Via service's oneshot + reboot_on_failure:
Although the example uses sh, it does not depend on it and can be replaced with any other Builtin Apps.
   ```
   on init
       exec_start mkdir_tmp
       ls /tmp

   service mkdir_tmp sh -c "mkdir /tmp"
       reboot_on_failure 0
       oneshot
   ```
Test
  - RC
  ```
      service console sh
          class core
          override
    >     reboot_on_failure 0
          restart_period 10000
  ```
  - Runtime
  ```
      [    0.150000] [ 3] [ 0] init_main: == Dump Services ==
      ...
      [    0.170000] [ 3] [ 0] init_main: Service 0x40486ea0 name 'console' path 'sh'
      [    0.170000] [ 3] [ 0] init_main:   pid: 0
      [    0.170000] [ 3] [ 0] init_main:   arguments:
      [    0.170000] [ 3] [ 0] init_main:       [0] 'service'
      [    0.170000] [ 3] [ 0] init_main:       [1] 'console'
      [    0.170000] [ 3] [ 0] init_main:       [2] 'sh'
      [    0.170000] [ 3] [ 0] init_main:   classes:
      [    0.170000] [ 3] [ 0] init_main:     'core'
      [    0.170000] [ 3] [ 0] init_main:   restart_period: 10000
    > [    0.170000] [ 3] [ 0] init_main:   reboot_on_failure: 0
      [    0.170000] [ 3] [ 0] init_main:   flags:
      [    0.170000] [ 3] [ 0] init_main:     'override'
      ...
      [    0.380000] [ 3] [ 0] init_main: started service 'console' pid 4
      ...
      nsh> kill -9 4
      [    8.060000] [ 3] [ 0] init_main: service 'console' flag 0x20000004 add 0x4
      [    8.060000] [ 3] [ 0] init_main:   -flag 'running'
      [    8.060000] [ 3] [ 0] init_main: service 'console' flag 0x20000000 add 0x8
      [    8.060000] [ 3] [ 0] init_main:   +flag 'restarting'
    > [    8.060000] [ 3] [ 0] init_main: Error reboot on failure of service 'console' reason 0
  ```
## Impact
This component (abbreviated as "NxInit") is specifically developed for NuttX and intended for system initialization. While we have adopted the Android Init Language among various syntax options, this is a brand-new implementation—it is not a port or variant of Android Init.

## Testing
- Please see logs above for selftest <--- @JianyuWang0623 maybe the test logs could be placed here instead?
- CI

### DIFF
--- a/system/nxinit/init.c
+++ b/system/nxinit/init.c
@@ -75,6 +75,7 @@ static void reap_process(FAR struct service_manager_s *sm,
   FAR const char *status;
   FAR struct service_s *service;
   int wstatus;
+  int ret;
   int pid;
 
   for (; ; )
@@ -93,12 +94,12 @@ static void reap_process(FAR struct service_manager_s *sm,
       else if (WIFEXITED(wstatus))
         {
           status = "status";
-          wstatus = WEXITSTATUS(wstatus);
+          ret = WEXITSTATUS(wstatus);
         }
       else if (WIFSIGNALED(wtatus))
         {
           status = "signal";
-          wstatus = WTERMSIG(wstatus);
+          ret = WTERMSIG(wstatus);
         }
       else
         {
@@ -115,12 +116,12 @@ static void reap_process(FAR struct service_manager_s *sm,
       if (service != NULL)
         {
           name = service->argv[1];
-          init_service_reap(service);
+          init_service_reap(service, ret);
         }
 
       init_log(service ? LOG_WARNING : LOG_DEBUG,
                "%s '%s' pid %d exited %s %d",
-               service ? "Service" : "Command", name, pid, status, wstatus);
+               service ? "Service" : "Command", name, pid, status, ret);
     }
 }
 
--- a/system/nxinit/service.c
+++ b/system/nxinit/service.c
@@ -25,6 +25,7 @@
  ****************************************************************************/
 
 #include <nuttx/clock.h>
+#include <sys/boardctl.h>
 
 #include <assert.h>
 #include <errno.h>
@@ -99,6 +100,10 @@ static int option_override(FAR struct service_manager_s *sm,
                            int argc, FAR char **argv);
 static int option_oneshot(FAR struct service_manager_s *sm,
                           int argc, FAR char **argv);
+#ifdef CONFIG_BOARDCTL_RESET
+static int option_reboot_on_failure(FAR struct service_manager_s *sm,
+                                    int argc, FAR char **argv);
+#endif
 
 /****************************************************************************
  * Private Data
@@ -111,6 +116,9 @@ static const struct cmd_map_s g_option[] =
   {"restart_period", 2, 2, option_restart_period},
   {"override", 1, 1, option_override},
   {"oneshot", 1, 1, option_oneshot},
+#ifdef CONFIG_BOARDCTL_RESET
+  {"reboot_on_failure", 2, 2, option_reboot_on_failure},
+#endif
 };
 
 #ifdef CONFIG_SYSTEM_NXINIT_DEBUG
@@ -159,7 +167,7 @@ static int kill_service(FAR struct service_s *service, int signo)
       ret = -errno;
       if (ret == -ESRCH)
         {
-          init_service_reap(service);
+          init_service_reap(service, 0);
         }
     }
 
@@ -260,6 +268,17 @@ static int option_oneshot(FAR struct service_manager_s *sm,
   add_flags(s, SVC_ONESHOT);
   return 0;
 }
+
+#ifdef CONFIG_BOARDCTL_RESET
+static int option_reboot_on_failure(FAR struct service_manager_s *sm,
+                                    int argc, FAR char **argv)
+{
+  FAR struct service_s *s = list_last_entry(&sm->services, struct service_s,
+                                            node);
+  s->reset_reason = atoi(argv[1]);
+  return 0;
+}
+#endif
 
 /****************************************************************************
  * Public Functions
@@ -371,8 +390,25 @@ init_service_find_by_pid(FAR struct service_manager_s *sm, const int pid)
   return NULL;
 }
 
-void init_service_reap(FAR struct service_s *service)
+void init_service_reap(FAR struct service_s *service, int status)
 {
+#ifdef CONFIG_BOARDCTL_RESET
+  int ret;
+
+  if (status && service->reset_reason >= 0)
+    {
+      init_err("Reboot on failure of service '%s' reason %d",
+               service->argv[1], service->reset_reason);
+      ret = boardctl(BOARDIOC_RESET, service->reset_reason);
+      if (ret < 0)
+        {
+          init_err("Reset failed %d", errno);
+        }
+    }
+#else
+  UNUSED(status);
+#endif
+
   remove_flags(service, SVC_RUNNING);
   if (check_flags(service, SVC_ONESHOT))
     {
@@ -404,6 +440,7 @@ int init_service_start(FAR struct service_s *service)
   if (ret != 0)
     {
       init_err("Starting service '%s': %d", service->argv[1], ret);
+      init_service_reap(service, ret);
       return -ret;
     }
 
@@ -536,6 +573,9 @@ int init_service_parse(FAR const struct parser_s *parser,
         }
 
       s->restart_period = CONFIG_SYSTEM_NXINIT_SERVICE_RESTART_PERIOD;
+#ifdef CONFIG_BOARDCTL_RESET
+      s->reset_reason = -1;
+#endif
       list_initialize(&s->classes);
       list_add_tail(&sm->services, &s->node);
     }
@@ -618,6 +658,9 @@ void init_dump_service(FAR struct service_s *s)
     }
 
   init_debug("  restart_period: %d", s->restart_period);
+#ifdef CONFIG_BOARDCTL_RESET
+  init_debug("  reboot_on_failure: %d", s->reset_reason);
+#endif
 
   if (s->flags)
     {
--- a/system/nxinit/service.h
+++ b/system/nxinit/service.h
@@ -100,6 +100,12 @@ struct service_s
   struct timespec time_kill;
   int restart_period;
   pid_t pid;
+
+  /* The "target" of service option "reboot_on_failure" */
+
+#ifdef CONFIG_BOARDCTL_RESET
+  int reset_reason;
+#endif
 };
 
 struct service_manager_s
@@ -112,7 +118,7 @@ struct service_manager_s
  ****************************************************************************/
 
 int  init_service_refresh(FAR struct service_manager_s *sm);
-void init_service_reap(FAR struct service_s *service);
+void init_service_reap(FAR struct service_s *service, int status);
 int  init_service_start(FAR struct service_s *service);
 int  init_service_stop(FAR struct service_s *service);
 int  init_service_start_by_class(FAR struct service_manager_s *sm,


linguini1 (dismissed):
The testing looks really thorough, great job! This is a very large PR though, with 7 commits. Is there any way it can broken up? Not sure if @jerpelea would find that easier when generating release notes, or if that only applies to the NuttX kernel. Otherwise looks good from me.

xiaoxiang781216 (dismissed):

acassis (changes requested):
nice work @JianyuWang0623 !!!
Please add an Documentation about this system init. Although a documentation for Android exists, NuttX will need a Documentation/ with specific case uses and configurations

cederom:
> nice work @JianyuWang0623 !!! Please add an Documentation about this system init. Although a documentation for Android exists, NuttX will need a Documentation/ with specific case uses and configurations

+1 :-)

xiaoxiang781216:
> @xiaoxiang781216 some months ago you commented about planing to add support to SystemV Init. Is this Android Init a replacement or do you still planing to add support to SystemV Init?

we decide to use Andriod init style since it's more simpler and clear. BTW,  the copyright used by Andriod is same as NuttX.

cederom:
> > @xiaoxiang781216 some months ago you commented about planing to add support to SystemV Init. Is this Android Init a replacement or do you still planing to add support to SystemV Init?
> 
> we decide to use Andriod init style since it's more simpler and clear. BTW, the copyright used by Andriod is same as NuttX.

Thanks @xiaoxiang781216 then I would advise to use "Android System Init" naming conventions in the all identifiers so things are self-explanatory and people know at first glance which system init standard is used? What do you think guys ? :-)

cederom (changes requested):
Thank you @JianyuWang0623 amazing work :-) 

I would advise using "Android System Init" naming convention as there are several different init types out there and the name / variables would be self-explanatory and maybe one day add other init types.

But if others are fine with this assumption that "system init" is "android system init" then I will follow :-)

system/nxinit/action.c
```diff
@@ -0,0 +1,346 @@
...
+        {
+          continue;
+        }
+
+      am->current = list_prepare_entry(am->current, &am->actions,
+                                       struct action_s, node);
+      list_for_every_entry_continue(am->current, &am->actions,
+                                    struct action_s, node)
+        {
+          if (am->current->event && !strcmp(am->current->event,
```
acassis, line 125: Please consider using strncmp() instead to avoid stack overflow

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+  int ms;
+
+  clock_gettime(CLOCK_MONOTONIC, &time);
+  clock_timespec_subtract(&time, &am->time_run, &time);
+  ms = TIMESPEC2MS(time);
+  if (ms > CONFIG_SYSTEM_INIT_ACTION_WARN_SLOW)
+    {
+      if (am->pid_running <= 0)
+        {
+          init_warn("command '%s' took %d ms", am->running->argv[0], ms);
```
acassis: Please always start error, warn, info message with Upper case, because if the text get mixed with other logs it will be easy to spot the beginning of the new message

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+        }
+      else
+        {
+          init_warn("command '%s' pid %d took %d ms", am->running->argv[0],
```
acassis: Ditto: Command

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+  FAR struct action_cmd_s *cmd;
+  FAR struct action_s *a;
+  int ret;
+
+  if (create)
+    {
+      ret = init_parse_arguments(buf, false, nitems(argv), argv);
+      if (ret < 2)
+        {
+          init_err("invalid argument: %s", buf);
```
acassis: Ditto

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+          return -EINVAL;
+        }
+
+      a = calloc(1, sizeof(*a));
+      if (a == NULL)
+        {
+          init_err("alloc");
```
acassis: Alloc action failed

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+          return -errno;
+        }
+
+      list_initialize(&a->cmds);
+
+      a->event = strdup(argv[--ret]);
+      if (a->event == NULL)
+        {
+          init_err("event trigger");
```
acassis: Event trigger

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+
+      list_add_tail(&am->actions, &a->node);
+      init_debug("add action %p(%s) to manager %p", a, argv[1], am);
+    }
+  else
+    {
+      cmd = calloc(1, sizeof(*cmd));
+      if (cmd == NULL)
+        {
+          init_err("alloc");
```
acassis: Alloc cmd failed!

There are many "alloc"  errors, it will be difficult to users discover where it is failing if you don't use unique error messages

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+          return -errno;
+        }
+
+      cmd->argc = init_parse_arguments(buf, true, nitems(cmd->argv) - 1,
+                                       cmd->argv);
+      if (cmd->argc < 1)
+        {
+          free(cmd);
+          init_err("invalid argument");
```
acassis: Invalid cmd argument

cederom (changes requested):
Just a feedback after dev@ mailing list discussion, thank you for understanding / considering @JianyuWang0623 :
1. We clearly need to use "Android System Init" naming here and update all kconfig/global variables names to clearly distinguish from current NuttX Init, Android System Init, and  SystemV Init (there is sill a change it will show up here one day), maybe others. "system_system" or "system_init" or "init_main" is really confusing and not self-explanatory, does not even point to "android_system_init". I know this is a bit longer but will provide important context and clear separation between completely different functionalities / implementations.
2. We need better documentation for this "Android System Init" as well as "Fastboot". These functionalities are welcome, people want to use it, but users / developers ask questions like: Why do we need another init system? What should happen with the previous ones? What does this new init system solve? How this relates to Fastboot? When, why, and how can I use it? What are best use cases with some examples? Please consider approach like explaining and convincing someone who wants to use it but has zero knowledge about the solution.

JianyuWang0623:
> Just a feedback after dev@ mailing list discussion, thank you for understanding / considering @JianyuWang0623 :
> 
> 1. We clearly need to use "Android System Init" naming here and update all kconfig/global variables names to clearly distinguish from current NuttX Init, Android System Init, and  SystemV Init (there is sill a change it will show up here one day), maybe others. "system_system" or "system_init" or "init_main" is really confusing and not self-explanatory, does not even point to "android_system_init". I know this is a bit longer but will provide important context and clear separation between completely different functionalities / implementations.
> 2. We need better documentation for this "Android System Init" as well as "Fastboot". These functionalities are welcome, people want to use it, but users / developers ask questions like: Why do we need another init system? What should happen with the previous ones? What does this new init system solve? How this relates to Fastboot? When, why, and how can I use it? What are best use cases with some examples? Please consider approach like explaining and convincing someone who wants to use it but has zero knowledge about the solution.

@cederom The following is part of the reply.
> Why do we need another init system? What should happen with the previous ones? What does this new init system solve? 

As I supplemented in the data for the PR https://github.com/apache/nuttx/pull/17216 , NSH, when used as an initialization program, may seem relatively bloated in scenarios where code size is a major concern. NSH and Init can coexist, allowing users to choose based on their own needs. Init addresses issues such as the excessive code size of NSH when used as an initialization program, and its lack of management (e.g., restarting) for daemons.
| Config Changes                     | ELF Size(bytes)  |
|:-----------------------------------|----------:|
| defconfig                          | 7,008,056 |
| `+ETC_ROMFS`                       | 7,066,048 |
| `+INIT`                            | 7,127,536 |
| `-TELENT`                          | 7,127,024 |
| `-NSH_LIBRARY` <br>`-SYSTEM_NSH`   | 6,509,392 |

For the ELF file, Init requires 61,488 bytes, while NSH and NSH_LIBRARY together require 617,632 bytes. Switching from NSH to Init can save a significant amount of space(556,144 bytes).

> How this relates to Fastboot? 

Init and Fastboot have nothing to do with each other; it's just that a daemon is needed as an example of a service.

> When, why, and how can I use it? 

Here are some simple examples: When code size is a major concern and no shell is needed, replacing NSH with Init can save a significant amount of space; management functions such as restarting are required when a daemon fails to start or exits abnormally; and so on. 

> What are best use cases with some examples?

An example has already been added to https://github.com/apache/nuttx/pull/17215.

system/init/action.h
```diff
@@ -0,0 +1,89 @@
...
+  struct list_node actions;                  /* struct action_s */
+  struct list_node ready_actions;            /* struct action_s */
+
+  FAR char *events[CONFIG_SYSTEM_INIT_ACTION_MANAGER_EVENT_MAX];
+  FAR struct action_s *current;
+
+  FAR struct action_cmd_s *running;
+  int pid_running;
+#if defined(CONFIG_SYSTEM_INIT_ACTION_WARN_SLOW) && \
+    CONFIG_SYSTEM_INIT_ACTION_WARN_SLOW > 0
```
tinnedkarma: We are always inline-ing the ``CONFIG_SYSTEM_INIT_ACTION_WARN_SLOW > 0`` in every check in the c files.
Would it not help to assert that here, in the header file? something like

```
#if defined(CONFIG_SYSTEM_INIT_ACTION_WARN_SLOW) && \
    CONFIG_SYSTEM_INIT_ACTION_WARN_SLOW > 0
#  error Threshold should be positive
#endif
```

Any other piece of code we can only use ``#if defined(CONFIG_SYSTEM_INIT_ACTION_WARN_SLOW)``

system/nxinit/action.c
```diff
@@ -0,0 +1,346 @@
...
+      list_add_tail(&a->cmds, &cmd->node);
+      init_debug("add command %p(%s) to action %p", cmd, cmd->argv[0], a);
+      init_dump_args(cmd->argc, cmd->argv);
+    }
+
+  return 0;
+}
+
+#ifdef CONFIG_SYSTEM_INIT_DEBUG
+void init_dump_actions(FAR struct list_node *head)
```
tinnedkarma, line 336: Can't this be moved directly to init.c?
It's only used in init.c, and it does not depend on anything from action.c specifically

* head is passed from init.c
* action is defined inside function scope
* struct action_s is the literal type
* node is the literal member

Nothing from the action.c is referenced here. Am I wrong?

system/nxinit/builtin.c
```diff
@@ -0,0 +1,237 @@
...
+  {"start", 2, 2, cmd_start},
+  {"stop", 2, 2, cmd_stop},
+  {"trigger", 2, 2, cmd_trigger},
+};
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+static int cmd_class_start(FAR struct action_manager_s *am,
```
tinnedkarma, line 88: This function just calls another function.
I suggest making it ``static inline``

system/init/builtin.c
```diff
@@ -0,0 +1,237 @@
...
+          init_info("executing command '%s'", argv[0]);
+          return g_builtin[i].func(am, argc, argv);
+        }
+    }
+
+  ret = posix_spawnp(&pid, argv[0], NULL, NULL, argv, NULL);
+  if (ret != 0)
+    {
+#ifdef CONFIG_SYSTEM_SYSTEM
+      char cmd[CONFIG_SYSTEM_INIT_RC_LINE_MAX];
```
tinnedkarma: This is also a pointer, should we add FAR?

system/nxinit/init.h
```diff
@@ -0,0 +1,93 @@
...
+#define init_warn(...)
+#endif
+
+#ifdef CONFIG_SYSTEM_INIT_ERR
+#define init_err(f, ...) syslog(LOG_ERR, "Error " f, ##__VA_ARGS__)
+#else
+#define init_err(...)
+#endif
+
+#define init_log(p, ...) \
```
tinnedkarma, line 71: Any reason why this should not be an ``switch case``?

system/init/action.h
JianyuWang0623: @tinnedkarma A value of zero is not an exception; rather, like being undefined, it indicates that this function is disabled.

SYSTEM_INIT_ACTION_WARN_SLOW
|Value|The function|
|:--:|:--:|
|Undefined| Disable|
|Zero| Disable|
|Greater than zero| Enable|

system/nxinit/action.c
JianyuWang0623: @tinnedkarma You're right, **but** from a functional standpoint, this is part of the action.

system/nxinit/builtin.c
JianyuWang0623: @tinnedkarma  I'm not sure if you've read the entire file.
```C
static const struct cmd_map_s g_builtin[] =
{
  {"class_start", 2, 2, cmd_class_start},
  {"class_stop", 2, 2, cmd_class_stop},
  {"exec", 3, 99, cmd_exec},
  {"exec_start", 2, 2, cmd_exec_start},
  {"start", 2, 2, cmd_start},
  {"stop", 2, 2, cmd_stop},
  {"trigger", 2, 2, cmd_trigger},
};
```

system/init/action.h
tinnedkarma: I think I've made a mistake by copying the lines. 
There should be ``CONFIG_SYSTEM_INIT_ACTION_WARN_SLOW < 0``.
In my example above, I forgot to change the sign from ">" to "<".
Anyway, maybe I did not understand correctly, we need to make sure the option have a positive integer for the value, right?

system/nxinit/init.h
JianyuWang0623: @tinnedkarma  Anything wrong? Are you sure that converting it to a switch-case is feasible or better?

system/nxinit/builtin.c
tinnedkarma: Oh, sorry, I may have mislead you with my previous suggestion.
I suggested to also add inline to the function signature, not drop the return value.

A function call creates a new stack, so here we have a stack for the cmd_class_start, and immediately after we create a new stack for init_service_start_by_class. By inline-ing the cmd_class_start we can avoid creating a useless stack.

Correct me if I am wrong.

system/nxinit/action.c
JianyuWang0623: @acassis Thank you for the reminder. We have also considered this issue. However, the good news is that the values of both `am->current->event` and `am->events[i]` are obtained from the `strdup()` function, which normally includes a terminating null byte.

system/nxinit/builtin.c
JianyuWang0623: @tinnedkarma  The function is only used for function pointers. What I'm wondering is whether 'inline' has any practical significance?

cederom:
> > Just a feedback after dev@ mailing list discussion, thank you for understanding / considering @JianyuWang0623 :
> > 
> > 1. We clearly need to use "Android System Init" naming here and update all kconfig/global variables names to clearly distinguish from current NuttX Init, Android System Init, and  SystemV Init (there is sill a change it will show up here one day), maybe others. "system_system" or "system_init" or "init_main" is really confusing and not self-explanatory, does not even point to "android_system_init". I know this is a bit longer but will provide important context and clear separation between completely different functionalities / implementations.
> > 2. We need better documentation for this "Android System Init" as well as "Fastboot". These functionalities are welcome, people want to use it, but users / developers ask questions like: Why do we need another init system? What should happen with the previous ones? What does this new init system solve? How this relates to Fastboot? When, why, and how can I use it? What are best use cases with some examples? Please consider approach like explaining and convincing someone who wants to use it but has zero knowledge about the solution.
> 
> @cederom The following is part of the reply.
> 
> > Why do we need another init system? What should happen with the previous ones? What does this new init system solve?
> 
> As I supplemented in the data for the PR [apache/nuttx#17216](https://github.com/apache/nuttx/pull/17216) , NSH, when used as an initialization program, may seem relatively bloated in scenarios where code size is a major concern. NSH and Init can coexist, allowing users to choose based on their own needs. Init addresses issues such as the excessive code size of NSH when used as an initialization program, and its lack of management (e.g., restarting) for daemons.
> Config Changes 	ELF Size(bytes)
> defconfig 	7,008,056
> `+ETC_ROMFS` 	7,066,048
> `+INIT` 	7,127,536
> `-TELENT` 	7,127,024
> `-NSH_LIBRARY`
> `-SYSTEM_NSH` 	6,509,392
> 
> For the ELF file, Init requires 61,488 bytes, while NSH and NSH_LIBRARY together require 617,632 bytes. Switching from NSH to Init can save a significant amount of space(556,144 bytes).
> 
> > How this relates to Fastboot?
> 
> Init and Fastboot have nothing to do with each other; it's just that a daemon is needed as an example of a service.
> 
> > When, why, and how can I use it?
> 
> Here are some simple examples: When code size is a major concern and no shell is needed, replacing NSH with Init can save a significant amount of space; management functions such as restarting are required when a daemon fails to start or exits abnormally; and so on.
> 
> > What are best use cases with some examples?
> 
> An example has already been added to [apache/nuttx#17215](https://github.com/apache/nuttx/pull/17215).


All this information information would be nice have in the documentation :-)


> Some supplementary explanations:
> 
>     * The naming "Init" is based on its functional perspective;
> 
>     * The Kconfig naming "SYSTEM_INIT" follows the conventions of existing components, with reference to the code path "apps/system/init" where "SYSTEM" corresponds to the "system" directory;
> 
>     * Compatibility with Android Init Language is considered for ease of use;
> 
>     * However, we are not directly porting Android Init, but rather implementing a new, lightweight Init that is compatible with most of its syntax.

I see, then there is nothing blocking the road to name it "Android System Init" right? We could place it in `apps/android/system/init` (maybe just `apps/android/init`) also the fastboot could be moved to `apps/android/fastboot`? Then we have clear separation of functionalities and implementations by name and location. `apps/system` looks more like "NuttX system applications" :-) Anyways we should put this "android" somewhere in the names and paths to clearly distinguish its and implementation coming from Android and not the NuttX internal solution :-)

Thanks again @JianyuWang0623 :-)

system/nxinit/builtin.c
tinnedkarma: @JianyuWang0623 To my knowledge yes, It's also my point, there is no "internal" logic to this function, so we could avoid incrementing the stack depth with another push-ret if it does not offer any practical significance.
Anyway, maybe another opinion will help.

JianyuWang0623:
> I see, then there is nothing blocking the road to name it "Android System Init" right? We could place it in `apps/android/system/init` (maybe just `apps/android/init`) also the fastboot could be moved to `apps/android/fastboot`? Then we have clear separation of functionalities and implementations by name and location. `apps/system` looks more like "NuttX system applications" :-) Anyways we should put this "android" somewhere in the names and paths to clearly distinguish its and implementation coming from Android and not the NuttX internal solution :-)
> 
> Thanks again @JianyuWang0623 :-)

@cederom I still have some questions. 

1. The reason for naming it "Init" is that it references Android Init, SystemV Init, and systemd. Considering the module's functionality, it's similar to how Linux's system initialization process is named init or systemd.
```
$ ls -l /usr/sbin/init
lrwxrwxrwx 1 root root 20 Jun  4 22:17 /usr/sbin/init -> /lib/systemd/systemd
```

2. The Kconfig is named SYSTEM_INIT to follow the existing modules in apps/system, defined according to the path: apps/system/init => CONFIG_SYSTEM_INIT.

3. Regarding the consideration of adding an ANDROID prefix because other init modules might be added in the future, my question is: for example, do other component in apps need to consider the same issue? If we need to account for possible alternative implementations, should all components have specific prefixes added, such as indicating their porting sources or which standards they are compatible with?

Thank you :-)

system/nxinit/init.h
tinnedkarma: Here is my reasoning:
* a do while(0) loop is used, but the loop wont ever "loop", so it's a do? :D Then is the same as using naked {}, which will create a scope and execute the instructions once.
* there is a set/fixed amount of log levels defined in the syslog.h, so it's arguably better to uses a switch case (look-up table) than "falling through" if-else.

I do see the same concept used throughout, so maybe there are other reasons that I may not know, hence why I addressed this as a question.

JianyuWang0623: @tinnedkarma 
> a do while(0) loop is used, but the loop wont ever "loop", so it's a do? :D Then is the same as using naked {}, which will create a scope and execute the instructions once.

Common usage; it can avoid syntax issues such as those occurring after expansion:
```C
#define LOG_ERROR(msg) \
    fprintf(stderr, "Error: %s\n", msg); \
    exit(1);

if (error)
    LOG_ERROR(...);

/* After preprocessing */

if (error)
    fprintf(...); exit(1);

```
As for runtime efficiency, the compiler will perform optimizations.
> there is a set/fixed amount of log levels defined in the syslog.h, so it's arguably better to uses a switch case (look-up table) than "falling through" if-else. 

This requires listing all levels; moreover, some code inspection tools will consider that some cases are missing breaks.
```C
switch (p)
  {
    case LOG_EMERG:
    case LOG_ALERT:
    case LOG_CRIT:
    case LOG_ERR:
        init_err(__VA_ARGS__);
        break;
    case LOG_WARNING:
        init_warn(__VA_ARGS__);
        break;
    /* ... ... */
  }
```

include/syslog.h:
```C
#define LOG_EMERG     0  /* System is unusable */
#define LOG_ALERT     1  /* Action must be taken immediately */
#define LOG_CRIT      2  /* Critical conditions */
#define LOG_ERR       3  /* Error conditions */
#define LOG_WARNING   4  /* Warning conditions */
#define LOG_NOTICE    5  /* Normal, but significant, condition */
#define LOG_INFO      6  /* Informational message */
#define LOG_DEBUG     7  /* Debug-level message */
```

tinnedkarma: > Common usage; it can avoid syntax issues such as those occurring after expansion:

Yes but the same issues can be avoided by adding the curly braces.

> As for runtime efficiency, the compiler will perform optimizations.

No, the compiler will not optimize the code in any way, is will generate predicable assembly.
Here is an example of code optimization [https://github.com/apache/nuttx/issues/17206](https://github.com/apache/nuttx/issues/17206).
Personally I would avoid intentionally offloading work to the compiler. It's our job to write predictable code. 

> This requires listing all levels; moreover, some code inspection tools will consider that some cases are missing breaks.

Yet, I still think it's a cleaner. :D

We are not getting anywhere. That is my personal view, it's fine if we do not agree, and anyway more input is always welcome.

linguini1: The do-while pattern is actually quite common, there are some scenarios where regular curly braces don't have the same semantics. That's why do-while (0) is used almost always for multiline macros: https://stackoverflow.com/questions/1067226/c-multi-line-macro-do-while0-vs-scope-block

It does get compiler-optimized out and I think it's clear enough to keep instead of curly braces.

tinnedkarma: It's settled then

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+#ifdef CONFIG_SYSTEM_INIT_DEBUG
+static void init_dump_action(FAR struct action_s *action)
+{
+  FAR struct action_cmd_s *cmd;
+
+  init_debug("action %p", action);
```
hartmannathan: Similar to @acassis suggestion elsewhere, starting new messages with capital letters makes it easier to see in the logs.

```suggestion
  init_debug("Action %p", action);
```

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+static void add_ready(FAR struct action_manager_s *am,
+                      FAR struct action_s *a)
+{
+  FAR struct action_s *ready;
+#ifdef CONFIG_SYSTEM_INIT_DEBUG
+  FAR struct action_cmd_s *cmd;
+
+  list_for_every_entry(&a->cmds, cmd, struct action_cmd_s, node)
+    {
+      init_debug("add ready %p %s", cmd, cmd->argv[0]);
```
hartmannathan: Ditto

```suggestion
      init_debug("Add ready %p %s", cmd, cmd->argv[0]);
```

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+    }
+#endif
+
+  list_for_every_entry(&am->ready_actions, ready, struct action_s,
+                       ready_node)
+    {
+      if (ready == a)
+        {
+          init_debug("action %p(%s) already on the queue", a,
```
hartmannathan: Ditto

```suggestion
          init_debug("Action %p(%s) already on the queue", a,
```

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+        {
+          if (am->current->event && !strcmp(am->current->event,
+                                            am->events[i]))
+            {
+              add_ready(am, am->current);
+              return;
+            }
+        }
+
+      init_debug("remove event [%zu] '%s'", i, am->events[i]);
```
hartmannathan: Ditto

```suggestion
      init_debug("Remove event [%zu] '%s'", i, am->events[i]);
```

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+    {
+      if (am->events[i])
+        {
+          continue;
+        }
+
+      am->events[i] = strdup(event);
+      if (am->events[i])
+        {
+          init_debug("add event [%zu] '%s'", i, am->events[i]);
```
hartmannathan: Ditto

```suggestion
          init_debug("Add event [%zu] '%s'", i, am->events[i]);
```

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+          return 0;
+        }
+
+      ret = -errno;
+      break;
+    }
+
+  init_warn("drop event '%s' %d", event, ret);
```
hartmannathan: Ditto.

```suggestion
  init_warn("Drop event '%s' %d", event, ret);
```

system/init/action.c
```diff
@@ -0,0 +1,346 @@
...
+ ****************************************************************************/
+
+int init_action_run_command(FAR struct action_manager_s *am)
+{
+  FAR struct action_s *ready;
+  int ret;
+
+  if (am->pid_running != -1)
+    {
+      init_debug("waiting '%s' pid %d", am->running->argv[0],
```
hartmannathan: Ditto

```suggestion
      init_debug("Waiting '%s' pid %d", am->running->argv[0],
```

system/init/action.c
hartmannathan: Ditto

```suggestion
          init_warn("Command '%s' took %d ms", am->running->argv[0], ms);
```

system/init/action.c
hartmannathan: Maybe say which alloc action failed? Might be other subsystems competing for the memory :-)

system/init/builtin.c
```diff
@@ -0,0 +1,237 @@
...
+
+static int cmd_exec_start(FAR struct action_manager_s *am,
+                          int argc, FAR char **argv)
+{
+  FAR struct service_s *service;
+
+  service = init_service_find_by_name(am->sm, argv[1]);
+  if (!service)
+    {
+      init_err("no such service '%s'", argv[1]);
```
hartmannathan: Ditto

```suggestion
      init_err("No such service '%s'", argv[1]);
```

system/init/builtin.c
```diff
@@ -0,0 +1,237 @@
...
+
+static int cmd_stop(FAR struct action_manager_s *am,
+                    int argc, FAR char **argv)
+{
+  FAR struct service_s *service;
+
+  service = init_service_find_by_name(am->sm, argv[1]);
+  if (!service)
+    {
+      init_err("no such service '%s'", argv[1]);
```
hartmannathan: Ditto

```suggestion
      init_err("No such service '%s'", argv[1]);
```

system/init/builtin.c
```diff
@@ -0,0 +1,237 @@
...
+ *   Execute the builtin command of Init, builtin App, or NSH builtin.
+ *
+ * Input Parameters:
+ *   am   - Instance of Action Manager.
+ *   argc - Argument count.
+ *   argv - A pointer to an array of string arguments. The end of the array
+ *          is indicated with a NULL entry.
+ *
+ * Returned Value:
+ *   Returns 0 (for Init builtin) or a PID(for builtin App or NSH builtin) on
```
hartmannathan:
```suggestion
 *   Returns 0 (for Init builtin) or a PID (for builtin App or NSH builtin) on
```

system/init/builtin.c
```diff
@@ -0,0 +1,237 @@
...
+      for (i = 0, cmd[i] = '\0'; i < argc; i++)
+        {
+          strlcat(cmd, argv[i], sizeof(cmd));
+          if (i < argc - 1)
+            {
+              strcat(cmd, " ");
+            }
+        }
+
+      init_debug("executing NSH command '%s'", cmd);
```
hartmannathan:
```suggestion
      init_debug("Executing NSH command '%s'", cmd);
```

system/init/builtin.c
```diff
@@ -0,0 +1,237 @@
...
+      ret = system(cmd);
+      if (WIFEXITED(ret))
+        {
+          init_debug("NSH command '%s' exited %d", cmd, WEXITSTATUS(ret));
+          return -WEXITSTATUS(ret);
+        }
+#endif
+
+      init_err("executing command '%s': %d", argv[0], ret);
```
hartmannathan:
```suggestion
      init_err("Executing command '%s': %d", argv[0], ret);
```

system/init/builtin.c
```diff
@@ -0,0 +1,237 @@
...
+      init_dump_args(argc, argv);
+      return -ret;
+    }
+
+  init_debug("executed command '%s' pid %d", argv[0], pid);
```
hartmannathan:
```suggestion
  init_debug("Executed command '%s' pid %d", argv[0], pid);
```

system/init/import.c
```diff
@@ -0,0 +1,60 @@
...
+  FAR char *argv[2];
+  int ret;
+
+  if (create)
+    {
+      init_parse_arguments(buf, false, nitems(argv), argv);
+      ret = init_parse_config_file(parser, argv[1]);
+      if (ret < 0)
+        {
+          init_err("import %s", argv[1]);
```
hartmannathan: Should the error say that this is a **parsing** error?

```suggestion
          init_err("Import %s", argv[1]);
```

system/init/import.c
```diff
@@ -0,0 +1,60 @@
...
+          return ret;
+        }
+    }
+  else
+    {
+      init_err("parse import");
```
hartmannathan: Needs more descriptive error message here. What went wrong?

```suggestion
      init_err("Parse import");
```

system/init/action.c
JianyuWang0623: done, thank you

system/init/action.c
JianyuWang0623: done

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Thank you, done.

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Done, thank you.

system/init/action.c
JianyuWang0623: Done, thank you.

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Done

system/init/action.c
JianyuWang0623: Done

system/init/builtin.c
JianyuWang0623: Done

system/init/builtin.c
JianyuWang0623: Done

system/init/builtin.c
JianyuWang0623: Thank you, done.

system/init/builtin.c
JianyuWang0623: Done

system/init/builtin.c
JianyuWang0623: Done

system/init/builtin.c
JianyuWang0623: Done

system/init/import.c
JianyuWang0623: Done

system/init/import.c
JianyuWang0623: Done

system/nxinit/init.h
```diff
@@ -0,0 +1,93 @@
...
+#include <syslog.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define TIMESPEC2MS(t) (((t).tv_sec * 1000) + (t).tv_nsec / 1000000)
+
+#ifdef CONFIG_SYSTEM_INIT_DEBUG
+#define init_debug(...) syslog(LOG_DEBUG, ##__VA_ARGS__)
```
linguini1, line 39: I would add the user space mask, `LOG_DEBUG | LOG_USER`. Not sure if there's a better subsystem mask to use but I don't think kernel is correct here.

system/nxinit/service.h
```diff
@@ -0,0 +1,138 @@
...
+
+#define SVC_REMOVE      (1 << 31)
+
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+struct service_class_s
+{
+  struct list_node node;
```
7rx8: for consistency you should document the expected structs everywhere.

system/nxinit/service.c
```diff
@@ -0,0 +1,622 @@
...
+
+  add_flags(service, SVC_DISABLED);
+  if (check_flags(service, SVC_RESTARTING))
+    {
+      remove_flags(service, SVC_RESTARTING);
+    }
+
+  if (check_flags(service, SVC_GENTLE_KILL))
+    {
+      init_debug("Gentel kill");
```
7rx8: little ooopsie typo :-)

system/nxinit/service.c
```diff
@@ -0,0 +1,622 @@
...
+{
+  FAR char *argv[CONFIG_SYSTEM_INIT_SERVICE_ARGS_MAX];
+  FAR struct service_manager_s *sm = parser->priv;
+  FAR struct service_s *s;
+  int argc;
+  size_t i;
+
+  if (create)
+    {
+      s = calloc(1, sizeof(*s));
```
7rx8, line 560: whats the best practice?
* sizeof of target variable?
* sizeof of intended type?

system/nxinit/action.c
7rx8: strncmp with a max length can still protect from bogus user input. I recommend doing something here. the boot process is critical and a user should not be able to mess things up with bogus data injected in the filesystem.

system/nxinit/builtin.c
xiaoxiang781216: this function is called through pointer, inline can't help here anyway.

system/init/builtin.c
xiaoxiang781216: char array doesn't add FAR since it allocate in stack directly

JianyuWang0623:
> Just a feedback after dev@ mailing list discussion, thank you for understanding / considering @JianyuWang0623 :
> 
> 1. We clearly need to use "Android System Init" naming here and update all kconfig/global variables names to clearly distinguish from current NuttX Init, Android System Init, and  SystemV Init (there is sill a change it will show up here one day), maybe others. "system_system" or "system_init" or "init_main" is really confusing and not self-explanatory, does not even point to "android_system_init". I know this is a bit longer but will provide important context and clear separation between completely different functionalities / implementations.
> 2. We need better documentation for this "Android System Init" as well as "Fastboot". These functionalities are welcome, people want to use it, but users / developers ask questions like: Why do we need another init system? What should happen with the previous ones? What does this new init system solve? How this relates to Fastboot? When, why, and how can I use it? What are best use cases with some examples? Please consider approach like explaining and convincing someone who wants to use it but has zero knowledge about the solution.

Hi @xiaoxiang781216 , @cederom, @acassis, and everyone following the naming discussion for this component,  

Thank you all for your suggestions. As I proposed in the mailing list (archive: https://www.mail-archive.com/dev@nuttx.apache.org/msg13730.html), what do you think of the name "NuttX Init"? If there are no objections, I will update the component name as soon as possible. Additionally, if you have time, I would greatly appreciate it if you could review the code and share your feedback.

The following is a copy of the original email content:
```
Dear All,

Thank you for your suggestions and ideas. Many of you have pointed out
that the name "init" may be too broad or lead to confusion, and I
believe this concern deserves our attention.

With reference to Jehudi’s suggestion, I propose naming this new
component "NuttX Init" (abbreviated as "NxInit", with the path:
apps/system/nxinit). This naming follows the style of existing
components, such as nxplayer and nxcamera.

Regarding names like "Android Init" or similar variations, it is
important to clarify that this component is specifically developed
for NuttX and intended for system initialization. While we have
adopted the Android Init Language among various syntax options,
this is a brand-new implementation—it is not a port or variant of
Android Init.

I agree that enhancing the recognizability of this component’s name
is necessary, but I suggest incorporating "NuttX" rather than
"Android" into the name to accurately reflect its affiliation and
nature.

Thank you again for your valuable input.
Jianyu Wang
```

cederom:
Copy-Paste from dev@ :-)

Okay so looks like there are two name propositions for this new application:
1. Android System Init -> apps/android/init -> android_init (or just ainit).
2. NXInit -> apps/system/nxinit -> nxinit.

If the discussion with ideas is concluded lets just VOTE on the mailing list so the community decides what name is best and merge the code named as decided by the community :-) For me both approaches are okay :-)

In order to stay compliant with https://www.apache.org/foundation/voting vote should be in dedicated thread marked [VOTE] prefix it should take 72h and we can receive votes for both options, +1 0 -1 replies are possible, the one that gets most count wins, -1 from PMC vetoes a choice unless vote update.

@JianyuWang0623  could you please start the voting thread (on dev@) and then make summary after 72h? 

Thanks :-)

system/nxinit/service.h
JianyuWang0623: done, thank you

system/nxinit/service.c
JianyuWang0623: thank you

system/nxinit/init.c
```diff
@@ -0,0 +1,254 @@
...
+    {
+      goto out;
+    }
+
+  init_dump_actions(&am.actions);
+  init_dump_services(&sm.services);
+
+  init_action_add_event(&am, "boot");
+
+  boardctl(BOARDIOC_INIT, 0);
```
7rx8, line 194: oh man boardctl does make so much more sense with a proper init system. I always found it clunky to do that kind of calls from nsh.

now it's great.

acassis: True! It was like a chicken-egg thing :-)

system/nxinit/Kconfig
```diff
@@ -0,0 +1,124 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config SYSTEM_NXINIT
+	tristate "System Init"
```
cederom: NXInit ?

system/nxinit/Kconfig
```diff
@@ -0,0 +1,124 @@
...
+	default n
+	depends on LIBC_EXECFUNCS
+	depends on SCHED_CHILD_STATUS
+	---help---
+		Enable system init.
```
cederom: NXInit application is a dedicated implementation of the Init application for NuttX with language syntax based on Android System Init. ?

system/nxinit/service.c
```diff
@@ -0,0 +1,682 @@
...
+  list_for_every_entry(&s->classes, c, struct service_class_s, node)
+    {
+      if (!strcmp(c->name, argv[1]))
+        {
+          return 0;
+        }
+    }
+
+  c = malloc(sizeof(*c) + len);
+  if (!c)
```
acassis:
```suggestion
  if (c == NULL)

File: system/nxinit/action.c [hunk 37]
```diff
@@ -0,0 +1,346 @@
...
+ */
+
+#define ACTION_ARGUMENTS_MAX 2
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+#ifdef CONFIG_SYSTEM_NXINIT_DEBUG
+static void init_dump_action(FAR struct action_s *action);
```
cederom, line 56: `init_*` -> `nxinit_*` refactor please it will be easier to debug :-)

system/nxinit/init.c
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+      if (pid == am->pid_running)
+        {
+          name = am->running->argv[0];
+          init_action_reap_command(am);
+        }
+
+      service = init_service_find_by_pid(sm, pid);
+      if (service)
```
acassis:
```suggestion
      if (service != NULL)

system/nxinit/Kconfig
JianyuWang0623: ok, thank you

system/nxinit/action.c
JianyuWang0623: @cederom Thank you, but I don't think this modification is necessary.

system/nxinit/service.c
JianyuWang0623: Thank you, but I don't think this modification is necessary.

system/nxinit/init.c
JianyuWang0623: Thank you, but I don't think this modification is necessary.

system/nxinit/service.c
acassis: It is more readable and also was compilers are become more restrictive, soon or later they can start raising warning about this kind of test

system/nxinit/init.c
acassis: same here

system/nxinit/service.c
JianyuWang0623: @acassis  ok, done

system/nxinit/init.c
JianyuWang0623: ok, done

7rx8:
Hello

I do not agree with this closure. This is a great addition to nuttx that was worked on for a long time and is almost ready. Why are you closing this?

@JianyuWang0623 has worked a lot of time to reply to community requests there is no point in closing this.

cederom:
@xiaoxiang781216 we can reopen this and related PRs, we have an active discussion on the mailing list, this functionality is useful and welcome by the community :-)

GUYS PLEASE USE DEV@ MAILING LIST TO BE PART OF THE COMMUNITY AND ACTIVELY DISCUSS CHANGES AND PLANS WITH THE NUTTX COMMUNITY THIS IS OUR MAIN COMMUNICATION CHANNEL! :-)

7rx8:
I definitely also have interest about this and for sure I would provide support and improvements because we're certainly going to use this, but it would be easier if it was upstreamed.

PR #3242, opened 2025-12-09
sh cle_fd returns EOF if there is no data be read.

yangsong8-a1:
## Summary

fix switchboot command spelling error.
sh cle_fd returns EOF if there is no data be read.

## Impact

when there is no data to be read. sh cle_fd returns EOF but not -EIO

## Testing

When CONFIG_NSH_CLE is enabled, after the adb shell service is destroyed, sh detects nread as 0 when reading data, but instead of returning EOF, it returns -EIO, thus preventing sh from exiting.
target： sim and xiaomi product


xiaoxiang781216 (approved):

linguini1 (approved):
Should be -EOF or is it expected to return a positive value everywhere?

xiaoxiang781216 (approved):

PR #3243, opened 2025-12-09
apps/system/composite: change cmake build name to conn and disconn

yangsong8-a1:
## Summary

* The executable file compiled by cmake was named `composite`, while the file compiled by make are named `conn` and `disconn`.
* The executable file compiled by cmake now has the same filename as that compiled by make.

## Impact

Uniform compiled application names between cmake and make.

## Testing

target: sim vela

```
Enabling CDCACM as an example when cmake

before fix:
$ composite 1

after fix:
$ conn 1
```

### DIFF
--- a/system/composite/CMakeLists.txt
+++ b/system/composite/CMakeLists.txt
@@ -21,5 +21,6 @@
 # ##############################################################################
 
 if(CONFIG_SYSTEM_COMPOSITE)
-  nuttx_add_application(NAME composite SRCS composite_main.c)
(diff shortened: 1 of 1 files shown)


yangsong8-a1:
> Is this referred to anywhere else? Like maybe in the docs somewhere that we should update?

There is nothing that needs updating. this doc usbdev board profile uses the term "conn".
https://nuttx.apache.org/docs/latest/platforms/sim/sim/boards/sim/index.html

Issue #3258, opened 2025-12-21
[FEATURE] Improve OS test extendability and readability


linguini1:
### Is your feature request related to a problem? Please describe.

OS Test is the primary testing mechanism for NuttX. It is used on virtually all PRs to check for regressions in kernel code changes, and is also used as a basis for verifying that a new board has successfully gotten NuttX ported and running.

However, the output from OS test is a little cryptic when looking at it for the first time. It is often difficult to determine what tests have been run, what the output means and which tests have even passed/failed. There are many components to OS test, provided by different contributors, and they all have different outputs in the console which lead to more confusion.

It is also difficult to determine what the tests are right away by looking at the code. For instance, the `aio.c` test file contains a large code block with tests separated by comments that say "Case 1", "Case 2", etc. It requires effort to read and understand, but could be easier.

### Describe the solution you'd like

My proposal is that OS Test be re-factored to use the [Unity testing framework](https://github.com/ThrowTheSwitch/Unity) already supported by NuttX. This is a very lightweight testing framework that provides a structured approach to adding, writing and running test cases, as well as a nice summary output in the console that makes it clear which tests have passed or failed.

Since Unity is lightweight, works well and is supported by NuttX, I think moving to this approach would provide a clearly defined way for devs to add test cases and review the results of running the tests easily. The separated structure of test cases also means it will be easy to look at the test modules like `aio.c` and determine what tests already exist, what they do and maybe help to identify gaps that could use more tests.

### Describe alternatives you've considered

I have considered just re-factoring the tests to have a more unified and structured output in the console, which would improve the ability of judging pass/fail results from logs, but that would require maintaining our own standard and I think it would be more work in the long run. Same for code-structure wise.

### Verification

- [x] I have verified before submitting the report.

PR #3260, opened 2025-12-22
testing/nest: NuttX unit testing collection

linguini1:
## Summary

This PR introduces 'nest', the NuttX unit test collection. It is a collection of
unit test suites built on top of the already integrated Unity test framework.
Running the application runs all of the enabled test cases with pass/fail/skip
results summarized in the shell.

Partially solves #3258, but OS test could probably use some tweaking to get the console output more readable.

Documentation for this application is coming soon, but I wanted to get feedback on the initial implementation first.

## Impact

Ideally, this will slowly phase out OS test and replace it with more structured
unit tests that have:
- Easily extendable code (simple to add new test cases with Unity)
- Configurable test cases (only need to run a small subset of tests depending on
your change)
- Easily readable test summaries in output
- Well-documented test-cases (docstrings above test case functions, descriptions
of suites on NuttX docs)

Since OS test is a favourite for checking regressions, hopefully nest can be
used to easily target subsystems to check for regressions and create a very
concise summary of the test results for reviewers.

## Testing

I tested compilation on a Linux machine. All of the initial tests I've written
as a starting point for nest were tested using `sim:nsh`. Only one test fails
(for `/dev/zero`) and that is because I believe I actually discovered a bug in
the `/dev/zero` implementation where its behaviour does not match Linux. I will
patch/discuss this later.

Here is the summary of the test results on sim that is produced by running `nest`:

```console
nsh> nest
nest_main.c:151:collections_cbuf__init_local:PASS
nest_main.c:152:collections_cbuf__init_malloc:PASS
nest_main.c:153:collections_cbuf__empty_local:PASS
nest_main.c:154:collections_cbuf__empty_malloc:PASS
nest_main.c:119:collections_hmap__create:PASS
nest_main.c:120:collections_hmap__create_r:PASS
nest_main.c:121:collections_hmap__create_zero:PASS
nest_main.c:122:collections_hmap__create_r_zero:PASS
nest_main.c:133:collections_list__empty_after_init:PASS
nest_main.c:134:collections_list__zerolen_after_init:PASS
nest_main.c:135:collections_list__notail_after_init:PASS
nest_main.c:136:collections_list__nohead_after_init:PASS
nest_main.c:137:collections_list__static_null_init:PASS
nest_main.c:238:devices_devascii__open_rdonly:PASS
nest_main.c:239:devices_devascii__open_rdwr:PASS
nest_main.c:240:devices_devascii__open_wronly:PASS
nest_main.c:242:devices_devascii__readzero:PASS
nest_main.c:243:devices_devascii__readline:PASS
nest_main.c:244:devices_devascii__readline_twice:PASS
nest_main.c:246:devices_devascii__writezero:PASS
nest_main.c:247:devices_devascii__writelarge:PASS
nest_main.c:155:devices_devconsole__open_rdonly:PASS
nest_main.c:156:devices_devconsole__open_rdwr:PASS
nest_main.c:157:devices_devconsole__open_wronly:PASS
nest_main.c:159:devices_devconsole__readzero:PASS
nest_main.c:160:devices_devconsole__writezero:PASS
nest_main.c:241:devices_devnull__open_rdonly:PASS
nest_main.c:242:devices_devnull__open_rdwr:PASS
nest_main.c:243:devices_devnull__open_wronly:PASS
nest_main.c:245:devices_devnull__readzero:PASS
nest_main.c:246:devices_devnull__readlarge:PASS
nest_main.c:248:devices_devnull__writezero:PASS
nest_main.c:249:devices_devnull__writelarge:PASS
nest_main.c:251:devices_devnull__wrrd:PASS
nest_main.c:252:devices_devurandom__open_rdonly:PASS
nest_main.c:253:devices_devurandom__open_rdwr:PASS
nest_main.c:254:devices_devurandom__open_wronly:PASS
nest_main.c:256:devices_devurandom__readzero:PASS
nest_main.c:257:devices_devurandom__writezero:PASS
nest_main.c:259:devices_devurandom__readlarge:PASS
nest_main.c:260:devices_devurandom__uniform:PASS
nest_main.c:256:devices_devzero__open_rdonly:PASS
nest_main.c:257:devices_devzero__open_rdwr:PASS
nest_main.c:258:devices_devzero__open_wronly:PASS
nest_main.c:260:devices_devzero__readzero:PASS
nest_main.c:138:devices_devzero__readlarge:FAIL: Expected 64 Was 0
nest_main.c:263:devices_devzero__writezero:PASS
nest_main.c:264:devices_devzero__writelarge:PASS
nest_main.c:266:devices_devzero__wrrd:PASS

-----------------------
49 Tests 1 Failures 0 Ignored
FAIL
```

I have also verified that enabling/disabling tests in Kconfig results in them
not being run. Disabling dependencies (i.e. setting `DEV_URANDOM=n` when
`TESTING_NEST_DEVICES_DEVURANDOM=y`) also causes the test cases dependent on
them to not be run (Kconfig automatically sets
`TESTING_NEST_DEVICES_DEVURANDOM=n`).

I am open to feedback if it would be preferred to replace tests missing
dependencies with a `TEST_IGNORE_MESSAGE("missing dependencies")` instead of
just excluding them from the build. I have opted for the latter to reduce code
sizes and catch missing dependencies at compilation instead of runtime. It also
uses Kconfig's existing features instead of the test-writer having to add more
`#ifdefs`.

## Other Limitations

This is meant to be a unit test collection. Some things in OS test are not
necessarily "unit tests", like test cases involving multi-threading and forcing
context switches, etc. I am still playing around with what can be done with
Unity; it may be possible to twist it into working with some multi-threaded test
cases. That being said, this is not intended as a direct replacement of OS test.
There is a reason that the test cases included in this PR are very easily
unit-testable subsystems.

The current organization of tests is not final! I think all three collections
(for instance) exist in separate branches of the NuttX source tree and may not
deserve to exist all together in one category. This is just a proof of concept
and I will probably re-arrange those tests later.

### DIFF
--- a/testing/unity/Kconfig
+++ b/testing/unity/Kconfig
@@ -17,7 +17,7 @@ config TESTING_UNITY_URL
 
 config TESTING_UNITY_VERSION
 	string "Version number"
-	default "2.5.2"
+	default "2.6.1"
 
 config TESTING_UNITY_EXCLUDE_SETJMP
 	bool "Exclude SETJMP"


testing/nest/nest_main.c
```diff
@@ -0,0 +1,105 @@
...
+/****************************************************************************
+ * nest_main
+ ****************************************************************************/
+
+int main(int argc, FAR char **argv)
+{
+  UNUSED(argc);
+  UNUSED(argv);
+
+  UNITY_BEGIN();
```
acassis, line 88: @linguini1 Amazing!!!

Question: is it possible to associate a Unique ID to each test and have a verbose (human understing) message about the test?

Of course it should be optional (in the Kconfig) because could add overhead (more text message in the firmware), but should be useful to have a high level understanding of what was tested, i.e.:

```
0001: "Circular Buffer Initialization"     => PASSED
0002: "Circular Buffer Memory Allocation"  => PASSED
0003: "Circular Buffer Memory Released"    => PASSED
...
0049: "Driver zerodev Read/Write Testing"  => PASSED
```

linguini1: I'm not certain if it the output would be as clear as that, but Unity has `TEST_PRINTF` which can display some text associated with a test. That could be used to print a description. Giving an ID to each test might be a lot harder.

PR #3269, opened 2025-12-23
net/wireless: fix compile warning issue for wireless

HedongGao:

## Summary

IEEE802154_STATUS_STRING is defined in the .h file, which will generate compilation warning information under some compilers. Moved the position of the variable to fix this issue. At the same time, changed the name of the variable to unify the code style.

## Impact

Rename static const char *IEEE802154_STATUS_STRING[] to FAR const char *g_ieee802154_status_string[]. Will not affect the functionality. 
This PR needs to be merged with another [PR ](https://github.com/apache/nuttx/pull/17652)

## Testing

gcc compiler has no warning infor for nuttx-apps/wireless/ieee802154/i8sak, but tasking compiler has warning information.

warnning log: unused variable "IEEE802154_STATUS_STRING"

After the issue is fixed, the warning dispeared, and the results compiled with both compilers are normal.

### DIFF
--- a/wireless/ieee802154/i8sak/i8sak_main.c
+++ b/wireless/ieee802154/i8sak/i8sak_main.c
@@ -126,7 +126,8 @@ static void i8sak_switch_instance(FAR char *ifname);
  * Public Data
  ****************************************************************************/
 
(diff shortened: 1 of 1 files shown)


linguini1:
It might be better to put this change in the same PR as before? That way there are no CI errors?

PR #3270, opened 2025-12-23
!testing/cmocka: Compile cmocka as a libary only by default

linguini1:
## Summary

**BREAKING CHANGE**

This commit introduces a Kconfig switch to include the cmocka binary built from cmocka_main.c. The default behaviour is now changed so that cmocka is built as a library only, which would be the desired behaviour for users creating their own cmocka projects.

To restore the old behaviour (where the cmocka program is compiled), add `CONFIG_TESTING_CMOCKA_PROG=y` to your legacy configuration.

## Impact

Cmocka is now only built as a library, so the `cmocka` binary is no longer
included by default.

PR (https://github.com/apache/nuttx/pull/17653) to the NuttX kernel preserves the compilation of the `cmocka` binary for the two existing defconfigs which use the cmocka application.

## Testing

Compilation of the cmocka libary without the binary using `sim:nsh` as a basis:

```console
$ make -j
Create version.h
LN: platform/board to /home/linguini/coding/nuttx-space/apps/platform/dummy
Register: dd
Register: nsh
Register: sh
Register: gcov
Register: hello
Register: dumpstack
Register: gpio
Register: ostest
CP:  /home/linguini/coding/nuttx-space/nuttx/include/nuttx/config.h
CP:  /home/linguini/coding/nuttx-space/nuttx/include/nuttx/fs/hostfs.h
LD:  nuttx
Pac SIM with dynamic libs..
'/usr/lib/libm.so.6' -> 'sim-pac/libs/libm.so.6'
'/usr/lib/libz.so.1' -> 'sim-pac/libs/libz.so.1'
'/usr/lib/libc.so.6' -> 'sim-pac/libs/libc.so.6'
'/usr/lib64/ld-linux-x86-64.so.2' -> 'sim-pac/libs/ld-linux-x86-64.so.2'
'/lib64/ld-linux-x86-64.so.2' -> 'sim-pac/ld-linux-x86-64.so.2'
SIM elf with dynamic libs archive in nuttx.tgz
```

Then, when disabling `CONFIG_TESTING_CMOCKA_ASLIB`, the compilation is:

```
$ make -j
Create version.h
LN: platform/board to /home/linguini/coding/nuttx-space/apps/platform/dummy
Register: cmocka
Register: hello
Register: nsh
Register: sh
Register: dd
Register: gcov
Register: gpio
Register: dumpstack
Register: ostest
CP:  /home/linguini/coding/nuttx-space/nuttx/include/nuttx/config.h
LD:  nuttx
Pac SIM with dynamic libs..
'/usr/lib/libm.so.6' -> 'sim-pac/libs/libm.so.6'
'/usr/lib/libz.so.1' -> 'sim-pac/libs/libz.so.1'
'/usr/lib/libc.so.6' -> 'sim-pac/libs/libc.so.6'
'/usr/lib64/ld-linux-x86-64.so.2' -> 'sim-pac/libs/ld-linux-x86-64.so.2'
'/lib64/ld-linux-x86-64.so.2' -> 'sim-pac/ld-linux-x86-64.so.2'
SIM elf with dynamic libs archive in nuttx.tgz
```

and I can run the cmocka application on sim:

```
nsh> cmocka --help
an elegant unit testing framework for C with support for mock objects
Usage: cmocka [OPTION [ARG]] ...
 -?, --help       show this help statement
 -l, --list       display only the names of testcases and testsuite,
                  don't execute them
 -t, --test A     only run cases where case function name matches A pattern
 -p, --skip B     don't run cases where case function name matches B pattern
 -s, --suite C    only run suites where PROGNAME matches C pattern
 -f, --output-path use xml report instead of standard output
 -d, --shuffle-seed shuffling test sequence,between 0 and 99999,
                   when seed is 0,use time(NULL) as the seed for
                   the random number generator
Example: cmocka --suite mm|sched --test Test* --skip TestNuttxMm0[123]

nsh>
```


testing/cmocka/Kconfig
```diff
@@ -12,6 +12,15 @@ config TESTING_CMOCKA
 
 if TESTING_CMOCKA
 
+config TESTING_CMOCKA_ASLIB
```
xiaoxiang781216: it's better to use TESTING_CMOCKA to enable the library, and use TESTING_CMOCKA_PROG to enable the program.

testing/cmocka/Kconfig
xiaoxiang781216:
```suggestion
	bool "Enable libcmocka"
```

testing/cmocka/Kconfig
```diff
@@ -12,6 +12,15 @@ config TESTING_CMOCKA
 
 if TESTING_CMOCKA
 
+config TESTING_CMOCKA_PROG
+	bool "Compile cmocka binary"
```
xiaoxiang781216:
```suggestion
	tristate "Compile cmocka binary"
```

testing/cmocka/Makefile
```diff
@@ -29,12 +29,14 @@ CFLAGS += -DHAVE_CLOCK_REALTIME=1
 CFLAGS += -DHAVE_STRUCT_TIMESPEC=1
 CFLAGS += ${INCDIR_PREFIX}$(APPDIR)/testing/cmocka
 
+MODULE = $(CONFIG_TESTING_CMOCKA)
```
xiaoxiang781216:
```suggestion
MODULE = $(CONFIG_TESTING_CMOCKA_PROG)
```

testing/cmocka/Makefile
```diff
@@ -29,12 +29,14 @@ CFLAGS += -DHAVE_CLOCK_REALTIME=1
...
+
+ifeq ($(CONFIG_TESTING_CMOCKA_PROG),y)
```
xiaoxiang781216:
```suggestion
ifneq ($(CONFIG_TESTING_CMOCKA_PROG),)
```

testing/cmocka/Kconfig
linguini1: How come tristate should be used here? Is it because of the `MODULE` variable in Makefile?

xiaoxiang781216 (approved):

linguini1:
Any ideas why the CI test is failing? It seems that the cmocka binary is present, just an assertion is failing.

acassis:
@linguini1 it is saying that boards/risc-v/qemu-rv/rv-virt/configs/citest/logs/ is not tracked, not sure if it needs to be.

Maybe @simbit18 or @lupyuen has some idea, they know more about the citest

simbit18:
Please excuse my silence, but I have some important family matters to resolve. Unfortunately, real life is not binary, 0 and 1, black and white, etc.,
but full of nuances, and its complexity requires time and patience.
When I have time, I will take a closer look at this PR.

cederom (approved):
Thank you @linguini1 :-)

jerpelea (approved):

linguini1:
So can this and the twin PR be merged?

acassis:
> So can this and the twin PR be merged?

Unfortunately none PR with error could be merged, unless it is a transient error in our CI. Otherwise all new PRs will fail because the CI catch the same error again and again.

I restarted the CI, so let see if the error get fixed, case the error persist I suggest you to update your upstream and rebase your branch to it, sometimes it is enough to fix the error ;-)

cederom:
Kinda corner case because we change testing stuff that affects ci and had cyclic dependency between two repos that also affects ci that in cycle is affected by the change :-P

We may force push to the upstream, but this kind of breaking changes should be avoided, maybe there is another way to introduce / implement the change? Maybe additional change in the ci setup here in this pr on gh in pair with this cmocka change?

If there is no other way we may need to force push on the one repo to enable the other.

acassis:
> > > So can this and the twin PR be merged?
> > 
> > 
> > Unfortunately none PR with error could be merged, unless it is a transient error in our CI. Otherwise all new PRs will fail because the CI catch the same error again and again.
> > I restarted the CI, so let see if the error get fixed, case the error persist I suggest you to update your upstream and rebase your branch to it, sometimes it is enough to fix the error ;-)
> 
> Hmmm, then it'll be impossible to merge this since it needs the NuttX PR to get merged in order to pass CI, and the NuttX PR needs this to merge to pass CI as well.
> 
> Any ideas how to get around that issue in the coupling? It's because this change directly affects the configuration the CI runner uses.

Ok, if the fail is just because circular dependence, then we can merge with error, assuming after both commits are integrated the error will be fixed

linguini1:
> Ok, if the fail is just because circular dependence, then we can merge with error, assuming after both commits are integrated the error will be fixed

That's correct. At least, things compile and run fine in my local testing.

xiaoxiang781216 (approved):

PR #3275, opened 2025-12-24
netutils/libwebsockets: add cmake support

yjq91115:
## Summary

Add CMake build support for `libwebsockets`, allowing it to be configured and built as part of NuttX using a CMake-based workflow. This change fills the gap where only other build systems were supported, by introducing/updating the necessary CMake configuration and target definitions so that compile options, include paths, and link dependencies are correctly resolved and integrated under CMake.

## Impact

- Build process: CMake-based builds can now build and link `libwebsockets` directly, reducing manual integration effort (e.g., custom Makefiles/scripts).

### DIFF
--- a/netutils/libwebsockets/CMakeLists.txt
+++ b/netutils/libwebsockets/CMakeLists.txt
@@ -1,0 +1,188 @@
+# ##############################################################################
+# apps/netutils/libwebsockets/CMakeLists.txt
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  The ASF licenses this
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

acassis:
> > @yjq91115 do you have some sample example that uses libwebsockets?
> 
> Currently, this patch only supports CMake compilation, which has the same capabilities as a Makefile. There are still some compilation issues in the libwebsocket library, which will be fixed later to ensure proper functionality. Do you think this patch should be merged in first? @acassis

Hi @yjq91115 yes both CMake and Makefile are equivalent, but NuttX needs both to work correctly, many users are still using Makefile because it was the only option for a long time (and I think it still the default option).

Please submit some sample code to let users to test it easily, i.e. integrated on sim board config.

PR #3293, opened 2025-12-31
nshlib/ls_handler: eliminate floating-point operations for human-readable sizes

Donny9:
## Summary
nshlib/ls_handler: eliminate floating-point operations for human-readable sizes

Replace floating-point arithmetic with fixed-point integer math to avoid linking soft-float library (~2-3KB Flash) when displaying human-readable file sizes (ls -lh command).

Changes:
- Use integer division and modulo to calculate size components
- Calculate decimal part: (remainder * 10) / unit for one decimal place
- Refactor duplicated code: consolidate GB/MB/KB logic into single path
- Remove CONFIG_HAVE_FLOAT dependency

This eliminates calls to __aeabi_f2d, __aeabi_fmul, __aeabi_i2f and other ARM EABI floating-point helpers, reducing Flash footprint for systems compiled with -mfloat-abi=soft.

## Impact

save flash size

## Testing

nsh ls

### DIFF
--- a/nshlib/nsh_fscmds.c
+++ b/nshlib/nsh_fscmds.c
@@ -518,24 +518,39 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
 
       if ((lsflags & LSFLAGS_SIZE) != 0)
         {
-#ifdef CONFIG_HAVE_FLOAT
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

linguini1:
Looks good! Can you add the output of the `ls` command to your testing section?

Issue #3294, opened 2026-01-01
[BUG]  trap when run "iptables -L" if configured CONFIG_NET_NAT while not configure CONFIG_NET_IPFILTER


masc2008:
### Description / Steps to reproduce the issue

when configure nuttx, configure CONFIG_NET_NAT while not configure CONFIG_NET_IPFILTER, 
then run: iptables -L

then trap happen, since iptables does set table type.

### On which OS does this issue occur?

[OS: Linux]

### What is the version of your OS?

ubuntu 24

### NuttX Version

master

### Issue Architecture

[Arch: all]

### Issue Area

[Area: System]

### Verification

- [x] I have verified before submitting the report.

PR #3295, opened 2026-01-01
system/iptables: avoid trap when run "iptables -L" if configured CONF…

masc2008:
avoid trap when run "iptables -L" if configured CONFIG_NET_NAT while not configured CONFIG_NET_IPFILTER
It's a trap happen everytime.

It's a trap happen everytime.

it can fix issue: https://github.com/apache/nuttx-apps/issues/3294

### DIFF
--- a/system/iptables/iptables.c
+++ b/system/iptables/iptables.c
@@ -506,7 +506,7 @@ int main(int argc, FAR char *argv[])
   else
 #endif
 #ifdef CONFIG_NET_NAT
-  if (strcmp(args.table, TABLE_NAME_NAT) == 0)
+  if (args.table != NULL && strcmp(args.table, TABLE_NAME_NAT) == 0)
     {
       ret = iptables_apply(&args, iptables_nat_command);
       if (ret < 0)


xiaoxiang781216:
@masc2008 please fix:
```
../nuttx/tools/checkpatch.sh -c -u -m -g 85539a1223c4770ee36e68817f5bfe91e6b49369..HEAD
❌ Commit subject too long > 80
❌ Missing Signed-off-by
Used config files:
    1: .codespellrc
```

system/iptables/iptables.c
```diff
@@ -506,7 +506,7 @@ int main(int argc, FAR char *argv[])
   else
 #endif
 #ifdef CONFIG_NET_NAT
-  if (strcmp(args.table, TABLE_NAME_NAT) == 0)
+  if (args.table == NULL || strcmp(args.table, TABLE_NAME_NAT) == 0)
```
hartmannathan: Just to make sure I understand correctly: If `args.table` is NULL, it's OK to call `iptables_apply()`? Note that in `iptables_apply()`, `COMMAND_FLUSH`, `COMMAND_LIST`, and `COMMAND_POLICY` will try to access `args->table`. It's possible other commands also will try, but I didn't dig further. If this was unintended, the line could be changed as follows, which will prevent calling `iptables_apply()` if `args.table` is NULL.
```suggestion
  if (args.table != NULL && strcmp(args.table, TABLE_NAME_NAT) == 0)
```

masc2008: this will be more near to the standard behavior, thanks!

xiaoxiang781216 (approved):

masc2008:
thanks, I thought the "Signed-off-by" info will be added automatically when it was merged.

PR #3313, opened 2026-01-13
netlib: Replace perror with nerr

zs39:
## Summary

Use the nerr interface for error print.

## Impact

Replacing only the log interface will not affect functionality.

## Testing

This function already uses nerr; it's simply replacing perror with the nerr interface, so no testing is needed.

### DIFF
--- a/netutils/netlib/netlib_checkhttpconnectivity.c
+++ b/netutils/netlib/netlib_checkhttpconnectivity.c
@@ -224,7 +224,7 @@ int netlib_check_httpconnectivity(FAR const char *host,
   ret = connect(sock, (struct sockaddr *)&server_addr, sizeof(server_addr));
   if (ret < 0)
     {
-      perror("Connection failed");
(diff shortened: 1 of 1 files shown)


netutils/netlib/netlib_checkhttpconnectivity.c
```diff
@@ -224,7 +224,7 @@ int netlib_check_httpconnectivity(FAR const char *host,
   ret = connect(sock, (struct sockaddr *)&server_addr, sizeof(server_addr));
   if (ret < 0)
     {
-      perror("Connection failed");
+      nerr("Connection failed");
```
anchao: `perror` is not equal to `nerr`, the `errno` should be added to the log message, even if logging is disabled.

zs39: The string corresponding to errno has been added for printing.

xiaoxiang781216 (approved):

PR #3324, opened 2026-01-14
libuv: fix sendmsg parameter type in NuttX port

extinguish:
## Summary

Update the sendmsg() stub implementation in the libuv NuttX port (src/unix/nuttx.c) to match the POSIX‑compliant signature by changing the msg parameter from struct msghdr * to const struct msghdr *. This change synchronizes the libuv port with the recent NuttX socket API update.

## Impact

1. Ensures the libuv NuttX port compiles cleanly with the updated NuttX socket headers.
2. Maintains type consistency between libuv and the underlying NuttX socket API.
3. No functional change—the stub still returns UV_ENOTSUP as before.

## Testing

Verified that the modified libuv patch applies successfully and compiles without type‑mismatch warnings.

### DIFF
--- a/system/libuv/0001-libuv-port-for-nuttx.patch
+++ b/system/libuv/0001-libuv-port-for-nuttx.patch
@@ -424,7 +424,7 @@ index 00000000..298bd2f8
 +    return 0;
 +  }
 +
-+  /* Make sure the memory is initiallized to zero using calloc() */
++  /* Make sure the memory is initialized to zero using calloc() */
 +  *addresses = uv__calloc(*count, sizeof(**addresses));
 +  if (*addresses == NULL) {
 +    freeifaddrs(addrs);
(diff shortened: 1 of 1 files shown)


system/libuv/0001-libuv-port-for-nuttx.patch
```diff
@@ -3487,7 +3487,7 @@ index bbc0c305..4e7996f5 100644
 -#if defined(__CYGWIN__) || defined(__MSYS__) || defined(__PASE__)
 +#if defined(__CYGWIN__) || defined(__MSYS__) || defined(__PASE__) || defined(__NuttX__)
  /* Cygwin crashes or hangs in socket() with too many AF_INET sockets.  */
- /* IBMi PASE timeout with too many AF_INET sockets.  */
+ /* IBMi PASS timeout with too many AF_INET sockets.  */
```
xiaoxiang781216: restore IBMi PASE is a trademark

extinguish: > restore IBMi PASE is a trademark

fixed in new patchset, but the precheck report error as:
```
/home/runner/work/nuttx-apps/nuttx-apps/apps/system/libuv/0001-libuv-port-for-nuttx.patch:3490: PASE ==> PASS, PACE, PARSE
```

should we skip this check on this patch?

cederom:
still some typos? :-P

PR #3330, opened 2026-01-14
ostest: add kconfig to set priority

zzby0:

## Summary
sometimes ostest fail because of other task with higher priority are running, this PR add kconfig to set ostest priority so that we can increase ostest priority easily.

## Impact

ostest priority

## Testing

ostest

### DIFF
--- a/testing/ostest/cancel.c
+++ b/testing/ostest/cancel.c
@@ -287,8 +287,8 @@ static FAR void *asynch_waiter(FAR void *parameter)
        ASSERT(false);
     }
 
-  /* Then wait a bit.  We should be canceled aynchronously while waiting, but
-   * the cancellation should pend because we are non-cancellable.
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (dismissed):

testing/ostest/mqueue.c
```diff
@@ -348,7 +341,7 @@ void mqueue_test(void)
       ASSERT(false);
     }
 
-  sparam.sched_priority = (prio_min + prio_mid) / 2;
+  sparam.sched_priority = PRIORITY - 10;
```
cederom: what if someone sets priority <= 10 for any reason? shouldn't we have a check here?

zzby0: thanks for review :)  i have added check here

PR #3365, opened 2026-01-21
fix(macro): fix CAPIO_FREQUENCY macro typo

comejv:
Going from CAPIOC_FREQUENCE to CAPIO_FREQUENCY according to https://github.com/apache/nuttx/pull/16925

## Summary

CAPIOC_FREQUENCE is a typo, should be FREQUENCY. nuttx/pull/16925 fixes this, so the app side should be fixed too.


simbit18:
OK, understood. I will force the merge.

PR #3406, opened 2026-02-19
Added changes for ls-l time support

### LINKED ISSUE #17063: [FEATURE] ls command doesn't show files date / timestamp
### Is your feature request related to a problem? Please describe.

NuttX "ls -l" command currently doesn't show the file timestamp, although this information is stored in some filesystems supported by NuttX.

### Describe the solution you'd like

Modify ls_handler() to print timestamp when it is available.

### Describe alternatives you've considered

_No response_

### Verification

Biancaa-R:
Fixes the issue : https://github.com/apache/nuttx/issues/17063

image

a) modification aims to add human-readable timestamp reporting to the ls command by utilizing the struct stat information. 
b) time_t st_mtime; / Time of last modification / from stat register to maintain time records.
c) Epoch entries replaced by ---.
d) ls -l (without options) implementation done here.

## Summary
  
* So basically displaying the last modified time that is stored in the stat struct in ls -l command.
* stat(fullpath, &buf): This fills the buf structure with metadata from the filesystem.
*  buf.st_mtime: This specifically targets the "Last Modified" timestamp.
* Safety Check: if (ret == 0 && buf.st_mtime > 0 ...) ensures that the stat call was successful and that the time value is valid before trying to process it.
 *  localtime_r(&t, &tm): This takes the Unix epoch time (seconds since 1970) and breaks it down into a struct tm (years, months, days, etc.).
 *  Formatting: %04d-%02d-%02d, which outputs the ISO 8601 standard (e.g., 2023-10-27).
## Impact

Previously the ls -l didnt display the time of last modification , this is a small fix for more "linux like" experience.

## Testing

* Apache NuttX Simulation Mode allows you to run the OS as a native Linux process.
* Building and running it and executing the ls -l command in the shell.

### DIFF
--- a/nshlib/nsh_fscmds.c
+++ b/nshlib/nsh_fscmds.c
@@ -44,8 +44,11 @@
 #include <libgen.h>
 #include <errno.h>
 #include <debug.h>
-
 #include "nsh.h"
+
+#ifdef CONFIG_ARCH_SIM
+#  include <time.h>
+#endif
 
 #if !defined(CONFIG_DISABLE_MOUNTPOINT)
 #  include <sys/mount.h>
@@ -364,7 +367,126 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
 
       memset(&buf, 0, sizeof(struct stat));
 
-      /* stat the file */
+      /* If entryp is provided, listing a directory and need to
+       * construct the full path to stat the file. Otherwise, dirpath
+       * is the target itself. (no separate file name as entryp)
+       */
+
+      if (entryp != NULL)
+        {
+          FAR char *fullpath = nsh_getdirpath(vtbl, dirpath, entryp->d_name);
+          ret = stat(fullpath, &buf);
+          free(fullpath);
+        }
+      else
+        {
+          ret = stat(dirpath, &buf);
+        }
+
+#ifdef CONFIG_CLOCK_TIMEKEEPING
+      /* manual epoch time to date calculation to reduce extra memory
+       * by using includes For boards with minimal flash.
+       */
+
+#  ifdef CONFIG_ARCH_SIM
+      struct timespec ts;
+
+      /* This pulls the ACTUAL current time from your Linux Host */
+
+      /* Sometime defaults to 2008 */
+
+      if (clock_gettime(CLOCK_REALTIME, &ts) == 0)
+        {
+          clock_settime(CLOCK_REALTIME, &ts);
+        }
+#  endif
+
+      /* for referencing /data : if not mounted the reference will fail. */
+
+#  if defined(CONFIG_ARCH_SIM) && defined(CONFIG_FS_HOSTFS)
+      static bool g_time_synced = false;
+
+      if (!g_time_synced)
+        {
+          struct stat hstat;
+
+          /* Data section always maintains the current time value. */
+
+          if (stat("/data", &hstat) == 0)
+            {
+              struct timespec ts_sync;
+
+              ts_sync.tv_sec  = hstat.st_mtime;
+              ts_sync.tv_nsec = 0;
+
+              /* This is the magic line that sets the system clock */
+
+              if (clock_settime(CLOCK_REALTIME, &ts_sync) == 0)
+                {
+                  g_time_synced = true;
+                }
+            }
+        }
+#  endif
+
+      if (ret == 0 && buf.st_mtime > 0 && buf.st_mtime < 0x7fffffff)
+        {
+          uint32_t seconds = (uint32_t)buf.st_mtime;
+          uint32_t minutes = seconds / 60;
+          uint32_t hours   = minutes / 60;
+          uint32_t days    = hours / 24;
+          uint32_t y       = 1970;
+          uint32_t m       = 0;
+
+          while (1)
+            {
+              uint32_t diy = (y % 4 == 0 && (y % 100 != 0 ||
+                              y % 400 == 0)) ? 366 : 365;
+
+              if (days < diy)
+                {
+                  break;
+                }
+
+              days -= diy;
+              y++;
+            }
+
+          static const uint8_t dim[] =
+          {
+            31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31
+          };
+
+          for (m = 0; m < 12; m++)
+            {
+              uint32_t d_m = dim[m];
+
+              if (m == 1 && (y % 4 == 0 && (y % 100 != 0 || y % 400 == 0)))
+                {
+                  d_m = 29;
+                }
+
+              if (days < d_m)
+                {
+                  break;
+                }
+
+              days -= d_m;
+            }
+
+          nsh_output(vtbl, " %04u-%02u-%02u %02u:%02u:%02u",
+                     (unsigned int)y, (unsigned int)m + 1,
+                     (unsigned int)days + 1, (unsigned int)(hours % 24),
+                     (unsigned int)(minutes % 60),
+                     (unsigned int)(seconds % 60));
+        }
+      else
+        {
+          /* Incase valid time entry was not maintained in the stat struct. */
+
+          nsh_output(vtbl, " ----/--/-- --:--");
+        }
+#endif
 
       if (entryp != NULL)
         {


cederom:
Thank you @Biancaa-R :-) Can you please provide output of `ls -l` command from Linux/Unix shell and `ls -l` from nsh after your change? Are these outputs comparable for parse (i.e. information order) :-)

cederom (changes requested):
CI build failed.

nshlib/nsh_fscmds.c
```diff
@@ -360,13 +364,52 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
...
 
-      if (entryp != NULL)
+      #ifdef CONFIG_LIBC_LOCALTIME
+      localtime_r(&t, &tm);
+      nsh_output(vtbl, " %04d-%02d-%02d %02d:%02d",
+               tm.tm_year + 1900,
+               tm.tm_mon + 1,
+               tm.tm_mday,
+               tm.tm_hour,
+               tm.tm_min);
```
acassis: @Biancaa-R please fix the alignment, all "tm.*" lines should start immediately after the "(" from previous line.

Biancaa-R:
> I read the review quickly and thought the issue was because of bss overflow and I initially suggested for it .. But since its flash overflow (as xip code/data) so I kept time module only for simulation case where there is a lot of memory to spare. I introduced a manual logic instead of libraries to reduce space for epoch calculation in microcontrollers , all the memory heavy implementations are conditional.
* Output result : Creating of new file also shows current time.
* Issue : Default time for a file without modification goes to epoch , getting of linux's internal time also sometime takes values like 2008 ,so I diverted the value taken by /data folder for accuracy in all cases (for sim with hostfs.) if present.
image

Checkpatch:
image

General output in linux :  (but in our case nuttx its fully single user for now.)
image

nshlib/nsh_fscmds.c
```diff
@@ -364,7 +367,126 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
...
+
+      /* This pulls the ACTUAL current time from your Linux Host */
+
+      /* Sometime defaults to 2008 */
+
+      if (clock_gettime(CLOCK_REALTIME, &ts) == 0)
+        {
+          clock_settime(CLOCK_REALTIME, &ts);
+        }
+#  endif
```
acassis, line 402: I think this part is the same for SIM and for native boards. Could you please confirm it in some ESP32 or STM32 board? If that is true we don't need this #ifdef CONFIG_ARCH_SIM

nshlib/nsh_fscmds.c
```diff
@@ -44,9 +44,12 @@
 #include <libgen.h>
 #include <errno.h>
 #include <debug.h>
-
 #include "nsh.h"
 
+#ifdef CONFIG_ARCH_SIM
+#  include <time.h>
+#endif
```
acassis, line 51: @Biancaa-R this could should be generic, so seems like you was guided incorrectly to fix it only for arch SIM. So, please remove this #ifdef / #else check

nshlib/nsh_fscmds.c
```diff
@@ -364,7 +367,126 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
...
+
+              /* This is the magic line that sets the system clock */
+
+              if (clock_settime(CLOCK_REALTIME, &ts_sync) == 0)
+                {
+                  g_time_synced = true;
+                }
+            }
+        }
+#  endif
```
acassis, line 430: I tested this hack on sim:nsh (that has CONFIG_FS_HOSTFS enabled by default) and it worked.

In fact my test was very simple, I modified the hello example to:

```
#include <time.h>
#include <sys/stat.h>

....

{
  struct stat hstat;
	
  /* Data section always maintains the current time value. */
	
  if (stat("/data", &hstat) == 0)
    {
       struct timespec ts_sync;
	
       ts_sync.tv_sec  = hstat.st_mtime;
       ts_sync.tv_nsec = 0;

       /* This is the magic line that sets the system clock */

      if (clock_settime(CLOCK_REALTIME, &ts_sync) == 0)
        {
		  return 0;
        }
    }

	return -1;
}
```

Then I ran ./nuttx and tested this way:

```
NuttShell (NSH) NuttX-12.12.0
nsh> date
Sun, Jun 01 00:00:01 2008
nsh> hello
nsh> echo $?
0
nsh> date
Mon, Feb 23 20:15:11 2026
nsh>
```

So, this hack should be in the #else of CONFIG_CLOCK_TIMEKEEPING

nshlib/nsh_fscmds.c
Biancaa-R: Sure sir Ill check it with a MC.

nshlib/nsh_fscmds.c
acassis: @Biancaa-R "/* This is the magic line that sets the system clock */" Sounds very AI generated, please use: "Now, set the system clock with the time value got /data"

nshlib/nsh_fscmds.c
Biancaa-R: > @Biancaa-R this could should be generic, so seems like you was guided incorrectly to fix it only for arch SIM. So, please remove this #ifdef / #else check

Yes , yes it can be generic but last time I added it and used a function the 128KB flash was overflowed by 4 bytes in the stm board . Thats the reason I tried to avoid that include and made my own implementation for epoch seconds to date time conversion, just so that there is no new mem overflow.

nshlib/nsh_fscmds.c
acassis: @Biancaa-R sorry, I just saw now that this is a hack to fix the 2008 issue. In fact it happened in my test (that don't have your PR applied), see below

nshlib/nsh_fscmds.c
acassis: Ok, makes sense. Please just confirm that your function works on some STM32 board ;-)

nshlib/nsh_fscmds.c
acassis: @Biancaa-R just remove all hacks code related SIM, just keep the code to display the date on "ls -l".
For native boards, I think when RTC and CONFIG_CLOCK_TIMEKEEPING is not enabled the date will be wrong, so in this case, instead of showing an old invalid date, just display "--/--/----"

acassis:
> lets just close this already okay?

Why to close it? She implement the "ls -l" that seems to be right

nshlib/nsh_fscmds.c
```diff
@@ -44,9 +44,12 @@
 #include <libgen.h>
 #include <errno.h>
 #include <debug.h>
-
 #include "nsh.h"
 
+#ifdef CONFIG_ARCH_SIM
```
xiaoxiang781216, line 49: why check CONFIG_ARCH_SIM

nshlib/nsh_fscmds.c
```diff
@@ -364,7 +367,126 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
...
+
+              if (clock_settime(CLOCK_REALTIME, &ts_sync) == 0)
+                {
+                  g_time_synced = true;
+                }
+            }
+        }
+#  endif
+
+      if (ret == 0 && buf.st_mtime > 0 && buf.st_mtime < 0x7fffffff)
```
xiaoxiang781216, line 432: call strftime instead

nshlib/nsh_fscmds.c
```diff
@@ -364,7 +367,126 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
 
       memset(&buf, 0, sizeof(struct stat));
 
-      /* stat the file */
+      /* If entryp is provided, listing a directory and need to
+       * construct the full path to stat the file. Otherwise, dirpath
+       * is the target itself. (no separate file name as entryp)
+       */
+
+      if (entryp != NULL)
```
xiaoxiang781216, line 375: why dup with line 494

nshlib/nsh_fscmds.c
```diff
@@ -364,7 +367,126 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
...
+#  ifdef CONFIG_ARCH_SIM
+      struct timespec ts;
+
+      /* This pulls the ACTUAL current time from your Linux Host */
+
+      /* Sometime defaults to 2008 */
+
+      if (clock_gettime(CLOCK_REALTIME, &ts) == 0)
+        {
+          clock_settime(CLOCK_REALTIME, &ts);
```
xiaoxiang781216, line 400: why do you change host time?

nshlib/nsh_fscmds.c
```diff
@@ -364,7 +367,126 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
...
+          if (stat("/data", &hstat) == 0)
+            {
+              struct timespec ts_sync;
+
+              ts_sync.tv_sec  = hstat.st_mtime;
+              ts_sync.tv_nsec = 0;
+
+              /* This is the magic line that sets the system clock */
+
+              if (clock_settime(CLOCK_REALTIME, &ts_sync) == 0)
```
xiaoxiang781216, line 424: ditto

nshlib/nsh_fscmds.c
```diff
@@ -364,7 +367,126 @@ static int ls_handler(FAR struct nsh_vtbl_s *vtbl, FAR const char *dirpath,
...
+
+              if (days < d_m)
+                {
+                  break;
+                }
+
+              days -= d_m;
+            }
+
+          nsh_output(vtbl, " %04u-%02u-%02u %02u:%02u:%02u",
```
xiaoxiang781216, line 477: should move just before file name like bash:
````
drwxr-xr-x 2 xiaoxiang xiaoxiang 4096 Feb 15 11:02 cmake_with_components
drwxr-xr-x 3 xiaoxiang xiaoxiang 4096 Feb 15 11:02 legacy_cmake_relpath
drwxr-xr-x 2 xiaoxiang xiaoxiang 4096 Feb 15 11:02 legacy_cmake_simple
```

xiaoxiang781216 (changes requested):

nshlib/nsh_fscmds.c
Biancaa-R: As I had mentioned , the logic I had built before ,including time.h for all cases was too much flash memory (xip code ,data ) for a particular stm board that had 128KB flash . So I replaced the logic for manual calculation of datetime from epoch instead of using localtime_r(&t, &tm) .

That logic will work in microcontrollers that have RTc and have kept TIMEKEEPING configured.
In simulation , it doesnt work because of  : https://github.com/apache/nuttx-apps/issues/3410 .

So I tried to fix it by gettime , settime functions (Only in case SIM is used) and since there is enough memory to handle it . 

But I have to remove it as it was pointed out by @linguini1 that its not the work of ls hander to display accurate date but , the date given by the system.

or we have to manually modify once before creating files in SIM.

```
nsh> date
Thu, Jan 22 18:08:41 2004 
nsh> date -s "Feb 23 22:35:00 2026"

```

nshlib/nsh_fscmds.c
Biancaa-R: No this one tries to set the time of /data section as the time pulled from the linux host in SIM defaults to 2004/ 2008 ,In case a hostfs is present /data always takes the accurate time , so I am keeping it as a reference here in this place. Look at : https://github.com/apache/nuttx-apps/pull/3406/changes/BASE..3ae3781f8844d1d3f6fa3b6a96b3515a48129f72#r2842897622

nshlib/nsh_fscmds.c
Biancaa-R: Sure Ill move it to the top so it happens once . This one gets executed in case of ls -l  specifically.

nshlib/nsh_fscmds.c
Biancaa-R: I dont know , but strftime() is a heavy function that pulls in a lot of the standard C library (libc). Many NuttX boards are configured with CONFIG_LIBC_LOCALTIME disabled to save space. The main reason I added a manual calculation for epoch to datetime is to save space . please look at : https://github.com/apache/nuttx-apps/pull/3406#issuecomment-3935605561

nshlib/nsh_fscmds.c
xiaoxiang781216: first, you should drop all sim specific code from nsh, it's wrong to add any arch specific code in the application. second, I don't understand why do you change realtime in ls command it's definitely wrong. before you make this hack, explain what's problem you encounter first.

nshlib/nsh_fscmds.c
xiaoxiang781216: but strftime doesn't require you enble CONFIG_LIBC_LOCALTIME,

nshlib/nsh_fscmds.c
Biancaa-R: Yes sir , as @linguini1 sir pointed we dont have to modify the displayed time in ls , even if its wrong. (only have ls functionality.) So I'll REMOVE SIM specific code . I 'll make the changes and push it.


For the issue I encountered please look at : https://github.com/apache/nuttx-apps/pull/3406#issuecomment-3947080063  
```

 nsh> echo "hey" > hello.txt

nsh> ls -l

/:

 ----/--/-- --:-- dr-xr-xr-x           0 bin/

 2026-02-23 16:58:38 drwxrwxr-x        4096 data/

 2008-06-01 00:00:00 drwxrwxrwx           0 dev/

 ----/--/-- --:-- dr-xr-xr-x           0 etc/

 2004-01-22 18:08:50 -rw-rw-rw-           4 hello.txt

 2004-01-22 18:08:46 -rw-rw-rw-           3 hi.txt

 ----/--/-- --:-- dr--r--r--           0 proc/

 ----/--/-- --:-- drw-rw-rw-           0 tmp/
``` 

Using clk_gettime for puling current time from Linux Host) sometimes gave me faulty results from the system clock ,where the system clock often defaults to the Unix epoch or a fixed build date (like 2008/2004).

* This is because in HostFS the data is dynamic and has access to system's clk -> So it will get correct time.
* All other sections and the files we create , assumes this base time (which is not accurate.)
* Unlike other microcontrollers we dont have a RTC here.

SO I'll remove the SIM specific code ,and push it again ,please give me some time ,Ill fix it and get back ,

nshlib/nsh_fscmds.c
xiaoxiang781216: it's simple to call ctime_r or strftime and print out the result, I don't understand why do you need any other code.

nshlib/nsh_fscmds.c
Biancaa-R: Oh okay , but it will still need <time.h> , Ill calculate the final size @xiaoxiang781216 , if It fits in the mem of 128KB flash of stm ,I'll add that instead . Thank you.

linguini1:
Any plans to work on this @Biancaa-R ?

Biancaa-R:
> Any plans to work on this @Biancaa-R ?

I m sorry for the delay Ill finish it by this weekend . I had some personal issues to deal with .

linguini1:
No rush, just checking

acassis:
@Biancaa-R @linguini1 I think the last patches from @Abhishekmishra2808 already fixed that, please double check his patches

Biancaa-R:
> @Biancaa-R @linguini1 I think the last patches from @Abhishekmishra2808 already fixed that, please double check his patches

Im sorry but I could only find these :
image

acassis:
> > @Biancaa-R @linguini1 I think the last patches from @Abhishekmishra2808 already fixed that, please double check his patches
> 
> Im sorry but I could only find these : image

You are right, he added only username and group listing. Sorry my confusion.

Biancaa-R:
> > > @Biancaa-R @linguini1 I think the last patches from @Abhishekmishra2808 already fixed that, please double check his patches
> > 
> > 
> > Im sorry but I could only find these : image
> 
> You are right, he added only username and group listing. Sorry my confusion.

Oh no issues ,Ill finish this pr asap .Thank you for your patience .

acassis:
@Biancaa-R someone already implemented it on mainline. Thank you!

PR #3415, opened 2026-02-28
Fix QOS arguments parsing for mqttc_pub

dmihai03:
## Summary

Modify the switch cases in parsearg function for QOS levels definition. The comparison was made between a long value and a char.

## Impact

All the 2 levels of QOS are now supported by the mqtt publisher example.

## Testing

Host used for build:
- OS: Linux 6.18.9-arch1-2
- CPU: AMD Ryzen 7 5800HS
- compiler: xtensa-esp32-elf-gcc

Target used for verification:
- board: ESP32C3
- config: esp32-devkitc:wifi

Connected the board to laptop's hotspot and dumped the packets in Wireshark. The broker used in testing is broker.hivemq.com.
For every QOS level the right handshakes are triggered.

1. Before:
    Before the changes, both QOS 1 and QOS 2 message publishing looked like this:

    image

    They were sent with QOS 0, fire and forget, because the QOS level was not modified in the packet.

2. After: 

    QOS 1 handshake:
    image

    QOS 2 handshake:
    image

### DIFF
--- a/examples/mqttc/mqttc_pub.c
+++ b/examples/mqttc/mqttc_pub.c
@@ -528,7 +528,7 @@ int main(int argc, FAR char *argv[])
                       NULL);
   if (mqtterr != MQTT_OK)
     {
-      printf("ERRPR! mqtt_init() failed.\n");
+      printf("ERROR! mqtt_init() failed.\n");
       goto err_with_conn;
     }
 


cederom:
Thank you @dmihai03 :-)
* Please update and sign git commit according to https://github.com/apache/nuttx/blob/master/CONTRIBUTING.md:
```
examples/mqttc: Fix QOS arguments parsing (char->long).

Modify the switch cases in parsearg function for QOS levels definition.
The comparison was made between a long value and a char.

Signed-off-by: ...
```
* It is always nice to show logs before and after the fix to show what the problem was and that you fixed it. For now we only see it is working but we do not see it was not working before :-)
* Other than that looks good :-)

cederom (changes requested):
todo: git commit fix :-)

acassis:
@dmihai03 nice finding! Thank you for submitting this fix.

BTW, could you please fix this typo that the CI found:

```
$ git grep ERRPR
examples/mqttc/mqttc_pub.c:      printf("ERRPR! mqtt_init() failed.\n");
```

acassis (approved):

acassis:
> Uhm, still some Linux build problem, CI restarted.



> Uhm, still some Linux build problem, CI restarted.

@cederom did you restarted the CI or did you restart only failing Jobs? Full restart can delay more, face similar issues and use more Runners

simbit18:
> I'm going to restart failing Jobs (I already did it two times).
> 
> @simbit18 seems like msvc is failing most of the time. BTW, the check is failing because MQTT CamelCase names, but let's ignore it

Hi @acassis, Runner images can fail for a thousand reasons, and sometimes it's not our fault but GitHub. This can happen for images for macOS, x64 Linux, and Windows images.

PR #3428, opened 2026-03-17
boot/nxboot: add flush barriers and CRC-validate primary before boot

neilberkman:
## Summary

Two hardening fixes for nxboot power-loss resilience:

1. **Flush barriers between critical partition operations** — Add flash_partition_flush() calls after copy_partition() completes in perform_update(). Without explicit barriers, writes may remain buffered in RAM when nxboot proceeds to the next phase. A power loss between phases can leave the recovery image uncommitted while the staging partition has already been consumed.

2. **Full CRC validation before booting primary** — Replace validate_image_header() with validate_image() in the final primary validation path of nxboot_perform_update(). The header-only check does not CRC-check the image body. After an interrupted update, a corrupt primary with an intact header would pass this check and be booted.

## Impact

- boot/nxboot only. No impact on other bootloaders or applications.
- Adds one fsync() call after each copy_partition() in the update path. On platforms without write buffering this is a no-op.
- Adds one full-image CRC computation at the end of the update path before booting. On SAMv7 this adds negligible boot time (confirmed by the nxboot maintainer).

## Testing

Tested with Renode emulation fault injection on nucleo-h743zi nxboot. The flush barriers and CRC validation together eliminate the persistent boot failure observed when FTL write buffering is enabled (92/94 failure rate reduced to 0/94).

### DIFF
--- a/boot/nxboot/loader/boot.c
+++ b/boot/nxboot/loader/boot.c
@@ -887,7 +887,6 @@ int nxboot_perform_update(bool check_only)
   int ret;
   int primary;
   struct nxboot_state state;
-  struct nxboot_img_header header;
 
   ret = nxboot_get_state(&state);
   if (ret < 0)


cederom (dismissed):
Thank you @neilberkman good catch! :-)

xiaoxiang781216 (dismissed):

xiaoxiang781216 (approved):

boot/nxboot/loader/boot.c
michallenc: Please also update this comment, it's no longer valid.

neilberkman: Fixed in 66c9805 — updated the comment to reflect that validate_image() now performs full CRC validation.

michallenc: This can also be now removed as `header` is not used. Sorry for double review, realized that only after I compiled the source.

neilberkman: No problem, fixed in 4772efa.

PR #3442, opened 2026-03-29
initial support for fixed-point data for sensors

raiden00pl:
## Summary

apps part for https://github.com/apache/nuttx/pull/17408

- system/sensorscope: support for fixed16 data 
- system/uorb: DEBUG_UORB depends on float data

## Impact

initial support for fixed16 sensor data

## Testing

https://github.com/apache/nuttx/pull/17408#issuecomment-4150426653

### DIFF
--- a/system/sensorscope/sensorscope_main.c
+++ b/system/sensorscope/sensorscope_main.c
@@ -50,6 +50,12 @@
 #define SENSOR_PATH_MAX    62
 #define SENSOR_CHNAME_MAX  16
 
+#ifdef CONFIG_SENSORS_USE_B16
+#  define NXSCOPE_DATA_TYPE NXSCOPE_TYPE_B16
+#else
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

cederom (approved):
Thank you @raiden00pl nice work! :-)

We may want to consider `DEBUG_UORB` also supporting fixed point variant? :-)

linguini1 (approved):

acassis (approved):

jerpelea (approved):

raiden00pl:
@cederom there is no printf support for fixed point. The only way is to convert fixed-point to float. This contradicts the idea of ​​introducing fixed points for sensors (getting rid of floats), but for debugging purposes, it makes sense and can be useful.

EDIT: or maybe we should support fixed point in printf as not standard extention

PR #3460, opened 2026-04-22
netlib: support checkipconnectivity when use reslov.conf

zhhyu7:
## Summary
Improve netlib_check_ipconnectivity() to support DNS nameservers from resolv.conf via dns_foreach_nameserver(), instead of relying on the compile-time CONFIG_NETDB_DNSSERVER_IPv4ADDR macro.

## Impact
- Add dns_ping_callback() to iterate over all configured DNS nameservers and ping each one until a reply is received.
- When ip is NULL, use dns_foreach_nameserver() to dynamically resolve and ping DNS servers at runtime, replacing the previous #ifdef CONFIG_NETDB_DNSSERVER_IPv4ADDR compile-time approach.
- When ip is provided, ping the specified address directly (behavior unchanged).
- Fix missing newline at end of file.

## Testing
Verify netlib_check_ipconnectivity(NULL, ...) pings DNS servers from resolv.conf and ram.

### DIFF
--- a/netutils/netlib/netlib_checkipconnectivity.c
+++ b/netutils/netlib/netlib_checkipconnectivity.c
@@ -35,6 +35,7 @@
 
 #include <arpa/inet.h>
 #include <nuttx/net/ip.h>
+#include <nuttx/net/dns.h>
 
 #include "netutils/icmp_ping.h"
 #include "netutils/netlib.h"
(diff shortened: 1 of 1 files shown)


jerpelea (approved):

masc2008:
@zhhyu7  it's not a good method to use ping to check if DNS server is reachable.  Here is the reason:
A DNS server is only required to answer DNS queries on port 53 over UDP/TCP. Replying to ping means replying to ICMP Echo, which is separate and optional.

  So a DNS server may:

  - answer DNS normally
  - ignore or block ping
  - be behind a firewall that drops ICMP

  If you want to test DNS reachability, use a DNS query like nslookup, dig, or host, not ping.

And we need move further:
It's not good practice to just check network connectivity via ping ipv4 address, the device may just have an IPv6 address only.

zhhyu7:
> @zhhyu7 it's not a good method to use ping to check if DNS server is reachable. Here is the reason: A DNS server is only required to answer DNS queries on port 53 over UDP/TCP. Replying to ping means replying to ICMP Echo, which is separate and optional.
> 
> So a DNS server may:
> 
> * answer DNS normally
> * ignore or block ping
> * be behind a firewall that drops ICMP
> 
> If you want to test DNS reachability, use a DNS query like nslookup, dig, or host, not ping.
> 
> And we need move further: It's not good practice to just check network connectivity via ping ipv4 address, the device may just have an IPv6 address only.

@masc2008, You're right. Regarding NTP, I think we should directly remove the netlib_check_ipconnectivity check from the ntpclient file, because we already have backoff behavior to avoid screen flooding. Meanwhile, I will add a new patch later to select a more appropriate IP address for the check, such as DRaddr, and also add support for IPv6 addresses.

masc2008:
thanks! Good job HongYu!👍

cederom (approved):
Thank you @zhhyu7 :-) I agree with @masc2008 remarks please todo in a free moment :-)

PR #3461, opened 2026-04-22
testing: `irtest` usage to test Espressif's RMT peripheral.

tmedicci:
## Summary

* system/irtest: Set `irtest` command line size to CONFIG_LINE_MAX
  - Let `irtest` command line size be equal to CONFIG_LINE_MAX so that bigger (or smaller) commands can be set.

* system/irtest: Avoids silently truncating long write_data commands
  - Long write_data commands can be truncated without any check due to the maximum size of CONFIG_SYSTEM_IRTEST_MAX_SIRDATA. This PR adds a check to avoid a silent fail.

* system/irtest: Fix issues regarding portability and buffer size
  - This PR fixes issues regarding portability by using variables with fixed width (and their format macro constants).

* examples: Remove deprecated `rmtchar` example application
  - The upper-half RMT driver is no longer available on NuttX. Instead, Espressif's RMT peripheral was bound directly to the lirc driver. For testing purposes, use the `irtest` application. NuttX OS PR https://github.com/apache/nuttx/pull/18654 removed the upper-half driver interface used by this application, so removing it does not break any existing feature on NuttX.

## Impact

Impact on user: Yes. Improve user's experience when trying `irqtest` testing application.

Impact on build: No.

Impact on hardware: No.

Impact on documentation: No.

Impact on security: No.

Impact on compatibility: No.

## Testing

Build the `rmt` defconfig of any Espressif SoC, interconnect output and input pins externally and run `irqtest`. For ESP32-C6-DevKitC-1 V1.2, for instance:

### Building

```
make -j distclean
./tools/configure.sh -S esp32c6-devkitc:rmt
kconfig-tweak -e SYSTEM_IRTEST
make olddefconfig
make flash ESPTOOL_PORT=/dev/ttyUSB0
picocom -b 115200 /dev/ttyUSB0
```

### Running

Connect GPIO 8 (RMT output pin) to GPIO 2 (RMT input pin) externally and run `irqtest`:

```
nsh> irtest
$open_device(/dev/lirc0)
$open_device(/dev/lirc1)
$write_data(1) 16777229 16 16777235 23
$read_data(0,4)
16777229, 16, 16777235, 23
```

### Results

Read data equals written data. Loopback test succeeded!


system/irtest/main.cxx
```diff
@@ -25,10 +25,15 @@
  ****************************************************************************/
 
 #include <stdio.h>
+#include <nuttx/config.h>
```
xiaoxiang781216, line 27: move before line 26

cederom:
Thank you @tmedicci :-)

Would it be possible to mark this change as breaking with a short note that `rmtchar` is removed and replaced by `irtest`?

We also need documentation update there are references to `rmtchar` see https://nuttx.apache.org/docs/latest/search.html?q=rmtchar&check_keywords=yes&area=default :-)

tmedicci:
Hi @cederom !

> Would it be possible to mark this change as breaking with a short note that `rmtchar` is removed and replaced by `irtest`?

I updated PR's description (and the respective commit) with the following message:

```
The upper-half RMT driver is no longer available on NuttX. Instead, Espressif's RMT peripheral was bound directly to the lirc driver. For testing purposes, use the irtest application.
```

I also added the `breaking change` tag, but it isn't really a breaking change because it just removed a deprecated application. The "breaking change" was introduced by https://github.com/apache/nuttx/pull/18654

```
NuttX OS PR https://github.com/apache/nuttx/pull/18654 removed the upper-half driver interface used by this application, so removing it does not break any existing feature on NuttX.
```

> We also need documentation update there are references to `rmtchar` see https://nuttx.apache.org/docs/latest/search.html?q=rmtchar&check_keywords=yes&area=default :-)

I'll prepare a PR for updating on NuttX OS repo, thanks!

acassis:
@xiaoxiang781216 this irtest in C++ will require more memory (normally just enabling libcxx will consume more 100KB even with almost all features disabled: exceptions, RTTI, etc). So it will a show stopper for many small MCUs.

Is it possible to convert it from C++ to C? I noticed the code is mostly simple C, except for some templates usage.

tmedicci:
Updated documentation PR at https://github.com/apache/nuttx/pull/18788.

simbit18 (approved):

cederom:
> @xiaoxiang781216 this irtest in C++ will require more memory (normally just enabling libcxx will consume more 100KB even with almost all features disabled: exceptions, RTTI, etc). So it will a show stopper for many small MCUs.
> 
> Is it possible to convert it from C++ to C? I noticed the code is mostly simple C, except for some templates usage.

+1 :-) This looks C :D

fdcavalcanti (approved):

cederom (approved):
Looks okay thank you @tmedicci :-) We can update C++ -> C in a free moment but we have working solution right now :-)

PR #3465, opened 2026-04-27
netlib: replace DNS ping with gateway ping for connectivity check

zhhyu7:
## Summary
Refactor netlib_check_ipconnectivity() to use gateway/router ping instead of DNS server ping for network connectivity checks. Add IPv6 support.

## Impact
- When ip is NULL, ping the default gateway of each interface (via getifaddrs) instead of the hardcoded DNS server address
- Add IPv6 ping support (icmpv6_ping) alongside existing IPv4 ping
- Add routing table fallback: ping routers from /proc/net/route and /proc/net/ipv6route when gateway ping fails
- For point-to-point/tunnel interfaces (no ARP), treat as reachable directly
- When ping is disabled at build time, fall back to checking network card status instead of always returning 1
- Auto-detect IPv4/IPv6 when a specific IP is provided
- netlib_checkipconnectivity.c no longer depends on CONFIG_NETUTILS_PING to compile
- Improve doxygen-style comments in netlib.h; fix typo "nagtive" → "negative"

## Testing
- Build tested with combinations of CONFIG_NETUTILS_PING, CONFIG_NETUTILS_PING6, CONFIG_NETDEV_IFINDEX, and HAVE_ROUTE_PROCFS enabled/disabled
- Verified ntpclient caller (netlib_check_ipconnectivity(NULL, 1, 1)) works correctly with the new gateway ping logic
- Tested on board with IPv4-only, IPv6-only, and dual-stack network interfaces
- Verified fallback path when ping is disabled at build time

### DIFF
--- a/include/netutils/netlib.h
+++ b/include/netutils/netlib.h
@@ -561,12 +561,49 @@ int netlib_getifstatistics(FAR const char *ifname,
 int netlib_check_ifconflict(FAR const char *ifname);
 #endif
 
-#ifdef CONFIG_NETUTILS_PING
+/****************************************************************************
+ * Name: netlib_check_ipconnectivity
+ *
+ * Description:
+ *   Check network connectivity by pinging a remote IP address.
+ *   If ip is NULL, ping the gateway of each network interface,
+ *   and optionally the routers from the routing table. If ping
+ *   is disabled, just check the status of the IP network card.
+ *
+ * Parameters:
+ *   ip       The ipv4 address to check, or NULL to ping gateways
+ *   timeout  The max timeout of each ping
+ *   retry    The retry times of ping
+ *
+ * Return:
+ *   nums of remote reply of ping; a negative or ZERO on failure
+ *
+ ****************************************************************************/
+
 int netlib_check_ipconnectivity(FAR const char *ip, int timeout, int retry);
+
+#ifdef CONFIG_NETUTILS_PING
+
+/****************************************************************************
+ * Name: netlib_check_ifconnectivity
+ *
+ * Description:
+ *   Check network connectivity by pinging the default gateway
+ *   of the specified network interface.
+ *
+ * Parameters:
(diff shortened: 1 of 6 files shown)


masc2008:
looks good to me.

jerpelea (dismissed):

include/netutils/netlib.h
```diff
@@ -561,12 +561,41 @@ int netlib_getifstatistics(FAR const char *ifname,
 int netlib_check_ifconflict(FAR const char *ifname);
 #endif
 
-#ifdef CONFIG_NETUTILS_PING
+/**
+ * @brief
```
acassis: Please fix, this is not the Description comment/decoration adopted by NuttX. See netutils/netlib/netlib_checkifconnectivity.c line 50 that you are editing for reference

include/netutils/netlib.h
```diff
@@ -561,12 +561,41 @@ int netlib_getifstatistics(FAR const char *ifname,
...
+ *   Check network connectivity by pinging the default gateway
+ *   of the specified network interface.
+ *
+ * @param  ifname  The name of the interface to use
+ * @param  timeout The timeout of ping
+ * @param  retry   The retry times of ping
+ *
+ * @return
+ *   nums of gateway reply of ping; a negative on failure.
+ */
```
acassis: Ditto

acassis (changes requested):

xiaoxiang781216 (dismissed):

include/netutils/netlib.h
zhhyu7: Done.

include/netutils/netlib.h
zhhyu7: Done.

acassis (approved):

xiaoxiang781216 (approved):

PR #3470, opened 2026-05-03
testing/ostest/wdog: replace sclock_t with clock_t

xiaoxiang781216:
Companion to apache/nuttx#18842.

The upstream nuttx change drops the `sclock_t` alias now that
`clock_t` is itself a signed 64-bit type (`int64_t`).  This patch is
the pure-mechanical rename of the only in-tree apps consumer
(`testing/ostest/wdog.c`).


linguini1 (changes requested):
Marking change request for PR template impact and testing info

PR #3504, opened 2026-05-26
system/nxinit: Add NuttX Init ("NxInit") reopened to be integrated on NuttX v13

acassis:
## Summary

Add NuttX Init ("NxInit")

This component (abbreviated as "NxInit") is specifically developed for NuttX and intended for system initialization. While we have adopted the Android Init Language among various syntax options, this is a brand-new implementation—it is not a port or variant of Android Init.

Refer to the implementation of Android Init Language, consists of five broad classes of statements: Actions, Commands, Services, Options, and Imports.

Actions support two types of triggers: event and action. Action triggers also support runtime triggering. Services support lifecycle management, including automatic restart (at specified intervals), and starting/stopping individually or by class. Import supports files or directories, and we may add a static method in the future. The following are some differences:

    The Android Init Language treats lines starting with # as comments, while we use a preprocessor to handle comments.
    For action commands, we can omit "exec" and directly execute built-in apps or nsh builtins.
    Regarding the property service, users can either adapt it by self or directly use the preset NVS-based properties.
    Only part of standard action commands and service options are implemented currlently.

To enable system/init:

-CONFIG_INIT_ENTRYPOINT="nsh_main"
+CONFIG_INIT_ENTRYPOINT="init_main"
+CONFIG_SYSTEM_NXINIT=y

## Impact

NuttX will be able to start applications and services without depending NSH and without using the app as NuttX entrypoint directly.

## Testing

```
   [    0.150000] [ 3] [ 0] init_main: == Dump Services ==
   ...
   [    0.160000] [ 3] [ 0] init_main: Service 0x40486aa8 name 'telnet' path 'telnetd'
   [    0.160000] [ 3] [ 0] init_main:   pid: 0
   [    0.160000] [ 3] [ 0] init_main:   arguments:
   [    0.160000] [ 3] [ 0] init_main:       [0] 'service'
   [    0.160000] [ 3] [ 0] init_main:       [1] 'telnet'
   [    0.160000] [ 3] [ 0] init_main:       [2] 'telnetd'
   [    0.160000] [ 3] [ 0] init_main:   classes:
   [    0.160000] [ 3] [ 0] init_main:     'test'
   [    0.170000] [ 3] [ 0] init_main:   restart_period: 3000
   [    0.170000] [ 3] [ 0] init_main:   reboot_on_failure: -1
   [    0.170000] [ 3] [ 0] init_main:   flags:
 > [    0.170000] [ 3] [ 0] init_main:     'oneshot'
   ...
   [    0.370000] [ 3] [ 0] init_main: starting service 'telnet' ...
   [    0.380000] [ 3] [ 0] init_main: service 'telnet' flag 0x2 add 0x4
   [    0.380000] [ 3] [ 0] init_main:   +flag 'running'
   [    0.380000] [ 3] [ 0] init_main: service 'telnet' flag 0x6 add 0x8
   [    0.380000] [ 3] [ 0] init_main:   -flag 'restarting'
   [    0.380000] [ 3] [ 0] init_main: service 'telnet' flag 0x6 add 0x1
   [    0.380000] [ 3] [ 0] init_main:   -flag 'disabled'
   [    0.380000] [ 3] [ 0] init_main: started service 'telnet' pid 9
   ...
   nsh> kill -9 9
   nsh> [    7.350000] [ 3] [ 0] init_main: service 'telnet' flag 0x6 add 0x4
   [    7.350000] [ 3] [ 0] init_main:   -flag 'running'
   [    7.350000] [ 3] [ 0] init_main: service 'telnet' flag 0x2 add 0x80000001
   [    7.350000] [ 3] [ 0] init_main:   +flag 'disabled'
   [    7.350000] [ 3] [ 0] init_main:   +flag 'remove'
   [    7.350000] [ 3] [ 0] init_main: service 'telnet' pid 9 exited status 1
 > [    7.360000] [ 3] [ 0] init_main: removing service 'telnet' ...
``

### DIFF
--- a/system/nxinit/Kconfig
+++ b/system/nxinit/Kconfig
@@ -1,0 +1,118 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config SYSTEM_NXINIT
+	tristate "NuttX Init"
+	default n
+	depends on LIBC_EXECFUNCS
+	depends on SCHED_CHILD_STATUS
+	---help---
+		Enable NxInit(NuttX Init) component for system initialization.
+		The script of NxInit(init.rc) is compatible with the Android Init Language syntax.
+
+if SYSTEM_NXINIT
+
+#
+# Basic
+#
+
+config SYSTEM_NXINIT_PRIORITY
+	int "Thread priority"
+	default 100
+
+config SYSTEM_NXINIT_STACKSIZE
+	int "Stack size"
+	default DEFAULT_TASK_STACKSIZE
+
+config SYSTEM_NXINIT_PROGNAME
+	string "Program name"
+	default "init"
+
+#
+# RC
+#
+
+config SYSTEM_NXINIT_RC_LINE_MAX
+	int "Max line length of RC file"
+	default 128
+	---help---
+		Maximum line length of RC file.
+		More details: https://android.googlesource.com/platform/system/core/+/master/init/README.md
+
+#
+# Action
+#
+
+config SYSTEM_NXINIT_ACTION_CMD_ARGS_MAX
+	int "Max number of command arguments"
+	default 8
+	---help---
+		Maximum number of command arguments.
+		Form:
+		```
+		on <trigger>
+		   <command>
+		   <command>
+		   <command>
+		   ...
+		```
+
+config SYSTEM_NXINIT_ACTION_MANAGER_EVENT_MAX
+	int "Max number of action manager events"
+	default 32
+	---help---
+		Maximum number of action manager events.
+		```
+		struct action_manager_s
+		{
+		  ...
+		  FAR char *events[CONFIG_SYSTEM_NXINIT_ACTION_MANAGER_EVENT_MAX];
+		  ...
+		};
+		```
+
+#
+# Service
+#
+
+config SYSTEM_NXINIT_SERVICE_ARGS_MAX
+	int "Max number of service arguments"
+	default 8
+	range 3 64
+	---help---
+		Maximum number of service arguments,
+		including "name", "pathname" and key word "service"(at least 3). Form:
+		```
+		service <name> <pathname> [ <argument> ]*
+		   <option>
+		   <option>
+		   ...
+		```
+
+config SYSTEM_NXINIT_SERVICE_RESTART_PERIOD
+	int "Service restart period in ms"
+	default 5000
+
+#
+# Log level
+#
+
+config SYSTEM_NXINIT_ERR
+	bool "Enable error log"
+	default !DEFAULT_SMALL
+
+config SYSTEM_NXINIT_WARN
+	bool "Enable warning log"
+	depends on SYSTEM_NXINIT_ERR
+
+config SYSTEM_NXINIT_INFO
+	bool "Enable info log"
+	depends on SYSTEM_NXINIT_WARN
+
+config SYSTEM_NXINIT_DEBUG
+	bool "Enable debug log"
+	depends on SYSTEM_NXINIT_INFO
+
+endif
--- a/system/nxinit/Kconfig
+++ b/system/nxinit/Kconfig
@@ -59,6 +59,13 @@ config SYSTEM_NXINIT_ACTION_CMD_ARGS_MAX
 		   ...
 		```
 
+config SYSTEM_NXINIT_ACTION_WARN_SLOW
+	int "Warn if command takes too long"
+	default 50
+	depends on SYSTEM_NXINIT_WARN
+	---help---
+		Warning if command took more than `SYSTEM_NXINIT_ACTION_WARN_SLOW` ms.
+
 config SYSTEM_NXINIT_ACTION_MANAGER_EVENT_MAX
 	int "Max number of action manager events"
 	default 32
--- a/system/nxinit/Kconfig
+++ b/system/nxinit/Kconfig
@@ -6,6 +6,7 @@
 config SYSTEM_NXINIT
 	tristate "NuttX Init"
 	default n
+	depends on EXPERIMENTAL
 	depends on LIBC_EXECFUNCS
 	depends on SCHED_CHILD_STATUS
 	---help---
--- a/system/nxinit/Kconfig
+++ b/system/nxinit/Kconfig
@@ -14,9 +14,7 @@ config SYSTEM_NXINIT
 
 if SYSTEM_NXINIT
 
-#
-# Basic
-#
+comment "NXInit Basic"
 
 config SYSTEM_NXINIT_PRIORITY
 	int "Thread priority"
@@ -30,9 +28,7 @@ config SYSTEM_NXINIT_PROGNAME
 	string "Program name"
 	default "init"
 
-#
-# RC
-#
+comment "NXInit Run Control(RC)"
 
 config SYSTEM_NXINIT_RC_LINE_MAX
 	int "Max line length of RC file"
@@ -41,9 +37,7 @@ config SYSTEM_NXINIT_RC_LINE_MAX
 		Maximum line length of RC file.
 		More details: https://android.googlesource.com/platform/system/core/+/master/init/README.md
 
-#
-# Action
-#
+comment "NXInit Action"
 
 config SYSTEM_NXINIT_ACTION_CMD_ARGS_MAX
 	int "Max number of command arguments"
@@ -60,7 +54,7 @@ config SYSTEM_NXINIT_ACTION_CMD_ARGS_MAX
 		```
 
 config SYSTEM_NXINIT_ACTION_WARN_SLOW
-	int "Warn if command takes too long"
+	int "Slow command warning timeout"
 	default 50
 	depends on SYSTEM_NXINIT_WARN
 	---help---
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

**Comment by acassis** [comment]
@simbit18 @lupyuen any idea what is this label error:

```
 label
Unhandled error: ReferenceError: keywords is not defined
```

File: system/nxinit/Kconfig [hunk 1]
```diff
@@ -0,0 +1,125 @@
...
+	depends on LIBC_EXECFUNCS
+	depends on SCHED_CHILD_STATUS
+	---help---
+		Enable NxInit(NuttX Init) component for system initialization.
+		The script of NxInit(init.rc) is compatible with the Android Init Language syntax.
+
+if SYSTEM_NXINIT
+
+#
+# Basic
```
linguini1: These should maybe be `comment "Basic"`, etc. to show up in the menu.

system/nxinit/Kconfig
```diff
@@ -0,0 +1,125 @@
...
+		```
+		on <trigger>
+		   <command>
+		   <command>
+		   <command>
+		   ...
+		```
+
+config SYSTEM_NXINIT_ACTION_WARN_SLOW
+	int "Warn if command takes too long"
```
linguini1: `int "Slow command warning timeout"` (slightly clearer to user)

system/nxinit/service.h
```diff
@@ -0,0 +1,144 @@
...
+#define SVC_REMOVE      (1 << 31)
+
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+struct service_class_s
+{
+  struct list_node node;     /* Service class list node */
+  FAR char name[0];
```
linguini1: What is 0 size array? Just a `char *`?

system/nxinit/service.c
```diff
@@ -0,0 +1,682 @@
...
+                                    int argc, FAR char **argv);
+#endif
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static const struct cmd_map_s g_option[] =
+{
+  {"class", 2, 99, option_class},
```
linguini1: Should max arguments not be limited by `SYSTEM_NXINIT_ACTION_CMD_ARGS_MAX`? Or is that for something else? 99 seems high

system/nxinit/init.h
```diff
@@ -0,0 +1,93 @@
...
+#include <syslog.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define TIMESPEC2MS(t) (((t).tv_sec * 1000) + (t).tv_nsec / 1000000)
+
+#ifdef CONFIG_SYSTEM_NXINIT_DEBUG
+#define init_debug(...) syslog(LOG_DEBUG, ##__VA_ARGS__)
```
linguini1: For all logs, can we add `| LOG_USER` to make it clear this is a user-space log?

system/nxinit/init.h
```diff
@@ -0,0 +1,93 @@
...
+#define init_warn(...)
+#endif
+
+#ifdef CONFIG_SYSTEM_NXINIT_ERR
+#define init_err(f, ...) syslog(LOG_ERR, "Error " f, ##__VA_ARGS__)
+#else
+#define init_err(...)
+#endif
+
+#define init_log(p, ...) \
```
linguini1, line 71: Why do we need this macro? We can just choose the appropriate other macro (`init_debug`, `init_err`, etc) for the log level.

system/nxinit/init.h
```diff
@@ -0,0 +1,93 @@
...
+#endif
+
+#ifdef CONFIG_SYSTEM_NXINIT_WARN
+#define init_warn(...) syslog(LOG_WARNING, ##__VA_ARGS__)
+#else
+#define init_warn(...)
+#endif
+
+#ifdef CONFIG_SYSTEM_NXINIT_ERR
+#define init_err(f, ...) syslog(LOG_ERR, "Error " f, ##__VA_ARGS__)
```
linguini1, line 66: Maybe splitting hairs, but I don't think we need to append "Error". Syslog can already do this "[ERROR]" prefix if users want. This will save some amount of space on strings (maybe negligible but still).

system/nxinit/init.c
```diff
@@ -0,0 +1,254 @@
...
+    {
+      goto out;
+    }
+
+  init_dump_actions(&am.actions);
+  init_dump_services(&sm.services);
+
+  init_action_add_event(&am, "boot");
+
+  boardctl(BOARDIOC_INIT, 0);
```
linguini1: Deprecated.

system/nxinit/init.c
```diff
@@ -0,0 +1,254 @@
...
+      ev[i].pfd = &pfds[i];
+      r = ev[i].init(&ev[i]);
+      if (r < 0)
+        {
+          init_err("Init event %zu", i);
+          goto out;
+        }
+    }
+
+  r = init_parse_config_file(parser, "/etc/init.d/init.rc");
```
linguini1, line 183: I think `/etc/init.d/init.rc` should be a configurable location.

Any way to bake init scripts into systems that don't want to add a whole file system? Maybe we can later add a build tool that compiles the script into a parse-able string instead?

linguini1 (changes requested):
I don't see anything really wrong with this, just some small comments.

Very excited for this in NuttX!

linguini1:
Also would be good to test on a couple devices!

system/nxinit/service.h
acassis: I think you found a BUG. I already saw structs with last item[0] to dynamically variable data, but I think this is not the case here. I create a MAX_NXINIT_SERVICE_NAME to handle it

system/nxinit/service.c
acassis: Ok, I created a NXINIT_ACTION_CMD_ARGS_MAX, but kept the value of 99. Since it is not allocated, not a big issue

system/nxinit/init.h
acassis: I will keep it this way, I think it was a particular developer taste to do this way instead of using an existing macro

system/nxinit/init.c
acassis: I think Xiaomi guys created an alternative to use, a kind of temporary FS to avoid requiring ROMFS for example. But I think have a standard place like /etc/init.d/init.rc is better, it avoids mess and fragmentation

xiaoxiang781216:
I would suggest that you don't copy nxinit because:

1. It is still in the active development
2. This old copy contain many bugs and lack many improvements

acassis:
> I would suggest that you don't copy nxinit because:
> 
>     1. It is still in the active development
> 
>     2. This old copy contain many bugs and lack many improvements

Hi @xiaoxiang781216 thank you for the warning, maybe you can share the existing fixes.

We could submit it as EXPERIMENTAL, this way more people will start to test it and will help to improve it. This way is better then developing in entirely on closed doors. Is it possible?

xiaoxiang781216:
> > I would suggest that you don't copy nxinit because:
> > ```
> > 1. It is still in the active development
> > 
> > 2. This old copy contain many bugs and lack many improvements
> > ```
> 
> Hi @xiaoxiang781216 thank you for the warning, maybe you can share the existing fixes.
> 
> We could submit it as EXPERIMENTAL, this way more people will start to test it and will help to improve it. This way is better then developing in entirely on closed doors. Is it possible?

it's better to download from openvela repo just like other 3rd party active project.

acassis:
> > > I would suggest that you don't copy nxinit because:
> > > ```
> > > 1. It is still in the active development
> > > 
> > > 2. This old copy contain many bugs and lack many improvements
> > > ```
> > 
> > 
> > Hi @xiaoxiang781216 thank you for the warning, maybe you can share the existing fixes.
> > We could submit it as EXPERIMENTAL, this way more people will start to test it and will help to improve it. This way is better then developing in entirely on closed doors. Is it possible?
> 
> it's better to download from openvela repo just like other 3rd party active project.

No, the idea is to make NXInit an integral part of NuttX, but a 3rd party.

So, I will change it to EXPERIMENTAL for now.

system/nxinit/service.c
linguini1: Hard to say without asking the original developer so I was just wondering!

system/nxinit/init.c
linguini1: So what will we do to use this program on boards without a file system?

Or, boards that want to mount an SD card like the Raspberry Pi, under a path like `/sd`? I think it's fine to use this value as a default but it must be configurable.

acassis: I think in this current stage it will require a persistent FS like ROMFS

linguini1: That's okay, it can be tackled in another PR. But I really insist that the init.rc path be configurable in this PR or it becomes very difficult to use on some devices.

acassis: I understood your point. We will make this customizable later, in the Kconfig.

system/nxinit/init.h
acassis: I think we can use existing macro, but I decided to keep the original code

linguini1 (approved):
For visibility, marking that init.rc path should be made configurable in the next patches

michallenc (approved):

cederom:
Ah just came back to the lab, but +1 from me, lets move forward making small steps, thank you @acassis :-)

lupyuen:
There's a typo in nxinit/Kconfig: `--help--` should be `---help---`. The fix is here:
- https://github.com/apache/nuttx-apps/pull/3507

PR #3511, opened 2026-05-29
system/popen: Avoid copying FILE

### LINKED ISSUE #2937: [BUG] system/popen memcpy FILE
### Description / Steps to reproduce the issue

for some reasons, system/popen memcpy a FILE.
it seems fragile at best to me.
it should not assume libc FILE implementation details.

### On which OS does this issue occur?

[OS: Mac]

### What is the version of your OS?

macOS 15.2

### NuttX Version

master

### Issue Architecture

[Arch: all]

### Issue Area

[Area: System]

### Verification

- [X] I have verified before submitting the report.

nightt5879:
﻿## Summary

Fixes #2937.

`system/popen` was embedding a copied `FILE` object in its private tracking container and copying it back in `pclose()`. That depends on libc `FILE` layout details.

This PR makes `popen()` return the actual stream from `fdopen()` and keeps a small private `FILE * -> pid` tracking list so `pclose()` can find the spawned shell process without copying `FILE`.

Scope:
- Does not change supported `popen()` modes.
- Does not change the spawn/file-action setup.
- Keeps `pclose()` responsible for closing the stream and waiting for the shell pid.
- Returns `EINVAL` if `pclose()` cannot find the stream in the private popen list.

## Impact

`popen()`/`pclose()` no longer rely on libc `FILE` internals.

- New feature: NO
- User adaptation required: NO
- Build process change: NO
- Hardware/architecture/board change: NO
- Documentation update required: NO
- Security impact: NO
- Compatibility impact: NO intended compatibility impact

## Testing

Host:
- Windows with WSL Ubuntu 24.04
- Target: `sim:nsh`

Checks:
- `git diff --check`: pass
- `checkpatch.sh -g HEAD~1..HEAD`: pass
- `sim:nsh` build with `CONFIG_SYSTEM_POPEN=y` and `CONFIG_EXAMPLES_POPEN=y`: pass
  - confirmed `CC: popen.c`
  - result: `SIM elf with dynamic libs archive in nuttx.tgz`
- `printf 'popen\npoweroff\n' | timeout 30s ./nuttx`: pass
  - confirmed `popen("help")` output is read
  - confirmed execution reaches `pclose()`

### DIFF
--- a/system/popen/popen.c
+++ b/system/popen/popen.c
@@ -37,16 +37,64 @@
  * Private Types
  ****************************************************************************/
 
-/* struct popen_file_s is a cast compatible version of FILE that contains
- * the additional PID of the shell processes needed by pclose().
- */
-
 struct popen_file_s
 {
-  FILE copy;
-  FILE *original;
+  int fd;
   pid_t shell;
 };
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: popen_file_read
+ ****************************************************************************/
+
+static ssize_t popen_file_read(FAR void *cookie, FAR char *buf,
+                               size_t size)
+{
+  FAR struct popen_file_s *filep = (FAR struct popen_file_s *)cookie;
+
+  return read(filep->fd, buf, size);
+}
+
+/****************************************************************************
+ * Name: popen_file_write
+ ****************************************************************************/
+
+static ssize_t popen_file_write(FAR void *cookie, FAR const char *buf,
+                                size_t size)
+{
+  FAR struct popen_file_s *filep = (FAR struct popen_file_s *)cookie;
+
+  return write(filep->fd, buf, size);
+}
+
+/****************************************************************************
+ * Name: popen_file_seek
+ ****************************************************************************/
+
+static off_t popen_file_seek(FAR void *cookie, FAR off_t *offset,
+                             int whence)
+{
+  set_errno(ESPIPE);
+  return ERROR;
+}
+
+/****************************************************************************
+ * Name: popen_file_close
+ ****************************************************************************/
+
+static int popen_file_close(FAR void *cookie)
+{
+  FAR struct popen_file_s *filep = (FAR struct popen_file_s *)cookie;
+  int ret;
+
+  ret = dpclose(filep->fd, filep->shell);
+  free(filep);
+  return ret < 0 ? ERROR : OK;
+}
 
 /****************************************************************************
  * Public Functions
@@ -93,6 +141,15 @@ struct popen_file_s
 FILE *popen(FAR const char *command, FAR const char *mode)
 {
   FAR struct popen_file_s *container;
+  FAR FILE *stream;
+  cookie_io_functions_t popen_io =
+    {
+      .read  = popen_file_read,
+      .write = popen_file_write,
+      .seek  = popen_file_seek,
+      .close = popen_file_close
+    };
+
   int oflag;
   int fd;
 
@@ -142,8 +199,9 @@ FILE *popen(FAR const char *command, FAR const char *mode)
 
   /* Wrap the raw fd in a FILE stream */
 
-  container->original = fdopen(fd, mode);
-  if (container->original == NULL)
+  container->fd = fd;
+  stream = fopencookie(container, mode, popen_io);
+  if (stream == NULL)
     {
       int errcode = errno;
       dpclose(fd, container->shell);
@@ -152,8 +210,7 @@ FILE *popen(FAR const char *command, FAR const char *mode)
       return NULL;
     }
 
-  memcpy(&container->copy, container->original, sizeof(FILE));
-  return &container->copy;
+  return stream;
 }
 
 /****************************************************************************
@@ -193,33 +250,12 @@ FILE *popen(FAR const char *command, FAR const char *mode)
  *   stream - The stream reference returned by a previous call to popen()
  *
  * Returned Value:
- *   The child termination status on success, or -1 (ERROR) on failure
- *   with errno set.
+ *   Zero (OK) is returned on success; otherwise -1 (ERROR) is returned and
+ *   errno is set appropriately.
  *
  ****************************************************************************/
 
 int pclose(FILE *stream)
 {
-  FAR struct popen_file_s *container = (FAR struct popen_file_s *)stream;
-  FILE *original;
-  pid_t shell;
-
-  original = container->original;
-
-  /* Set the state of the original file descriptor to the state of the
-   * working copy
-   */
-
-  memcpy(original, &container->copy, sizeof(FILE));
-
-  /* Then close the original and free the container (saving the PID of the
-   * shell process).  Pass -1 to dpclose since fclose already closed the fd.
-   */
-
-  fclose(original);
-
-  shell = container->shell;
-  free(container);
-
-  return dpclose(-1, shell);
+  return fclose(stream);
 }


acassis (dismissed):

acassis:
@nightt5879 it would be nice to have some test coverage to test it too.

nightt5879:
﻿Thanks, I added a small `testing/libc/popen` test in commit 23a698fde. @acassis 

The test opens two `popen(..., "r")` streams, verifies the output read from both streams, and closes the first stream before the second. That covers reading from the returned `popen()` stream and also exercises the private stream-to-pid lookup/removal path added by this PR when the stream is not the first entry in the tracking list.

I verified the added test with:

- `checkpatch.sh -g HEAD~1..HEAD`
- `sim:nsh` build with `CONFIG_SYSTEM_POPEN=y` and `CONFIG_TESTING_POPEN_TEST=y`
- `printf 'popen_test\npoweroff\n' | timeout 30s ./nuttx`

The simulator run prints `popen_test: successful`.

acassis:
> Thanks, I added a small `testing/libc/popen` test in commit [23a698f](https://github.com/apache/nuttx-apps/commit/23a698fde9de2f8b41c3e6868a415f5a13eed6d1). @acassis
> 
> The test opens two `popen(..., "r")` streams, verifies the output read from both streams, and closes the first stream before the second. That covers reading from the returned `popen()` stream and also exercises the private stream-to-pid lookup/removal path added by this PR when the stream is not the first entry in the tracking list.
> 
> I verified the added test with:
> 
>     * `checkpatch.sh -g HEAD~1..HEAD`
> 
>     * `sim:nsh` build with `CONFIG_SYSTEM_POPEN=y` and `CONFIG_TESTING_POPEN_TEST=y`
> 
>     * `printf 'popen_test\npoweroff\n' | timeout 30s ./nuttx`
> 
> 
> The simulator run prints `popen_test: successful`.

My bad, I looked only the first commit. So, everything is fine!

nightt5879:
> > Thanks, I added a small `testing/libc/popen` test in commit [23a698f](https://github.com/apache/nuttx-apps/commit/23a698fde9de2f8b41c3e6868a415f5a13eed6d1). @acassis
> > The test opens two `popen(..., "r")` streams, verifies the output read from both streams, and closes the first stream before the second. That covers reading from the returned `popen()` stream and also exercises the private stream-to-pid lookup/removal path added by this PR when the stream is not the first entry in the tracking list.
> > I verified the added test with:
> > ```
> > * `checkpatch.sh -g HEAD~1..HEAD`
> > 
> > * `sim:nsh` build with `CONFIG_SYSTEM_POPEN=y` and `CONFIG_TESTING_POPEN_TEST=y`
> > 
> > * `printf 'popen_test\npoweroff\n' | timeout 30s ./nuttx`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The simulator run prints `popen_test: successful`.
> 
> My bad, I looked only the first commit. So, everything is fine!

Just to clarify: I added the additional commit after seeing your suggestion.
Thanks for reviewing it again!

system/popen/popen.c
```diff
@@ -45,17 +46,78 @@
...
+  FAR FILE *stream;
   pid_t shell;
 };
 
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static sem_t g_popen_sem = SEM_INITIALIZER(1);
+static FAR struct popen_file_s *g_popen_files;
```
xiaoxiang781216: but it isn't good (or even worse) to add a global list to fix a minor issue.

nightt5879: Thanks for the review. I trust your judgment here.

The global list was added because once `popen()` returns the real `fdopen()` stream, `pclose(FILE *stream)` still needs a way to recover the shell pid for `waitpid()`.

The alternatives I can see are:
1. keep external `stream/fd -> pid` state, as this PR does;
2. store popen metadata in NuttX `FILE`/stdio internals;
3. keep the old copied-`FILE` approach;
4. make a broader libc/helper API change.

I chose (1) to avoid copying `FILE` or modifying libc internals, but I understand this may not be the tradeoff you want for this issue.

What approach would you suggest here?

xiaoxiang781216: @michallenc added fopencookie before, you may attach the private data to FILE by this function:
https://github.com/apache/nuttx/pulls?q=fopencookie

nightt5879: Thanks for the pointer. I followed this direction and reworked the implementation to use fopencookie(): popen now stores the pipe fd and shell pid in the stream's private cookie instead of a global FILE-to-pid list. The cookie callbacks forward read/write/close to the fd, and pclose() retrieves the pid from the cookie before fclose()/waitpid(). The popen test remains in the second commit and passes on sim:nsh.

system/popen/popen.c
```diff
@@ -399,31 +458,30 @@ FILE *popen(FAR const char *command, FAR const char *mode)
...
 
-  DEBUGASSERT(container != NULL && container->original != NULL);
-  original = container->original;
-
-  /* Set the state of the original file descriptor to the state of the
-   * working copy
-   */
-
-  memcpy(original, &container->copy, sizeof(FILE));
+  DEBUGASSERT(stream != NULL);
```
xiaoxiang781216: why debug here and check stream at line 470

system/popen/popen.c
```diff
@@ -399,31 +458,30 @@ FILE *popen(FAR const char *command, FAR const char *mode)
...
+    {
+      errno = EINVAL;
+      return ERROR;
+    }
 
   shell = container->shell;
-  free(container);
+  fclose(stream);
 
 #ifdef CONFIG_SCHED_WAITPID
```
xiaoxiang781216: why not move the follow code into popen_file_close and call fclose in pclose directly

system/popen/popen.c
xiaoxiang781216: change to errout_with_container

system/popen/popen.c
```diff
@@ -341,14 +396,18 @@ FILE *popen(FAR const char *command, FAR const char *mode)
   posix_spawnattr_destroy(&attr);
 
 errout_with_stream:
-  fclose(container->original);
+  fclose(stream);
+  container = NULL;
 
 errout_with_pipe:
   close(fd[0]);
   close(fd[1]);
```
xiaoxiang781216: double close fd too

system/popen/popen.c
```diff
@@ -341,14 +396,18 @@ FILE *popen(FAR const char *command, FAR const char *mode)
...
 
 errout_with_container:
-  free(container);
+  if (container != NULL)
```
xiaoxiang781216: remove the check, free will do it

nightt5879:
@xiaoxiang781216 Thanks, I addressed the latest review comments in commit 6a16a4d4d.

The follow-up changes:
- changed the invalid-mode path to jump to errout_with_container, since no pipe fd exists there;
- avoided closing the stream fd twice after fopencookie() has taken ownership;
- removed the redundant NULL check before free();
- moved fd close, waitpid(), and cookie free into popen_file_close(), so pclose() now calls fclose() directly.

One caveat I noticed while moving waitpid() into the fopencookie close callback: fclose() treats any non-OK close callback return as an EOF/error result. Because of that, popen_file_close() now returns OK after close() and waitpid() themselves succeed, and reports only close()/waitpid() failures as errors. Otherwise a normal nonzero shell termination status could be converted into an fclose() failure.

I re-tested with:
- checkpatch.sh -f apps/system/popen/popen.c
- sim:nsh build with CONFIG_SYSTEM_POPEN=y and CONFIG_TESTING_POPEN_TEST=y
- popen_test, which prints successful

system/popen/popen.c
```diff
@@ -97,11 +97,37 @@ static off_t popen_file_seek(FAR void *cookie, FAR off_t *offset,
...
+          errcode = errno;
+        }
+    }
+#endif
+
   free(filep);
-  return ret;
+  if (errcode != OK)
+    {
+      errno = errcode;
```
xiaoxiang781216: why need set errno manually?

system/popen/popen.c
```diff
@@ -397,17 +425,20 @@ FILE *popen(FAR const char *command, FAR const char *mode)
 
 errout_with_stream:
   fclose(stream);
-  container = NULL;
+  close(newfd[0]);
```
xiaoxiang781216: fclose will close newfd for us?

system/popen/popen.c
nightt5879: Done in 433405a4a. close() and waitpid() already leave errno on failure, so popen_file_close() now returns ERROR directly without saving/restoring errno manually.

system/popen/popen.c
nightt5879: fclose(stream) only closes retfd, which is owned by the returned FILE stream. newfd[0] is the child-side descriptor prepared for dup2(), so it is still owned by popen() on this error path and still needs to be closed separately. I kept the close and added a short comment to make that ownership split explicit.

xiaoxiang781216:
@nightt5879 please rebase your patch to fix the conflict, thanks.

nightt5879:
@xiaoxiang781216 Rebased onto the latest master and force-pushed the branch.

The conflict was resolved by keeping the upstream dpopen()/dpclose() implementation and the new testing/libc/popen test, then reapplying only the fopencookie() change in popen.c to avoid copying FILE.

I re-tested with:
- git diff --check upstream/master..HEAD
- checkpatch.sh -f ../apps/system/popen/popen.c
- sim:nsh build with CONFIG_SYSTEM_POPEN=y and CONFIG_TESTING_POPEN=y
- popen_test, which reports 0 failed

xiaoxiang781216 (approved):

xiaoxiang781216:
> @xiaoxiang781216 Rebased onto the latest master and force-pushed the branch.
> 
> The conflict was resolved by keeping the upstream dpopen()/dpclose() implementation and the new testing/libc/popen test, then reapplying only the fopencookie() change in popen.c to avoid copying FILE.
> 
> I re-tested with:
> 
> * git diff --check upstream/master..HEAD
> * checkpatch.sh -f ../apps/system/popen/popen.c
> * sim:nsh build with CONFIG_SYSTEM_POPEN=y and CONFIG_TESTING_POPEN=y
> * popen_test, which reports 0 failed

Thanks, the new version is much simple now.

xiaoxiang781216:
@acassis please review this patch

jerpelea (approved):

PR #3524, opened 2026-06-04
system/uorb: Show listener data without CONFIG_DEBUG_UORB

### LINKED ISSUE #3201: [BUG] uorb_listener doesn't report data unless DEBUG_UORB is true
### Description / Steps to reproduce the issue

Default configuration with `DEBUG_UORB` set to false
```
nsh> uorb_listener

Monitor objects num:1
object_name:sensor_baro, object_instance:0
```

Setting `DEBUG_UORB` to true
```
nsh> uorb_listener

Monitor objects num:1
object_name:sensor_baro, object_instance:0
sensor_baro(now:7352094):timestamp:7352065,pressure:1013.250000,temperature:15.000000
sensor_baro(now:7371640):timestamp:7371618,pressure:1013.250000,temperature:15.000000
sensor_baro(now:7402419):timestamp:7402395,pressure:1013.250000,temperature:15.000000
```

I would expect `uorb_listener` to always output data without being contained by DEBUG_UORB

### On which OS does this issue occur?

[OS: Mac]

### What is the version of your OS?

MacOs 15.6.1

### NuttX Version

releases/12.2

### Issue Architecture

[Arch: simulator]

### Issue Area

[Area: System]

### Host information

_No response_

### Verification

Zepp-Hanzj:
*Note: Please adhere to [Contributing Guidelines](https://github.com/apache/nuttx-apps/blob/master/CONTRIBUTING.md).*

## Summary

Currently, format strings (`o_format`), `orb_info()`, `orb_fprintf()`, and `orb_sscanf()` are guarded by `#ifdef CONFIG_DEBUG_UORB`. This prevents `uorb_listener` from displaying sensor data when debug output is disabled.

Introduce a new `CONFIG_UORB_FORMAT` Kconfig option to control whether format strings are compiled in. `UORB_LISTENER`, `UORB_GENERATOR`, and `DEBUG_UORB` all `select UORB_FORMAT` automatically, so format strings are included when any of these features are enabled. Users who don't need any of these features get no image size increase.

Fixes [apache/nuttx-apps#3201](https://github.com/apache/nuttx-apps/issues/3201).

## Impact

- Is new feature?: YES (new Kconfig option CONFIG_UORB_FORMAT)
- Impact on user?: YES (users can now see sensor data with `uorb_listener` without enabling `CONFIG_DEBUG_UORB`)
- Impact on build?: NO (only adds a new Kconfig option, no build system changes)
- Impact on hardware?: NO
- Impact on documentation?: NO
- Impact on security?: NO
- Impact on compatibility?: NO (existing configs with `CONFIG_DEBUG_UORB` or `CONFIG_UORB_LISTENER` continue to work as before)

## Testing

Host: Linux x86_64, sim:nsh

Config:
```
CONFIG_USENSOR=y
CONFIG_UORB=y
CONFIG_UORB_LISTENER=y
CONFIG_UORB_GENERATOR=y
CONFIG_SENSORS=y
CONFIG_SENSORS_FAKESENSOR=y
CONFIG_LINE_MAX=256
CONFIG_NSH_MAXARGUMENTS=16
```

Steps:
```
nsh> uorb_generator -n 100000 -r 10 -s -t sensor_accel0 timestamp:1,x:1,y:1,z:1,temperature:1 &
nsh> sleep 3
nsh> uorb_listener -T -l
```

Before fix (with CONFIG_DEBUG_UORB disabled):
```
current objects: 1
NAME                           INST #SUB RATE #Q SIZE
sensor_accel                    0    0    0  0   24
```

After fix:
```
current objects: 1
NAME                           INST #SUB RATE #Q SIZE
sensor_accel                    0    0    8  1   24
```

Also verified: with only `CONFIG_UORB` enabled (no LISTENER/GENERATOR/DEBUG_UORB), `CONFIG_UORB_FORMAT` is not selected and format strings are excluded from the build.

### DIFF
--- a/system/uorb/Kconfig
+++ b/system/uorb/Kconfig
@@ -18,12 +18,18 @@ config UORB_STACKSIZE
 	int "stack size"
 	default DEFAULT_TASK_STACKSIZE
 
+config UORB_FORMAT
+	bool
+	default n
+
 config UORB_LISTENER
 	bool "uorb listener"
+	select UORB_FORMAT
 	default n
 
 config UORB_GENERATOR
 	bool "uorb generator"
+	select UORB_FORMAT
 	default n
 
 config UORB_TESTS
@@ -46,6 +52,7 @@ endif # UORB_TESTS
 config DEBUG_UORB
 	bool "uorb debug output"
 	select LIBC_PRINT_EXTENSION
+	select UORB_FORMAT
 	depends on LIBC_FLOATINGPOINT && SENSORS_USE_FLOAT
 	default n
 
--- a/system/uorb/sensor/step_counter.c
+++ b/system/uorb/sensor/step_counter.c
@@ -30,7 +30,7 @@
  * Private Functions
  ****************************************************************************/
 
-#ifdef CONFIG_DEBUG_UORB
+#ifdef CONFIG_UORB_FORMAT
 static const char sensor_step_counter_format[] =
   "timestamp:%" PRIu64 ",event:%" PRIu32 ",cadence:%" PRIu32 "";
 #endif
--- a/system/uorb/listener.c
+++ b/system/uorb/listener.c
@@ -49,7 +49,7 @@
 #define ORB_TOP_WAIT_TIME  1000
 #define ORB_DATA_DIR       "/data/uorb/"
 
-#if defined(CONFIG_DEBUG_UORB) && !defined(CONFIG_LIBC_FLOATINGPOINT)
+#if defined(CONFIG_UORB_FORMAT) && !defined(CONFIG_LIBC_FLOATINGPOINT)
 #error "Enable CONFIG_LIBC_FLOATINGPOINT, required to see debug output"
 #endif
 
@@ -551,7 +551,7 @@ static int listener_print(FAR const struct orb_metadata *meta, int fd)
   int ret;
 
   ret = orb_copy(meta, fd, buffer);
-#ifdef CONFIG_DEBUG_UORB
+#ifdef CONFIG_UORB_FORMAT
   if (ret == OK && meta->o_format != NULL)
     {
       orb_info(meta->o_format, meta->o_name, buffer);
@@ -774,7 +774,7 @@ static int listener_record(FAR const struct orb_metadata *meta, int fd,
   int ret;
 
   ret = orb_copy(meta, fd, buffer);
-#ifdef CONFIG_DEBUG_UORB
+#ifdef CONFIG_UORB_FORMAT
   if (ret == OK && meta->o_format != NULL)
     {
       ret = orb_fprintf(file, meta->o_format, buffer);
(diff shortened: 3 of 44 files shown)


system/uorb/sensor/step_counter.c
```diff
@@ -30,10 +30,8 @@
  * Private Functions
  ****************************************************************************/
 
-#ifdef CONFIG_DEBUG_UORB
```
xiaoxiang781216, line 33: but it will increase the image size and no method to remove it.

RogerNext: You're right, I didn't think this through carefully. Removing all #ifdef CONFIG_DEBUG_UORB guards unconditionally will indeed increase the image size with no way to opt out.

Here's a revised approach I'd like to propose:

Introduce a new CONFIG_UORB_FORMAT option to control whether format strings are compiled in:

1.Add config UORB_FORMAT in system/uorb/Kconfig
2.CONFIG_DEBUG_UORB will select UORB_FORMAT (so debug output still works as before)
3.CONFIG_UORB_LISTENER will also select UORB_FORMAT (so listener can display data)
4.All #ifdef CONFIG_DEBUG_UORB guards around format strings, o_format field, orb_info()/orb_fprintf()/orb_sscanf() will be changed to #ifdef CONFIG_UORB_FORMAT

This way:

1.Debug + listener users: format strings are automatically included (same as before)
2.Neither debug nor listener users: format strings are excluded, no image size increase
3.Users can also manually disable UORB_FORMAT even with listener enabled, if they prefer smaller images and don't need formatted output

The changes would be ~6 locations in uORB.h, uORB.c, listener.c, Kconfig, plus 43 sensor files (simple CONFIG_DEBUG_UORB → CONFIG_UORB_FORMAT replacement).

Would this approach be acceptable?

xiaoxiang781216: yes, it's a better approach.

linguini1:
> Verified on sim:nsh with CONFIG_DEBUG_UORB disabled

How? There is no uORB accelerometer on `sim:nsh`

RogerNext:
Good catch. Fixed by adding `select UORB_FORMAT` to `UORB_GENERATOR` in Kconfig. Generator needs format strings to parse input, so it can't work without them.

Also updated the PR to follow the template. Updated code and PR body are now pushed.

RogerNext:
We use `CONFIG_SENSORS_FAKESENSOR=y` which provides a fake accelerometer sensor on sim. Then use `uorb_generator` to publish data to it.

system/uorb/Kconfig
```diff
@@ -18,12 +18,18 @@ config UORB_STACKSIZE
 	int "stack size"
 	default DEFAULT_TASK_STACKSIZE
 
+config UORB_FORMAT
+	bool "uorb format strings"
```
xiaoxiang781216:
```suggestion
	bool
```

xiaoxiang781216 (approved):

jerpelea (approved):

PR #3532, opened 2026-06-08
netutils/dropbear: initial Dropbear SSH server port for NuttX

FelipeMdeO:
## Summary

This PR adds two related changes that together bring up an SSH server
on the ESP32-C3 DevKit board using the Dropbear application:

**boards/risc-v/esp32c3/esp32c3-devkit/configs/dropbear**

A new `dropbear` defconfig is introduced for the ESP32-C3 DevKit board.
It wires up the Dropbear SSH server application together with:

- Wi-Fi STA mode with DHCP client (WAPI tooling included for link
  bring-up at boot).
- urandom device for key material.
- SPIFFS on SPI flash (`/data` mountpoint) to persist the host key and
  the password database.
- `FSUTILS_PASSWD` pointing to `/data/passwd` as the credential store,
  replacing a previous Dropbear-specific password-file path.
- ECDSA host key stored at `/data/dropbear_ecdsa_host_key`.
- NSH autostart of the `dropbear` task on every boot.
- PTY support and Ctrl-C signal delivery enabled for interactive
  sessions.
- `CONFIG_NETUTILS_DROPBEAR_STACKSIZE` pinned to 65536 bytes; the
  default 32 KiB overflows during key exchange on this RISC-V target.
- `CONFIG_NETUTILS_DROPBEAR_LISTEN_RETRY_MAX=120` so the daemon keeps
  retrying until the Wi-Fi link is fully up.

Wi-Fi credentials (`myssid` / `mypasswd`) are placeholders and must be
set via `menuconfig` before flashing.

**crypto: expose ChaCha20 stream helpers**

Dropbear uses the `chacha20-poly1305@openssh.com` cipher, which requires
a stateful, multi-call ChaCha20 stream interface rather than the single-
block interface currently exposed by `crypto/chachapoly.c`. Three helpers
and a context struct are added:

- `struct chacha20_stream_ctx` — opaque wrapper around `chacha_ctx`.
- `chacha20_stream_setkey()` — initialise the key.
- `chacha20_stream_ivctr64()` — set IV and 64-bit counter.
- `chacha20_stream_crypt()` — encrypt/decrypt an arbitrary-length buffer.

All three functions are thin wrappers over the existing `chacha_*`
primitives; no new algorithm code is introduced.

## Impact

- **New board configuration**: the `dropbear` defconfig is additive and
  does not affect any existing configuration.
- **New public API**: three functions and one struct are added to
  `include/crypto/chachapoly.h`. The change is purely additive; existing
  users of `chacha20_setkey` / `chacha20_crypt` are unaffected.
- **Build**: no impact on boards or configurations that do not select
  `CONFIG_NETUTILS_DROPBEAR`.
- **Security**: host keys and credentials live on a SPIFFS partition
  under `/data`; they are generated at first run and persist across
  reboots. Wi-Fi credentials must be provisioned by the user before
  flashing.

## Testing

**Host:** Linux x86_64, GCC RISC-V toolchain
**Board:** ESP32-C3 DevKit (rev 0.4)

Build:
```
./tools/configure.sh esp32c3-devkit:dropbear
make -j$(nproc)
make flash ESPTOOL_PORT=/dev/ttyUSB0
```

First-time user provisioning (serial console):

The NuttX passwd file lives on SPIFFS (`/data/passwd`) and is empty on a
fresh flash. Before the first SSH login, create a user from the NSH
serial console:
```
nsh> useradd root <password>
```
The ECDSA host key is generated automatically on first boot.

Boot log shows Dropbear listening after Wi-Fi association:
```
NuttShell (NSH) NuttX-12.6.0
nsh> loaded ECDSA P-256 host key from /data/dropbear_ecdsa_host_key
     using NuttX passwd auth at /data/passwd
     dropbear: listening on port 2222
```

SSH connection from the host:
```
$ ssh -p 2222 root@<board-ip>
root@<board-ip>'s password: <password>
NuttShell (NSH) NuttX-12.6.0
nsh>
```

### DIFF
--- a/netutils/dropbear/Kconfig
+++ b/netutils/dropbear/Kconfig
@@ -1,0 +1,120 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+menuconfig NETUTILS_DROPBEAR
+	tristate "Dropbear SSH server"
+	default n
+	depends on NET && NET_TCP
+	depends on !DISABLE_PSEUDOFS_OPERATIONS
+	depends on !DISABLE_PTHREAD
+	depends on SCHED_WAITPID
+	depends on NSH_LIBRARY
+	depends on FSUTILS_PASSWD
+	depends on PSEUDOTERM
+	depends on SERIAL
+	depends on ARCH_HAVE_RNG
+	depends on DEV_RANDOM
+	depends on DEV_URANDOM
+	depends on LIBC_NETDB
+	depends on LIBC_GAISTRERROR
+	select CRYPTO
+	select CRYPTO_RANDOM_POOL
+	---help---
+		Enable a minimal Dropbear SSH server port for NuttX.  This initial
+		port is based on the ESP-IDF MCU test port and provides a single
+		foreground SSH server process with SSH sessions backed by NSH.
+
+if NETUTILS_DROPBEAR
+
+config NETUTILS_DROPBEAR_STACKSIZE
+	int "Dropbear main stack size"
+	default 65536
+	---help---
+		Stack size for the Dropbear server built-in.
+		This is architecture-specific, so adjust it according to your setup.
+
+config NETUTILS_DROPBEAR_PRIORITY
+	int "Dropbear main priority"
+	default 100
+
+config NETUTILS_DROPBEAR_SHELL_PRIORITY
+	int "Dropbear NSH session priority"
+	default 100
+
+config NETUTILS_DROPBEAR_PROGNAME
+	string "Dropbear program name"
+	default "dropbear"
+	---help---
+		This is the name of the program that will be used when the NSH ELF
+		program is installed.
+
+config NETUTILS_DROPBEAR_LISTEN_RETRIES
+	int "Dropbear listen retries"
+	default 0
+	---help---
+		Number of times to retry listen setup when no listen socket could
+		be opened.  Zero means to retry forever.
+
+config NETUTILS_DROPBEAR_LISTEN_RETRY_MAX
+	int "Dropbear maximum listen retry interval"
+	default 120
+	range 1 3600
+	---help---
+		Maximum number of seconds to wait between listen setup retries.
+		The retry delay starts at one second and doubles until it reaches
+		this value.
+
+config NETUTILS_DROPBEAR_SHELL_STACKSIZE
+	int "Dropbear NSH session task stack size"
+	default 8192
+
+config NETUTILS_DROPBEAR_PORT
+	int "Dropbear listen port"
+	default 2222
+
+config NETUTILS_DROPBEAR_HOSTKEY_PATH
+	string "Dropbear ECDSA P-256 host key path"
+	default "/etc/dropbear/dropbear_ecdsa_host_key"
+	---help---
+		Path to the persistent ECDSA P-256 host key used by the Dropbear
+		server.  The file is stored in Dropbear's native host key format.
+
+config NETUTILS_DROPBEAR_GENERATE_HOSTKEY
+	bool "Generate host key if missing"
+	default y
+	---help---
+		Pass -R so Dropbear generates an ECDSA P-256 host key on demand and
+		persists it at NETUTILS_DROPBEAR_HOSTKEY_PATH when it does not exist.
+		Product builds can disable this and provision the host key
+		externally (loaded with -r).
+
+config NETUTILS_DROPBEAR_COMPRESSION
+	bool "Enable SSH compression (zlib)"
+	default n
+	depends on LIB_ZLIB
+	---help---
+		Enable zlib compression for SSH sessions.  Requires the zlib
+		library (LIB_ZLIB).  When disabled, Dropbear is built with
+		DISABLE_ZLIB and negotiates no compression.
+
+		WARNING: each session allocates a zlib deflate state of about
+		256 KiB (DROPBEAR_ZLIB_WINDOW_BITS=15, DROPBEAR_ZLIB_MEM_LEVEL=8),
+		and the state is allocated even for the delayed zlib@openssh.com
+		method, right after key exchange.
+
+config NETUTILS_DROPBEAR_SYSLOG
+	bool "Log via syslog"
+	default n
+	---help---
+		Route Dropbear log messages through syslog().  When disabled,
+		Dropbear is built with DISABLE_SYSLOG.
+
+config NETUTILS_DROPBEAR_COMMIT
+	string "Dropbear upstream commit"
+	default "54ef47adf8c99b422be6a8f694f2866e62f88b9e"
+	---help---
+		Upstream Dropbear (mkj/dropbear) commit to download and build.
+
+endif
--- a/netutils/dropbear/dropbear_main.c
+++ b/netutils/dropbear/dropbear_main.c
@@ -1,0 +1,329 @@
+/****************************************************************************
+ * apps/netutils/dropbear/dropbear_main.c
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <sys/socket.h>
+#include <setjmp.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "includes.h"
+#include "algo.h"
+#include "crypto_desc.h"
+#define dropbear_main dropbear_multi_entry
+#include "dbutil.h"
+#undef dropbear_main
+#include "dbrandom.h"
+#include "netio.h"
+#include "runopts.h"
+#include "session.h"
+#include "signkey.h"
+#include "gensignkey.h"
+#include "ssh.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define DROPBEAR_PORT_STRING_HELPER(n) #n
+#define DROPBEAR_PORT_STRING(n) DROPBEAR_PORT_STRING_HELPER(n)
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+typedef void (*dropbear_exit_handler_t)(int exitcode, FAR const char *format,
+                                        va_list param) ATTRIB_NORETURN;
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static jmp_buf g_session_exit_jmp;
+static int g_session_exitcode;
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+static void dropbear_session_exit(int exitcode, FAR const char *format,
+                                  va_list param) noreturn_function;
+static void dropbear_session_exit(int exitcode, FAR const char *format,
+                                  va_list param)
+{
+  char exitmsg[150];
+  char fullmsg[300];
+  char fromaddr[60];
+  int signal_pipe[2];
+
+  vsnprintf(exitmsg, sizeof(exitmsg), format, param);
+
+  fromaddr[0] = '\0';
+
+  if (svr_ses.addrstring != NULL)
+    {
+      snprintf(fromaddr, sizeof(fromaddr), " from <%s>", svr_ses.addrstring);
+    }
+
+  if (!ses.init_done)
+    {
+      snprintf(fullmsg, sizeof(fullmsg), "Early exit%s: %s", fromaddr,
+               exitmsg);
+    }
+  else if (ses.authstate.authdone)
+    {
+      snprintf(fullmsg, sizeof(fullmsg), "Exit (%s)%s: %s",
+               ses.authstate.pw_name, fromaddr, exitmsg);
+    }
+  else if (ses.authstate.pw_name != NULL)
+    {
+      snprintf(fullmsg, sizeof(fullmsg),
+               "Exit before auth%s: (user '%s', %u fails): %s",
+               fromaddr, ses.authstate.pw_name, ses.authstate.failcount,
+               exitmsg);
+    }
+  else
+    {
+      snprintf(fullmsg, sizeof(fullmsg), "Exit before auth%s: %s", fromaddr,
+               exitmsg);
+    }
+
+  dropbear_log(LOG_INFO, "%s", fullmsg);
+
+  signal_pipe[0] = ses.signal_pipe[0];
+  signal_pipe[1] = ses.signal_pipe[1];
+  session_cleanup();
+
+  if (signal_pipe[0] > STDERR_FILENO)
+    {
+      m_close(signal_pipe[0]);
+    }
+
+  if (signal_pipe[1] > STDERR_FILENO && signal_pipe[1] != signal_pipe[0])
+    {
+      m_close(signal_pipe[1]);
+    }
+
+  memset(&ses, 0, sizeof(ses));
+  memset(&svr_ses, 0, sizeof(svr_ses));
+
+  g_session_exitcode = exitcode;
+  longjmp(g_session_exit_jmp, 1);
+
+  while (1)
+    {
+    }
+}
+
+static void dropbear_setup(FAR const char *port)
+{
+  /* Load the host key with -r so it lives in the persistent (epoch 0)
+   * allocation rather than being created lazily inside a per-session malloc
+   * epoch (-R), which would free it when the first session ends and leave a
+   * dangling svr_opts.hostkey for the next connection.
+   */
+
+  FAR char *argv[] =
+  {
+    "dropbear",
+    "-F",
+    "-p",
+    (FAR char *)port,
+    "-r",
+    CONFIG_NETUTILS_DROPBEAR_HOSTKEY_PATH,
+    NULL
+  };
+
+  _dropbear_exit = svr_dropbear_exit;
+  _dropbear_log = svr_dropbear_log;
+
+  disallow_core();
+  svr_getopts(6, argv);
+  seedrandom();
+  crypto_init();
+
+#ifdef CONFIG_NETUTILS_DROPBEAR_GENERATE_HOSTKEY
+  /* Generate the ECDSA host key now if it is missing.  signkey_generate()
+   * with skip_exist=1 silently succeeds when the file already exists.
+   */
+
+  if (signkey_generate(DROPBEAR_SIGNKEY_ECDSA_NISTP256, 0,
+                       CONFIG_NETUTILS_DROPBEAR_HOSTKEY_PATH, 1)
+      == DROPBEAR_FAILURE)
+    {
+      dropbear_exit("failed to generate host key");
+    }
+#endif
+
+  load_all_hostkeys();
+
+  if (dropbear_auth_initialize() < 0)
+    {
+      dropbear_exit("failed to initialize password auth");
+    }
+}
+
+static void dropbear_run_session(int childsock)
+{
+  dropbear_exit_handler_t saved_exit;
+
+  saved_exit = _dropbear_exit;
+  _dropbear_exit = dropbear_session_exit;
+
+  m_malloc_set_epoch(1);
+
+  if (setjmp(g_session_exit_jmp) == 0)
+    {
+      svr_session(childsock, -1);
+    }
+
+  m_malloc_free_epoch(1, 1);
+  m_malloc_set_epoch(0);
+
+  _dropbear_exit = saved_exit;
+
+  if (g_session_exitcode != EXIT_SUCCESS)
+    {
+      dropbear_log(LOG_WARNING, "session exited with status %d",
+                   g_session_exitcode);
+    }
+}
+
+static size_t dropbear_listen_sockets(FAR int *socks, size_t sockcount,
+                                      FAR int *maxfd)
+{
+  FAR char *errstring = NULL;
+  size_t sockpos = 0;
+  unsigned int i;
+
+  for (i = 0; i < svr_opts.portcount; i++)
+    {
+      int nsock;
+      unsigned int n;
+
+      nsock = dropbear_listen(svr_opts.addresses[i], svr_opts.ports[i],
+                              &socks[sockpos], sockcount - sockpos,
+                              &errstring, maxfd, svr_opts.interface);
+
+      if (nsock < 0)
+        {
+          dropbear_log(LOG_WARNING, "failed listening on '%s': %s",
+                       svr_opts.ports[i],
+                       errstring != NULL ? errstring : "unknown error");
+          m_free(errstring);
+          errstring = NULL;
+          continue;
+        }
+
+      for (n = 0; n < (unsigned int)nsock; n++)
+        {
+          set_sock_priority(socks[sockpos + n], DROPBEAR_PRIO_LOWDELAY);
+        }
+
+      sockpos += (size_t)nsock;
+    }
+
+  return sockpos;
+}
+
+static size_t dropbear_wait_listen_sockets(FAR int *socks, size_t sockcount,
+                                           FAR int *maxfd)
+{
+  int retries = 0;
+  int retry_delay = 1;
+
+  while (1)
+    {
+      size_t count;
+
+      *maxfd = -1;
+      count = dropbear_listen_sockets(socks, sockcount, maxfd);
+
+      if (count > 0)
+        {
+          return count;
+        }
+
+#if CONFIG_NETUTILS_DROPBEAR_LISTEN_RETRIES > 0
+      if (retries++ >= CONFIG_NETUTILS_DROPBEAR_LISTEN_RETRIES)
+        {
+          dropbear_exit("no listening ports available");
+        }
+#else
+      retries++;
+#endif
+
+      dropbear_log(LOG_WARNING, "Retry %d in %d seconds...", retries,
+                   retry_delay);
+      sleep(retry_delay);
+
+      retry_delay *= 2;
+
+      if (retry_delay > CONFIG_NETUTILS_DROPBEAR_LISTEN_RETRY_MAX)
+        {
+          retry_delay = CONFIG_NETUTILS_DROPBEAR_LISTEN_RETRY_MAX;
+        }
+    }
+}
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+int main(int argc, FAR char *argv[])
+{
+  FAR const char *port = DROPBEAR_PORT_STRING(CONFIG_NETUTILS_DROPBEAR_PORT);
+  int listensocks[MAX_LISTEN_ADDR];
+  int maxfd = -1;
+
+  if (argc > 1)
+    {
+      port = argv[1];
+    }
+
+  dropbear_setup(port);
+
+  dropbear_wait_listen_sockets(listensocks, MAX_LISTEN_ADDR, &maxfd);
+
+  printf("dropbear: listening on port %s\n", port);
+
+  while (1)
+    {
+      struct sockaddr_storage remoteaddr;
+      socklen_t remoteaddrlen = sizeof(remoteaddr);
+      FAR char *remote_host = NULL;
+      FAR char *remote_port = NULL;
+      int childsock;
+
+      childsock = accept(listensocks[0], (FAR struct sockaddr *)&remoteaddr,
+                         &remoteaddrlen);
+
+      if (childsock < 0)
+        {
+          continue;
+        }
+
+      getaddrstring(&remoteaddr, &remote_host, &remote_port, 0);
+      dropbear_log(LOG_INFO, "connection from %s:%s",
+                   remote_host != NULL ? remote_host : "?",
+                   remote_port != NULL ? remote_port : "?");
+      m_free(remote_host);
+      m_free(remote_port);
+
+      seedrandom();
+      dropbear_run_session(childsock);
+      close(childsock);
+    }
+
+  return EXIT_SUCCESS;
+}
--- a/netutils/dropbear/patch/0002-use-nuttx-passwd-auth.patch
+++ b/netutils/dropbear/patch/0002-use-nuttx-passwd-auth.patch
@@ -1,0 +1,87 @@
+--- a/src/svr-authpasswd.c
++++ b/src/svr-authpasswd.c
+@@ -33,6 +33,77 @@
+ 
+ #if DROPBEAR_SVR_PASSWORD_AUTH
+ 
++#if DROPBEAR_NUTTX_PASSWD
++
++/* Process a password auth request, sending success or failure messages as
++ * appropriate */
++void svr_auth_password(int valid_user) {
++
++	char * password = NULL;
++	unsigned int passwordlen;
++	unsigned int changepw;
++	int auth_ok = 0;
++
++	/* check if client wants to change password */
++	changepw = buf_getbool(ses.payload);
++	if (changepw) {
++		/* not implemented by this server */
++		send_msg_userauth_failure(0, 1);
++		return;
++	}
++
++	password = buf_getstring(ses.payload, &passwordlen);
++	if (valid_user && passwordlen <= DROPBEAR_MAX_PASSWORD_LEN &&
++			strlen(password) == passwordlen) {
++		auth_ok = dropbear_verify_password(ses.authstate.pw_name, password);
++	}
++	m_burn(password, passwordlen);
++	m_free(password);
++
++	/* After we have got the payload contents we can exit if the username
++	is invalid. Invalid users have already been logged. */
++	if (!valid_user) {
++		send_msg_userauth_failure(0, 1);
++		return;
++	}
++
++	if (passwordlen > DROPBEAR_MAX_PASSWORD_LEN) {
++		dropbear_log(LOG_WARNING,
++				"Too-long password attempt for '%s' from %s",
++				ses.authstate.pw_name,
++				svr_ses.addrstring);
++		send_msg_userauth_failure(0, 1);
++		return;
++	}
++
++	if (auth_ok == DROPBEAR_SUCCESS) {
++		if (svr_opts.multiauthmethod && (ses.authstate.authtypes & ~AUTH_TYPE_PASSWORD)) {
++			/* successful password authentication, but extra auth required */
++			dropbear_log(LOG_NOTICE,
++					"Password auth succeeded for '%s' from %s, extra auth required",
++					ses.authstate.pw_name,
++					svr_ses.addrstring);
++			ses.authstate.authtypes &= ~AUTH_TYPE_PASSWORD; /* password auth ok, delete the method flag */
++			send_msg_userauth_failure(1, 0);  /* Send partial success */
++		} else {
++			/* successful authentication */
++			dropbear_log(LOG_NOTICE,
++					"Password auth succeeded for '%s' from %s",
++					ses.authstate.pw_name,
++					svr_ses.addrstring);
++			send_msg_userauth_success();
++		}
++	} else {
++		dropbear_log(LOG_WARNING,
++				"Bad password attempt for '%s' from %s",
++				ses.authstate.pw_name,
++				svr_ses.addrstring);
++		send_msg_userauth_failure(0, 1);
++	}
++}
++
++#else
++
+ /* not constant time when strings are differing lengths. 
+  string content isn't leaked, and crypt hashes are predictable length. */
+ static int constant_time_strcmp(const char* a, const char* b) {
+@@ -131,4 +202,6 @@
+ 	}
+ }
+ 
++#endif /* DROPBEAR_NUTTX_PASSWD */
++
+ #endif
--- a/netutils/dropbear/port/config.h
+++ b/netutils/dropbear/port/config.h
@@ -1,0 +1,20 @@
+/****************************************************************************
+ * apps/netutils/dropbear/port/config.h
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ****************************************************************************/
+
+#ifndef __APPS_NETUTILS_DROPBEAR_PORT_CONFIG_H
+#define __APPS_NETUTILS_DROPBEAR_PORT_CONFIG_H
+
+/* Dropbear looks for a file named config.h. Keep this wrapper name stable;
+ * NuttX-specific autoconf replacements live in nuttx_config.h.
+ */
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include "nuttx_config.h"
+
+#endif /* __APPS_NETUTILS_DROPBEAR_PORT_CONFIG_H */
--- a/netutils/dropbear/port/localoptions.h
+++ b/netutils/dropbear/port/localoptions.h
@@ -1,0 +1,81 @@
+/****************************************************************************
+ * apps/netutils/dropbear/port/localoptions.h
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ****************************************************************************/
+
+#ifndef __APPS_NETUTILS_DROPBEAR_PORT_LOCALOPTIONS_H
+#define __APPS_NETUTILS_DROPBEAR_PORT_LOCALOPTIONS_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define DROPBEAR_TRACKING_MALLOC 1
+
+/* Persist the native Dropbear ECDSA host key at the configured path. */
+
+#define ECDSA_PRIV_FILENAME CONFIG_NETUTILS_DROPBEAR_HOSTKEY_PATH
+
+#define DROPBEAR_SVR_DROP_PRIVS 0
+
+#ifdef CONFIG_SCHED_USER_IDENTITY
+#  define DROPBEAR_SVR_MULTIUSER 1
+#else
+#  define DROPBEAR_SVR_MULTIUSER 0
+#endif
+
+#define DROPBEAR_SVR_PASSWORD_AUTH 1
+#define DROPBEAR_SVR_PUBKEY_AUTH 0
+#define DROPBEAR_SVR_PUBKEY_OPTIONS 0
+
+#define DROPBEAR_REEXEC 0
+#define DROPBEAR_SMALL_CODE 1
+#define DROPBEAR_USER_ALGO_LIST 0
+
+#define DROPBEAR_X11FWD 0
+#define DROPBEAR_SVR_AGENTFWD 0
+#define DROPBEAR_SVR_LOCALTCPFWD 0
+#define DROPBEAR_SVR_REMOTETCPFWD 0
+#define DROPBEAR_SVR_LOCALSTREAMFWD 0
+#define DROPBEAR_SVR_REMOTESTREAMFWD 0
+
+#define DROPBEAR_DSS 0
+#define DROPBEAR_RSA 0
+#define DROPBEAR_ECDSA 1
+#define DROPBEAR_ED25519 0
+#define DROPBEAR_SK_KEYS 0
+#define DROPBEAR_ECC_256 1
+#define DROPBEAR_ECC_384 0
+#define DROPBEAR_ECC_521 0
+
+#define DROPBEAR_AES128 1
+#define DROPBEAR_AES256 0
+#define DROPBEAR_CHACHA20POLY1305 1
+#define DROPBEAR_ENABLE_CBC_MODE 0
+#define DROPBEAR_ENABLE_GCM_MODE 0
+
+#define DROPBEAR_SHA1_HMAC 0
+#define DROPBEAR_SHA2_256_HMAC 1
+#define DROPBEAR_SHA2_512_HMAC 0
+#define DROPBEAR_SHA1_96_HMAC 0
+
+#define DROPBEAR_CURVE25519 1
+#define DROPBEAR_DH_GROUP14_SHA1 0
+#define DROPBEAR_DH_GROUP14_SHA256 0
+#define DROPBEAR_DH_GROUP16 0
+#define DROPBEAR_DH_GROUP1 0
+#define DROPBEAR_ECDH 0
+#define DROPBEAR_SNTRUP761 0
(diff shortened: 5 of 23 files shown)


acassis:
@FelipeMdeO nice work! Kudos!!!

Some suggestions to improve this PR:
* many of these patches are just to fix the UNUSED()s, moving it from function parameters to inside at the end of the function, please talk with the original author, maybe he accepts to move it to the end of the function;

nshlib/nsh_dropbear.c
```diff
@@ -26,45 +26,51 @@
...
-  if (buf == NULL || len == 0)
-    {
-      return lrc;
-    }
+  DEBUGASSERT(pstate != NULL);
 
-  for (i = 0; i < len; i++)
+  ninfo("Starting the Dropbear SSH server\n");
+
+  ret = nsh_parse(&pstate->cn_vtbl, cmdline);
```
acassis, line 65: Why to use it instead of doing like Telnet does: "nsh_session(pstate, NSH_LOGIN_TELNET, argc, argv);" ?

netutils/dropbear/port/nuttx_config.h
```diff
@@ -0,0 +1,139 @@
...
+#define DISABLE_LASTLOG 1
+#define DISABLE_PAM 1
+#define DISABLE_PUTUTLINE 1
+#define DISABLE_PUTUTXLINE 1
+#define DISABLE_SYSLOG 1
+#define DISABLE_UTMP 1
+#define DISABLE_UTMPX 1
+#define DISABLE_WTMP 1
+#define DISABLE_WTMPX 1
+#define DISABLE_ZLIB 1
```
acassis: I think these variable could be controlled from Kconfig definitions

netutils/dropbear/port/nuttx_config.h
```diff
@@ -0,0 +1,139 @@
...
+#define HAVE_STDLIB_H 1
+#define HAVE_STRINGS_H 1
+#define HAVE_STRING_H 1
+#define HAVE_STRLCAT 1
+#define HAVE_STRLCPY 1
+#define HAVE_STRUCT_ADDRINFO 1
+#define HAVE_STRUCT_IN6_ADDR 1
+#define HAVE_STRUCT_SOCKADDR_IN6 1
+#define HAVE_STRUCT_SOCKADDR_STORAGE 1
+#define HAVE_STRUCT_SOCKADDR_STORAGE_SS_FAMILY 1
```
acassis, line 78: Some of these definitions cannot be hard-code, i.e.: HAVE_STRUCT_SOCKADDR_IN6, HAVE_STRUCT_SOCKADDR_IN6, could be disabled if IPv6 support on NuttX is not enabled.

netutils/dropbear/Kconfig
```diff
@@ -0,0 +1,102 @@
...
+	depends on FSUTILS_PASSWD
+	depends on PSEUDOTERM
+	depends on SERIAL
+	depends on ARCH_HAVE_RNG
+	select CRYPTO
+	select CRYPTO_RANDOM_POOL
+	select DEV_RANDOM
+	select DEV_URANDOM
+	select LIBC_NETDB
+	select LIBC_GAISTRERROR
```
xiaoxiang781216: change select to depends on

netutils/dropbear/Kconfig
```diff
@@ -0,0 +1,102 @@
...
+		Enable a minimal Dropbear SSH server port for NuttX.  This initial
+		port is based on the ESP-IDF MCU test port and provides a single
+		foreground SSH server process with SSH sessions backed by NSH.
+
+if NETUTILS_DROPBEAR
+
+config NETUTILS_DROPBEAR_STACKSIZE
+	int "Dropbear main stack size"
+	default 65536 if ARCH_CHIP_ESP32C3
+	default 32768
```
xiaoxiang781216: why not fix to 65536

fsutils/passwd/passwd_adduser.c
```diff
@@ -67,6 +69,15 @@ int passwd_adduser(FAR const char *username, FAR const char *password)
 
   /* Check if the username already exists */
 
+  stream = fopen(CONFIG_FSUTILS_PASSWD_PATH, "a");
```
xiaoxiang781216: why need open stream and close it immediately

FelipeMdeO: This was only to create file if it doesn't exist, but this is not necessary anymore, I am removing this change.

netutils/dropbear/port/nuttx_config.h
FelipeMdeO: Hello Alan, good point. Most of that DISABLE_* block isn't really configurable, for example: DISABLE_PAM, DISABLE_LASTLOG, but other configs can be controlled by user, like CONFIG_LIB_ZLIB.
I will move to Kconfig what can be controlled by user.

netutils/dropbear/port/nuttx_config.h
FelipeMdeO: About HAVE_STRUCT_SOCKADDR_IN6:
The `fake-rfc2553.h` dropbear file has the source code below:

```
#ifndef HAVE_STRUCT_IN6_ADDR
struct in6_addr {
	u_int8_t	s6_addr[16];
};
#endif /* !HAVE_STRUCT_IN6_ADDR */

```

The issue is that the nuttx has the same struct in6_addr, so I need "disable" this code sector to be able compile.

Look NuttX file  `netinet/in.h`, it always define s6_addr:

```
#define s6_addr               in6_u.u6_addr8
#define s6_addr16             in6_u.u6_addr16
#define s6_addr32             in6_u.u6_addr32
```

I am using this macro equal 1 because s6_addr always exist in nuttx.

netutils/dropbear/Kconfig
FelipeMdeO: done

netutils/dropbear/Kconfig
FelipeMdeO: You're right. Previously I used esp32s3 which uses less stack. During the development I keep this if-else to allow test in both HW.
I am keep the highest size and user can adjust to specific HW.

FelipeMdeO:
I am reviewing your comments and doing improvement in the source code, I will ping you all to review again in the next days. tks.

nshlib/nsh_dropbear.c
FelipeMdeO: Hello Alan, nsh_dropbearstart() mirrors nsh_telnetstart() (nsh_telnetd.c), not nsh_telnetmain(). This is a bit tricky, but let me try to explain.

SSH can't reuse the nsh_session() path: unlike Telnet's line protocol, each SSH connection multiplexes channels and the shell must run on a PTY (for SIGINT/Ctrl-C, window size, raw mode). So the per-session NSH is spawned inside the Dropbear app (dropbear_nshsession.c) over a PTY pair, with no equivalent nsh_session() call at the nshlib layer.

We have this diff "approach" because telnet can hand the socket straight to nsh_session() because it's a plain-text line protocol while SSH transfer data uses encrypted data.

FelipeMdeO:
> @FelipeMdeO nice work! Kudos!!!
> 
> Some suggestions to improve this PR:
> 
>     * many of these patches are just to fix the UNUSED()s, moving it from function parameters to inside at the end of the function, please talk with the original author, maybe he accepts to move it to the end of the function;

Hello @acassis, @xiaoxiang781216. Could you please confirm whether you would like me to contact the original author? I am concerned about his position regarding Xiaoxiang’s PR: https://github.com/mkj/dropbear/pull/437

FelipeMdeO:
Guys, I opened the following discussion in the dropbear's repo: https://github.com/mkj/dropbear/issues/440

netutils/dropbear/patch/0002-use-nuttx-ecdsa-hostkey-sign.patch
```diff
@@ -0,0 +1,103 @@
...
+ 
++#ifndef DROPBEAR_ECC_256
+ #define DROPBEAR_ECC_256 (DROPBEAR_ECC)
++#endif
++#ifndef DROPBEAR_ECC_384
+ #define DROPBEAR_ECC_384 (DROPBEAR_ECC)
++#endif
++#ifndef DROPBEAR_ECC_521
+ #define DROPBEAR_ECC_521 (DROPBEAR_ECC)
++#endif
```
mkj: This can go upstream

netutils/dropbear/patch/0004-fix-nuttx-compile-warnings.patch
```diff
@@ -0,0 +1,107 @@
...
+ /* */
+ #endif /* !DROPBEAR_SVR_REMOTETCPFWD */
+ 
++#if DROPBEAR_SVR_REMOTETCPFWD
+ static int svr_cancelremotetcp(void);
+ static int svr_remotetcpreq(int *allocated_listen_port);
++#endif
++#if DROPBEAR_SVR_LOCALTCPFWD
+ static int newtcpdirect(struct Channel * channel);
++#endif
```
mkj: This has since changed upstream so might now be OK. https://github.com/mkj/dropbear/pull/414

netutils/dropbear/patch/0008-fix-strict-prototypes-warnings.patch
mkj, line 1: The func(void) changes make sense to go upstream

netutils/dropbear/dropbear_main.c
```diff
@@ -0,0 +1,305 @@
...
+    {
+      snprintf(fullmsg, sizeof(fullmsg), "Exit before auth%s: %s", fromaddr,
+               exitmsg);
+    }
+
+  dropbear_log(LOG_INFO, "%s", fullmsg);
+
+  signal_pipe[0] = ses.signal_pipe[0];
+  signal_pipe[1] = ses.signal_pipe[1];
+  session_cleanup();
```
mkj, line 105: `session_cleanup()` _should_ be OK to clean up the session in the clean exit case (the fuzzer tests for leaks), but memory won't be freed when dropbear_exit() occurs for an error case. Those error cases can readily be triggered by network traffic.

Dropbear's fuzzing harness had the same problem. As a usable hack it sets `DROPBEAR_TRACKING_MALLOC` and then the harness calls `m_malloc_free_epoch()`.
https://github.com/mkj/dropbear/blob/master/FUZZER-NOTES.md#malloc-wrapper

I don't want to add any more complexity/API like that to upstream Dropbear (it was never intended as a library), but you might be to patch something similar in the NuttX wrapper? If NuttX has arena allocators (or similar) that might be another option.

netutils/dropbear/dropbear_main.c
```diff
@@ -0,0 +1,305 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "includes.h"
+#include "algo.h"
+#include "crypto_desc.h"
+#define dropbear_main dropbear_multi_entry
+#include "dbutil.h"
+#undef dropbear_main
```
mkj, line 25:
```suggestion
#undef dropbear_main
#include "dbutil.h"
```

mkj (commented):
Good work getting it going! I suspect there are probably a few more cleanup edge cases that need dealing with, upstream Dropbear relies on exit() cleanup a fair bit. At least most of the postauth code should be in your own dropbear_nshsession.c to handle specially there.

I've added a few notes about upstream parts.

I don't think `UNUSED()` changes make sense to go upstream. It seems like it would be simpler to `#undef UNUSED` somewhere in an appropriate nuttx-dropbear header file? (Not sure where the NuttX definition comes from)

acassis:
> Good work getting it going! I suspect there are probably a few more cleanup edge cases that need dealing with, upstream Dropbear relies on exit() cleanup a fair bit. At least most of the postauth code should be in your own dropbear_nshsession.c to handle specially there.
> 
> I've added a few notes about upstream parts.
> 
> I don't think `UNUSED()` changes make sense to go upstream. It seems like it would be simpler to `#undef UNUSED` somewhere in an appropriate nuttx-dropbear header file? (Not sure where the NuttX definition comes from)

Hi @mkj thank you very much for your review and suggestions.

I think undefining UNUSED macro and defining it again could work, but could we agree at least in moving the UNUSED from been used in the function prototype and move it to end of the function, like here:

https://github.com/apache/nuttx/blob/master/drivers/syslog/syslog_channel.c#L304

This way works better for old compilers (that don't know about __attribute__((unused)) ), since NuttX is used for retro-computing as well (not only microcontrollers).

mkj:
I'm not keen on changing Dropbear's style. On other compilers Dropbear makes it a no-op.

```c
#ifdef UNUSED 
#elif defined(__GNUC__) 
# define UNUSED(x) UNUSED_ ## x __attribute__((unused)) 
#elif defined(__LCLINT__) 
# define UNUSED(x) /*@unused@*/ x 
#else 
# define UNUSED(x) x 
#endif
```

xiaoxiang781216:
> Guys, I opened the following discussion in the dropbear's repo: [mkj/dropbear#440](https://github.com/mkj/dropbear/issues/440)

@FelipeMdeO I try a porting in the last week but doesn't require so many patch/hack to dropbear code base. So could you split the real change require to port dropbear from the pure improvement?

acassis:
> > Guys, I opened the following discussion in the dropbear's repo: [mkj/dropbear#440](https://github.com/mkj/dropbear/issues/440)
> 
> @FelipeMdeO I try a porting in the last week but doesn't require so many patch/hack to dropbear code base. So could you split the real change require to port dropbear from the pure improvement?

Yes, I think we don't need to change the UNUSED() macro neither its position, since it won't be included into NuttX code base, but just compiled as external app.

netutils/dropbear/patch/0004-fix-nuttx-compile-warnings.patch
FelipeMdeO: Thank you for highlighting this. I am updating dropbear commit and reduce patch needs.

FelipeMdeO: Solved in commit sha: db01ca86005c0acef3cafeb8139c7d6f02bef632

netutils/dropbear/patch/0008-fix-strict-prototypes-warnings.patch
FelipeMdeO: Great, PR opened : https://github.com/mkj/dropbear/pull/447/changes

If you want / prefer, I can update other files to follow this standard.

netutils/dropbear/dropbear_main.c
FelipeMdeO: Sorry, I can't apply this change.

I tried this change and unfortunately it breaks the build. See Below:

```
CC:  inode/fs_inoderemove.c <command-line>: error: conflicting types for 'dropbear_main'; have 'int(int,  char **)'
dropbear_main.c:258:5: note: in expansion of macro 'main'
  258 | int main(int argc, FAR char *argv[])
      |     ^~~~
In file included from dropbear_main.c:24:
/Users/felipemouradeoliveira/nuttxspace/apps/netutils/dropbear/dropbear/src/dbutil.h:110:5: note: previous declaration of 'dropbear_main' with type 'int(int,  char **, const char *)'
  110 | int dropbear_main(int argc, char ** argv, const char * multipath);
      |     ^~~~~~~~~~~~~
```

The NuttX build system injects -Dmain=dropbear_main at compile time, but dbutil.h already declares dropbear_main with a different signature (3 args: argc, argv, multipath). That causes a conflicting types error when int main(int argc, ...) gets expanded.

The #define dropbear_main dropbear_multi_entry before the include is a trick to silently rename that declaration to dropbear_multi_entry during preprocessing, so the 3-arg version stays out of the way.

netutils/dropbear/patch/0002-use-nuttx-ecdsa-hostkey-sign.patch
FelipeMdeO: Thank you for catch that.
I opened a PR : https://github.com/mkj/dropbear/pull/448

netutils/dropbear/dropbear_main.c
FelipeMdeO: Hello mkj, nice catch. 
This type of problem is critical.

NuttX has "tools" to handle memory properly but I will follow your suggestion to use DROPBEAR_TRACKING_MALLOC because it is more simple to apply.

You can check fix in commit sha: 7f37a59d8fef9648fadf8418ea36e76ae047454f

FelipeMdeO:
> > > Guys, I opened the following discussion in the dropbear's repo: [mkj/dropbear#440](https://github.com/mkj/dropbear/issues/440)
> > 
> > 
> > @FelipeMdeO I try a porting in the last week but doesn't require so many patch/hack to dropbear code base. So could you split the real change require to port dropbear from the pure improvement?
> 
> Yes, I think we don't need to change the UNUSED() macro neither its position, since it won't be included into NuttX code base, but just compiled as external app.

I tried the #undef UNUSED approach but it doesn't work in our setup. The problem is that dropbear_nshsession.c includes both Dropbear headers and nsh_console.h, which uses UNUSED() in a function body. I cannot find other solution.

FelipeMdeO:
> @FelipeMdeO I try a porting in the last week but doesn't require so many patch/hack to dropbear code base. So could you split the real change require to port dropbear from the pure improvement?

Hello @xiaoxiang781216,

A significant portion of the changes (patches 0003–0005) are dedicated to replacing Dropbear's built-in crypto with NuttX's crypto libraries. Specifically: SHA-256/HMAC uses NuttX's mbedtls digest API, ChaCha20-Poly1305 uses NuttX's libtomcrypt state types directly, and password authentication delegates to NuttX's PAM/shadow stack. 

In other words the port will be more simple if we use crypto libs from Dropbear instead NuttX libs, but we decided use NuttX libs.
I am updating the patch descriptions to make each change clearer and easier to review.

netutils/dropbear/port/nuttx_config.h
acassis: @FelipeMdeO although the macro exist, I think it could cause some (silent?) error on dropbear if it assumes NuttX has the IPv6 support enabled, but it is disabled. What do you think @xiaoxiang781216 @mkj ?

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
```diff
@@ -0,0 +1,267 @@
...
++static int void_start(int cipher, const unsigned char *IV,
++			const unsigned char *key,
++			int keylen, int num_rounds, void *cipher_state) {
++	UNUSED(cipher);
++	UNUSED(IV);
++	UNUSED(key);
++	UNUSED(keylen);
++	UNUSED(num_rounds);
++	UNUSED(cipher_state);
++
```
acassis, line 50: @FelipeMdeO Since this kind of code applies to dropbear, you don't need to modify to follow the NuttX style

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
```diff
@@ -0,0 +1,267 @@
...
++static int dropbear_chachapoly_start(int cipher, const unsigned char *IV,
+ 			const unsigned char *key, int keylen,
+-			int UNUSED(num_rounds), dropbear_chachapoly_state *state) {
++			int num_rounds, dropbear_chachapoly_state *state) {
+ 	int err;
+
++	UNUSED(cipher);
++	UNUSED(IV);
++	UNUSED(num_rounds);
++
```
acassis: @FelipeMdeO Since this kind of code applies to dropbear, you don't need to modify to follow the NuttX style

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
```diff
@@ -0,0 +1,267 @@
...
++static void *dropbear_zalloc(void *opaque, uInt items, uInt size) {
++	UNUSED(opaque);
++
+ 	return m_calloc(items, size);
+ }
+
+-static void dropbear_zfree(void* UNUSED(opaque), void* ptr) {
++static void dropbear_zfree(void *opaque, void *ptr) {
++	UNUSED(opaque);
++
```
acassis: Ditto

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
```diff
@@ -0,0 +1,267 @@
...
+@@ -37,9 +37,11 @@
+
+ static int dropbear_gcm_start(int cipher, const unsigned char *IV,
+ 			const unsigned char *key, int keylen,
+-			int UNUSED(num_rounds), dropbear_gcm_state *state) {
++			int num_rounds, dropbear_gcm_state *state) {
+ 	int err;
+
++	UNUSED(num_rounds);
++
```
acassis: Ditto

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
```diff
@@ -0,0 +1,267 @@
...
++++ b/src/ltc_prng.c
+@@ -32,8 +32,10 @@
+   @param prng     [out] The PRNG state to initialize
+   @return CRYPT_OK if successful
+ */
+-int dropbear_prng_start(prng_state* UNUSED(prng))
++int dropbear_prng_start(prng_state *prng)
+ {
++   UNUSED(prng);
++
```
acassis: Ditto

netutils/dropbear/patch/0008-fix-strict-prototypes-warnings.patch
```diff
@@ -0,0 +1,76 @@
...
+ 	struct Channel *chan = get_any_ready_channel();
+
+--- a/src/dbrandom.c
++++ b/src/dbrandom.c
+@@ -129,7 +129,7 @@ static void addrandom(const unsigned char * buf, unsigned int len) {
+ 	sha256_done(&hs, hashpool);
+ }
+
+-static void write_urandom()
++static void write_urandom(void)
```
acassis: I think this commit could be submitted to dropbear mainline, right @mkj ?

netutils/dropbear/patch/0008-fix-strict-prototypes-warnings.patch
```diff
@@ -0,0 +1,76 @@
...
+ 	char *expand_fn = NULL;
+--- a/src/svr-runopts.c
++++ b/src/svr-runopts.c
+@@ -38,7 +38,7 @@ svr_runopts svr_opts; /* GLOBAL */
+ static void printhelp(const char * progname);
+ static void addportandaddress(const char* spec);
+ static void loadhostkey(const char *keyfile, int fatal_duplicate);
+ static void addhostkey(const char *keyfile);
+-static void load_banner();
++static void load_banner(void);
```
acassis: Ditto

netutils/dropbear/patch/0008-fix-strict-prototypes-warnings.patch
FelipeMdeO: This change is already done in dropbear mainline: https://github.com/mkj/dropbear/pull/447
Next review corrections round will use the last dropbear version and solve this.

netutils/dropbear/patch/0008-fix-strict-prototypes-warnings.patch
FelipeMdeO: This change is already done in dropbear mainline: https://github.com/mkj/dropbear/pull/447
Next review corrections round will use the last dropbear version and solve this.

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
FelipeMdeO: Done.

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
FelipeMdeO: Done.

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
FelipeMdeO: Done.

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
FelipeMdeO: Done.

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
FelipeMdeO: Done.

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
FelipeMdeO: Sorry, I did a mistake.
We need these functions. For example void_start is used  dropbear_mode_none. So I need keep it in the patch.

Please see common-algo.c:85

netutils/dropbear/port/nuttx_config.h
FelipeMdeO: Hello @acassis , ff IPv6 is disabled, NuttX's getaddrinfo() only returns IPv4 addresses, so dropbear never attempts to open IPv6 sockets. If IPv6 is enabled, getaddrinfo() returns both families and dropbear handles both correctly since NuttX's struct in6_addr is fully compatible.

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
acassis: I mean: since void_start() is not a function defined inside nuttx-apps source code like these files here;

image

So, you don't need to change the function definition, I think UNUSED() will work as well when used directly on parameters definition of the function. Right?

acassis (changes requested):
@FelipeMdeO please move the Documentation to nuttx/Documentation/ repository. All Documentation there will be included in the official Documentation, i.e.: https://nuttx.apache.org/docs/latest/applications/graphics/jpegresizetool/index.html

acassis:
@xiaoxiang781216 PTAL

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
FelipeMdeO: Hello.

The void_start is implemented dropbear/src/common-algo.c:43. This is part of dropbear source code.
In netutils/dropbear/Makefile:96 this file is added in the compilation.

If I try compile without change this function I am facing the following build issue:
```
dropbear/src/common-algo.c:43:42: error: expected declaration specifiers or '...' before numeric constant
   43 |                 unsigned long len, void* UNUSED(cipher_state)) {
      |                                          ^~~~~~
dropbear/src/common-algo.c:50:27: error: expected declaration specifiers or '...' before numeric constant
CC:  dropbear/libtomcrypt/src/pk/dsa/dsa_make_key.c tatic int void_start(int UNUSED(cipher), const unsigned char* UNUSED(IV),
      |                           ^~~~~~
dropbear/src/common-algo.c:50:43: error: expected ';', ',' or ')' before 'const'
   50 | static int void_start(int UNUSED(cipher), const unsigned char* UNUSED(IV),
```

FelipeMdeO:
> @FelipeMdeO please move the Documentation to nuttx/Documentation/ repository. All Documentation there will be included in the official Documentation, i.e.: https://nuttx.apache.org/docs/latest/applications/graphics/jpegresizetool/index.html

Should I open a dedicated PR for this documentation, or can I add documento in the following PR: https://github.com/apache/nuttx/pull/19062?

acassis:
> > @FelipeMdeO please move the Documentation to nuttx/Documentation/ repository. All Documentation there will be included in the official Documentation, i.e.: https://nuttx.apache.org/docs/latest/applications/graphics/jpegresizetool/index.html
> 
> Should I open a dedicated PR for this documentation, or can I add documento in the following PR: [apache/nuttx#19062](https://github.com/apache/nuttx/pull/19062)?

Hi Felipe, it is up to you, both ways are fine. Maybe adding it in a separated PR is better because you can have better granularity. For example, we could revert the esp32 PR without removing the Documentation about Dropbear

netutils/dropbear/patch/0007-use-nuttx-unused-macro.patch
acassis: @FelipeMdeO understood, thank you. So NuttX UNUSED() is not compatible with Dropbear.

@mkj so the way UNUSED is used in Dropbear only works if the compiler supports it, the suggestion you propose to detect compiler and create a UNUSED macro for old compilers will not work.

@FelipeMdeO ok, since we don't have other option here, let's to live with more this .patch

FelipeMdeO:
Hello @acassis , PR documentation is available here: https://github.com/apache/nuttx/pull/19190

netutils/dropbear/patch/0001-use-nuttx-unused-macro.patch
```diff
@@ -0,0 +1,291 @@
+--- a/src/chachapoly.c
++++ b/src/chachapoly.c
+@@ -43,11 +43,15 @@
+ const struct dropbear_cipher dropbear_chachapoly =
+ 	{&dummy, CHACHA20_KEY_LEN*2, CHACHA20_BLOCKSIZE};
+ 
+-static int dropbear_chachapoly_start(int UNUSED(cipher), const unsigned char* UNUSED(IV),
```
xiaoxiang781216: let's fix UNUSED macro by update the definition on nuttx git for gcc/clang branch:
```
#define UNUSED(x) UNUSED_ ## x __attribute__((unused))
```

netutils/dropbear/patch/0002-use-nuttx-ecdsa-hostkey-sign.patch
```diff
@@ -0,0 +1,103 @@
...
++#include "nuttx_hostkey.h"
++#endif
+ 
+ #if DROPBEAR_ECDSA
+ 
+@@ -67,10 +70,14 @@
+ 			, bit_size);
+ 	}
+ 
++#ifdef DROPBEAR_NUTTX
```
xiaoxiang781216: could you split the optimized change from basic porting to new pr?

netutils/dropbear/patch/0003-guard-environ-declaration-for-nuttx.patch
```diff
@@ -0,0 +1,14 @@
+--- a/src/includes.h
++++ b/src/includes.h
+@@ -184,7 +184,11 @@ typedef u_int32_t uint32_t;
+ #include <dlfcn.h>
+ #endif
+ 
++#ifdef DROPBEAR_NUTTX
++/* NuttX provides environ as a macro in stdlib.h. */
++#else
+ extern char** environ;
```
xiaoxiang781216: why not drop environ declaration directly and upstream to the community? most implementation prototype in in unistd.h.

netutils/dropbear/patch/0004-fix-nuttx-compile-warnings.patch
```diff
@@ -0,0 +1,107 @@
+--- a/src/common-kex.c
++++ b/src/common-kex.c
+@@ -117,7 +117,7 @@ void send_msg_kexinit() {
+ 
+ }
+ 
+-static void switch_keys() {
++static void switch_keys(void) {
```
xiaoxiang781216: remove this patch? already merge into the upstream

netutils/dropbear/patch/0003-use-nuttx-sha256-hmac.patch
```diff
@@ -0,0 +1,45 @@
+--- a/libtomcrypt/src/headers/tomcrypt_hash.h
++++ b/libtomcrypt/src/headers/tomcrypt_hash.h
```
xiaoxiang781216: move to new pr

netutils/dropbear/patch/0004-use-nuttx-chachapoly-state.patch
```diff
@@ -0,0 +1,84 @@
...
+ 
+ /* Fortuna PRNG */
+ #define LTC_FORTUNA
+--- a/libtomcrypt/src/headers/tomcrypt_mac.h
++++ b/libtomcrypt/src/headers/tomcrypt_mac.h
+@@ -113,6 +113,9 @@ void pmac_shift_xor(pmac_state *pmac);
+ #endif /* PMAC */
+ 
+ #ifdef LTC_POLY1305
++#ifdef DROPBEAR_NUTTX_CHACHAPOLY
```
xiaoxiang781216: move to new pr

netutils/dropbear/patch/0002-use-nuttx-passwd-auth.patch
```diff
@@ -0,0 +1,87 @@
+--- a/src/svr-authpasswd.c
++++ b/src/svr-authpasswd.c
+@@ -33,6 +33,77 @@
+ 
+ #if DROPBEAR_SVR_PASSWORD_AUTH
+ 
++#if DROPBEAR_NUTTX_PASSWD
```
xiaoxiang781216, line 7: move to new pr

netutils/dropbear/port/config.h
```diff
@@ -0,0 +1,20 @@
...
+
+/* Dropbear looks for a file named config.h. Keep this wrapper name stable;
+ * NuttX-specific autoconf replacements live in nuttx_config.h.
+ */
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include "nuttx_config.h"
```
xiaoxiang781216, line 18: why not move the  content of nuttx_config.h into here and remove nuttx_config.h

netutils/dropbear/port/default_options_guard.h
```diff
@@ -0,0 +1,693 @@
+/****************************************************************************
+ * apps/netutils/dropbear/port/default_options_guard.h
```
xiaoxiang781216: remove, should be generated by:
```
       $(Q) sh $(DROPBEAR_UNPACKNAME)/src/ifndef_wrapper.sh \
               < $(DROPBEAR_UNPACKNAME)/src/default_options.h > $@
```

netutils/dropbear/port/localoptions.h
```diff
@@ -0,0 +1,21 @@
...
+/* Dropbear looks for a file named localoptions.h when
+ * LOCALOPTIONS_H_EXISTS is defined. Keep this wrapper name stable and place
+ * the NuttX-specific option choices in nuttx_localoptions.h.
+ */
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include "nuttx_localoptions.h"
```
xiaoxiang781216: merge the content of nuttx_localoptions.h into here and remove nuttx_localoptions.h

netutils/dropbear/port/nuttx_compat.c
```diff
@@ -0,0 +1,78 @@
...
+
+  return NULL;
+}
+
+void dropbear_endusershell(void)
+{
+  g_shell_returned = 0;
+}
+
+int dropbear_getgroups(int size, gid_t list[])
```
xiaoxiang781216: see https://github.com/apache/nuttx/pull/19182

netutils/dropbear/dropbear_main.c
```diff
@@ -0,0 +1,305 @@
+/****************************************************************************
+ * apps/netutils/dropbear/dropbear_main.c
```
xiaoxiang781216, line 2: why not use the implementation from dropbear directly?

netutils/dropbear/port/nuttx_compat.c
```diff
@@ -0,0 +1,78 @@
...
+{
+  (void)size;
+  (void)list;
+
+  errno = ENOSYS;
+  return -1;
+}
+
+int link(FAR const char *path1, FAR const char *path2)
```
xiaoxiang781216: why add the dummy implementation which already provide by nuttx kernel?

netutils/dropbear/port/nuttx_compat.c
```diff
@@ -0,0 +1,78 @@
...
+ ****************************************************************************/
+
+static char g_dropbear_shell[] = "/bin/sh";
+static int g_shell_returned;
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+void dropbear_setusershell(void)
```
xiaoxiang781216: remove HAVE_GETUSERSHELL from config.h

netutils/dropbear/patch/0001-use-nuttx-unused-macro.patch
acassis: @xiaoxiang781216 it will fix the case where GCC is used, but the code still need to be modified to support other compilers, like the used bu eZ80, etc

netutils/dropbear/patch/0003-guard-environ-declaration-for-nuttx.patch
FelipeMdeO: I will open PR in dropbear repo. I will send later the PR link.

netutils/dropbear/patch/0004-fix-nuttx-compile-warnings.patch
FelipeMdeO: The last PR version removed this patch: commit sha: 02d925cde

netutils/dropbear/port/nuttx_compat.c
FelipeMdeO: Dropbear uses link() in gensignkey.c:160 to atomically write the host key: it writes to a temp file, then calls link(tmp, final) expecting a hard link so both names reference the same inode. When tmp is later removed, the data survives through final. With a symlink instead, final would become a dangling pointer after tmp is deleted — silently losing the host key.

Therefore, the safe fallback (non-atomic write via buf_writefile) is only triggered when link() returns an error such as ENOSYS. If the symlink-based link() returns success, Dropbear proceeds assuming hard link semantics and the key gets lost.

For more details, please look for the code below netutils/dropbear/dropbear/src/gensignkey.c:154

(1) Key content is written to a temporary file only
`ret = buf_writefile(buf, fn_temp, 0);
`
(2) link() is expected to create a hard link:
  both fn_temp and filename reference the same inode.
  If link() fails with ENOSYS/EPERM/EACCES, Dropbear falls back
  to writing filename directly (buf_writefile).
  If link() succeeds, Dropbear assumes filename already holds the
  key via the shared inode — no additional write is done.
```
if (link(fn_temp, filename) < 0) {
    if (errno == EPERM || errno == EACCES || errno == ENOSYS) {
        /* safe fallback: write content directly to filename */
        ret = buf_writefile(buf, filename, skip_exist);
    }
    goto out;
}
```

success path: key is accessible via filename through the hard link
`fsync_parent_dir(filename);`

out:
    (3) fn_temp is always deleted here.
        With a hard link: filename still holds the inode — key survives.
        With a symlink:   filename points to fn_temp, which is now
         deleted — dangling symlink, key is lost.
`    unlink(fn_temp);`

netutils/dropbear/dropbear_main.c
FelipeMdeO: Because dropbear_main.c replaces main_noinetd() with a single persistent task that uses setjmp/longjmp to intercept dropbear_exit() and return to the accept loop after each session.

fork() is the central difference. svr-main.c:308 calls fork() to create a child process per connection. NuttX (flat build) does not support fork(), and the entire main_noinetd() is structured around it — without fork, the whole model breaks down.

netutils/dropbear/port/localoptions.h
FelipeMdeO: Done.

netutils/dropbear/port/nuttx_compat.c
FelipeMdeO: Can we solve this in next PR round?

netutils/dropbear/port/config.h
FelipeMdeO: Done

netutils/dropbear/port/nuttx_compat.c
FelipeMdeO: done

netutils/dropbear/patch/0003-guard-environ-declaration-for-nuttx.patch
FelipeMdeO: waiting for @mkj review: https://github.com/mkj/dropbear/pull/453

netutils/dropbear/port/default_options_guard.h
FelipeMdeO: done

netutils/dropbear/patch/0003-guard-environ-declaration-for-nuttx.patch
FelipeMdeO: Done.

netutils/dropbear/patch/0002-use-nuttx-passwd-auth.patch
FelipeMdeO: Sorry Xiang, I can't remove this patch.
Dropbear's native path does crypt(password, pw_passwd) against the /etc/shadow hash from getpwnam(), but NuttX has no shadow/crypt() store — credentials live in fsutils/passwd and are checked with passwd_verify() (verify-only), so the native path can't authenticate.

FelipeMdeO:
Hello @xiaoxiang781216 , I did most part of you requested, I tested and everything is ok.
I saved a backup here to open new PR moving crypto from dropbear to NuttX crypto later.
Also I used [https://github.com/apache/nuttx/pull/19184](url) but not [https://github.com/apache/nuttx/pull/19182#pullrequestreview-4556426127](url) yet, I can add this in the next PR.

netutils/dropbear/patch/0001-use-nuttx-unused-macro.patch
xiaoxiang781216: other compilers are better to disable the unused variable warning directly instead modifying the origin code base.

netutils/dropbear/dropbear_main.c
xiaoxiang781216: Yes, the block porting issue for MMU less system is fork, @mkj do you think it's a good idea to add an option which launch the new instance through posix_spawn? I can provide a general patch(work on major POSIX OS include NuttX), if you think it's a right direction.

netutils/dropbear/dropbear_main.c
xiaoxiang781216: @mkj https://github.com/mkj/dropbear/pull/437 could fix this issue, please reconsider it.

netutils/dropbear/port/nuttx_compat.c
xiaoxiang781216: Ok, since NuttX VFS doesn't support the hard link yet, so it's better to set errno to ENOSYS at https://github.com/apache/nuttx/blob/master/fs/vfs/fs_link.c#L63.

xiaoxiang781216:
> Hello @xiaoxiang781216 , I did most part of you requested, I tested and everything is ok. I saved a backup here to open new PR moving crypto from dropbear to NuttX crypto later. Also I used [https://github.com/apache/nuttx/pull/19184](url) but not [https://github.com/apache/nuttx/pull/19182#pullrequestreview-4556426127](url) yet, I can add this in the next PR.

Thanks.

netutils/dropbear/port/nuttx_compat.c
xiaoxiang781216: but why need wait the next pr? it is a simple change.

netutils/dropbear/dropbear_main.c
mkj: I think it's better to keep a patch in nuttx. `-Dmain` is pretty unusual.

netutils/dropbear/dropbear_main.c
mkj: I don't think posix_spawn could do the equivalent of fexecve that Dropbear uses at the moment? (That's used so the spawned process is the same as what's running, to avoid having to worry about incompatibility if the binary gets upgraded). Could keep a posix_spawn version in nuttx though.

netutils/dropbear/patch/0002-use-nuttx-passwd-auth.patch
acassis: @FelipeMdeO please open an Issue at NuttX github to add support to it.

acassis (dismissed):

acassis:
@xiaoxiang781216 PTAL

netutils/dropbear/port/nuttx_compat.c
FelipeMdeO: Ok, I will take a look

netutils/dropbear/port/nuttx_compat.c
FelipeMdeO: Ok, I believe this PR will be merged in the next hours, after this I will change my code.

netutils/dropbear/patch/0002-use-nuttx-passwd-auth.patch
FelipeMdeO: Sure, I will and I can take care this.

netutils/dropbear/port/nuttx_compat.c
FelipeMdeO: Sorry @xiaoxiang781216, I tried but couldn't apply this request.

That getgroups() is only used by Dropbear's non-multiuser sanity check, which expects it to fail with ENOSYS to confirm a non-multiuser kernel. getgrouplist() can't replace it. It's a group-database lookup that always returns at least the primary group passed in by the caller. 
In short, I want to check if the kernel is single user, and getgrouplist returns other thing.

FelipeMdeO: Hello. Thinking better, I solve this issue using other approach and as result I can remove this stub.
I will use CONFIG_SCHED_USER_IDENTITY to know if we are running in single or multi user mode, this is all that I need.

netutils/dropbear/port/nuttx_compat.c
FelipeMdeO: Hello @xiaoxiang781216, can you check this PR please: https://github.com/apache/nuttx/pull/19212

FelipeMdeO:
Hello @xiaoxiang781216 , I finished the changes, however we need merge first the link change:
https://github.com/apache/nuttx/pull/19212

acassis (approved):

acassis:
@xiaoxiang781216 could you please verify if everything is fine now?

PR #3535, opened 2026-06-09
benchmarks/mtd: use uniform output style for numbers

sumpfralle:
## Summary

Previously the output of the `mtd` benchmark tool looked like this:

```
FLASH device parameters:
   Sector size:         256
   Sector count:       2048
   Erase block:        1000
   Total size:       524288
```

The third line sneakily deviated from the format of the other lines:
- The value `1000` is a hexadecimal number.
- The three other values are decimal numbers.

There is no good reason for using the hexadecimal representation for this specific value.

## Impact

The change influences only the human readable output of the interactive `mtd` benchmark tool.

## Testing

I tried the `mtd` tool before and after on my `rp2040` board.
The numbers looked much better afterwards.

### DIFF
--- a/benchmarks/mtd/mtd.c
+++ b/benchmarks/mtd/mtd.c
@@ -102,10 +102,10 @@ int main(int argc, FAR char *argv[])
   /* Report the device structure */
 
   printf("FLASH device parameters:\n");
-  printf("   Sector size:  %10d\n", info.sectorsize);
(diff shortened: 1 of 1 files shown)


cederom (approved):
Thank you @sumpfralle :-) Please to remember to attach build and runtime logs after change :-)

xiaoxiang781216 (approved):

PR #3536, opened 2026-06-11
testing/sig_sp_test: add test for SP context restore on signal return

cshung:
## Summary

Add a test application that verifies modifying `REG_R13` (SP) in saved register context during a signal handler is honored on exception return.

**Test scenario** (simulates a managed runtime unwinding past a trampoline):
1. Push values 1 and 2 onto the stack (simulating trampoline pushes)
2. Wait for `SIGALRM` (async delivery from timer interrupt)
3. Handler advances SP by 4 (emulates "pop" / unwind past one frame)
4. Handler redirects PC to a resume function
5. Resume function pops one value — verifies it equals 1 (not 2)

This exercises the HW exception frame relocation fix in `arm_exception.S`.

**Dependencies:**
- Requires `CONFIG_ARMV7M_SP_CONTEXT_RESTORE=y` or `CONFIG_ARMV8M_SP_CONTEXT_RESTORE=y`
- Flat build only (`BUILD_FLAT`) — accesses `nxsched_self()` for `saved_regs`
- ARM architecture only

Companion OS fix PR: https://github.com/apache/nuttx/pull/19123

## Impact

- New test application only, no impact on existing code
- Gated by Kconfig dependencies (will not appear unless SP context restore is enabled)

## Testing

**Host:** Ubuntu 22.04 x86_64, arm-none-eabi-gcc 13.3, QEMU 8.2.2 (via Docker)

**Targets tested:**
- `lm3s6965-ek:qemu-flat` (ARMv7-M, Cortex-M3)
- `mps2-an521:nsh` (ARMv8-M, Cortex-M33)

**Test log (ARMv7-M):**
```
nsh> sig_sp_test
sig_sp_test: Signal SP restore test
sig_sp_test: push 1, push 2, alarm, handler SP+=4, pop => 1
sig_sp_test: handler - PC=0x00022770 SP=0x2000a9a8
sig_sp_test: handler - new SP=0x2000a9ac PC=0x000226ac
sig_sp_test: popped value = 1 (expected 1)
sig_sp_test: PASS
```

**Test log (ARMv8-M):**
```
nsh> sig_sp_test
sig_sp_test: Signal SP restore test
sig_sp_test: push 1, push 2, alarm, handler SP+=4, pop => 1
sig_sp_test: handler - PC=0x1002b2b8 SP=0x38007b08
sig_sp_test: handler - new SP=0x38007b0c PC=0x1002b1fc
sig_sp_test: popped value = 1 (expected 1)
sig_sp_test: PASS
```

**Sanity check (without SP adjustment in handler):**
```
sig_sp_test: popped value = 2 (expected 1)
sig_sp_test: FAIL - expected 1, got 2
```

### DIFF
--- a/testing/sig_sp_test/sig_sp_test_main.c
+++ b/testing/sig_sp_test/sig_sp_test_main.c
@@ -1,0 +1,210 @@
+/****************************************************************************
+ * apps/testing/sig_sp_test/sig_sp_test_main.c
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Test: Verify that modifying SP (REG_R13) in the saved register context
+ * is honored on context restore (exception return path).
+ *
+ * Scenario (simulates runtime unwinding past a trampoline):
+ *   1. Push values 1 and 2 onto the stack (simulating a trampoline push)
+ *   2. Busy-wait for a timer signal (SIGALRM)
+ *   3. In the handler, "emulate a pop" by advancing SP by 4
+ *      (simulating a runtime deciding the top frame was a stub)
+ *   4. Redirect PC to resume_after_signal
+ *   5. After signal return, pop a value — it should be 1
+ *
+ * This exercises:
+ *   - The SP relocation fix in arm_exception.S
+ *   - The backward sliding copy (new SP > old SP, stack shrinks)
+ *   - A realistic use case: unwinding past a trampoline frame
+ *
+ * Requirements:
+ *   - Flat build only (accesses nxsched_self() for saved_regs)
+ *   - ARM architecture (uses ARMv7-M register layout)
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <signal.h>
+#include <string.h>
+#include <stdint.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <nuttx/arch.h>
+#include <nuttx/sched.h>
+#include <arch/irq.h>
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static volatile int g_result = -1;
+static volatile int g_ready = 0;
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: verify_result
+ *
+ * Description:
+ *   Called with the popped value in r0. Verifies it equals 1.
+ ****************************************************************************/
+
+void __attribute__((noinline, used)) verify_result(uint32_t value)
+{
+  printf("sig_sp_test: popped value = %lu (expected 1)\n",
+         (unsigned long)value);
+
+  if (value == 1)
+    {
+      printf("sig_sp_test: PASS\n");
+      g_result = 0;
+    }
+  else
+    {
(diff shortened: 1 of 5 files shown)


testing/sig_sp_test/sig_sp_test_main.c
```diff
@@ -0,0 +1,198 @@
...
+static void sigalrm_handler(int signo, siginfo_t *info, void *ucontext)
+{
+  struct tcb_s *rtcb = nxsched_self();
+  uint32_t *regs;
+
+  (void)signo;
+  (void)info;
+  (void)ucontext;
+
+  regs = rtcb->xcp.saved_regs;
```
xiaoxiang781216, line 117: it's better to pass regs context through ucontext, and remove depends on BUILD_FLAT

cshung: Thanks for the suggestion. I looked into passing the register context through ucontext (the 3rd argument to the signal handler).

Currently, `sig_deliver.c` passes NULL for ucontext on all three dispatch paths (lines 142, 164, 173). To pass the saved register context there, we'd need to address a few concerns:

1. **Protected/kernel builds**: `saved_regs` resides in kernel memory and cannot be directly exposed to userspace. Supporting ucontext in protected mode would require copying the context to the user stack before dispatch and copying it back after — adding unconditional overhead to every signal delivery, even when the handler doesn't use ucontext.

2. **NULL on synchronous signals**: `saved_regs` is only populated on the async path (interrupt → `arm_schedulesigaction`). For synchronous signals (e.g., `raise()`), it is NULL, so the handler cannot assume ucontext is always valid.

3. **Cross-architecture scope**: `sig_deliver.c` is arch-independent. Changing the ucontext passing behavior affects all architectures, which seems like a larger change than this test PR should drive.

Given these constraints, I think keeping the `BUILD_FLAT` dependency is appropriate for now. The test validates the SP restoration fix using `nxsched_self()->xcp.saved_regs`, which is the simplest correct approach for flat builds. A proper ucontext implementation could be a separate effort if there's broader interest.

xiaoxiang781216: but saved_regs should point to the user stack?

xiaoxiang781216 (dismissed):

simbit18:
@cshung please fix

```
../nuttx/tools/checkpatch.sh -c -u -m -g ee4727fdbacd0d847f3c197fcf7ed06f14328289..HEAD
ERROR __main__.py:618: Check failed: /home/runner/work/nuttx-apps/nuttx-apps/apps/testing/sig_sp_test/CMakeLists.txt

cmake-format check failed, run following command to update the style:
  $ cmake-format <src> -o <dst>

ERROR __main__.py:618: Check failed: /home/runner/work/nuttx-apps/nuttx-apps/apps/testing/sig_sp_test/CMakeLists.txt
Used config files:
    1: .codespellrc
Error: Process completed with exit code 1.
```

testing/sig_sp_test/sig_sp_test_main.c
cshung: You're right, I apologize for the confusion. I was incorrectly applying my understanding of Cortex-A (where context save/restore is entirely software-managed) to Cortex-M. On ARMv7-M/ARMv8-M, the hardware pushes the exception frame onto PSP (the user stack), and the software saves the remaining registers there as well — so `saved_regs` does indeed point to user-accessible memory.

xiaoxiang781216 (approved):

acassis (approved):

PR #3539, opened 2026-06-12
examples/thttpd: use procfs for tasks CGI

### LINKED ISSUE #245: Illegal call to nxsched_foreach() in examples/thttpd
As noted by @gregory-nutt in commit # 3abdf30e8c62f1a82fc48a0763392ba6c8b0ee2d the logic in examples/thttpd/content/tasks/tasks.c is calling OS internal function nxsched_foreach() which violates the POSIX OS interface.

This internal function is being called to generate a list of tasks for display in the webpage served by the web server example.

Although this is an example program, or perhaps more so _because_ it is an example, we should not be teaching the wrong way to do things!

nightt5879:
## Summary

- Replace the thttpd tasks CGI use of `nxsched_foreach()` with procfs-based task enumeration.
- Parse `/proc/<pid>/status` for task name, type, state, priority, and scheduler policy.
- Keep the tasks page generated through standard file APIs instead of scheduler internals.

## Why

The thttpd example was reading task data through NuttX scheduler internals (`struct tcb_s` via `nxsched_foreach()`), which violates the application/kernel interface boundary. procfs already exposes the same task information through a standard file interface.

Fixes #245.

## Testing

- `git diff --check`
- `gcc -x c -std=c99 -Wall -Wextra -Werror -Wno-unused-parameter -fsyntax-only ... examples/thttpd/content/tasks/tasks.c` with a temporary local NuttX config/arch shim for syntax checking

Full board build was not run because this local checkout does not currently have a configured NuttX `.config`.

### DIFF
--- a/examples/thttpd/content/tasks/tasks.c
+++ b/examples/thttpd/content/tasks/tasks.c
@@ -26,48 +26,36 @@
 
 #include <nuttx/config.h>
 
+#include <ctype.h>
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
 #include <stdio.h>
-#include <stdlib.h>
+#include <string.h>
 #include <unistd.h>
-#include <sched.h>
-
-#include <nuttx/sched.h>
 
 /****************************************************************************
- * Private Data
+ * Pre-processor Definitions
  ****************************************************************************/
 
-static const char *g_statenames[] =
-{
-  "INVALID ",
-  "PENDING ",
-  "READY   ",
-  "RUNNING ",
-  "INACTIVE",
-  "WAITSEM ",
-#ifndef CONFIG_DISABLE_MQUEUE
-  "WAITSIG ",
+#ifndef THTTPD_PROCFS_MOUNTPOINT
+#  define THTTPD_PROCFS_MOUNTPOINT "/proc"
 #endif
-#ifndef CONFIG_DISABLE_MQUEUE
-  "MQNEMPTY",
-  "MQNFULL "
-#endif
-};
 
-static const char *g_ttypenames[4] =
-{
-  "TASK   ",
-  "PTHREAD",
-  "KTHREAD",
-  "--?--  "
-};
+#define THTTPD_STATUS_SIZE 512
 
-static FAR const char *g_policynames[4] =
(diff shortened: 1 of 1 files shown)


examples/thttpd/content/tasks/tasks.c
```diff
@@ -78,38 +66,210 @@ static FAR const char *g_policynames[4] =
...
+  char filepath[sizeof(THTTPD_PROCFS_MOUNTPOINT) + NAME_MAX +
+                sizeof("/status") + 1];
+  ssize_t nread;
+  size_t total = 0;
+  int ret;
+  int fd;
+
+  ret = snprintf(filepath, sizeof(filepath), "%s/%s/status",
+                 THTTPD_PROCFS_MOUNTPOINT, pid);
+  if (ret < 0 || (size_t)ret >= sizeof(filepath))
```
xiaoxiang781216:
```suggestion
  if (ret < 0 || ret >= sizeof(filepath))
```

examples/thttpd/content/tasks/tasks.c
```diff
@@ -78,38 +66,210 @@ static FAR const char *g_policynames[4] =
...
+          ret = -errno;
+          close(fd);
+          return ret;
+        }
+      else if (nread == 0)
+        {
+          break;
+        }
+
+      total += (size_t)nread;
```
xiaoxiang781216:
```suggestion
      total += nread;
```

examples/thttpd/content/tasks/tasks.c
```diff
@@ -78,38 +66,210 @@ static FAR const char *g_policynames[4] =
...
+{
+  FAR struct dirent *entryp;
+  DIR *dirp;
+
+  dirp = opendir(THTTPD_PROCFS_MOUNTPOINT);
+  if (dirp == NULL)
+    {
+      printf("Unable to open %s (is procfs mounted?): %d\n",
+             THTTPD_PROCFS_MOUNTPOINT, errno);
+      return ERROR;
```
xiaoxiang781216:
```suggestion
      return -errno;
```

xiaoxiang781216 (approved):

acassis (approved):

PR #3556, opened 2026-06-24
industry/scpi: add unit parser config options

### LINKED ISSUE #3087: [FEATURE] scpi-parser: allow configuration of build flags
### Is your feature request related to a problem? Please describe.

I am using the [scpi-parser](https://github.com/apache/nuttx-apps/blob/master/industry/scpi/) app.

The upstream software (`scpi-parser`) provides [a few defines](https://github.com/j123b567/scpi-parser/blob/698eabe62eca0d9df5663a2ebb3fcab80e72d5d3/libscpi/inc/scpi/config.h#L125) for configuring the build.

I would like to enable some of these (e.g. `USE_UNITS_TIME=1`).

But the [Makefile of the app](https://github.com/apache/nuttx-apps/blob/master/industry/scpi/Makefile) does not expose configurable build flags.

I tried to assemble a list of potential solutions for my need.
Maybe you can help me commenting on the proper way.
I plan to prepare a PR, if a change in `nuttx-apps` could solve my problem.

# Approach A: add a raw CFLAGS configuration option for this app

I tried to add a new configuration variable to the app (`CONFIG_SCPI_PARSER_CFLAGS`), but I had trouble getting rid of the quoting (i.e. multiple `-DFOO=1` arguments were magically quoted into a single argument). Not even something like the following helped with the quoting:
```make
CFLAGS += $(patsubst "%",%,$(CONFIG_SCPI_PARSER_CFLAGS))
```
And I am not sure, whether this kind of "raw" injection of CFLAGS would be the proper way to do this.

# Approach B: mirror each upstream build option in nuttx's configuration

The obvious alternative could be to simply replicate *all* configuration options from upstream into the app's configuration (`Kconfig`). But this feels like a bit too much for me. For `scpi-parser` this would be 14 options simply for unit conversion features.

# Approach C: gate *all* features of the app behind a single nuttx configuration switch

This obviously does not expose the full flexibility of the upstream app's build configuration. But at least for my use-case it would be sufficient.

# Approach D: some tricky way to sneak CFLAGS into an app without forking?

Maybe there is some way, I could influence the CFLAGS of an unrelated app.
I would prefer this approach, since it wouldn't require any changes in `nuttx-apps`.
But I am not aware of a way to do this.

### Verification

nightt5879:
## Summary
- Add explicit Kconfig options for scpi-parser's optional unit groups.
- Map each option to the matching upstream `USE_UNITS_*` compile-time define.
- Keep the existing default unit set while allowing users to enable groups such as `USE_UNITS_TIME=1` without raw CFLAGS injection.

Fixes #3087

## Testing
- `git diff upstream/master...HEAD --check`
- Parsed `industry/scpi/Kconfig` with `kconfiglib` and verified all new unit symbols are present.


acassis (changes requested):
@nightt5879 please update the SCPI documentation since you are working/testing it. Currently it is empty: https://nuttx.apache.org/docs/latest/applications/industry/scpi/index.html

nightt5879:
Added the requested SCPI documentation in apache/nuttx#19207: https://github.com/apache/nuttx/pull/19207

xiaoxiang781216 (approved):

linguini1 (commented):
Did you try using the scpi tool after these changes?

Also why use $(call) in the Makefile?

acassis (approved):

PR #3557, opened 2026-06-24
!nshlib: Require explicit fixed login password at build time

Abhishekmishra2808:
## Summary

This PR is the nuttx-apps companion to [apache/nuttx#19209](https://github.com/apache/nuttx/pull/19209), which adds the host `mkpasswd` tool and ROMFS `/etc/passwd` build integration. Here, runtime password handling moves from reversible TEA to one-way PBKDF2-HMAC-SHA256 using a modular crypt string stored in `/etc/passwd`. The change adds `apps/crypto/pbkdf2` with a small self-contained SHA-256 and HMAC implementation (Mbed TLS PKCS5 is used when available), rewrites `passwd_encrypt` and `passwd_verify` with shared base64url helpers and password complexity checks, and adds `pbkdf2_test` for RFC 6070 SHA-256 vector coverage. The round-trip portion of that test is skipped automatically when the passwd file is read-only or when `DEV_URANDOM` is not enabled. NSH also changes the default fixed-login username to `root` and removes insecure fixed-login password defaults.

**BREAKING CHANGE:** TEA-encoded `/etc/passwd` entries no longer verify after this upgrade. Every entry must be regenerated with NSH `passwd` or `useradd` once both nuttx and nuttx-apps are updated. If `CONFIG_NSH_LOGIN_FIXED=y` is used, `CONFIG_NSH_LOGIN_PASSWORD` must be set explicitly in the board defconfig or menuconfig; there is no default password.

## Impact

This change modifies existing password file behavior and is not backward compatible with TEA-encoded hashes. Users must regenerate stored passwords. The build flow itself is unchanged in nuttx-apps; the ROMFS build-time password prompt lives in the paired nuttx PR. Documentation updates are included in that nuttx PR rather than here.

## Testing

Testing was done on `sim:login` with NSH console login against a build-time PBKDF2 hash, and login succeeded. 
Login was also verified on ESP32-C3 using the `esp32c3-devkit:login` configuration over USB serial. Check- https://github.com/apache/nuttx/pull/19209

`pbkdf2_test` passed the RFC 6070 SHA-256 vectors on `sim:login`, and the passwd encrypt/verify round-trip passed with a writable passwd path at `/tmp/passwd`, readonly disabled, and `DEV_URANDOM` enabled. 
image


nshlib/Kconfig
```diff
@@ -1265,17 +1222,22 @@ endchoice # Verification method
...
+	default "root"
 	depends on !NSH_LOGIN_PASSWD
 	---help---
-		Login user name.  Default: "admin"
+		Login user name.  Default: "root"
 
 config NSH_LOGIN_PASSWORD
 	string "Login password"
-	default "Administrator"
+	default ""
```
xiaoxiang781216: remove the default value

nshlib/Makefile
```diff
@@ -114,6 +110,26 @@ ifeq ($(CONFIG_NSH_LOGIN_PASSWD),y)
 CSRCS += nsh_passwdcmds.c
 endif
 
+ifeq ($(CONFIG_NSH_LOGIN_FIXED),y)
```
xiaoxiang781216: why not remove the default value in the case of CONFIG_NSH_LOGIN_FIXED directly and remove this warning check, so the user must enter the password in defconfig.

ThePassionate:
Thanks for this work. One suggestion:
 
the NuttX kernel already provides CRYPTO_PBKDF2_HMAC_SHA256 (defined in 
cryptodev.h) with a complete software implementation in cryptosoft.c (swcr_pbkdf2), accessible via /dev/crypto.

Would it be possible for fsutils/passwd to use the kernel capability directly via /dev/crypto ioctl, rather than introducing a new apps/crypto/pbkdf2 module? The PBKDF2 logic already lives in the kernel — the app layer only needs to open /dev/crypto, call CIOCGSESSION with CRYPTO_PBKDF2_HMAC_SHA256, then CIOCCRYPT. This would avoid duplicating the implementation and keep apps/crypto/ clean.

Abhishekmishra2808:
Thanks for the review @ThePassionate 
host build tool tools/mkpasswd.c (in the NuttX PR) will still keep a standalone PBKDF2 implementation since it runs on the build machine, not on the target, Right?

I'll remove apps/crypto/pbkdf2 and have fsutils/passwd use kernel PBKDF2 via /dev/crypto. The host build tool tools/mkpasswd.c (nuttx PR) will keep a standalone implementation since it runs on the build machine.

xiaoxiang781216:
> Thanks for the review @ThePassionate host build tool tools/mkpasswd.c (in the NuttX PR) will still keep a standalone PBKDF2 implementation since it runs on the build machine, not on the target, Right?
> 
> I'll remove apps/crypto/pbkdf2 and have fsutils/passwd use kernel PBKDF2 via /dev/crypto. The host build tool tools/mkpasswd.c (nuttx PR) will keep a standalone implementation since it runs on the build machine.

it's more simple to convert host tool from c to python, since we don't need compile the tool and could support more OS.

fsutils/passwd/Kconfig
```diff
@@ -6,8 +6,15 @@
 config FSUTILS_PASSWD
 	bool "Password file support"
 	default n
+	select CRYPTO
+	select ALLOW_BSD_COMPONENTS
+	select CRYPTO_CRYPTODEV
```
xiaoxiang781216: let's use depends on or skip PBKDF2-HMAC-SHA256 if crypto isn't enabled

fsutils/passwd/Kconfig
```diff
@@ -23,20 +30,14 @@ config FSUTILS_PASSWD_IOBUFFER_SIZE
...
-	default 0x9abcdef0
-
-config FSUTILS_PASSWD_KEY3
-	hex "Encryption key value 3"
-	default 0x12345678
-
-config FSUTILS_PASSWD_KEY4
-	hex "Encryption key value 4"
-	default 0x9abcdef0
+	config FSUTILS_PASSWD_PBKDF2_ITERATIONS
```
xiaoxiang781216: remove indent

fsutils/passwd/passwd_base64.c
```diff
@@ -26,137 +26,162 @@
...
  * Pre-processor Definitions
  ****************************************************************************/
 
-#ifndef CONFIG_EXAMPLES_RNG90_DEVPATH
-#  define CONFIG_EXAMPLES_RNG90_DEVPATH "/dev/rng0"
-#endif
+/* RFC 4648 section 5 base64url alphabet (no padding). */
 
-#define RNG90_MAX_BYTES 32
+static const char g_base64url[] =
```
xiaoxiang781216: should we call netutils/codecs/base64.c?

testing/crypto/passwd/pbkdf2_test.c
```diff
@@ -0,0 +1,287 @@
...
+  0x1c
+};
+
+static const uint8_t g_vector6[] =
+{
+  0x89, 0xb6, 0x9d, 0x05, 0x16, 0xf8, 0x29, 0x89,
+  0x3c, 0x69, 0x62, 0x26, 0x65, 0x0a, 0x86, 0x87
+};
+
+static const struct pbkdf2_vector_s g_vectors[] =
```
xiaoxiang781216, line 110: should we move pdkdf2 to testing/drivers/crypto/pbkdf2.c

testing/pbkdf2/pbkdf2_test.c
```diff
@@ -0,0 +1,287 @@
+/****************************************************************************
+ * apps/testing/pbkdf2/pbkdf2_test.c
```
xiaoxiang781216: move to testing/crypto/passwd

fsutils/passwd/Kconfig
Abhishekmishra2808: Can we make NSH_CONSOLE_LOGIN depend on FSUTILS_PASSWD / cryptodev so login always uses PBKDF2, and treat NSH_LOGIN_FIXED as an explicit opt-in for legacy boards only?

Abhishekmishra2808:
> it's more simple to convert host tool from c to python, since we don't need compile the tool and could support more OS.

Previously C was chosen to avoid a Python dependency. If maintainers prefer Python for Windows/CI simplicity, I can switch the nuttx PR to mkpasswd.py before merge

fsutils/passwd/Kconfig
Abhishekmishra2808: Can we keep something like this:
image

Because if we use "depends on," the user might go for the fixed password most of the time, which is not recommended.

acassis:
> > Thanks for the review @ThePassionate host build tool tools/mkpasswd.c (in the NuttX PR) will still keep a standalone PBKDF2 implementation since it runs on the build machine, not on the target, Right?
> > I'll remove apps/crypto/pbkdf2 and have fsutils/passwd use kernel PBKDF2 via /dev/crypto. The host build tool tools/mkpasswd.c (nuttx PR) will keep a standalone implementation since it runs on the build machine.
> 
> it's more simple to convert host tool from c to python, since we don't need compile the tool and could support more OS.

@xiaoxiang781216 the problem with Python it because it is a moving target, we don't know when will be the next version transition that will break everything (like 2 to 3). So if we can keep it in plain and old C it is safe.

But I'm not against using Python, I just suggest avoid depending on external libraries, this way it will survive longer and we will avoid listening complains from new users because NuttX is not working

cederom:
> > > Thanks for the review @ThePassionate host build tool tools/mkpasswd.c (in the NuttX PR) will still keep a standalone PBKDF2 implementation since it runs on the build machine, not on the target, Right?
> > > I'll remove apps/crypto/pbkdf2 and have fsutils/passwd use kernel PBKDF2 via /dev/crypto. The host build tool tools/mkpasswd.c (nuttx PR) will keep a standalone implementation since it runs on the build machine.
> > 
> > 
> > it's more simple to convert host tool from c to python, since we don't need compile the tool and could support more OS.
> 
> @xiaoxiang781216 the problem with Python it because it is a moving target, we don't know when will be the next version transition that will break everything (like 2 to 3). So if we can keep it in plain and old C it is safe.
> 
> But I'm not against using Python, I just suggest avoid depending on external libraries, this way it will survive longer and we will avoid listening complains from new users because NuttX is not working

Actually every minor Python release now is kinda self-incompatible and changes syntax / internals.. unless we use really simple syntax :-)

acassis:
> > > > Thanks for the review @ThePassionate host build tool tools/mkpasswd.c (in the NuttX PR) will still keep a standalone PBKDF2 implementation since it runs on the build machine, not on the target, Right?
> > > > I'll remove apps/crypto/pbkdf2 and have fsutils/passwd use kernel PBKDF2 via /dev/crypto. The host build tool tools/mkpasswd.c (nuttx PR) will keep a standalone implementation since it runs on the build machine.
> > > 
> > > 
> > > it's more simple to convert host tool from c to python, since we don't need compile the tool and could support more OS.
> > 
> > 
> > @xiaoxiang781216 the problem with Python it because it is a moving target, we don't know when will be the next version transition that will break everything (like 2 to 3). So if we can keep it in plain and old C it is safe.
> > But I'm not against using Python, I just suggest avoid depending on external libraries, this way it will survive longer and we will avoid listening complains from new users because NuttX is not working
> 
> Actually every minor Python release now is kinda self-incompatible and changes syntax / internals.. unless we use really simple syntax :-)

I think implementing in C is the safer approach. The compile time is not an issue, so it is better to wait 0.5s to finish the compilation than getting an incompatibility error.

fsutils/passwd/Kconfig
Abhishekmishra2808: update: I have used "depends on".

xiaoxiang781216 (dismissed):

jerpelea (dismissed):

cederom:
In order for CI to pass we need to merge https://github.com/apache/nuttx-apps/pull/3557 and then https://github.com/apache/nuttx/pull/19209 correct? :-)

jerpelea (dismissed):

acassis (dismissed):

linguini1:
I think this needs a rebase

xiaoxiang781216 (approved):

jerpelea (approved):

Abhishekmishra2808:
@xiaoxiang781216 , PTAL !!

nshlib/Kconfig
```diff
@@ -1230,32 +1230,30 @@ config NSH_LOGIN
 config NSH_CONSOLE_LOGIN
 	bool "Console Login"
 	default n
+	depends on FSUTILS_PASSWD
```
michallenc, line 1233: I am getting a little late to this (we've updated our NuttX repositories and found this change now), but why is the dependency enforced? Why does the password need to be verified against the file? I understand the benefits, but why enforce the option. This should be optional, not mandatory. @Abhishekmishra2808

Abhishekmishra2808: The intent was to close a gap where `NSH_CONSOLE_LOGIN` could be enabled without a real password ever being configured (empty/effectively-bypassable login), similar to the hard-coded password issue I fixed. But you're right that this also blocks the `NSH_LOGIN_PLATFORM` path, which doesn't need a password file at all. I think a better fix would be to scope the dependency to depend on FSUTILS_PASSWD || NSH_LOGIN_PLATFORM (or move the check into the NSH_LOGIN_PASSWD choice itself) so platform-verification boards aren't forced to enable it unnecessarily.

PR #3566, opened 2026-06-28
Replace O_RDOK with O_RDONLY after alias removal

xiaoxiang781216:
## Summary

The non-standard `O_RDOK`/`O_WROK` aliases are being removed from NuttX's `include/fcntl.h` (separate nuttx PR). They were simply aliases for `O_RDONLY`/`O_WRONLY`. This change replaces all remaining `O_RDOK` usage in the apps repository with the standard `O_RDONLY`.

Affected files:
- `examples/configdata/configdata_main.c`
- `nshlib/nsh_script.c`
- `platform/mikroe-stm32f4/mikroe_configdata.c`

## Impact

No functional change. `O_RDOK` was defined as `O_RDONLY`, so this is a pure text substitution that keeps the same behavior while using the standard POSIX flag name. Safe to merge independently — `O_RDONLY` is and remains defined in upstream nuttx.

## Testing

- `tools/checkpatch.sh -f` passes on all three changed files.
- Builds cleanly against apache/nuttx master.

### DIFF
--- a/examples/configdata/configdata_main.c
+++ b/examples/configdata/configdata_main.c
@@ -664,7 +664,7 @@ int main(int argc, FAR char *argv[])
 
   /* Open the /dev/config device */
 
-  g_fd = open("/dev/config", O_RDOK);
+  g_fd = open("/dev/config", O_RDONLY);
   if (g_fd == -1)
     {
       printf("ERROR: Failed to open /dev/config %d\n", -errno);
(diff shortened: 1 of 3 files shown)


Donny9 (approved):

cederom (approved):
Thank you @xiaoxiang781216 :-)

PR #3561, opened 2026-06-25
netutils/dropbear: initial Dropbear SSH server port for NuttX

FelipeMdeO:
## Summary

This PR is the first step of the Dropbear port to Apache NuttX.

At this stage, the port still uses Dropbear’s internal cryptographic library. Only the password-checking routines have been adapted to use native NuttX mechanisms.

As a second step, this application will be improved to use the cryptographic libraries provided by NuttX instead of Dropbear’s own crypto implementation.

With this application, a user can establish SSH connections to the hardware, allowing remote access to the system.

The documentation is already added to the mainline: [PR#19190](https://github.com/apache/nuttx/pull/19190)
Also, the defconfig to you be able test is available in the following commit: https://github.com/apache/nuttx/pull/19062

This PR is a "reboot" of https://github.com/apache/nuttx-apps/pull/3532. The first try added step 1 and 2 in the same PR adding a long discussion, so I drop the old PR and opened a clean PR here, ready to review. 

## Impact

New board configuration: the dropbear defconfig is additive and does not affect any existing configuration.
Is not expected any issue added by this code.

## Testing

Host: Linux x86_64, GCC RISC-V toolchain
Board: ESP32-C3 DevKit (rev 0.4)

Build:

```
./tools/configure.sh esp32c3-devkit:dropbear
make -j$(nproc)
<setup wifi ssid and pass>
make flash ESPTOOL_PORT=/dev/ttyUSB0
```

First-time user provisioning (serial console):

The NuttX passwd file lives on SPIFFS (/data/passwd) and is empty on a
fresh flash. Before the first SSH login, create a user from the NSH
serial console:

`nsh> useradd root <password>`

The ECDSA host key is generated automatically on first boot.

Boot log shows Dropbear listening after Wi-Fi association:

```
NuttShell (NSH) NuttX-12.6.0
nsh> loaded ECDSA P-256 host key from /data/dropbear_ecdsa_host_key
     using NuttX passwd auth at /data/passwd
     dropbear: listening on port 2222
```

SSH connection from the host:

```
$ ssh -p 2222 root@<board-ip>
root@<board-ip>'s password: <password>
NuttShell (NSH) NuttX-12.6.0
nsh>
```

PR #3573, opened 2026-06-29
games/NXDoom: Initial port of Chocolate DOOM to NuttX.

linguini1:
## Summary

This commit includes a (highly) modified version of Chocolate DOOM which can run on NuttX. The majority of the modifications were made to pass the NuttX style check. Some small modifications have been added to support keyboard input, render graphics to frame buffers and directly use the POSIX interfaces NuttX supplies, stripping out Windows/Mac stuff and any references to SDL.

**NOTE:** Sound is currently not supported in any capacity, nor is the networking stuff. A lot of Chocolate DOOM code was stripped out since it was unused. If there is a need/desire to add it back later, the original Chocolate DOOM source can be used as a reference.

**WARNING:** The NuttX keyboard codec is incredibly non-standard and so there are problems translating from X11 keys to NuttX ones to play DOOM. Right now, the CTRL key for firing doesn't work because the NuttX codec has no concept of it. The NuttX codec should be modified (and other input devices supported), but at the time of this port I am not sufficiently comfortable doing so since I am afraid of breaking other things in the kernel.

**NOTE:** This port (and likely the original DOOM) seems to be written with 32-bit computers in mind. As such, most things are given the type of natural `int`, even when a single byte might do. There are significant size optimizations that could be made to make this more suited to embedded devices that NuttX typically runs on.

This is part of my GSoC proposal; the final milestone of adding some cool user demos: https://github.com/apache/nuttx/issues/18507

The twin PR to the kernel containing a defconfig for playing NXDoom on sim and documenting the port is available here: https://github.com/apache/nuttx/pull/19239

## Impact

Play DOOM on NuttX!

## Testing

Played on NuttX simulator:

doom-on-sim


games/NXDoom/pcsound/pcsound.c
```diff
@@ -0,0 +1,129 @@
+/****************************************************************************
+ * apps/games/NXDoom/pcsound/pcsound.c
```
xiaoxiang781216, line 2: should we download doom from the original repo

acassis (dismissed):

games/NXDoom/pcsound/pcsound.c
linguini1: No, there haven't been any meaningful updates to Chocolate DOOM in several years. I also plan that NXDoom will become its own unique port that diverges quite heavily from the original Chocolate DOOM port in order to run well on embedded devices. The patches that would need to be applied to the original source to get it to run well would be really cumbersome to apply after downloading from the original repo.

I've kept some of the stuff (like pcsound) as a reference for when I (or others) go to implement the sound support. But I also gutted a ton of other stuff (SDL support, different sound mixers, support for game variants like "Hexen" and "Strife).

I plan to make other changes, like replace the checksum, hashing, fixed-point math, etc. stuff with NuttX-native libraries. So this port will really become more "nuttx-ified" over time, hence my plan to maintain a separate port.

xiaoxiang781216: Ok, could you add the Chocolate DOOM source code as the initial patch? so, we can track what's change you apply to the original code base.

linguini1: How might I do that? As in one commit is just the entire source of Chocolate Doom and then the patches after? Is there an easy way to pull in the chocolate Doom source as the first patch that I can use? That commit will fail CI btw.

xiaoxiang781216: in the same pr, this first patch is the origin code, the second one is your change.

acassis: @linguini1 please take a look at https://github.com/apache/nuttx-apps/pull/3561 he did it for Dropbear

linguini1: Thank you both! I'll try to do that today.

linguini1:
Okay @xiaoxiang781216, @acassis ; the first commit in the history is now the original Chocolate DOOM source before any modifications were applied!

games/NXDoom/.devcontainer/.gitignore
```diff
@@ -1,4 +0,0 @@
-*.xauth
```
xiaoxiang781216, line 1: could you avoid removing the source files which you don't support now? since other people may want to support it later.

linguini1: No because it is super time consuming already to modify all the source files to pass CI. Most of what I removed is not intended to be supported as part of the DOOM port (i.e., Hexen, Strife, etc.). In the future, if someone would like to bring in support for those other things, it should be quite easy to cherry pick the appropriate files from the Chocolate DOOM repo and modify them. Otherwise they just sit uncompiled but I have to modify them all to pass CI checks.

EDIT: also, a lot of what I removed I have no way to test (like the various sound libraries for different platforms) and are highly likely to not ever be compatible with NuttX. I want a minimal port here :)

cederom (approved):
Great work @linguini1 Big Thank You :-)
Lets just gather some more feedback and testing :-)

acassis (approved):

linguini1:
> Lets just gather some more feedback and testing :-)

It seems like it's already been run on some new targets besides the simulator! I think @aviralgarg05 got it working on ESP32 and @speccy88 on an RP2350 board this morning!

PR #3588, opened 2026-07-05
canutils/lely: fix unreliable download/unpack/patch logic

raiden00pl:
## Summary

- canutils/lely: fix unreliable download/unpack/patch logic

  Rebuild from a clean tree in one recipe so a stale state no longer breaks the build.
  This commit also simplify some build logic.
  
## Impact

more reliable build with make

## Testing

local build with master-slave examples for lely (not yet upstream)


xiaoxiang781216 (approved):

cederom (approved):
Thank you @raiden00pl :-)

PR #3593, opened 2026-07-05
examples: lp503x: Fix command argument handling

Old-Ding:
## Summary
- treat an empty `lp503x` colour command argument as the query case
- keep bank colour and brightness command values signed until range checks run
- avoid silently truncating invalid brightness values to 8 bits
- guard the `readline_stream()` result before using it as a buffer index

## Testing
- git diff --check origin/master..HEAD
- git show --check --stat --oneline HEAD

### DIFF
--- a/examples/lp503x/lp503x_main.c
+++ b/examples/lp503x/lp503x_main.c
@@ -119,9 +119,9 @@ static struct lp503x_cmd_s g_lp503x_cmds[] =
   },
   {
     "c",
-    "hex",
+    "[hex]",
     lp503x_cmd_setcolour,
-    LP503X_HELP_TEXT("Hex RGB colour required")
+    LP503X_HELP_TEXT("show or set current RGB colour")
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (dismissed):

examples/lp503x/lp503x_main.c
cederom, line 546: we may want to add information that NULL parameter is a query request?

cederom, line 524: In some places comparison is against `255` and here `MAX_BRIGHTNESS` please unify :-)

cederom:
Thank you @Old-Ding, please update function description/comments to match changes, also please provide Testting section with build and runtime confirmation as required by MAX_BRIGHTNESS :-)

cederom (approved):

xiaoxiang781216 (approved):

PR #3595, opened 2026-07-05
system/vncviewer: Drain truncated desktop names

Old-Ding:
## Summary
- keep the original RFB ServerInit desktop-name length separate from the copied length
- drain any desktop-name bytes that do not fit in `conn->name`
- reuse a small discard helper for skipped RFB payloads

## Why
The RFB ServerInit desktop name is a length-prefixed field. When the server advertises a name longer than the local buffer, vncviewer truncated the copy length before reading from the socket. That left the remaining name bytes queued as if they were the next RFB message, which can desynchronize subsequent framebuffer update parsing.

## Testing
- GitHub Actions `Lint` passed on the current head: https://github.com/apache/nuttx-apps/actions/runs/28947812560/job/85886182859
- attempted locally via Git Bash: `../nuttx/tools/checkpatch.sh -f system/vncviewer/rfb_protocol.c`; blocked because this Windows environment does not have `../nuttx/tools/nxstyle` built and `make` is not available
- runtime VNC repro was not run in this local Windows environment

### DIFF
--- a/system/vncviewer/rfb_protocol.c
+++ b/system/vncviewer/rfb_protocol.c
@@ -411,6 +411,36 @@ static int rfb_recv_exact(int sockfd, void *buf, size_t len)
 }
 
 /****************************************************************************
+ * Name: rfb_discard_exact
+ *
+ * Description:
+ *   Receive and discard exactly 'len' bytes from socket.
+ *
+ ****************************************************************************/
+
+static int rfb_discard_exact(int sockfd, size_t len)
+{
+  uint8_t buf[256];
+  size_t chunk;
+  int ret;
+
+  while (len > 0)
+    {
+      chunk = len > sizeof(buf) ? sizeof(buf) : len;
+
+      ret = rfb_recv_exact(sockfd, buf, chunk);
+      if (ret < 0)
+        {
+          return ret;
+        }
+
+      len -= chunk;
+    }
+
+  return OK;
+}
+
+/****************************************************************************
  * Name: rfb_send_exact
  *
  * Description:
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

cederom (changes requested):
Thank you @Old-Ding, looks good, just please update git commit signature email to valid address :-)

cederom:
@JianyuWang0623 could you please take a look? :-)

JianyuWang0623 (approved):

JianyuWang0623:
@Old-Ding Thanks for the fix — the analysis of the desktop-name desync looks correct to me.

One concern about the **Testing** section: the listed checks (`git diff --check`, `git show --check`) only validate formatting/whitespace, not the actual behavior this PR fixes. Since the bug is a runtime protocol-desync issue, could you share how it was reproduced and verified?

Specifically it would help to have:

1. **Repro setup** — which VNC server was used, and how a desktop name longer than `sizeof(conn->name) - 1` was produced (server name, or how you forced an oversized `desktop-name-length`).
2. **Before-fix evidence** — logs showing the desync, e.g. the `vncviewer: unknown message type ...` / `-EPROTO` path or a garbled framebuffer update caused by the leftover name bytes being parsed as the next message.
3. **After-fix evidence** — the same session working, including the `vncviewer: server desktop: "..." WxH` line and a stable framebuffer stream afterwards.

Also, since this touches parsing of an attacker-controllable length field, it would be good to confirm the behavior when the server advertises a very large `name-length` (the code now drains `name_len - copy_len` bytes unconditionally). A short note on whether that path was exercised would be reassuring.

Finally, please run `./tools/checkpatch.sh -f system/vncviewer/rfb_protocol.c` once — the Testing note says it was skipped due to no bash/WSL, but the new `rfb_discard_exact` block should be verified against the coding style before merge.

cederom (approved):

PR #3596, opened 2026-07-05
examples: nrf24l01_btle: Fix RX terminator bounds

Old-Ding:
## Summary
- reserve one extra byte for the RX buffer terminator
- keep the nRF24L01 read length capped at `NRF24L01_MAX_PAYLOAD_LEN`
- dump only the bytes actually received in the debug path

## Why
`nrf24_read()` appends a NUL terminator after reading a packet. The local buffer was exactly 32 bytes, which is the nRF24L01 maximum payload length. A full-size payload can therefore return 32 and make `rbuf[ret] = '\0'` write one byte past the buffer.

## Testing
- local compile/checkpatch not run: this Windows environment has no usable `gcc`/`clang`/`make` or WSL/bash shell

### DIFF
--- a/examples/nrf24l01_btle/nrf24l01_btle.c
+++ b/examples/nrf24l01_btle/nrf24l01_btle.c
@@ -484,9 +484,9 @@ int nrf24_read(int wl_fd)
 {
   int ret;
   uint32_t pipeno;
-  uint8_t rbuf[32];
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

cederom (changes requested):
Thank you @Old-Ding, looks good, please update git commit signature to valid email :-)

cederom (approved):

PR #3597, opened 2026-07-05
examples: Fix network echo receive terminators

Old-Ding:
## Summary

The netloop and poll TCP echo examples receive up to `IOBUFFER_SIZE` bytes and then append a NUL terminator before logging the received data. A full-size receive can therefore write one byte past the input buffer.

Reserve one extra byte for the buffers that are terminated after `recv()`, while keeping the receive limit and echo length capped at `IOBUFFER_SIZE`.

## Validation

- `git ls-files --eol examples/netloop/lo_main.c examples/netloop/lo_listener.c examples/poll/host.c examples/poll/net_listener.c examples/poll/net_reader.c`

### DIFF
--- a/examples/netloop/lo_listener.c
+++ b/examples/netloop/lo_listener.c
@@ -64,7 +64,7 @@ struct net_listener_s
   struct sockaddr_in addr;
   fd_set          master;
   fd_set          working;
-  char            buffer[IOBUFFER_SIZE];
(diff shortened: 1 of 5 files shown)


xiaoxiang781216 (approved):

cederom (changes requested):
Thank you @Old-Ding, please update your Singature to valid email in case we need to contact you :-)

cederom (approved):

PR #3599, opened 2026-07-05
testing: nettest: Fix TCP server terminator bounds

Old-Ding:
## Summary

The TCP nettest server receives up to `TEST_BUFFER_SIZE` bytes and then appends a NUL terminator before checking for the `exit` command. A full-size receive can therefore write one byte past the receive buffer.

Reserve one extra byte for the terminator while keeping the `recv()` limit and echo length capped at `TEST_BUFFER_SIZE`.

## Validation

- `git ls-files --eol testing/nettest/utils/nettest_tcpserver.c`

### DIFF
--- a/testing/nettest/utils/nettest_tcpserver.c
+++ b/testing/nettest/utils/nettest_tcpserver.c
@@ -47,7 +47,7 @@ struct nettest_listener_s
 {
   fd_set master;
   fd_set working;
-  char   buffer[TEST_BUFFER_SIZE];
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

cederom (changes requested):
Thank you @Old-Ding, looks good, please just updaye git commit signature to valid email :-)

cederom (approved):

PR #3600, opened 2026-07-05
netutils: smtp: Treat recv EOF as an error

Old-Ding:
## Summary

`recv()` returns zero when the peer closes a TCP connection cleanly. The SMTP response path only treated negative returns as errors, so an EOF could leave stale or empty response data to be checked as if a server reply had arrived.

Return `ERROR` on zero-length receives at each SMTP response point.

## Validation

- `git ls-files --eol netutils/smtp/smtp.c`

### DIFF
--- a/netutils/smtp/smtp.c
+++ b/netutils/smtp/smtp.c
@@ -3,7 +3,8 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  * SPDX-FileCopyrightText: 2015, 2020 Gregory Nutt. All rights reserved.
- * SPDX-FileCopyrightText: 2007, 2009, 2011, Gregory Nutt. All rights reserved.
+ * SPDX-FileCopyrightText: 2007, 2009, Gregory Nutt. All rights reserved.
+ * SPDX-FileCopyrightText: 2011, Gregory Nutt. All rights reserved.
  * SPDX-FileCopyrightText: 2004, Adam Dunkels. All rights reserved.
  * SPDX-FileContributor: Gregory Nutt <gnutt@nuttx.org>
  * SPDX-FileContributor: Adam Dunkels <adam@dunkels.com>
@@ -116,7 +117,7 @@ static const char g_smtpto[]             = "To: ";
 
 static inline int smtp_send_message(int sockfd, struct smtp_state *psmtp)
 {
-  if (recv(sockfd, psmtp->buffer, SMTP_INPUT_BUFFER_SIZE, 0) < 0)
+  if (recv(sockfd, psmtp->buffer, SMTP_INPUT_BUFFER_SIZE, 0) <= 0)
     {
       return ERROR;
     }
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (dismissed):

xiaoxiang781216:
@Old-Ding please fix:
```
../nuttx/tools/checkpatch.sh -c -u -m -g 4d5f60005c6d91c76e4a585fe3102aab50119ec6..HEAD
Error: /home/runner/work/nuttx-apps/nuttx-apps/apps/netutils/smtp/smtp.c:6:79: error: Long line found
Used config files:
    1: .codespellrc
```

netutils/smtp/smtp.c
```diff
@@ -116,7 +116,7 @@ static const char g_smtpto[]             = "To: ";
 
 static inline int smtp_send_message(int sockfd, struct smtp_state *psmtp)
 {
-  if (recv(sockfd, psmtp->buffer, SMTP_INPUT_BUFFER_SIZE, 0) < 0)
+  if (recv(sockfd, psmtp->buffer, SMTP_INPUT_BUFFER_SIZE, 0) <= 0)
```
cederom, line 120: Hmm wouldn't that return error on empty message? This looks like a functional change?

cederom (changes requested):
Thank you @Old-Ding :-)
* When no data received (recv() returns 0) do we want to return error are you sure? It looks there should be no processing but no error too? This looks like a functional change?
* Please add valid email as git commit signature.

Old-Ding:
Updated in `0c9459b2`: wrapped the long SPDX copyright line that checkpatch reported and added `Signed-off-by` to the follow-up commit. Both commits now use the same Old-Ding GitHub noreply identity.

For the `recv() == 0` behavior: this function expects a required SMTP server reply before every next protocol step. A zero return from `recv()` means the peer performed an orderly shutdown, so no reply was received for that step. Treating that as `ERROR` keeps the transaction from continuing after EOF and avoids checking whatever was already in `psmtp->buffer` from an earlier reply.

cederom:
* Thank you @Old-Ding for clarification that `0` means `EOT`.
* Please take a look at https://github.com/apache/nuttx/blob/master/CONTRIBUTING.md:
  * Please provide build and runtime logs confirming the fix is operational and works as expected.
  * Each git commit needs a message.
  * Each git commit needs a valid email signature.

xiaoxiang781216 (approved):

Old-Ding:
Updated in `ab9dc070`:

- Amended the follow-up commit so it has a descriptive message body and `Signed-off-by` using the Old-Ding GitHub noreply address.
- `check`, `Lint`, `labeler`, and the Build setup/select jobs are passing now; the remaining Build matrix jobs are still running/queued.

I do not have a local NuttX build/runtime environment on this Windows host (`gcc`, `make`, and `cmake` are not installed), so I am waiting for the GitHub Actions build logs. A runtime log still needs a real SMTP test environment; I have not attached one yet.

cederom:
"Old-Ding GitHub noreply address" is not a valid email address, please use email address that anyone can contact you when needed.

linguini1:
> I do not have a local NuttX build/runtime environment on this Windows host

Please try to follow the workflow documentation to setup a NuttX build/runtime environment. In order to contribute non-trivial changes to NuttX, you will need to be able to perform tests.

cederom (approved):

PR #3604, opened 2026-07-05
examples: nrf24l01_term: reserve RX terminator

Old-Ding:
## Summary
- Limit nrf24l01_term RX reads to `sizeof(buff) - 1`.
- Keep the trailing byte available for the `buff[ret] = '\0'` terminator.

## Root cause
`buff` is declared as `NRF24L01_MAX_PAYLOAD_LEN + 1`, but `read_pkt()` requested `sizeof(buff)` bytes before appending a NUL terminator. A full-length read could write the terminator one byte past `buff`.

## Testing
- `git diff --check`

Not run: local compile, because this environment does not have `make`, `cmake`, `gcc`, `clang`, or `cc` available.

### DIFF
--- a/examples/nrf24l01_term/nrf24l01_term.c
+++ b/examples/nrf24l01_term/nrf24l01_term.c
@@ -221,7 +221,7 @@ int read_pkt(int wl_fd)
   int ret;
   uint32_t pipeno;
 
-  ret = read(wl_fd, buff, sizeof(buff));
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

cederom (changes requested):
Thank you @Old-Ding , looks good, please add git commit message and valid email signature :-)

cederom (approved):

PR #3606, opened 2026-07-05
examples: wgetjson: check allocations before use

Old-Ding:
## Summary
- Check wgetjson response, JSON path, request buffer, and POST body allocations before use.
- Return or propagate allocation failures instead of dereferencing NULL buffers.

## Root cause
Several `malloc()` results in the wgetjson example were used immediately by `memcpy()`, `snprintf()`, `web_posts_str()`, `web_post_str()`, or `webclient_perform()` without first checking whether allocation succeeded. Low-memory paths could therefore crash before the existing cleanup logic ran.

## Testing
- `git diff --check`

Not run: local compile, because this environment does not have `make`, `cmake`, `gcc`, `clang`, or `cc` available.

### DIFF
--- a/examples/wgetjson/wgetjson_main.c
+++ b/examples/wgetjson/wgetjson_main.c
@@ -110,9 +110,14 @@ static int wgetjson_callback(FAR char **buffer, int offset, int datend,
       return 0;
     }
 
-  if (!g_json_buff)
+  if (g_json_buff == NULL)
     {
       g_json_buff = malloc(len + 1);
+      if (g_json_buff == NULL)
+        {
+          return -ENOMEM;
+        }
+
       memcpy(g_json_buff, &((*buffer)[offset]), len);
       g_json_buff[len] = 0;
       g_json_bufflen = len;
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (dismissed):

xiaoxiang781216:
@Old-Ding please fix:
```
c1da0ee1f examples: wgetjson: check allocations before use
../nuttx/tools/checkpatch.sh -c -u -m -g 4d5f60005c6d91c76e4a585fe3102aab50119ec6..HEAD
❌ Missing git commit message
Used config files:
    1: .codespellrc
Error: Process completed with exit code 1.
```

examples/wgetjson/wgetjson_main.c
```diff
@@ -113,6 +113,11 @@ static int wgetjson_callback(FAR char **buffer, int offset, int datend,
   if (!g_json_buff)
     {
       g_json_buff = malloc(len + 1);
+      if (!g_json_buff)
```
cederom: Do we prefer `g_json_buff == NULL` here @acassis ? :-)

cederom (changes requested):
Thank you @Old-Ding, please git commit message and valid email as git commit signature.

examples/wgetjson/wgetjson_main.c
acassis: Yes, since it it not boolean, it is better to test with the right type

cederom (approved):

xiaoxiang781216 (approved):

PR #3609, opened 2026-07-05
system: audio: bound command filenames

Old-Ding:
## Summary
- Tokenize the nxplayer `playraw` filename argument into a `PATH_MAX` buffer before parsing optional raw audio parameters.
- Tokenize the nxrecorder `recordraw` and `record` filename arguments into `PATH_MAX` buffers before parsing optional raw audio parameters.
- Preserve the existing optional raw audio parameter parsing behavior.

## Testing
- `git diff --check`
- GitHub Actions `check`, `Lint`, `Fetch-Source`, and `labeler` passed after the latest update.

## CI notes
- Current `Linux (sim-02)` failure is in NuttX `sim/login`: `CONFIG_BOARD_ETC_ROMFS_PASSWD_PASSWORD is not set`. That configuration is outside this PR's `nxplayer` / `nxrecorder` parsing diff; later matrix jobs were cancelled after that failure.
- Thread-aware review collection shows no active unresolved review threads after the `PATH_MAX` tokenization update, although GitHub can continue to show the prior changes-requested decision until reviewer state refreshes.

### DIFF
--- a/system/nxplayer/nxplayer_main.c
+++ b/system/nxplayer/nxplayer_main.c
@@ -317,10 +317,21 @@ static int nxplayer_cmd_playraw(FAR struct nxplayer_s *pplayer, char *parg)
   int bpsamp = 0;
   int samprate = 0;
   int chmap = 0;
-  char filename[128];
+  size_t len;
+  char filename[PATH_MAX];
 
-  sscanf(parg, "%s %d %d %d %d", filename, &channels, &bpsamp,
-                                 &samprate, &chmap);
+  parg += strspn(parg, " \t\r\n");
+  len = strcspn(parg, " \t\r\n");
+  if (len >= sizeof(filename))
+    {
+      len = sizeof(filename) - 1;
+    }
+
+  memcpy(filename, parg, len);
+  filename[len] = '\0';
+
+  sscanf(parg + len, "%d %d %d %d", &channels, &bpsamp,
+         &samprate, &chmap);
 
   /* Try to play the file specified */
 
--- a/system/nxrecorder/nxrecorder_main.c
+++ b/system/nxrecorder/nxrecorder_main.c
@@ -190,10 +190,21 @@ static int nxrecorder_cmd_recordraw(FAR struct nxrecorder_s *precorder,
   int bpsamp = 0;
   int samprate = 0;
   int chmap = 0;
-  char filename[128];
+  size_t len;
+  char filename[PATH_MAX];
 
-  sscanf(parg, "%s %d %d %d %d", filename, &channels, &bpsamp,
-                                 &samprate, &chmap);
+  parg += strspn(parg, " \t\r\n");
+  len = strcspn(parg, " \t\r\n");
+  if (len >= sizeof(filename))
+    {
+      len = sizeof(filename) - 1;
+    }
+
+  memcpy(filename, parg, len);
+  filename[len] = '\0';
+
+  sscanf(parg + len, "%d %d %d %d", &channels, &bpsamp,
+         &samprate, &chmap);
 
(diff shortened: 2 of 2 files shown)


system/nxplayer/nxplayer_main.c
```diff
@@ -312,15 +312,34 @@ static int nxplayer_cmd_play(FAR struct nxplayer_s *pplayer, char *parg)
...
+  FAR char *params;
   int ret;
   int channels = 0;
   int bpsamp = 0;
   int samprate = 0;
   int chmap = 0;
+  size_t filename_len;
   char filename[128];
 
-  sscanf(parg, "%s %d %d %d %d", filename, &channels, &bpsamp,
```
xiaoxiang781216, line 322: what's problem to call sscanf

cederom: looks like `%s` overflow protection first, then using `sscanf()` with only `%d` parameters? :-)

cederom (dismissed):
Thank you @Old-Ding, good catch, please use valid email in git commit signature :-)

acassis (dismissed):

system/nxplayer/nxplayer_main.c
xiaoxiang781216: could protected by %xxs

xiaoxiang781216 (dismissed):

acassis (approved):

system/nxplayer/nxplayer_main.c
```diff
@@ -319,8 +319,8 @@ static int nxplayer_cmd_playraw(FAR struct nxplayer_s *pplayer, char *parg)
   int chmap = 0;
   char filename[128];
 
-  sscanf(parg, "%s %d %d %d %d", filename, &channels, &bpsamp,
-                                 &samprate, &chmap);
+  sscanf(parg, "%127s %d %d %d %d", filename, &channels, &bpsamp,
```
acassis: @Old-Ding I think instead using hard-code filename[128] we should use CONFIG_PATH_MAX

acassis (changes requested):

system/nxplayer/nxplayer_main.c
acassis: I think there are many places in the source tree where we are not obeying the limits that could be defined in the menuconfig, here is just one of them

acassis: I think boards/arm/lc823450/lc823450-xgevk/configs/audio/defconfig defines CONFIG_PATH_MAX=765 for this reason, but this code is not following it.

system/nxplayer/nxplayer_main.c
```diff
@@ -317,10 +317,11 @@ static int nxplayer_cmd_playraw(FAR struct nxplayer_s *pplayer, char *parg)
   int bpsamp = 0;
   int samprate = 0;
   int chmap = 0;
-  char filename[128];
+  char format[32];
+  char filename[PATH_MAX];
 
-  sscanf(parg, "%s %d %d %d %d", filename, &channels, &bpsamp,
-                                 &samprate, &chmap);
+  snprintf(format, sizeof(format), "%%%ds %%d %%d %%d %%d", PATH_MAX - 1);
```
xiaoxiang781216: why not use PATH_MAX directly

xiaoxiang781216:
@Old-Ding let's squash into one patch.

acassis (approved):

xiaoxiang781216 (approved):

xiaoxiang781216:
> Thank you @Old-Ding, good catch, please use valid email in git commit signature :-)

does the author decide not to publish the personal information? @cederom .

PR #3611, opened 2026-07-06
{bp-3589} apps/nsh: Don't disable error by default

jerpelea:
## Summary

NSH error shouldn't be disabled by default, even when in SMALL Memory because it will make difficult to discover why things are not working. It should be disabled manually by experienced dev.

## Impact

RELEASE

## Testing

CI


jerpelea:
@cederom ping

acassis (approved):

PR #3616, opened 2026-07-07
games/NXDoom: use default glob implementation

linguini1:
## Summary

Uses the DOOM glob implementation based on dirent. This removes the compile time
warning for no native glob implementation.

## Impact

More featured DOOM port, and removed a compiler warning during build.

## Testing

Tested using the `sim:nxdoom` config:

image

### DIFF
--- a/games/NXDoom/Kconfig
+++ b/games/NXDoom/Kconfig
@@ -9,6 +9,7 @@ config GAMES_NXDOOM
 	depends on ALLOW_GPL_COMPONENTS
 	depends on VIDEO_FB
 	depends on CRYPTO
+	depends on LIBC_LOCALE
 	---help---
 		Play DOOM on NuttX!
 
--- a/games/NXDoom/src/i_glob.c
+++ b/games/NXDoom/src/i_glob.c
@@ -33,15 +33,10 @@
 #include "i_glob.h"
 #include "m_misc.h"
 
-#if defined(HAVE_DIRENT_H)
+#include <dirent.h>
+
 #include <dirent.h>
 #include <sys/stat.h>
-#elif defined(__WATCOMC__)
-/* Watcom has the same API in a different header. */
-#include <direct.h>
-#else
-#define NO_DIRENT_IMPLEMENTATION
-#endif
 
 /****************************************************************************
  * Pre-processor Definitions
@@ -72,8 +67,6 @@ struct glob_s
  * Private Functions
  ****************************************************************************/
 
-#ifndef NO_DIRENT_IMPLEMENTATION
-
 /* Only the fields d_name and (as an XSI extension) d_ino are specified
  * in POSIX.1.  Other than Linux, the d_type field is available mainly
  * only on BSD systems.  The remaining fields are available on many, but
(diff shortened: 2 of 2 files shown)


cederom (approved):
Thank you @linguini1 =)

acassis (approved):

linguini1:
FYI @cederom, this patch is necessary to get the NXDoom port running on the RPi4B. It also appears I will need to fix the EMMC2 write/read size limit first as well, as the WAD lumps are big enough that they trigger the same error in https://github.com/apache/nuttx/issues/17255

So I am now splitting my attention between the SDMMC issue and getting PCM audio working.

games/NXDoom/Kconfig
```diff
@@ -9,6 +9,7 @@ config GAMES_NXDOOM
 	depends on ALLOW_GPL_COMPONENTS
 	depends on VIDEO_FB
 	depends on CRYPTO
+	depends on LIBC_LOCALE
```
acassis, line 12: @linguini1 why wasn't it necessary before? How much will it increase the code?

linguini1: It was, but I believe `sim` includes this by default so I did not get any compile-time issues in my earlier testing. I am not certain what the increase is in code size with this addition but I believe its relatively small. Other optimizations are more important.

cederom:
> FYI @cederom, this patch is necessary to get the NXDoom port running on the RPi4B. It also appears I will need to fix the EMMC2 write/read size limit first as well, as the WAD lumps are big enough that they trigger the same error in [apache/nuttx#17255](https://github.com/apache/nuttx/issues/17255)
> 
> So I am now splitting my attention between the SDMMC issue and getting PCM audio working.

Ah I remember we were testing that during initial SD implementation hope it wont be too hard to fix! :-P

cederom:
Btw @linguini1  I use https://github.com/apache/nuttx/issues/17303 for flashing NuttX to rPI-4B without SD card just over the usb-bootloader and the `rpiboot` on the host side.. maybe that would help evade SD problems just for comparison or when needed? :-)

linguini1:
> Ah I remember we were testing that during initial SD implementation hope it wont be too hard to fix! :-P

I hope so too; I would think there should be some way to limit the size of the write chunks but I couldn't find one yet. If I spend more time on it I might get somewhere :)

> Btw @linguini1 I use https://github.com/apache/nuttx/issues/17303 for flashing NuttX to rPI-4B without SD card just over the usb-bootloader and the rpiboot on the host side.. maybe that would help evade SD problems just for comparison or when needed? :-)

That is useful, but I don't think it will work around this issue. The problem is because the EMMC2 hardware can only handle a maximum block size of 1023 at a time. So I will need to split larger chunks into smaller ones, no matter what is connected. I should maybe look into getting one of those eventually; it's tiring to swap the SD card every time haha!

cederom:
Yeah, its very handy, just need some bootloader setup prior that to enable usb boot :-)

Unfortunately usb enumeration of lots of devices is non linear and hard to know which device is what when you have like 30 of them at one.. and uhubctl still is not fully functional.. so testing is still kind of manual.. but this is another story :D

PR #3624, opened 2026-07-09
examples: Implement drivers for microwindows

Acfboy:
Hello the community! Here is the work I have done for the first half of GSoC, now complete and ready for review.

## Work Completed

The GDI-level Microwindows core, drivers, and `mwdemo` are functional on QEMU x86-64 and the simulator.

### Build System Integration

- Based on the initial proof-of-concept branch by @ppisa, the Microwindows core (engine, drivers, precompiled bitmap fonts) has been successfully compiled into the NuttX apps library.
- Uses `MWCONFIG_FILE="mwconfig.nuttx"` to inject NuttX-specific configuration into Microwindows without touching upstream headers.
- Image resources used by `mwdemo.c` are compiled into binary using the `convbmp` tool during the build.

### NuttX Hardware Drivers

Framebuffer driver (`scr_nuttx.c`), keyboard drivers (`kbd_nuttx_event.c` for event-mode and `kbd_nuttx_raw.c` for raw byte-stream), mouse driver (`mou_nuttx_mouse.c`), and touchscreen driver (`mou_nuttx_ts.c`) have been implemented and submitted to upstream Microwindows mainline https://github.com/ghaerr/microwindows/pull/193. The `graphics/microwindows/Makefile` selects the appropriate drivers via `ARCH=NUTTX` and Kconfig-driven `KEYBOARD`/`MOUSE` variables.

### Kconfig Options

Flexible Kconfig for keyboard driver: event-mode (`/dev/kbd`), raw byte-stream via kbd_codec (`/dev/kbda`), none, or custom. For mouse/touchscreen: relative mouse (`/dev/mouse0`), touchscreen (`/dev/input0`), none, or custom. Custom options allow BSP/app-level overrides without modifying the Microwindows build logic.

### Demo Application

Ported `mwdemo.c` (a complex Windows-like demo) into a standalone NuttX example application. The demo runs successfully on both `qemu-intel64:mw` and the NuttX simulator (`sim:mw`).

I have also tested the demo with `#define CONTROL 1` enabled in `mwdemo.c`, which represents a more complex use case.

### Upstream Bug Fixes

Several issues were discovered, analysed, and most of them already fixed:

| Issue | Root Cause | Fix |
|-------|-----------|-----|
| Animation freezing in mwdemo | `MwSelect()` blocked forever when a timer had already expired, because the code assumed no timer existed when `GdGetNextTimeout` returned false. | Patch by ppisa, correctly handle the case where `timeout == 0` and return immediately. |
| Cursor leaving a blue ghost image | `#if defined(ELKS)` in `mwtypes.h` forced `MWPIXELVALHW` to `unsigned char` even when `ELKS=0` (intended to disable it). This truncated 32-bit pixel values read from the framebuffer. | Changed to `#if ELKS` |
| System freeze when dragging a window caption | During drag, `MwSelect` switches to polling (`select(…, timeout=0)`). The USB input thread had a lower priority (50) than the application (100), so it never got CPU time to generate events. | Short-term: adjusted thread priorities in defconfigs. A permanent solution in Microwindows may be discussed later. |
| Mouse drift in simulator after dragging | `TOUCH_POS_VALID` flag was not being checked in the touchscreen driver, causing uninitialized coordinate values to be used. | Fixed in `mou_nuttx.c` by checking `TOUCH_POS_VALID` before using touch coordinates. |
| KEYCODE_xxx and ASCII range conflict |NuttX's `KEYCODE_xxx` macros (e.g., `KEYCODE_PAUSE = 0x20`) overlap with ASCII characters, making it impossible to distinguish special keys from normal input in the event-mode keyboard driver.| **Still under discussions**  https://github.com/apache/nuttx/issues/19527|

### qemu-intel64:mw

```
make distclean
./tools/configure.sh qemu-intel64:mw
make -j$(nproc)
```

Then create a bootable disk following the [documentation](https://nuttx.apache.org/docs/12.5.0/platforms/x86_64/qemu/boards/qemu-intel64/index.html) and run:

```
qemu-system-x86_64 -cpu host -enable-kvm -m 2G \
        -cdrom boot_mw.iso \
        -serial stdio -monitor none -no-reboot -device qemu-xhci -device usb-kbd -device usb-mouse
```

### sim:mw

```
make distclean
./tools/configure.sh sim:mw
make -j$(nproc)
./nuttx
```

### Notice

- You can set `#define CONTROL 1` manually in `mwdemo.c` to test more complex case.
- Most of the `nxstyle` problems in mwdemo.c are fixed, except the names of Win32 API.
- A draft document for microwindows is at https://github.com/Acfboy/nuttx/commit/f4d9316acd629f60783fc83df7f8c031c9754c19
- The Nano-X X11 supprot is still work in progress.

### DIFF
--- a/examples/microwindows/Make.defs
+++ b/examples/microwindows/Make.defs
@@ -1,0 +1,25 @@
+############################################################################
+# apps/examples/microwindows/Make.defs
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifneq ($(CONFIG_EXAMPLES_MICROWINDOWS),)
+CONFIGURED_APPS += $(APPDIR)/examples/microwindows
+endif
--- a/examples/microwindows/Makefile
+++ b/examples/microwindows/Makefile
@@ -1,0 +1,55 @@
+############################################################################
+# apps/examples/microwindows/Makefile
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+PROGNAME  = $(CONFIG_EXAMPLES_MICROWINDOWS_PROGNAME)
+PRIORITY  = $(CONFIG_EXAMPLES_MICROWINDOWS_PRIORITY)
+STACKSIZE = $(CONFIG_EXAMPLES_MICROWINDOWS_STACKSIZE)
+MODULE    = $(CONFIG_EXAMPLES_MICROWINDOWS)
+
+MAINSRC = mwdemo_main.c
+CSRCS += mwdemo.c
+MW_IMAGES = penguin microwin
+CSRCS += $(MW_IMAGES:%=$(MW_IMGDIR)/%.c)
+
+# Generate image .c files from .bmp
+MW_SRC = $(APPDIR)/graphics/microwindows/microwindows/src
+MW_IMGDIR = images
+MW_CONVBMP = convbmp$(HOSTEXEEXT)
+
+$(MW_CONVBMP): $(MW_SRC)/images/tools/convbmp.c
+	$(HOSTCC) $(HOSTCFLAGS) -I$(MW_SRC)/include $< -o $@
+
+$(MW_IMGDIR)/%.c: $(MW_SRC)/images/demos/mwin/%.bmp $(MW_CONVBMP)
+	mkdir -p $(MW_IMGDIR)
+	$(CURDIR)/$(MW_CONVBMP) -o $@ $<
+
+clean::
+	$(call DELFILE, $(MW_IMAGES:%=$(MW_IMGDIR)/%.c))
+	$(Q) rmdir $(MW_IMGDIR) 2>/dev/null || true
+	$(call DELFILE, $(MW_CONVBMP))
+
+CFLAGS += -Wno-undef 
+CFLAGS += -DNOGDI
+
+include $(APPDIR)/Application.mk
--- a/examples/microwindows/mwdemo.c
+++ b/examples/microwindows/mwdemo.c
@@ -1,0 +1,333 @@
+/****************************************************************************
+ * apps/examples/microwindows/mwdemo.c
+ *
+ * Copyright (c) 1999, 2000, 2001, 2010 Greg Haerr <greg@censoft.com>
+ * Portions Copyright (c) 2002 by Koninklijke Philips Electronics N.V.
+ *
+ * Demo program for Microwindows
+ *
+ *  GB: 10-14-2004: Modified to store degrees data on each window,
+ *                  for new Timers features.
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <stdlib.h>
+#include <uni_std.h>
+#define MWINCLUDECOLORS
+#include <windows.h>
+#include <device.h>
+#include <graph3d.h>
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+typedef struct
+{
+  int startdegrees;
+  int enddegrees;
+} demoWndData;
+
+typedef demoWndData *pdemoWndData;
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+extern MWIMAGEHDR image_penguin;
+PMWIMAGEHDR g_image = &image_penguin;
+
+#define APPCLASS    "test"
+#define APPCHILD    "test2"
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+LRESULT CALLBACK WndProc(HWND hwnd, UINT uMsg, WPARAM wp, LPARAM lp);
+
+LRESULT CALLBACK ChildWndProc(HWND hwnd, UINT uMsg, WPARAM wp, LPARAM lp);
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+int MwUserInit(int ac, char **av)
+{
+  /* test user init procedure - do nothing */
+
+  return 0;
+}
+
+int RegisterAppClass(HINSTANCE hInstance)
+{
+  WNDCLASS wc;
+
+  /* register builtin controls and dialog classes */
+
+  MwInitializeDialogs(hInstance);
+
+  wc.style = CS_DBLCLKS | CS_VREDRAW | CS_HREDRAW;
+  wc.lpfnWndProc = (WNDPROC) WndProc;
+  wc.cbClsExtra = 0;
+  wc.cbWndExtra = sizeof(LONG_PTR);
+  wc.hInstance = 0;
+  wc.hIcon = 0;                 /* LoadIcon(GetHInstance(), MAKEINTRESOURCE(
+                                 * 1)); */
+  wc.hCursor = LoadCursor(NULL, IDC_ARROW);
+  wc.hbrBackground = (HBRUSH) GetStockObject(LTGRAY_BRUSH);
+  wc.lpszMenuName = NULL;
+  wc.lpszClassName = APPCLASS;
+  RegisterClass(&wc);
+
+  wc.lpfnWndProc = (WNDPROC) ChildWndProc;
+  wc.lpszClassName = APPCHILD;
+  return RegisterClass(&wc);
+}
+
+HWND CreateAppWindow(void)
+{
+  HWND hwnd;
+  static int nextid = 1;
+  int width;
+  int height;
+  RECT r;
+  GetWindowRect(GetDesktopWindow(), &r);
+  width = height = r.right / 2;
+
+  hwnd = CreateWindowEx(0L, APPCLASS,
+                        "Microwindows Application",
+                        WS_OVERLAPPEDWINDOW | WS_VISIBLE,
+                        CW_USEDEFAULT, CW_USEDEFAULT,
+                        width, height,
+                        NULL, (HMENU)(LONG_PTR)nextid++, NULL, NULL);
+
+  if (hwnd && (nextid & 03) != 2)
+    {
+      CreateWindowEx(0L, APPCHILD,
+                     "",
+                     WS_BORDER | WS_CHILD | WS_VISIBLE,
+                     4, 4, width / 3 - 6, height / 3,
+                     hwnd, (HMENU)2, NULL, NULL);
+      CreateWindowEx(0L, APPCHILD,
+                     "",
+                     WS_BORDER | WS_CHILD | WS_VISIBLE,
+                     width / 3, height / 3,
+                     width / 3 - 6, height / 3,
+                     hwnd, (HMENU)3, NULL, NULL);
+      CreateWindowEx(0L, APPCHILD,
+                     "",
+                     WS_BORDER | WS_CHILD | WS_VISIBLE,
+                     width * 3 / 5, height * 3 / 5,
+                     width * 2 / 3, height * 2 / 3,
+                     hwnd, (HMENU)4, NULL, NULL);
+      CreateWindowEx(0L, "EDIT",
+                     "OK",
+                     WS_BORDER | WS_CHILD | WS_VISIBLE,
+                     width * 5 / 8, 10,
+                     100, 18,
+                     hwnd, (HMENU)5, NULL, NULL);
+      CreateWindowEx(0L, "PROGBAR",
+                     "OK",
+                     WS_BORDER | WS_CHILD | WS_VISIBLE,
+                     width * 5 / 8, 32,
+                     100, 18,
+                     hwnd, (HMENU)6, NULL, NULL);
+
+      HWND hlist = CreateWindowEx(0L, "LISTBOX",
+                                  "OK",
+                                  WS_HSCROLL | WS_VSCROLL |
+                                  WS_BORDER | WS_CHILD |
+                                  WS_VISIBLE,
+                                  width * 5 / 8, 54,
+                                  100, 48,
+                                  hwnd, (HMENU)7, NULL, NULL);
+      SendMessage(hlist, LB_ADDSTRING, 0,
+                  (LPARAM)(LPSTR)"Cherry");
+      SendMessage(hlist, LB_ADDSTRING, 0,
+                  (LPARAM)(LPSTR)"Apple");
+      SendMessage(hlist, LB_ADDSTRING, 0,
+                  (LPARAM)(LPSTR)"Orange");
+
+      CreateWindowEx(0L, "BUTTON",
+                     "Cancel",
+                     BS_AUTOCHECKBOX | WS_CHILD | WS_VISIBLE,
+                     width * 5 / 8 + 50, 106,
+                     50, 14,
+                     hwnd, (HMENU)8, NULL, NULL);
+    }
+
+  return hwnd;
+}
+
+LRESULT CALLBACK ChildWndProc(HWND hwnd, UINT msg, WPARAM wp, LPARAM lp)
+{
+  RECT rc;
+  PAINTSTRUCT ps;
+
+  switch (msg)
+    {
+    case WM_PAINT:
+      BeginPaint(hwnd, &ps);
+      GetClientRect(hwnd, &rc);
+
+      HDC hdcMem;
+      HBITMAP hbmp, hbmpOrg;
+
+      /* redirect painting to offscreen dc, then use blit function */
+
+      hdcMem = CreateCompatibleDC(ps.hdc);
+
+      /* Note: rc.right, rc.bottom happens to be smaller than image
+       * width/height.  We use the image size, so we can stretchblit
+       * from the whole image.
+       */
+
+      hbmp = CreateCompatibleBitmap(hdcMem, g_image->width,
+                                    g_image->height);
+      hbmpOrg = SelectObject(hdcMem, hbmp);
+
+      /* draw onto offscreen dc */
+
+      DrawDIB(hdcMem, 0, 0, g_image);
+
+      /* blit offscreen with physical screen */
+
+      StretchBlt(ps.hdc, 0, 0, rc.right * 4 / 5,
+                 rc.bottom * 4 / 5, hdcMem,
+                 0, 0, g_image->width, g_image->height, MWROP_COPY);
+      DeleteObject(SelectObject(hdcMem, hbmpOrg));
+      DeleteDC(hdcMem);
+      EndPaint(hwnd, &ps);
+      break;
+    default:
+      return DefWindowProc(hwnd, msg, wp, lp);
+    }
+
+  return 0;
+}
+
+LRESULT CALLBACK WndProc(HWND hwnd, UINT msg, WPARAM wp, LPARAM lp)
+{
+  PAINTSTRUCT ps;
+  HDC hdc;
+  RECT rc;
+  static int countup = 1;
+  int id;
+  static vec1 gx;
+  static vec1 gy;
+  static POINT mousept;
+  pdemoWndData pData;
+
+  switch (msg)
+    {
+    case WM_CREATE:
+      pData = (pdemoWndData) malloc(sizeof(demoWndData));
+      SetWindowLongPtr(hwnd, 0, (LONG_PTR) pData);
+      mousept.x = 60;
+      mousept.y = 20;
+      pData->startdegrees = 0;
+      pData->enddegrees = 30;
+      SetTimer(hwnd, 1, (50 + random() % 250), NULL);
+      break;
+
+    case WM_TIMER:
+      pData = (pdemoWndData)(LONG_PTR)GetWindowLongPtr(hwnd, 0);
+      GetClientRect(hwnd, &rc);
+      if (countup)
+        {
+          mousept.y += 20;
+          if (mousept.y >= rc.bottom)
+            {
+              mousept.y -= 20;
+              countup = 0;
+            }
+        }
+      else
+        {
+          mousept.y -= 20;
+          if (mousept.y < 20)
+            {
+              mousept.y += 20;
+              countup = 1;
+            }
+        }
+
+      SendMessage(hwnd, WM_MOUSEMOVE, 0,
+                  MAKELONG(mousept.x, mousept.y));
+      break;
+    case WM_DESTROY:
+      KillTimer(hwnd, 1);
+      pData = (pdemoWndData)(LONG_PTR)GetWindowLongPtr(hwnd, 0);
+      free(pData);
+      SetWindowLongPtr(hwnd, 0, (LONG_PTR)0);
+      break;
+    case WM_SIZE:
+      break;
+    case WM_MOVE:
+      break;
+
+    case WM_ERASEBKGND:
+      if ((GetWindowLong(hwnd, GWL_ID) & 03) == 1)
+        {
+          return 1;
+        }
+
+      return DefWindowProc(hwnd, msg, wp, lp);
+    case WM_PAINT:
+      hdc = BeginPaint(hwnd, &ps);
+      id = (int)GetWindowLong(hwnd, GWL_ID) & 03;
+      init3(hdc, id == 1 ? hwnd : NULL);
+      switch (id)
+        {
+        case 0:
+          rose(1.0, 7, 13);
+          break;
+        case 1:
+          look3(-2 * gx, -2 * gy, 1.2);
+          drawgrid(-8.0, 8.0, 10, -8.0, 8.0, 10);
+          break;
+        case 2:
+          setcolor3(BLACK);
+          circle3(1.0);
+          break;
+        case 3:
+          setcolor3(BLUE);
+          daisy(1.0, 20);
+          break;
+        }
+
+      paint3(hdc);
+      EndPaint(hwnd, &ps);
+      break;
+
+    case WM_MOUSEMOVE:
+      pData = (pdemoWndData)(LONG_PTR)GetWindowLongPtr(hwnd, 0);
+      if ((GetWindowLong(hwnd, GWL_ID) & 03) == 1)
+        {
+          POINT pt;
+
+          POINTSTOPOINT(pt, lp);
+          GetClientRect(hwnd, &rc);
+          gx = (vec1) pt.x / rc.right;
+          gy = (vec1) pt.y / rc.bottom;
+          InvalidateRect(hwnd, NULL, FALSE);
+          mousept.x = pt.x;
+          mousept.y = pt.y;
+        }
+      break;
+    case WM_LBUTTONDOWN:
+      break;
+    case WM_LBUTTONUP:
+      break;
+    case WM_RBUTTONDOWN:
+      break;
+    default:
+      return DefWindowProc(hwnd, msg, wp, lp);
+    }
+
+  return 0;
+}
--- a/examples/microwindows/mwdemo_main.c
+++ b/examples/microwindows/mwdemo_main.c
@@ -1,0 +1,101 @@
+/****************************************************************************
+ * apps/examples/microwindows/mwdemo_main.c
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include <time.h>
+#include <stdlib.h>
+#include <uni_std.h>
+#include <windows.h>
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+extern MWIMAGEHDR image_microwin;
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+extern int RegisterAppClass(HINSTANCE hInstance);
+extern HWND CreateAppWindow(void);
+
+#define APPCLASS    "test"
+#define APPCHILD    "test2"
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+int main(int argc, FAR char *argv[])
+{
+  MSG msg;
+  HWND hwnd;
+  RECT rc;
+
+  invoke_WinMain_Start(argc, argv);
+
+  srandom(time(NULL));
+  RegisterAppClass(0);
+  GetWindowRect(GetDesktopWindow(), &rc);
+
+  /* create penguin window */
+
+  CreateWindowEx(0L, APPCHILD, "",
+                 WS_BORDER | WS_VISIBLE,
+                 rc.right - 130 - 1, rc.bottom - 153 - 1,
+                 130, 153,
+                 GetDesktopWindow(), (HMENU)1000, NULL, NULL);
+
+  CreateAppWindow();
+  CreateAppWindow();
+  CreateAppWindow();
+  CreateAppWindow();
+  CreateAppWindow();
+  CreateAppWindow();
+  CreateAppWindow();
+  CreateAppWindow();
+
+  /* set background wallpaper */
+
+  MwSetDesktopWallpaper(&image_microwin);
+  hwnd = CreateAppWindow();
+  GetWindowRect(hwnd, &rc);
+  OffsetRect(&rc, 50, 50);
+  MoveWindow(hwnd, rc.left, rc.top,
+             rc.bottom - rc.top,
+             rc.right - rc.left, TRUE);
+
+  while (GetMessage(&msg, NULL, 0, 0))
+    {
+      TranslateMessage(&msg);
+      DispatchMessage(&msg);
+    }
+
+  invoke_WinMain_End();
+
+  return 0;
+}
--- a/graphics/microwindows/Make.defs
+++ b/graphics/microwindows/Make.defs
@@ -1,0 +1,33 @@
+############################################################################
+# apps/graphics/microwindows/Make.defs
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifneq ($(CONFIG_GRAPHICS_MICROWINDOWS),)
+CONFIGURED_APPS += $(APPDIR)/graphics/microwindows
+
+# It allows `microwindows/src/include` import.
+
+CFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/include
+CXXFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/include
+
+CFLAGS += -DMWCONFIG_FILE='"mwconfig.nuttx"'
+
+endif
--- a/graphics/microwindows/Makefile
+++ b/graphics/microwindows/Makefile
@@ -1,0 +1,154 @@
+############################################################################
+# apps/graphics/microwindows/Makefile
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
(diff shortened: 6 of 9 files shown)


ppisa:
Thanks to @Acfboy for the work and this draft request for discussion and @ghaerr for consultations and support on Microwidows side.

I am leaving for four days now, but there are some my thoughts, the NuttX Microwidows screen, mouse and keyboard drivers should be submitted with appropriate build enable options to the mainline Microwindows https://github.com/ghaerr/microwindows before the final nuttx-apps pull request.

examples/microwindows/Make.defs
```diff
@@ -1,5 +1,5 @@
 ############################################################################
-# apps/examples/rng90/Make.defs
+# apps/examples/hello/Make.defs
```
acassis: @Acfboy hello -> microwindows

acassis:
@Acfboy please fix the issues on mwdemo:
image

simbit18:
Hi @Acfboy , could you also add support for building with CMake?

Acfboy:
> Hi @Acfboy , could you also add support for building with CMake?

Thank you for the reminder. I will add CMake support later.

acassis:
@Acfboy could you spend some time fixing and improving this PR? It still as Draft. It is important to get it merged and let more people to test and review it. Also please remember to include a proper nuttx/Documentation about nanox/microwindows support on NuttX

Acfboy:
Hi @acassis , thank for the reminder.  I agree I should submit the previous work as a proper PR for the community to review. However, for it to be a proper PR, I think I need to merge the existing NuttX drivers into the Microwindows mainline first. I apologize for postponing this part of the work earlier. I will submit a PR for upstream MW review soon.

ppisa:
Yes, this what I have proposed and expect as well.

You should set something like `SCREEN=NUTTX` or `SCREEN=NUTTX_FB` in the local `graphics/microwindows/Makefile` and then Microwindows with merged code should add correct driver sources when you include `microwindows/src/drivers/Objects.rules`.

Same for the `MOUSE` and `KEYBOARD`. The question is if the mouse should be a single driver or one for relative motion mouse device and other for touchscreen.

As for the keyboard, there can be space to think how to map it the best way. The NnuttX provides its `KEYCODE_xxx` macros in [kbd_codec.h](https://github.com/apache/nuttx/blob/master/include/nuttx/input/kbd_codec.h#L44) for special keys. But only for these and rest is expected to be ASCII. Microwindows works with `MWKEY` and `MWSCANCODE` which aligns with concept of real X11 where the keyboards provide "position" of pressed key by scancode and it can be remapped to symbol even latter in application. But usually `MWKEY` code is used and its definitions are provided in [src/include/mwtypes.h](https://github.com/ghaerr/microwindows/blob/master/src/include/mwtypes.h#L1402).

You provide translation in `translate_keycode` when keyboard is in `g_kbd_mode` set to `KBD_MODE_EVENT` but there is heuristics during open. I think that there could be two drivers, one for event type and another for raw keyboards. This could be configured through `Kconfig`, same would be useful for device name. So it is necessary to think a little what is the best option there.

You should try even if Nano-X server and clients can be run on NuttX. I do not see X11 graphics option as the main priority, but it would prove that even complex setup can be build and check if it is stable on NuttX.

Acfboy:
> The NnuttX provides its KEYCODE_xxx macros in [kbd_codec.h](https://github.com/apache/nuttx/blob/master/include/nuttx/input/kbd_codec.h#L44) for special keys. But only for these and rest is expected to be ASCII.

Hi, @ppisa . I have a question about this statement. It seems to me that the ranges of the `KEYCODE_xxx` macros and ASCII are conflicting, which makes it impossible for my driver to distinguish whether it's a character or a special key. In my `sim:mw` setup, this results in being unable to type a space.

>  I think that there could be two drivers, one for event type and another for raw keyboards. This could be configured through Kconfig, same would be useful for device name.

Two drivers makes the code cleaner. But specify the device path/name via Kconfig risks inconsistency with the real device. Alternatively, if Microwindows auto-selects the device based on NuttX config, then we have to keep both sides in sync — which adds maintenance overhead.

> You should try even if Nano-X server and clients can be run on NuttX.

Okay, I'll try server and client of NanoX.

ppisa:
> > The NnuttX provides its KEYCODE_xxx macros in [kbd_codec.h](https://github.com/apache/nuttx/blob/master/include/nuttx/input/kbd_codec.h#L44) for special keys. But only for these and rest is expected to be ASCII.
> 
> Hi, @ppisa . I have a question about this statement. It seems to me that the ranges of the `KEYCODE_xxx` macros and ASCII are conflicting, which makes it impossible for my driver to distinguish whether it's a character or a special key. In my `sim:mw` setup, this results in being unable to type a space.

Yes, you are right, I expected that for  `KEYCODE_NORMAL`  the ASCII value is returned in another field of the event structure, but there is only 
```
struct keyboard_event_s
{
  uint32_t type;
  uint32_t code;
};
```
where `type` is  `KEYBOARD_PRESS` or `KEYBOARD_RELEASE`. But into `code` sent to `keyboard_event()` is ASCII code for example in [g_km_keymap](https://github.com/apache/nuttx/blob/master/boards/arm/common/stm32/src/stm32_kmatrix_gpio.c#L100) in STM32 matrix keyboard but some other drivers call `keyboard_event()` even with X11 encoded KeySim ([arch/sim/src/sim/posix/sim_x11eventloop.c](https://github.com/apache/nuttx/blob/master/arch/sim/src/sim/posix/sim_x11eventloop.c#L126)) and some with keys translated to `KEYCODE_xxx` where is no space for ASCII.

So this seems to be call for priority issue.

@acassis @gregory-nutt Please, do you have some some insight what is right?

The simple fix is to push `KEYCODE_FWDDEL` in [include/nuttx/input/kbd_codec.h](https://github.com/apache/nuttx/blob/master/include/nuttx/input/kbd_codec.h#L44) to be above ASCII, for example 128 or 129. But may it be it can break some targets where it would result in some large arrays, etc. When keyboards with interpreted national symbols are used then it would worth to push special codes even further after basic unicode...
 
> > I think that there could be two drivers, one for event type and another for raw keyboards. This could be configured through Kconfig, same would be useful for device name.
> 
> Two drivers makes the code cleaner. But specify the device path/name via Kconfig risks inconsistency with the real device. Alternatively, if Microwindows auto-selects the device based on NuttX config, then we have to keep both sides in sync — which adds maintenance overhead.

Yes but NuttX is based on configuration and it is better to fail then to do lot of testing of different device names and even deciding if it is raw driver or event based. But in general it seems that there is pace for some discussion and making KBD simple on the NuttX side.

> > You should try even if Nano-X server and clients can be run on NuttX.
> 
> Okay, I'll try server and client of NanoX.

Thanks.

acassis:
> > > The NnuttX provides its KEYCODE_xxx macros in [kbd_codec.h](https://github.com/apache/nuttx/blob/master/include/nuttx/input/kbd_codec.h#L44) for special keys. But only for these and rest is expected to be ASCII.
> > 
> > 
> > Hi, @ppisa . I have a question about this statement. It seems to me that the ranges of the `KEYCODE_xxx` macros and ASCII are conflicting, which makes it impossible for my driver to distinguish whether it's a character or a special key. In my `sim:mw` setup, this results in being unable to type a space.
> 
> Yes, you are right, I expected that for `KEYCODE_NORMAL` the ASCII value is returned in another field of the event structure, but there is only
> 
> ```
> struct keyboard_event_s
> {
>   uint32_t type;
>   uint32_t code;
> };
> ```
> 
> where `type` is `KEYBOARD_PRESS` or `KEYBOARD_RELEASE`. But into `code` sent to `keyboard_event()` is ASCII code for example in [g_km_keymap](https://github.com/apache/nuttx/blob/master/boards/arm/common/stm32/src/stm32_kmatrix_gpio.c#L100) in STM32 matrix keyboard but some other drivers call `keyboard_event()` even with X11 encoded KeySim ([arch/sim/src/sim/posix/sim_x11eventloop.c](https://github.com/apache/nuttx/blob/master/arch/sim/src/sim/posix/sim_x11eventloop.c#L126)) and some with keys translated to `KEYCODE_xxx` where is no space for ASCII.
> 
> So this seems to be call for priority issue.
> 
> @acassis @gregory-nutt Please, do you have some some insight what is right?
> 
> The simple fix is to push `KEYCODE_FWDDEL` in [include/nuttx/input/kbd_codec.h](https://github.com/apache/nuttx/blob/master/include/nuttx/input/kbd_codec.h#L44) to be above ASCII, for example 128 or 129. But may it be it can break some targets where it would result in some large arrays, etc. When keyboards with interpreted national symbols are used then it would worth to push special codes even further after basic unicode...
> 
> > > I think that there could be two drivers, one for event type and another for raw keyboards. This could be configured through Kconfig, same would be useful for device name.
> > 
> > 
> > Two drivers makes the code cleaner. But specify the device path/name via Kconfig risks inconsistency with the real device. Alternatively, if Microwindows auto-selects the device based on NuttX config, then we have to keep both sides in sync — which adds maintenance overhead.
> 
> Yes but NuttX is based on configuration and it is better to fail then to do lot of testing of different device names and even deciding if it is raw driver or event based. But in general it seems that there is pace for some discussion and making KBD simple on the NuttX side.
> 
> > > You should try even if Nano-X server and clients can be run on NuttX.
> > 
> > 
> > Okay, I'll try server and client of NanoX.
> 
> Thanks.

Hi @ppisa I don't know the right approach to keyboard input keys symbols. Actually @linguini1 faced similar issue when porting Doom to NuttX. Unfortunately NuttX doesn't have a standard keyboard symbols

linguini1:
One option suggested by @cederom was to adopt the EVDEV codec: https://en.wikipedia.org/wiki/Evdev

ppisa:
> Hi @ppisa I don't know the right approach to keyboard input keys symbols. Actually @linguini1 faced similar issue when porting Doom to NuttX. Unfortunately NuttX doesn't have a standard keyboard symbols

@linguini1 and @acassis, I propose simple fix [include/nuttx/input/kbd_codec.h](https://github.com/apache/nuttx/blob/master/include/nuttx/input/kbd_codec.h#L44)

```diff
-  KEYCODE_FWDDEL,
+  KEYCODE_FWDDEL = 128,     
```

Or even larger shift to something like 0xF800 as used on Microwindows.

This would solve problem with overlap of letters and other ASCII codes with special keys.

I can open issue for this. The question is if there is something, some other keyboard driver or application, which breaks by this shift of `KEYCODE_xxx`. It would be problem if somebody uses the codes as index into some array etc... 128 or 129 shift is not so big so it less expensive if somebody uses arrays for some key code transformation.

I think that that this is reasonable solution for NuttX and current model.

linguini1:
I agree, and this was a solution I considered but ultimately did not have enough time to verify if it would break other things in the kernel. I think that is at least a good interim solution which is less invasive.

ghaerr:
> Or even larger shift to something like 0xF800 as used on Microwindows.

Looking ahead at Unicode, Microwindows moved its key code base to F800 because that belongs to the Unicode BMP Private Use Area. This won't overlap other Unicode values if/when NuttX applications move to Unicode.

> It would be problem if somebody uses the codes as index into some array etc... 128 or 129 shift is not so big

Possibly good idea, but pushes off actively looking through source for problems now, and not having to change them again perhaps later. FWIW, myself knowing very little about NuttX, if a small shift is thought better for now, choosing a base of 256 at least allows space for using the upper half of various 256-byte multilingual code pages for non-US (e.g. European) accented characters, etc. OTOH, if keycodes are stored in byte arrays, then 128/129 would likely be mandated.

acassis (dismissed):

graphics/microwindows/Make.defs
```diff
@@ -20,6 +20,21 @@
...
+CXXFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/include
+
+CFLAGS += -DHAVE_FNT_SUPPORT=1
+CFLAGS += -DUNIX=1
+
+CFLAGS += -DSCREEN_PIXTYPE=MWPF_TRUECOLORARGB
+CFLAGS += -DSCREEN_DEPTH=8
+
+CFLAGS += -DKEYBOARD=NOKBD
+CFLAGS += -DMOUSE=NOMOUSE
```
xiaoxiang781216: should we move the config to Makefile

graphics/microwindows/Makefile
```diff
@@ -0,0 +1,89 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+# LVGL graphic library
```
xiaoxiang781216: fix the comment

graphics/microwindows/Makefile
```diff
@@ -0,0 +1,89 @@
...
+
+MICROWINDOWS_DIR = .
+MICROWINDOWS_DIR_NAME = microwindows
+
+#-include ./lvgl/lvgl.mk
```
xiaoxiang781216: ditto

graphics/microwindows/Makefile
```diff
@@ -0,0 +1,89 @@
...
+
+# Set up build configuration and environment
+
+WD := ${shell echo $(CURDIR) | sed -e 's/ /\\ /g'}
+
+MICROWINDOWS_COMMIT_HASH := 8809effdd0dfe6c2ac8fd0f455a31407e8495a93
+
+#CONFIG_GRAPH_MICROWINDOWS_URL ?= "https://codeload.github.com/ghaerr/microwindows/zip/refs/heads/master"
```
xiaoxiang781216: remove

graphics/microwindows/Makefile
```diff
@@ -0,0 +1,89 @@
...
+
+MW_DIR_OBJ = $(WD)/$(MICROWINDOWS_DIR_NAME)/src
+
+-include microwindows/src/engine/Objects.rules
+-include microwindows/src/drivers/Objects.rules
+-include microwindows/src/fonts/Objects.rules
+
+CSRCS += $(MW_CORE_OBJS:%.o=%.c)
+
+#$(error EXTOBJS=$(EXTOBJS))
```
xiaoxiang781216: remove

graphics/microwindows/Makefile
```diff
@@ -0,0 +1,89 @@
...
+
+include $(APPDIR)/Application.mk
+
+ifeq ($(wildcard $(MICROWINDOWS_UNPACKNAME)/.git),)
+distclean::
```
xiaoxiang781216: move after line 63

examples/microwindows/Make.defs
```diff
@@ -0,0 +1,25 @@
+############################################################################
+# apps/examples/hello/Make.defs
```
xiaoxiang781216: fix the comment

examples/microwindows/Makefile
```diff
@@ -0,0 +1,34 @@
...
+# under the License.
+#
+############################################################################
+
+include $(APPDIR)/Make.defs
+
+PROGNAME  = $(CONFIG_EXAMPLES_MICROWINDOWS_PROGNAME)
+PRIORITY  = $(CONFIG_EXAMPLES_MICROWINDOWS_PRIORITY)
+STACKSIZE = $(CONFIG_EXAMPLES_MICROWINDOWS_STACKSIZE)
+MODULE    = y
```
xiaoxiang781216: y->$(CONFIG_EXAMPLES_MICROWINDOWS)

examples/microwindows/Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+MAINSRC = mwexample.c
+
+CFLAGS += -Wno-undef
```
xiaoxiang781216: fix the code

examples/microwindows/mwexample.c
```diff
@@ -0,0 +1,21 @@
+#include <stdio.h>
```
xiaoxiang781216: add copyright

examples/microwindows/mwexample.c
```diff
@@ -0,0 +1,21 @@
...
+#include <unistd.h>
+#include <nuttx/compiler.h>
```
xiaoxiang781216: add blank line

graphics/microwindows/mwin-local/scr_nuttx.c
```diff
@@ -0,0 +1,104 @@
+#include <fcntl.h>
```
xiaoxiang781216: remove, use upstream version instead

graphics/microwindows/Make.defs
```diff
@@ -27,13 +27,21 @@ CONFIGURED_APPS += $(APPDIR)/graphics/microwindows
 
 CFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/include
 CXXFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/include
+CFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/drivers
+
+ifeq ($(CONFIG_VIDEO_FB),)
```
xiaoxiang781216: add the dependence to Kconfig

graphics/microwindows/Make.defs
```diff
@@ -27,13 +27,21 @@ CONFIGURED_APPS += $(APPDIR)/graphics/microwindows
...
+$(error Microwindows requires CONFIG_VIDEO_FB=y)
+endif   
 
 CFLAGS += -DHAVE_FNT_SUPPORT=1
 CFLAGS += -DUNIX=1
 
 CFLAGS += -DSCREEN_PIXTYPE=MWPF_TRUECOLORARGB
 CFLAGS += -DSCREEN_DEPTH=8
 
+KEYBOARD = NOKBD
```
xiaoxiang781216: merge to the first patch

examples/microwindows/mwexample.c
```diff
@@ -1,21 +1,148 @@
...
+
+    draw_textbox();
+
+    printf("Mouse+keyboard test. Click to draw, type to input, ESC to quit.\n");
+
+    fd_set fds;
+    struct timeval timeout;
+    MWCOORD mx, my;
+    int buttons, last_buttons = 0;
+    int maxfd = 0;
```
xiaoxiang781216: move to beginning

examples/microwindows/Makefile
xiaoxiang781216, line 21: @Acfboy please clean up your patch before merging, @acassis don't merge this patch before I approve.

xiaoxiang781216 (changes requested):

examples/microwindows/mwexample.c
Acfboy: This file, `mwexample.c`, comes from an intermediate commit; it has already been removed in the final PR.

graphics/microwindows/mwin-local/scr_nuttx.c
Acfboy: The mwin-local folder exists only in an intermediate commit. In the final PR, the drivers have been merged into Microwindows upstream.

graphics/microwindows/Make.defs
Acfboy: Yes, the pr moves these to Makefile and mwconfig.nuttx.

graphics/microwindows/Makefile
Acfboy: Thanks for the careful review!

graphics/microwindows/Make.defs
Acfboy: thanks!

examples/microwindows/Makefile
Acfboy: The header file of microwindows use lots of ` #if MW_FEATURE_xxx`, which cause warnings, so we add the `-Wno-undef`, should I remove the `CFLAGS`?

Acfboy:
Hi @xiaoxiang781216 @acassis  thanks for the thorough review! I've addressed all your feedback and fixed the non-standard commit messages flagged by checkpatch.

xiaoxiang781216:
> Hi @xiaoxiang781216 @acassis thanks for the thorough review! I've addressed all your feedback and fixed the non-standard commit messages flagged by checkpatch.

please merge your temp change.

ppisa:
@Acfboy please, squash, merge commits as @xiaoxiang781216 suggests into single commit or some small logical set of incremental commits, there is no use for keeping the new component development history, you ca keep it on some branch of your repository as work progress documentation. So for logical commits series, I can imagine one commit which introduces Microwindows with basic Kconfig then another one which adds some demos, Kconfig options, etc. But the code has to be clean, adhere NutttX style and requirements and  build after each incremental commit.

As for the decision, which defines should go into `Makefile` and which to `Make.defs`, I think that only these defines which can be used by potential applications directly or through public Microwindows header files should be present in  Make.defs, Sofor example, if `SCREEN_PIXTYPE` is made configurable through  Kconfig and then used by header files, then it should be defined in  `Make.defs`, same i required for additional include locations. On the other hand defines provided only to choose correct behavior during library build i.e. through `Objects.rules` should be enough to define only in related NuttX `Makefile`. But I think that current state is mostly aligned with these requirements.

Acfboy:
Thanks for pointing this out, @ppisa @xiaoxiang781216 . I've reorganized the changes into two commits.

graphics/microwindows/Makefile
```diff
@@ -0,0 +1,142 @@
...
+MOUSE = CUSTOM
+endif
+
+-include microwindows/src/engine/Objects.rules
+-include microwindows/src/drivers/Objects.rules
+-include microwindows/src/fonts/Objects.rules
+
+CSRCS += $(MW_CORE_OBJS:%.o=%.c)
+
+ifneq ($(CONFIG_EXAMPLES_MICROWINDOWS),)
```
xiaoxiang781216: why check CONFIG_EXAMPLES_MICROWINDOWS in library makefile?

graphics/microwindows/Makefile
```diff
@@ -0,0 +1,142 @@
...
+context:: $(MICROWINDOWS_UNPACKNAME)
+
+distclean::
+	$(call DELDIR, $(MICROWINDOWS_UNPACKNAME))
+	$(call DELFILE, $(MICROWINDOWS_TARBALL))
+endif
+
+CFLAGS += -DRTEMS=0 -DPSP=0 -D__ECOS=0 -D__MINGW32__=0
+CFLAGS += -DELKS=0 -D_MINIX=0 -DMSDOS=0 -DLINUX=0
+CFLAGS += -DMACOSX=0 -DTRIMEDIA=0
```
xiaoxiang781216: why do we need define other OS macro to zero here?

graphics/microwindows/Make.defs
```diff
@@ -0,0 +1,50 @@
...
+############################################################################
+
+ifneq ($(CONFIG_GRAPHICS_MICROWINDOWS),)
+CONFIGURED_APPS += $(APPDIR)/graphics/microwindows
+
+# It allows `microwindows/src/include` import.
+
+CFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/include
+CXXFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/include
+CFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/drivers
```
xiaoxiang781216: the private search path and macro should move to Makefile

examples/microwindows/mwdemo_main.c
```diff
@@ -0,0 +1,32 @@
+/****************************************************************************
+ * apps/examples/microwindows/mwdemo_main.c
+ *
+ * NuttX entry point for mwdemo (Microwindows Win32 API demo)
```
xiaoxiang781216: follow nuttx copyright banner

examples/microwindows/mwdemo_main.c
```diff
@@ -0,0 +1,32 @@
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include "windows.h"
```
xiaoxiang781216: <>

examples/microwindows/mwdemo_main.c
```diff
@@ -0,0 +1,32 @@
...
+ ****************************************************************************/
+
+extern int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance,
+                          LPSTR lpCmdLine, int nShowCmd);
+
+int main(int argc, FAR char *argv[])
+{
+  int ret;
+
+  invoke_WinMain_Start(argc, (char **)argv);
```
xiaoxiang781216: remove the cast

examples/microwindows/mwdemo_main.c
```diff
@@ -0,0 +1,32 @@
...
+
+  ret = WinMain(0, NULL, (LPSTR)(argc > 0 ? argv[0] : ""), 0);
```
xiaoxiang781216: remove the cast

examples/microwindows/mwdemo.c
```diff
@@ -0,0 +1,640 @@
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <time.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include "uni_std.h"
```
xiaoxiang781216: change all "xxx.h" to <xxx.h>

examples/microwindows/mwdemo.c
```diff
@@ -0,0 +1,640 @@
...
+#define MWINCLUDECOLORS
+#include "windows.h"
+#include "device.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#if DOS_TURBOC
+unsigned _stklen = 4096;
```
xiaoxiang781216: why add?

examples/microwindows/mwdemo.c
```diff
@@ -0,0 +1,640 @@
...
+
+/* test timer system with either demo */
+
+#define TIMERDEMO   1       /* set for WM_TIMER demo */
+
+/* not used */
+
+#define CLIENT3D    0       /* old client draw test */
+
+#if RTEMS
```
xiaoxiang781216: why add

examples/microwindows/mwdemo.c
```diff
@@ -0,0 +1,640 @@
...
+
+#if GRAPH3D
+#  include "graph3d.h"
+#endif
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+extern MWIMAGEHDR image_car8;
```
xiaoxiang781216: global variables add g_ prefix

examples/microwindows/mwdemo.c
```diff
@@ -0,0 +1,640 @@
...
+
+  return 0;
+}
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+int WINAPI
+WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine,
```
xiaoxiang781216: merge the function body into main and remove WinMain

examples/microwindows/mwdemo.c
```diff
@@ -0,0 +1,640 @@
...
+    default:
+      return DefWindowProc(hwnd, msg, wp, lp);
+    }
+
+  return (0);
+}
+
+#endif
+
+typedef struct
```
xiaoxiang781216: move to private type section

examples/microwindows/mwdemo.c
xiaoxiang781216: remove ALL unused variables

graphics/microwindows/Makefile
ppisa: I think that it is correct. It allows to build Microwindows library without examples when you do not need them and thy would only enlarge NuttX image when registered applications are pull in.

graphics/microwindows/Makefile
ppisa: They should not pollute global NuttX CFLAGS. So they can be and are defined in local `Makefile` and are used only in private parts of the Microwindows. If they are used in some public header file file then they would be required even for applications because Microwindows use only `#if MSDOS` etc. not `#ifdef` and undefined macros would lead to warnings. It has not been problem in the past but `-Wundef` is defined in NuttX `Toolchain.defs` and may be it is even enabled in modern GCC by default by `-Wall`, etc.

Generally, I see as the best solution do not need for these to be defined by build system to zero. One option is to protects `#if` parts by `#if defined(MSDOS) && MSDOS` which prevents (tested on GCC 14.2.0)
```
warning: "MSDOS" is not defined, evaluates to 0 [-Wundef]
```
This needs to be done in whole Microwindows and care has to be taken to not introduce some flip in some logic.

Another option is to put these zeroing defines in `microwindows/src/include/mwconfig.nuttx` and all other specific includes.

When I look at the possible usecases of Microwindows header files, I have noticed that  [microwindows/src/include/mwtypes.h](https://github.com/ghaerr/microwindows/blob/master/src/include/mwtypes.h) uses some config defines, i.e. `MWPIXEL_FORMAT` but seems to not include `mwconfig.h` directly or indirectly. There is one exception
```
#if __ECOS
#include <ecosmwconfig.h>	/*include the eCos configuration "translation" header */
#endif
```

This meas that if this header file is included without preceding `mwconfig.h` include, then there could be a problem. May it be, it has purpose to not include whole `mwconfig.h` there but including defines part in addition to or instead of `__ECOS` could be there
```
#ifdef MWCONFIG_FILE
#define INCFILE(name)   name
#include INCFILE(MWCONFIG_FILE)
#endif
```
On the other hand target specific include fragments do not have protection against multiple inclusion and redefines.

So I have plea to @ghaerr to discuss which direction he suggest.

graphics/microwindows/Make.defs
ppisa: These are required even for other applications using Microwindows which is main goal for integration.

There is global problem with NuttX with components includes because they are not propagated, header files copied` into `make export` archive. So there should be in `Make.defs` but ideally added to something like `APP_COMPONENT_INCLUDE_DIRS +=` which is then converted by substitution into `CFLAGS` and `CXXFLAGS` for in tree applications builds and iterated during `make export` which copies such directories into target archive `include` directory.

We have this problem with our SHV, uLUt and it would be problem with more other components which can be build as part of standard NuttX build and used through more sub-directories in APPS tree but are broken after export.

examples/microwindows/mwdemo_main.c
ppisa: The one of Microwindows APIs provides some basic compatibility with Windows GDI API and it is then possible to port applications from Windows there. But it is possible to use GDI even fully without this partial compatibility layer same as X11 APIs etc.   So this example shows how this compatibility can be used. There is attempt to show how such minimal demo can be added to NuttX. On the other hand, demos can be build directly from Microwindows sources so we can left this only there. Windows API substitution file is provided by `microwindows/src/include/windows.h`. It would worth to think about setup where microwindows header files are provided in some subdirectory to `include` that application can decide if it wants or does not want access to these...

So this is for discussion. But large part of Microwindows value is providing APIs which allows to port applications programmed against different massively used graphic full featured stacks to embedded systems if reasobale subset of APIs is used.

examples/microwindows/mwdemo.c
ppisa: Such and other OS specific part can/should be removed from example included in NuttX.

Acfboy:
Thanks @xiaoxiang781216 for the very thorough review, and @ppisa for the clarification and discussion. I have updated the code under your guidance. 

Regarding whether all include paths in Make.defs should be moved to the Makefile, I agree with @ppisa that this would require more changes to the build system, so I haven't made that change for now.

graphics/microwindows/Makefile
xiaoxiang781216: why it's correct? how to build Microwindows source code if I don't enable CONFIG_EXAMPLES_MICROWINDOWS, but write my demo apps.

xiaoxiang781216:
> Thanks @xiaoxiang781216 for the very thorough review, and @ppisa for the clarification and discussion. I have updated the code under your guidance.
> 
> Regarding whether all include paths in Make.defs should be moved to the Makefile, I agree with @ppisa that this would require more changes to the build system, so I haven't made that change for now.

@Acfboy please always address the comment in the origin patch instead creating new one.

graphics/microwindows/Makefile
xiaoxiang781216: since -Wno-def is added https://github.com/apache/nuttx-apps/pull/3624/changes#diff-103e40a66bf18382d3e6ef792c8861476c568886fbf4906501affb88576941beR52
let's remove these macro definition.

graphics/microwindows/Make.defs
xiaoxiang781216: Does ALL the follow definition is required for compiling application source files:
```
CFLAGS += ${INCDIR_PREFIX}$(APPDIR)/graphics/microwindows/microwindows/src/drivers

HAVE_FNT_SUPPORT = Y

CFLAGS += -DMWCONFIG_FILE='"mwconfig.nuttx"'

CFLAGS += -DNUTTX_FB_PATH=\"$(CONFIG_MICROWINDOWS_FB_PATH)\"
ifdef CONFIG_MICROWINDOWS_KBD_EVENT_PATH
CFLAGS += -DNUTTX_KBD_EVENT_PATH=\"$(CONFIG_MICROWINDOWS_KBD_EVENT_PATH)\"
endif
ifdef CONFIG_MICROWINDOWS_KBD_RAW_PATH
CFLAGS += -DNUTTX_KBD_RAW_PATH=\"$(CONFIG_MICROWINDOWS_KBD_RAW_PATH)\"
endif
ifdef CONFIG_MICROWINDOWS_MOUSE_PATH
CFLAGS += -DNUTTX_MOUSE_PATH=\"$(CONFIG_MICROWINDOWS_MOUSE_PATH)\"
endif
ifdef CONFIG_MICROWINDOWS_TS_PATH
CFLAGS += -DNUTTX_TOUCHSCREEN_PATH=\"$(CONFIG_MICROWINDOWS_TS_PATH)\"
endif
```
or just need when compile microwindows self.

graphics/microwindows/Makefile
ppisa: Microwidows is APIs providing library. So if it is build then elsewhere in the source tree can be application which uses Microwindows and combines it with something else, some real value program.

For example, we have used low level GDI API which allowed us to run medical syringe pump firmware build against RTEMS and RETMS Microwindows port on real hardware with all the check functions and communication with another MCU which control delivery. [Related RTEMS page](https://www.rtems.org/applications/hardware/amv_technic_i/) and our [page](https://www.pikron.com/pages/devel/medinst.html). The same complex application source code with out upper level libraries [SuiTk](https://gitlab.com/pikron/sw-base/suitk/-/wikis/home) could be run against Microwidows library compiled for Linux X11 in window emulating target screen, in this case with stubs for syringe motor, etc., and we have even run same code on Linux framebuffer directly. Same libraries could be ran even against Microwidows on the Cortex-M without operating system, in this case limited to single thread.

So if you build Microwidows without its included application then it can still be used on NuttX. My personal preferred NuttX applications build is to export NuttX and use it as library for our applications. Then application from same sources can be build for GNU/Linux and linked with GLibc and other libraries, but with switch of rules to build for NuttX and linked with it into firmware image or build against RTEMS development environment and BSPs installation and I obtain RTEMS system firmware image. Our OMK build system even allows to have source tree with multiple applications which are built as separate executables for GNU/Linux and build together into single NuttX image and registered as builtins aromatically or even compiled as loadable ELF files to be loaded and linked into NuttX dynamically.

I expect that there could be even more configuration options which select which Microwidows programs (in NuttX without MMU builtins) should be provided, i.e., request to build Nano-X X11 like server, separate applications and demos.

At least this is my idea and where I see value of Microwidows. Ability to write portable POSIX applications which can even use single graphic API on more operating systems.

graphics/microwindows/Makefile
ppisa: There is question if there are some warning even from exported header files used outside `apps/graphics/microwindows` directory. But in such case `-Wno-def` would be needed globally not only in local `Makefile` and this is not the right direction. I think that we should solve problem with this (correctly) more strict diagnostics in Microwidows sources. It is question if in this or some latter iteration.

graphics/microwindows/Make.defs
ppisa:
```
CFLAGS += -DMWCONFIG_FILE='"mwconfig.nuttx"'
```
is required for Microwidows use outside this directory for sure.

`src/drivers` would be probably require only for custom keyboard or mouse drivers, 
`graphics/microwindows/microwindows/src/include` is required externally. Mouse and keyboard options and related paths can be moved to `Makefile`, they are required only inside of Microwidows drivers build. For teh custom case, they can be prepared locally from  `CONFIG_MICROWINDOWS_KBD_` etc. So this can be limited to local directory only, i.e to `Makefile`for now.

```
HAVE_FNT_SUPPORT = Y
```
is provided by as `#define HAVE_FNT_SUPPORT    1` by `src/include/mwconfig.nuttx`. So I think that it can be removed from `Make.defs` and it should not be needed even in `Makefile`. But needs to test.

ghaerr:
> Generally, I see as the best solution do not need for these to be defined by build system to zero.

On the Microwindows side, for various reasons the idiom of `#if RTEMS` etc is used a lot more than `#ifdef RTEMS`, mostly so that the same mechanism can be used to turn options on or off without changing an \#ifdef to an \#if. However, it was not the intent to force developers to find all cases of \#if SYM and pass -DSYM=0 everywhere. Thus, I do not agree the best solution is to set otherwise undefined symbols to 0.

It seems this may be a problem because of the use of -Wundef in NuttX?

> There is one exception
> #if __ECOS

> This meas that if this header file is included without preceding mwconfig.h include, then there could be a problem. 

The ECOS port is quite old, and probably doesn't follow some of the more recent ideas of keeping platform build-specific options in a separate configuration file. This could be fixed, but I hesitate to change code that I or others can't easily test in the main repo.

ppisa:
> > Generally, I see as the best solution do not need for these to be defined by build system to zero.
> 
> On the Microwindows side, for various reasons the idiom of `#if RTEMS` etc is used a lot more than `#ifdef RTEMS`, mostly so that the same mechanism can be used to turn options on or off without changing an #ifdef to an #if. However, it was not the intent to force developers to find all cases of #if SYM and pass -DSYM=0 everywhere. Thus, I do not agree the best solution is to set otherwise undefined symbols to 0.

OK, I have tried to add checks for defines presence in Microwindows code. I agree that there is some added complexity. I have repeated some defines sequences because with checks for define presence the conditions testing multiple target OS alternatives are hard to write such that warning is prevented

https://github.com/ghaerr/microwindows/pull/194

graphics/microwindows/Makefile
Acfboy: Yes, I should provide a config option that allows users to decide whether to enable the Win32 API (or x11 support in the future), instead of only allowing the example apps to use it.

Acfboy:
Thank you for your careful review, detailed discussion, and guidance. I have made the revisions.

xiaoxiang781216 (approved):

xiaoxiang781216:
@acassis @ppisa is ready for merging.

jerpelea (approved):

ppisa:
It seems that I am late to this party/iteration as the work is merged.

@Acfboy  has done sound work and as the Microwidows are fully optional component and the main work on NuttX drivers API related code is already integrated into mainline Microwidows, then merge can be base for next @Acfboy work and good opportunity for others to see the integration as preview. On the other hand I do not consider this as production ready/final version.

@Acfboy has even prepared documentation and configs with examples configurations qemu-intel64:mw and sim:mw on related NuttX fork https://github.com/Acfboy/nuttx/commits/add-microwindows/

It would worth to be integrated through pull-request as well. But I would like to discuss some more options for example configurations. I would suggest to start the Microwidows demo through NSH to allow console access into running system for debugging, etc. But these are details which will be sorted out.

As for the maim Microwindows integration, there worth to be done mone polishment to find reasonable way how to resolve warnings, solve https://github.com/apache/nuttx/issues/19527 (hope to find time to propose patch today). Then the test of more complex demos and applications based for example on X11 API should be tested. We will se where we find problems or some NuttX specific variants would be required... But as both Microwindows and NuttX use POSIX as main runtime model, it should be non-intrusive on both sides.

I would be happy if components headers export for NuttX `make export` is solve in future which is useful not only for Microwidows but more general way.

xiaoxiang781216:
> It seems that I am late to this party/iteration as the work is merged.
> 
> @Acfboy has done sound work and as the Microwidows are fully optional component and the main work on NuttX drivers API related code is already integrated into mainline Microwidows, then merge can be base for next @Acfboy work and good opportunity for others to see the integration as preview. On the other hand I do not consider this as production ready/final version.
> 
> @Acfboy has even prepared documentation and configs with examples configurations qemu-intel64:mw and sim:mw on related NuttX fork https://github.com/Acfboy/nuttx/commits/add-microwindows/
> 
> It would worth to be integrated through pull-request as well. But I would like to discuss some more options for example configurations. I would suggest to start the Microwidows demo through NSH to allow console access into running system for debugging, etc. But these are details which will be sorted out.
> 
> As for the maim Microwindows integration, there worth to be done mone polishment to find reasonable way how to resolve warnings, solve [apache/nuttx#19527](https://github.com/apache/nuttx/issues/19527) (hope to find time to propose patch today). Then the test of more complex demos and applications based for example on X11 API should be tested. We will se where we find problems or some NuttX specific variants would be required... But as both Microwindows and NuttX use POSIX as main runtime model, it should be non-intrusive on both sides.
> 
> I would be happy if components headers export for NuttX `make export` is solve in future which is useful not only for Microwidows but more general way.

Thanks for the detailed summary and future plan. Do you have Nano X11 progress @ppisa ?

ppisa:
> Thanks for the detailed summary and future plan. Do you have Nano X11 progress @ppisa?

This is more to @Acfboy , I try to help where I have knowledge and as I find a time. I expect that it can be relatively straightforward. But who know. We need UNIX sockets or other local connection. But it should be available on NuttX.

Acfboy:
Yes, I'm working on porting the nano-X X11 API and will likely submit a new PR soon. After that, I'll do more thorough testing, including on real hardware, and address all warnings to make it production-ready.

PR #3632, opened 2026-07-11
interpreters/berry: Fix raw console input

speccy88:
## Summary

Fix the Berry interactive console on raw NuttX terminals, including USB CDC ACM consoles that do not provide canonical input, CR-to-LF conversion, or driver echo.

The pinned Berry default port uses `fgets()`. On a raw console, Enter commonly arrives as CR rather than LF, so the REPL and `input()` wait indefinitely. Routing Berry through the shared NuttX readline implementation also leaks NSH history and tab completion into the interpreter.

This change:

- adds an opt-in `USE_NUTTX_CONSOLE` reader to the pinned Berry source;
- accepts CR, LF, and CRLF line endings;
- handles a queued LF left by a CRLF command that launched Berry from NSH;
- provides termios-aware local echo and BS/DEL editing without double echo;
- treats ETX or an interrupted read as console termination so Ctrl-C returns to NSH;
- keeps Berry input isolated from NSH readline history, completion, and prompt state; and
- enables the reader in both Make and CMake builds.

The normal Berry `fgets()` path remains unchanged when `USE_NUTTX_CONSOLE` is not defined.

The equivalent source fix is also proposed upstream in [berry-lang/berry#539](https://github.com/berry-lang/berry/pull/539). This local patch is needed for the Berry commit currently pinned by NuttX Apps.

## Impact

- User impact: Berry REPL and `input()` work on raw serial and USB CDC terminals.
- Build impact: Berry alone receives one additional compile definition and pinned-source patch.
- Other applications: unchanged; `system/readline` is not modified.
- Compatibility: the existing stdio fallback is preserved.

## Verification

Fruit Jam RP2350 hardware, minimal NuttX image over raw USB CDC ACM:

- CR-only, LF-only, and CRLF input all complete.
- A CRLF launch no longer leaves an empty line for the first `input()` call.
- Piped `berry\r\n6*7\r\n` preserves the first expression byte and evaluates to `42`.
- CRLF `10+5` evaluates to `15` with one following prompt.
- A following CR-only `2+3` evaluates to `5`.
- LF-only empty input remains valid.
- Berry input is not recalled by NSH history, and Tab does not expose NSH completions.
- Ctrl-C exits Berry and returns to the NSH prompt.

Build and local checks:

- Exact patched `default/be_port.c` blob: `7d52aebe055b098febec472f32690c1ef71780e3`.
- Strict standalone Berry build with `-Wall -Wextra -std=c99 -pedantic-errors -DUSE_NUTTX_CONSOLE` passed.
- `sim:berry` Make build passed; `berry -e "print(6*7)"` returned `42`.
- CMake FetchContent applied both patches and compiled the final `be_port.c` successfully; the full `sim:berry` CMake/Ninja path passed before the final hardware-only handoff addition.
- Pipe and raw-PTY tests covered CR, LF, delayed CRLF, non-LF after CR, empty LF input, BS, DEL, ETX, queued first input, echo, and `input()` prompt flushing.
- `git diff --check` passed.
- NuttX `checkpatch.sh` and `cmake-format` passed.

### DIFF
--- a/interpreters/berry/0002-Use-NuttX-console-reader.patch
+++ b/interpreters/berry/0002-Use-NuttX-console-reader.patch
@@ -1,0 +1,134 @@
+diff --git a/default/be_port.c b/default/be_port.c
+index be19394..7d52aeb 100644
+--- a/default/be_port.c
++++ b/default/be_port.c
+@@ -13,0 +14,7 @@
++#if defined(USE_NUTTX_CONSOLE)
++#include <errno.h>
++#include <poll.h>
++#include <termios.h>
++#include <unistd.h>
++#endif
++
+@@ -17,0 +25,20 @@
++#if defined(USE_NUTTX_CONSOLE)
++static pid_t console_pid = -1;
++static int skip_lf_after_cr;
++
++static void console_write(const char *buffer, size_t length)
++{
++    while (length > 0) {
++        ssize_t result = write(STDOUT_FILENO, buffer, length);
++        if (result > 0) {
++            buffer += result;
++            length -= (size_t)result;
++        } else if (result < 0 && errno == EINTR) {
++            continue;
++        } else {
++            break;
++        }
++    }
++}
++#endif
++
+@@ -24,0 +52,98 @@ BERRY_API char* be_readstring(char *buffer, size_t size)
++#if defined(USE_NUTTX_CONSOLE)
++    struct termios attr = { 0 };
++    size_t length = 0;
++    int echo = 0;
++    int map_newline = 0;
++    int pending = -1;
++    int tty = 0;
++
++    if (size > 0) {
++        buffer[0] = '\0';
++    }
++    if (size < 2) {
++        return NULL;
++    }
++    if (tcgetattr(STDIN_FILENO, &attr) == 0) {
++        tty = 1;
++        echo = (attr.c_lflag & ECHO) == 0;
++        map_newline = (attr.c_oflag & (OPOST | ONLCR)) ==
++            (OPOST | ONLCR);
++    }
++    be_fflush(stdout);
++    if (console_pid != getpid()) {
++        console_pid = getpid();
++        skip_lf_after_cr = 0;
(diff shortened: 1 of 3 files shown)


acassis (approved):

xiaoxiang781216:
@speccy88 could you explain more why the process in serial driver framework doesn't work as expect? The basic process(\r, \n, DEL, BACK) should be handled by driver directly:
https://github.com/apache/nuttx/blob/master/drivers/serial/serial.c#L1001-L1125

linguini1 (approved):
Hi @speccy88 , just a notice of the contribution guide update; if you're using generative tooling to help your development please be sure to indicate the model using the `Assisted-by: ` commit field! Ignore if you're not :)

speccy88:
Thanks for questioning the scope of this change. I retraced the reproducer and found that the behavior was real in my downstream Fruit Jam image, but the root cause was in that integration rather than in Berry or stock NuttX.

My downstream NSH console patch configured the shared USB CDC descriptor with `c_iflag = 0` and `c_lflag = ISIG` so readline could own input and echo permanently. Berry then inherited the same raw descriptor, without canonical buffering or CR-to-LF translation, and its normal `fgets()` input appeared to hang.

I have corrected the terminal lifecycle at the point where I introduced the regression in [speccy88/NuttX-RP2350#4](https://github.com/speccy88/NuttX-RP2350/pull/4): the terminal remains canonical and echo-enabled for applications, while NSH readline temporarily takes byte-at-a-time/no-driver-echo ownership and restores the complete saved mode before launching a command. The Berry-specific reader has been removed downstream.

This PR is therefore not pertinent to NuttX Apps, and I am closing it. Thank you for catching the scope problem before merge.

For transparency, Codex (GPT-5) assisted with diagnosis, implementation, and review. I reviewed the result and remain responsible for the contribution. Future AI-assisted Apache commits from me will include the required `Assisted-by: Codex:GPT-5` trailer alongside my own human sign-off.

PR #3642, opened 2026-07-16
system/nxpkg: Make the storage root configurable and crash-safe.

aviralgarg05:

## Summary

This PR has been reduced to the storage layer. The rest of what it used to
contain is now in #3718, #3719 and #3720, in that order.

The package store, the catalog and the temporary download area were fixed
paths spread across `/etc`, `/var/lib` and `/var/cache`. A board rarely has
all three, and nothing let it put them somewhere it does have, such as a
removable card. They are now derived from `CONFIG_SYSTEM_NXPKG_ROOT`.

Every file the package manager owns was also written in place, so a reset
partway through a write left a truncated catalog or database behind. Writes
now go to a temporary file that is renamed into place.

Locking used only the file modification time. A card whose clock does not
follow `CLOCK_REALTIME` reports a file created a moment ago as decades old,
so a live lock looked stale to the next caller and was taken away from its
owner. Each lock now records a per-boot token and the owning task, so a
contender can tell a live owner from an abandoned one, and reclaims the
lock only in the second case.

Diagnostics go to syslog, since a package operation started by a supervisor
has no console to print to. `pkg_runtime_compat()` also falls back to
`CONFIG_ARCH_BOARD_CUSTOM_NAME`, because a board with a custom board
directory defines no `CONFIG_ARCH_BOARD` at all, which does not build.

## Impact

- New feature: NO, this is the foundation the other three parts build on.
- User adaptation: YES. The storage root moves to
  `CONFIG_SYSTEM_NXPKG_ROOT`, default `/var/lib/nxpkg`. A board that mounts
  writable storage elsewhere sets that symbol.
- Build: NO.
- Hardware: NO.
- Documentation: The companion documentation is apache/nuttx#18875.
- Security: NO new boundary.
- Compatibility: A store created under the old fixed paths is not migrated;
  point `CONFIG_SYSTEM_NXPKG_ROOT` at the old location to keep it.

## Testing

(`esp-14.2.0_20251107`).

  every changed file
- all `system/nxpkg` sources compiled for the target
- this part builds on `master` on its own; the three that follow declare it
  with `Depends-On`

On the target: a lock held by a running task was kept while a second caller
asked for it, and locks left by an exited task and by an earlier boot were
both reclaimed.

## PR verification Self-Check

### DIFF
--- a/system/nxpkg/Kconfig
+++ b/system/nxpkg/Kconfig
@@ -29,4 +29,15 @@ config SYSTEM_NXPKG_STACKSIZE
 	int "'nxpkg' stack size"
 	default 16384
 
+config SYSTEM_NXPKG_ROOT
+	string "'nxpkg' storage root"
+	default "/var/lib/nxpkg"
+	---help---
+		Base directory used by nxpkg for its local index, installed
+		metadata, temporary downloads, and package payload storage.
+		The default follows the conventional persistent application-data
+		location.  A board must mount persistent storage at /var or set
+		this to another persistent location, such as /mnt/sdcard/nxpkg,
+		if packages need to survive a reset.
+
 endif
--- a/system/nxpkg/Makefile
+++ b/system/nxpkg/Makefile
@@ -29,6 +29,7 @@ MODULE    = $(CONFIG_SYSTEM_NXPKG)
 
 CSRCS = pkg_compat.c pkg_hash.c pkg_install.c pkg_log.c pkg_manifest.c
 CSRCS += pkg_metadata.c pkg_store.c pkg_txn.c
+CSRCS += pkg_repo.c
 MAINSRC = pkg_main.c
 
 include $(APPDIR)/Application.mk
--- a/system/nxpkg/pkg.h
+++ b/system/nxpkg/pkg.h
@@ -27,30 +27,93 @@
  * Included Files
  ****************************************************************************/
 
+#include <nuttx/config.h>
+
 #include <limits.h>
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdio.h>
+#include <stdlib.h>
 
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 
-#define PKG_REPO_DIR          "/etc/nxpkg"
-#define PKG_REPO_INDEX        "/etc/nxpkg/index.json"
-#define PKG_REPO_INSTALLED    "/var/lib/nxpkg/installed.json"
-#define PKG_STORE_DIR         "/var/lib/nxpkg/pkgs"
-#define PKG_TMP_DIR           "/var/cache/nxpkg"
-#define PKG_TMP_PKG_DIR       "/var/cache/nxpkg/pkg"
+#define PKG_ROOT_DIR          CONFIG_SYSTEM_NXPKG_ROOT
+#define PKG_REPO_DIR          PKG_ROOT_DIR
+#define PKG_REPO_INDEX        PKG_ROOT_DIR "/index.jsn"
+#define PKG_REPO_SOURCE       PKG_ROOT_DIR "/repo.url"
+#define PKG_REPO_INSTALLED    PKG_ROOT_DIR "/instpkg.jsn"
+#define PKG_STORE_DIR         PKG_ROOT_DIR "/pkgs"
+#define PKG_TMP_DIR           PKG_ROOT_DIR "/tmp"
+#define PKG_TMP_PKG_DIR       PKG_ROOT_DIR "/tmp/pkg"
 
 #define PKG_NAME_MAX          63
 #define PKG_VERSION_MAX       31
 #define PKG_ARCH_MAX          31
 #define PKG_COMPAT_MAX        63
+#define PKG_DESCRIPTION_MAX   127
+#define PKG_CATEGORY_MAX      31
 #define PKG_HASH_HEX_LEN      64
-#define PKG_INDEX_MAX         32
+/* Each manifest slot is ~1.7KB (dominated by PKG_LAUNCH_ARGS_MAX slots).
+ * Keep the catalog bounded so repository-provided metadata cannot cause
+ * unbounded memory use.  Callers should allocate struct pkg_index_s from
+ * the application heap rather than placing it on a small task stack.
+ */
+
+#define PKG_INDEX_MAX         16
 #define PKG_INSTALLED_MAX     16
 #define PKG_INSTALLED_VERSIONS_MAX 8
+#define PKG_LAUNCH_ARGS_MAX   8
+#define PKG_LAUNCH_ARG_MAX    127
+
+/* Caps against a malicious/compromised HTTP server: without these, an
+ * oversized response can exhaust SD-card space (downloads) or force an
+ * unbounded single heap allocation sized directly off attacker-controlled
+ * content (pkg_store_read_text).  Text/metadata files (index.jsn,
+ * instpkg.jsn) are always small; artifact downloads cover the largest
+ * real payloads seen in practice (a multi-MB WAD, a ~1MB game ELF) with
+ * generous headroom.
+ */
+
+#define PKG_TEXT_MAX_SIZE     (256 * 1024)
+#define PKG_DOWNLOAD_MAX_SIZE (32 * 1024 * 1024)
+
+/* nxpkg is a one-shot CLI, not a daemon, so a lock file older than this
+ * cannot belong to a still-running install under normal use (even a full
+ * multi-MB artifact over a slow link finishes well within this window) -
+ * it can only be left over from a process that was killed or a device
+ * that lost power mid-install.  Reclaiming it is what makes install/
+ * update/rollback usable again after the crash/power-loss scenarios this
+ * target is prone to, instead of failing with EBUSY forever.
+ */
+
+#define PKG_LOCK_STALE_SECONDS (600)
+
+static inline void *pkg_malloc(size_t size)
+{
+  return malloc(size);
+}
+
+static inline void *pkg_zalloc(size_t size)
+{
+  return calloc(1, size);
+}
+
+static inline void *pkg_realloc(void *ptr, size_t size)
+{
+  return realloc(ptr, size);
+}
+
+static inline void pkg_free(void *ptr)
+{
+  free(ptr);
+}
+
+static inline FAR char *pkg_path_alloc(void)
+{
+  return pkg_malloc(PATH_MAX);
+}
 
 /****************************************************************************
  * Public Types
@@ -83,7 +146,12 @@ struct pkg_manifest_s
   char compat[PKG_COMPAT_MAX + 1];
   char artifact[PATH_MAX];
   char sha256[PKG_HASH_HEX_LEN + 1];
+  char launch_args[PKG_LAUNCH_ARGS_MAX][PKG_LAUNCH_ARG_MAX + 1];
+  char description[PKG_DESCRIPTION_MAX + 1];
+  char category[PKG_CATEGORY_MAX + 1];
+  char icon[PATH_MAX];
   enum pkg_payload_type_e type;
+  size_t launch_argc;
 };
 
 struct pkg_index_s
@@ -116,6 +184,7 @@ struct pkg_installed_db_s
 
 const char *pkg_manifest_type_str(enum pkg_payload_type_e type);
 int pkg_manifest_validate(FAR const struct pkg_manifest_s *manifest);
+bool pkg_validate_path_component(FAR const char *value);
 int pkg_manifest_parse_type(FAR const char *value,
                             FAR enum pkg_payload_type_e *type);
 
@@ -124,6 +193,7 @@ int pkg_store_ensure_package_root(FAR const char *name);
 int pkg_store_ensure_version_dir(FAR const char *name,
                                  FAR const char *version);
 int pkg_store_format_index_path(FAR char *buffer, size_t size);
+int pkg_store_format_repo_source_path(FAR char *buffer, size_t size);
 int pkg_store_format_installed_path(FAR char *buffer, size_t size);
 int pkg_store_format_package_root(FAR char *buffer, size_t size,
                                   FAR const char *name);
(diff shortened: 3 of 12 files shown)


system/nxpkg/Kconfig
```diff
@@ -29,4 +29,13 @@ config SYSTEM_NXPKG_STACKSIZE
 	int "'nxpkg' stack size"
 	default 16384
 
+config SYSTEM_NXPKG_ROOT
+	string "'nxpkg' storage root"
+	default "/tmp/nxpkg"
```
acassis: I think keeping it in /tmp is fine for this first version, but the final version should be in some non-volatile directory like /var, etc, to resist during the board reset

system/nxpkg/README.txt
linguini1, line 1: Why is there a readme.txt? This is not how things are documented in NuttX.

linguini1 (changes requested):

system/nxpkg/pkg_install.c
```diff
@@ -91,6 +107,82 @@ static int pkg_install_acquire_lock(FAR const char *name, FAR char *path,
   return 0;
 }
 
+/****************************************************************************
+ * Name: pkg_install_acquire_installed_lock
+ *
+ * Description:
```
halyssonJr: Shorten the function description. Specific implementation details might be more valuable in separate documentation.

system/nxpkg/pkg_repo.c
```diff
@@ -0,0 +1,612 @@
...
+ ****************************************************************************/
+
+bool pkg_source_is_url(FAR const char *source)
+{
+  if (source == NULL)
+    {
+      return false;
+    }
+
+  return strncasecmp(source, "http://", 7) == 0 ||
```
halyssonJr: I recommend creating a macro and using `strlen` to return the number of characters.

```
#define PKG_REPO_HTTP "http://"
#define PKG_REPO_HTTPS "https://"

strncasecmp(source, PKG_REPO_HTTP, strlen(PKG_REPO_HTTP)) 
```

system/nxpkg/pkg_repo.c
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+  index = pkg_zalloc(sizeof(*index));
+  tmp = pkg_path_alloc();
+  index_path = pkg_path_alloc();
+  source_path = pkg_path_alloc();
+  if (index == NULL || tmp == NULL || index_path == NULL ||
+      source_path == NULL)
+    {
+      pkg_free(index);
```
halyssonJr: What do you think about using `go to` in this case?
For example
```
ret = pkg_store_prepare_layout();
  if (ret < 0)
{
 goto exit:
}
...
exit:
  pkg_store_remove_file(tmp);
  pkg_free(text);
  pkg_free(index);
  pkg_free(tmp);
  pkg_free(index_path);
  pkg_free(source_path);
     
  return ret;
```
cc:  @acassis  @linguini1

system/nxpkg/pkg_store.c
```diff
@@ -329,70 +368,167 @@ int pkg_store_read_text(FAR const char *path, FAR char **buffer)
...
+  ret = pkg_store_write_all(fd, text, strlen(text));
   if (ret < 0)
     {
+      close(fd);
+      unlink(path);
       return ret;
     }
 
-  if ((size_t)ret >= sizeof(tmp))
+  if (close(fd) < 0)
```
halyssonJr: Please use `ret` to receive the function's return value, and then you can check it.

system/nxpkg/pkg_store.c
```diff
@@ -409,6 +545,22 @@ int pkg_store_write_text_atomic(FAR const char *path, FAR const char *text)
...
+  /* Force the write through to disk before renaming.  nxpkg writes
+   * several small, unrelated files back-to-back during install (per-
+   * package txn state, then the shared installed-packages database);
+   * without an explicit sync here, the FAT driver's single shared
+   * sector cache can still hold a not-yet-committed buffer for this
+   * file when the very next atomic write starts touching a different
+   * file, corrupting one or both.
+   */
+
+  if (fsync(fd) < 0)
```
halyssonJr: THe same here.

system/nxpkg/Kconfig
aviralgarg05: Updated the default to `/var/lib/nxpkg`. The Kconfig help now states that the board must mount persistent storage there or override `CONFIG_SYSTEM_NXPKG_ROOT`.

system/nxpkg/README.txt
aviralgarg05: Removed the in-tree README. The companion user documentation is now in apache/nuttx#18875.

system/nxpkg/pkg_install.c
aviralgarg05: Shortened the description to the purpose of the installed-database lock and its serialized read-modify-write operation.

system/nxpkg/pkg_repo.c
aviralgarg05: Updated the protocol checks to use `PKG_REPO_HTTP` and `PKG_REPO_HTTPS`, with `strlen()` providing their lengths.

system/nxpkg/pkg_repo.c
aviralgarg05: Refactored `pkg_sync()` to use one `out:` path that owns temporary-file and allocated-state cleanup.

system/nxpkg/pkg_store.c
aviralgarg05: Updated the atomic-write path to assign the operation result to `ret` before checking it, while preserving `errno` before cleanup.

system/nxpkg/pkg_store.c
aviralgarg05: Applied the same `ret` handling to the close and rename paths.

aviralgarg05:
Rebased onto current `master` (`2aebae740`) and squashed to a single commit.

The squash also removes an artefact of the previous history: the first commit added `system/nxpkg/README.txt` and a later commit removed it again after review, which made `tools/checkpatch.sh` fail on the intermediate commit while trying to resolve a path that no longer exists. All checks pass on the squashed commit.

The diff is limited to `system/nxpkg/`. Keeping this as a draft until I attach the full HTTP lifecycle transcript (sync, install, update, rollback, remove, and checksum rejection) from the ESP32-S3 target.

system/nxpkg/Makefile
```diff
@@ -29,6 +29,7 @@ MODULE    = $(CONFIG_SYSTEM_NXPKG)
 
 CSRCS = pkg_compat.c pkg_hash.c pkg_install.c pkg_log.c pkg_manifest.c
 CSRCS += pkg_metadata.c pkg_store.c pkg_txn.c
+CSRCS += pkg_repo.c
```
xiaoxiang781216: merge to previous line

linguini1:
This is a 2.5k LoC patch with 1 commit. Your PR lists a bunch of functionally different features in this PR (i.e adding an icon, adding SHA checks). Please split this into multiple commits, and ideally open a few PRs.

acassis:
> This is a 2.5k LoC patch with 1 commit. Your PR lists a bunch of functionally different features in this PR (i.e adding an icon, adding SHA checks). Please split this into multiple commits, and ideally open a few PRs.

I suggest dividing in more commit inside this PR, because he has already many PRs opened and the final evaluation date is arriving

linguini1:
It won't make much difference except that the patches will be easier to review if split into many.

aviralgarg05:
> It won't make much difference except that the patches will be easier to review if split into many.

I will split and stack the PRs, it will be easy for review then

linguini1:
Thank you Aviral!

halyssonJr (approved):
Good progress, @aviralgarg05

aviralgarg05:
Split as asked. This PR now carries only the storage layer, and the rest went into three follow-ups:

- #3718 catalog and database validation
- #3719 network fetch
- #3720 install lifecycle and command line

Each declares the one before it with `Depends-On`, so CI applies them in order. Only this one builds on `master` by itself, which is why it is first.

The content is unchanged from what you already reviewed; it is the same tree, cut into six commits across four PRs.

system/nxpkg/pkg.h
```diff
@@ -27,30 +27,93 @@
...
+ * real payloads seen in practice (a multi-MB WAD, a ~1MB game ELF) with
+ * generous headroom.
+ */
+
+#define PKG_TEXT_MAX_SIZE     (256 * 1024)
+#define PKG_DOWNLOAD_MAX_SIZE (32 * 1024 * 1024)
+
+/* nxpkg is a one-shot CLI, not a daemon, so a lock file older than this
+ * cannot belong to a still-running install under normal use (even a full
+ * multi-MB artifact over a slow link finishes well within this window) -
```
linguini1: How slow of a link? Did you test this on something or is this hypothesizing. 10 minutes seems pretty reasonable for now, but I would remove this claim.

system/nxpkg/pkg.h
```diff
@@ -27,30 +27,93 @@
...
+
+static inline void pkg_free(void *ptr)
+{
+  free(ptr);
+}
+
+static inline FAR char *pkg_path_alloc(void)
+{
+  return pkg_malloc(PATH_MAX);
+}
```
linguini1: Why

system/nxpkg/pkg_compat.c
```diff
@@ -40,7 +40,16 @@ const char *pkg_runtime_arch(void)
...
 const char *pkg_runtime_compat(void)
 {
+#ifdef CONFIG_ARCH_BOARD
   return CONFIG_ARCH_BOARD;
+#elif defined(CONFIG_ARCH_BOARD_CUSTOM_NAME)
+  if (CONFIG_ARCH_BOARD_CUSTOM_NAME[0] != '\0')
+    {
+      return CONFIG_ARCH_BOARD_CUSTOM_NAME;
+    }
+#endif
```
linguini1: Why check this if you'll just return the empty string anyways?

system/nxpkg/pkg_log.c
```diff
@@ -26,20 +26,28 @@
...
+  int ret;
+
+  ret = vsnprintf(message, sizeof(message), fmt, ap);
+  if (ret < 0)
+    {
+      return;
+    }
+
+  syslog(strcmp(level, "error") == 0 ? LOG_ERR : LOG_INFO,
+         "nxpkg: %s: %s", level, message);
```
linguini1: Don't add level to the message. There is already a syslog option that allows level to be logged.

I would suggest a macro that just compile time prefixes the message with `nxpkg` if you need that, but there is also a syslog option that allows logging the process name as well which is a better choice.

system/nxpkg/pkg_store.c
```diff
@@ -35,10 +40,84 @@
...
+ ****************************************************************************/
+
+#define PKG_LOCK_RECORD_MAGIC "NXPKG1"
+#define PKG_LOCK_RECORD_SIZE  64
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static pthread_once_t g_pkg_lock_boot_once = PTHREAD_ONCE_INIT;
```
xiaoxiang781216: could we remove the global changed variable from nxpkg? so multiple instances could run concurrently.

system/nxpkg/pkg_store.c
```diff
@@ -188,6 +272,119 @@ int pkg_store_prepare_layout(void)
...
+      (now - st.st_mtime) < PKG_LOCK_STALE_SECONDS)
+    {
+      return;
+    }
+
+  pkg_error("reclaiming legacy stale lock '%s' (age %ld s)",
+            path, (long)(now - st.st_mtime));
+  unlink(path);
+}
+
```
linguini1: Is compatibility necessary? This application is quite new and unfinished anyways.

system/nxpkg/pkg_store.c
```diff
@@ -266,21 +468,52 @@ int pkg_store_format_previous_path(FAR char *buffer, size_t size,
...
 {
-  return pkg_store_format(buffer, size, PKG_TMP_PKG_DIR "/%s-%s.npkg", name,
-                          version);
+  int ret;
+
+  UNUSED(name);
+  UNUSED(version);
+
+  /* This used to be "PKG_TMP_PKG_DIR/name-version.pkg", which breaks on
+   * this SD card's short-name-only FAT mount as soon as name+version
```
linguini1: Don't mention "this SD card". Just refer to short name FAT. This is an application

system/nxpkg/pkg_store.c
```diff
@@ -266,21 +468,52 @@ int pkg_store_format_previous_path(FAR char *buffer, size_t size,
...
+   * exceeds the 8.3 8-character base-name limit - e.g. "nxdoom-1" (8
+   * chars) fits and installs fine, but "nxdoom-10" or "nxdoom-9.1" (9+
+   * chars) fails the open(O_CREAT) in pkg_repo_fetch_url() with
+   * -EINVAL, surfacing as "acquire source failed: -22" for any
+   * multi-character version - independent of name/version length here,
+   * unlike pkg_store_make_tmp_path()'s already-FAT-safe scheme.  The
+   * pid is small, bounded, and unique per concurrently running install
+   * (each `nxpkg install` is its own process with its own per-name
+   * lock), so it can't collide the way a single fixed name would if
+   * two different packages were being installed at once.
```
linguini1: This function just checks a length. We don't need to know about another function's safe naming scheme here, it's confusing. Check the AI generated comments for relevance.

linguini1 (dismissed):

system/nxpkg/pkg_store.c
aviralgarg05: Fixed

xiaoxiang781216:
@aviralgarg05 sqaush your temp change into patch.

halyssonJr:
@aviralgarg05, 

Use this way: 
```
git rebase -i HEAD~ N
git push origin your-branch -f
```

xiaoxiang781216 (approved):

acassis (approved):

PR #3646, opened 2026-07-16
interpreters: add Tcl support

csanchezdll:
## Summary

Tcl is added using The Jim Interpreter. Jim is an opensource small-footprint implementation of the Tcl programming language. It implements a large subset of Tcl and adds new features like references with garbage collection, closures, built-in Object Oriented Programming system, Functional Programming commands, first-class arrays and UTF-8 support.

## Impact

Build should be straightforward, just enable the application inside the "Interpreters" section of the configuration. You need to make sure you have `setjmp` support, so consider enabling it in "Build Setup" -> "Customize Header Files".

## Testing

Tested on master, with a minimal custom board with a STM32H5. Planning to test on a nucleo as soon as I can get one.

```
NuttShell (NSH) NuttX-13.0.0
nsh> jimsh
Welcome to Jim version 0.83
. puts "hello world"
. exit
hello world
nsh> jimsh
Welcome to Jim version 0.83
. foreach i {a b c} {
{> puts $i
{> }
. exit
a
b
c
nsh> 
```

### DIFF
--- a/interpreters/jimtcl/Kconfig
+++ b/interpreters/jimtcl/Kconfig
@@ -1,0 +1,12 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config INTERPRETERS_JIMTCL
+	bool "The Jim (TCL) Interpreter"
+	default n
+	---help---
+		Enable support for The Jim interpreter.
+		Jim is an opensource small-footprint implementation of the
+		Tcl programming language.
--- a/interpreters/jimtcl/Makefile
+++ b/interpreters/jimtcl/Makefile
@@ -1,0 +1,64 @@
+include $(APPDIR)/Make.defs
+
+PROGNAME  = jimsh
+PRIORITY  = 100
+STACKSIZE = 8192
+
+JIMTCL_VERSION  = 0.83
+JIMTCL_TARBALL  = $(JIMTCL_VERSION).tar.gz
+JIMTCL_UNPACK   = jimtcl
+JIMTCL_URL_BASE = https://github.com/msteveb/jimtcl/archive/refs/tags
+JIMTCL_URL      = $(JIMTCL_URL_BASE)/$(JIMTCL_TARBALL)
+JIMTCL_SRC      = $(JIMTCL_UNPACK)
+
+MAINSRC = $(JIMTCL_SRC)$(DELIM)jimsh.c
+
+# Obtain object list from jimtcl Makefile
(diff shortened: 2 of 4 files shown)


interpreters/jimtcl/Makefile
```diff
@@ -0,0 +1,64 @@
+include $(APPDIR)/Make.defs
+
+PROGNAME  = jimsh
+PRIORITY  = 100
+STACKSIZE = 8192
+
+JIMTCL_VERSION  = 0.83
+JIMTCL_TARBALL  = $(JIMTCL_VERSION).tar.gz
+JIMTCL_UNPACK   = jimtcl
+JIMTCL_URL_BASE = https://github.com/msteveb/jimtcl/archive/refs/tags
```
xiaoxiang781216, line 10: it's possible to add Cmake too?

csanchezdll: Sure! I'll add it.

interpreters/jimtcl/Kconfig
```diff
@@ -0,0 +1,12 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config INTERPRETERS_JIMTCL
+	bool "The Jim (TCL) Interpreter"
+	default n
```
linguini1, line 8: You should add `depends on` here for the config enabling setjmp support if that's a requirement as you mentioned

linguini1:
Cool!

acassis:
Nice work @csanchezdll ! Jim is used on OpenOCD too. We have many Interpreters on NuttX, I'm curious to know which one has better performance to some specific tasks (doing some similar task as comparing the time each one delay)

interpreters/jimtcl/Kconfig
csanchezdll: In theory, setjmp support usually comes from the toolchain. That's why by default ARCH_SETJMP is off. Some architectures might be able to use it from the toolchain directly (no need to select ARCH_SETJMP) but others do require it (this is the case for arm-none-eabi I am using, because toolchain setjmp comes from newlib implementation in libc, and it not linked in NuttX).
Making a depend here will force to use the NuttX alternative even for architectures where the toolchain provide a working setjmp.

linguini1: That makes sense!

linguini1 (dismissed):

interpreters/jimtcl/Makefile
csanchezdll: CMake support added. I think the PR is ready now, unmarking draft.

acassis (approved):

xiaoxiang781216 (approved):

PR #3658, opened 2026-07-23
netutils/telnetd: fix silent listen-socket loss and per-connection daemon death

ricardgb:
## Summary

Two robustness fixes for the telnet daemon (`netutils/telnetd/telnetd_daemon.c`). As it stands, the standard NuttX telnet console can be disabled by a single aborted connection, and on some configurations it silently serves nothing from boot.

**1. Keep the listen socket out of the standard-stream range.**
When the daemon starts without open standard streams — exactly what happens when `nsh_telnetstart()` spawns `telnetd &` before the console device exists (e.g. `CONFIG_NSH_USBCONSOLE` boards, where `nsh_initialize()` runs before the USB console is connected) — `socket()` returns a descriptor in 0..2. The accept loop's own "go silent" `close(0)..close(2)` then destroys the listen socket on the first iteration; every subsequent `accept4()` fails and the daemon serves nothing, while appearing alive in the task list. Fix: `F_DUPFD_CLOEXEC` the descriptor above 2 before use.

**2. Survive transient connection errors instead of exiting.**
Previously any single failed connection killed the whole daemon:
- a peer resetting before the accept completes (a port scanner, or a plain `nc -z` probe) surfaces as an `accept4()` error such as `ECONNABORTED` and took the errout path;
- the per-connection error paths after a successful accept (setsockopt, `/dev/telnet` open, `SIOCTELNET`, session-device open, session spawn) likewise exited the daemon.

One bad or aborted connection and the telnet console is dead until reboot — a trivial remote way to take out the console of any reachable NuttX device. Fix: treat these as per-connection failures (drop the connection, keep accepting). Genuinely unrecoverable `accept4()` errors (`EBADF`/`ENOTSOCK`/`EINVAL`/`EOPNOTSUPP`) still exit loudly, and a short pause on repeated transient failures avoids busy-spinning while an interface is down. Daemon setup errors (socket/bind/listen) exit as before.

## Impact

Any board using `netutils/telnetd` (directly or via `CONFIG_NSH_TELNET`). Bug 1 affects configurations where the daemon can start before its standard streams exist (USB-console boards being the common case). Bug 2 affects everyone: a subnet `nmap` sweep kills every reachable NuttX telnet console.

## Testing

Reproduced and verified on hardware: Raspberry Pi Pico 2 W (RP2350), composite USB CDC-ACM console + CDC-NCM network, `CONFIG_NSH_USBCONSOLE`, telnetd auto-started via `nsh_telnetstart()`.

- Before (bug 1): daemon task alive but port 23 refused all connections from boot; the RTOS-aware debugger thread list showed it looping on the failed accept with the listen socket closed by its own `close(0)`.
- Before (bug 2): a single `nc -zv <addr> 23` probe permanently killed a manually started, correctly listening daemon.
- After: daemon listens from boot (within 5 s of enumeration), survives repeated `nc -z` probe barrages, and interactive nsh telnet sessions work.

`tools/nxstyle.c` passes on the modified file.

## Note

These changes were developed with the assistance of an AI agent (Claude) and have been human-reviewed and tested on real hardware as described above.

### DIFF
--- a/netutils/telnetd/telnetd_daemon.c
+++ b/netutils/telnetd/telnetd_daemon.c
@@ -126,12 +126,11 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
     }
 
   /* If the daemon was started without standard streams (e.g. spawned by
-   * nsh_telnetstart() before a USB console device exists), socket() may
-   * have returned a descriptor in 0..2.  The "go silent"
-   * close(0)..close(2) at the top of the accept loop below would then
-   * destroy the listen socket: every subsequent accept4() fails and the
-   * daemon serves nothing.  Move the descriptor above the
-   * standard-stream range.
+   * nsh_telnetstart() before a USB console exists), socket() may have
+   * returned a descriptor in 0..2.  The "go silent" close(0)..close(2)
+   * at the top of the accept loop below would then destroy the listen
+   * socket: every subsequent accept4() fails and the daemon serves
+   * nothing.  Move the descriptor above the standard-stream range.
    */
 
   if (listensd <= 2)
@@ -224,17 +223,33 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
       acceptsd = accept4(listensd, &addr.generic, &addrlen, SOCK_CLOEXEC);
       if (acceptsd < 0)
         {
-          /* Just continue if a signal was received */
+          int err = errno;
 
-          if (errno == EINTR)
+          /* A bad listen socket can never recover - exiting loudly beats
+           * spinning forever while serving nothing.
+           */
+
+          if (err == EBADF || err == ENOTSOCK || err == EINVAL ||
+              err == EOPNOTSUPP)
             {
-              continue;
-            }
-          else
-            {
-              nerr("ERROR: accept failed: %d\n", errno);
+              nerr("ERROR: accept failed fatally: %d\n", err);
               goto errout_with_socket;
             }
+
+          /* Everything else is transient: a peer that reset before the
+           * accept completed (a port scan, nc -z, ECONNABORTED), a
+           * signal, or the interface bouncing.  None of those may take
+           * the daemon down; the pause keeps a persistent transient
+           * (interface down) from busy-spinning.
+           */
+
+          nerr("ERROR: accept failed: %d\n", err);
+          if (err != EINTR)
+            {
+              usleep(100 * 1000);
+            }
+
+          continue;
         }
 
 #ifdef CONFIG_NET_SOLINGER
@@ -247,8 +262,14 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
       if (setsockopt(acceptsd, SOL_SOCKET, SO_LINGER,
                      &ling, sizeof(struct linger)) < 0)
         {
+          /* Per-connection failure (e.g. the peer already reset the
+           * connection - a port scan does this).  Drop the connection and
+           * keep accepting; exiting here lets any probe kill the daemon.
+           */
+
           nerr("ERROR: setsockopt failed: %d\n", errno);
-          goto errout_with_acceptsd;
+          close(acceptsd);
+          continue;
         }
 #endif
 
@@ -258,7 +279,8 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
       if (drvrfd < 0)
         {
           nerr("ERROR: open(/dev/telnet) failed: %d\n", errno);
-          goto errout_with_acceptsd;
+          close(acceptsd);
+          continue;
         }
 
       /* Create a character device to "wrap" the accepted socket descriptor */
@@ -271,9 +293,10 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
       if (ioctl(drvrfd, SIOCTELNET,
                 (unsigned long)((uintptr_t)&session)) < 0)
         {
-          nerr("ERROR: open(/dev/telnet) failed: %d\n", errno);
+          nerr("ERROR: SIOCTELNET failed: %d\n", errno);
           close(drvrfd);
-          goto errout_with_acceptsd;
+          close(acceptsd);
+          continue;
         }
 
       close(drvrfd);
@@ -284,8 +307,12 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
       drvrfd = open(session.ts_devpath, O_RDWR);
       if (drvrfd < 0)
         {
+          /* The socket now belongs to the telnet driver instance; just
+           * keep accepting.
+           */
+
           nerr("ERROR: Failed to open %s: %d\n", session.ts_devpath, errno);
-          goto errout_with_socket;
+          continue;
         }
 
       /* Use this driver as stdin, stdout, and stderror */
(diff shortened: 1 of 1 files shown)


acassis (approved):

xiaoxiang781216 (approved):

netutils/telnetd/telnetd_daemon.c
```diff
@@ -125,6 +125,27 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
       goto errout;
     }
 
+  /* If the daemon was started without standard streams (e.g. spawned by
+   * nsh_telnetstart() before a USB console exists), socket() may have
+   * returned a descriptor in 0..2.  The "go silent" close(0)..close(2)
```
xiaoxiang781216, line 130: could we fix the original problem directly @ricardgb

ricardgb: Something like this?

```diff
diff --git a/netutils/telnetd/telnetd_daemon.c b/netutils/telnetd/telnetd_daemon.c
index 3662fa4aa..e0a7df5be 100644
--- a/netutils/telnetd/telnetd_daemon.c
+++ b/netutils/telnetd/telnetd_daemon.c
@@ -81,6 +81,7 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
   socklen_t addrlen;
   int listensd;
   int acceptsd;
+  int nullfd;
 #ifdef CONFIG_NET_SOCKOPTS
   int optval;
 #endif
@@ -115,6 +116,31 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
     }
 #endif /* CONFIG_SCHED_HAVE_PARENT && CONFIG_ENABLE_ALL_SIGNALS */
 
+  /* If the daemon was started without standard streams (e.g. spawned by
+   * nsh_telnetstart() before a USB console exists), descriptors 0..2 are
+   * free and any descriptor created below -- the listen socket, an
+   * accepted socket or a session driver -- could be allocated in the
+   * standard-stream range, where the per-session dup2(0..2) would clobber
+   * it.  Anchor any free slot in 0..2 to /dev/null first so that can
+   * never happen.
+   */
+
+  for (; ; )
+    {
+      nullfd = open("/dev/null", O_RDWR);
+      if (nullfd < 0)
+        {
+          nerr("ERROR: open(/dev/null) failed: %d\n", errno);
+          goto errout;
+        }
+
+      if (nullfd > 2)
+        {
+          close(nullfd);
+          break;
+        }
+    }
+
   /* Create a new TCP socket to use to listen for connections */
 
   listensd = socket(config->d_family, SOCK_STREAM | SOCK_CLOEXEC, 0);
@@ -125,27 +151,6 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
       goto errout;
     }
 
-  /* If the daemon was started without standard streams (e.g. spawned by
-   * nsh_telnetstart() before a USB console exists), socket() may have
-   * returned a descriptor in 0..2.  The "go silent" close(0)..close(2)
-   * at the top of the accept loop below would then destroy the listen
-   * socket: every subsequent accept4() fails and the daemon serves
-   * nothing.  Move the descriptor above the standard-stream range.
-   */
-
-  if (listensd <= 2)
-    {
-      int highsd = fcntl(listensd, F_DUPFD_CLOEXEC, 3);
-      if (highsd < 0)
-        {
-          nerr("ERROR: F_DUPFD_CLOEXEC failed: %d\n", errno);
-          goto errout_with_socket;
-        }
-
-      close(listensd);
-      listensd = highsd;
-    }
-
 #ifdef CONFIG_NET_SOCKOPTS
   /* Set socket to reuse address */
 
@@ -211,11 +216,24 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
 #endif
       int drvrfd;
 
-      /* Now go silent. */
+      /* Now go silent: detach from the previous session's driver (or the
+       * startup streams) by re-anchoring 0..2 to /dev/null instead of
+       * closing them.  Closing would leave 0..2 free, and accept4() or
+       * open() below could then allocate into the standard-stream range
+       * where the dup2(0..2) for the next session would clobber it.
+       */
 
-      close(0);
-      close(1);
-      close(2);
+      nullfd = open("/dev/null", O_RDWR);
+      if (nullfd >= 0)
+        {
+          dup2(nullfd, 0);
+          dup2(nullfd, 1);
+          dup2(nullfd, 2);
+          if (nullfd > 2)
+            {
+              close(nullfd);
+            }
+        }
 
       ninfo("Accepting connections on port %d\n", ntohs(config->d_port));
 
```

This removes the listen-socket relocation entirely and instead (1) anchors any free slot in 0..2 to `/dev/null` before the socket is created, and (2) re-anchors 0..2 to `/dev/null` at the top of the accept loop instead of closing them — so no descriptor (listen, accept or driver) can ever be allocated into the standard-stream range in the first place. Compile-tested (rp2350 telnet config, builds clean, nxstyle clean); I'd re-run the on-hardware test before pushing it. If you prefer this shape I'll update the PR.

netutils/telnetd/telnetd_daemon.c
```diff
@@ -224,17 +223,33 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
...
+           * the daemon down; the pause keeps a persistent transient
+           * (interface down) from busy-spinning.
+           */
+
+          nerr("ERROR: accept failed: %d\n", err);
+          if (err != EINTR)
             {
-              nerr("ERROR: accept failed: %d\n", errno);
-              goto errout_with_socket;
+              usleep(100 * 1000);
```
xiaoxiang781216, line 249: why sleep

netutils/telnetd/telnetd_daemon.c
```diff
@@ -224,17 +223,33 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
...
-          /* Just continue if a signal was received */
+          int err = errno;
 
-          if (errno == EINTR)
+          /* A bad listen socket can never recover - exiting loudly beats
+           * spinning forever while serving nothing.
+           */
+
+          if (err == EBADF || err == ENOTSOCK || err == EINVAL ||
+              err == EOPNOTSUPP)
```
xiaoxiang781216, line 233: why not ignore the error and continue directly

netutils/telnetd/telnetd_daemon.c
```diff
@@ -126,12 +126,11 @@ int telnetd_daemon(FAR const struct telnetd_config_s *config)
     }
 
   /* If the daemon was started without standard streams (e.g. spawned by
-   * nsh_telnetstart() before a USB console device exists), socket() may
-   * have returned a descriptor in 0..2.  The "go silent"
-   * close(0)..close(2) at the top of the accept loop below would then
-   * destroy the listen socket: every subsequent accept4() fails and the
-   * daemon serves nothing.  Move the descriptor above the
-   * standard-stream range.
+   * nsh_telnetstart() before a USB console exists), socket() may have
```
xiaoxiang781216, line 129: move to previous patch if you really want to modify

ricardgb: Thanks @xiaoxiang781216 . Good questions on the accept-loop error handling —
I'd like to hold this one briefly before I revise it. I'm coordinating a
related change with the maintainers and want the daemon's error handling to
line up with it rather than reworking this twice. I'll follow up here with the
updates (the fatal-vs-transient split, the sleep, and folding the comment
change into the earlier commit) once that's settled. Appreciate the review.

linguini1 (changes requested):
As always, please include logs from your runtime testing :)

PR #3665, opened 2026-07-25
xipfs: compaction command, test suite and NXFLAT execute-in-place demo

casaroli:
## Summary

The applications that go with **XIPFS**, the writable execute-in-place file
system added in the companion PR:

**Companion PR: apache/nuttx#19536**

This one depends on that one and cannot be built without it: every Kconfig
option here depends on `CONFIG_FS_XIPFS`.

Three additions:

* **`system/xipfs`** — a command that compacts a volume and reports how its
  erase blocks are laid out. Because a XIPFS file is a single contiguous
  extent, an allocation can fail for lack of a contiguous run while plenty of
  free space remains, and compaction is never automatic: the file system
  returns `-ENOSPC` and the caller decides. This command is that decision made
  by hand. `-n` surveys without moving anything and prints the per-file table,
  a block map, and the share of free space a single allocation cannot reach —
  which is the number a caller facing `-ENOSPC` actually wants. The pass is
  asked for through a descriptor for the mountpoint directory, so no file is
  held open while it runs and one pass reaches every extent.

* **`testing/fs/xipfs`** — the test suite, 14 sections selectable by name
  because the power-loss sweeps run for minutes while everything else runs in
  seconds. Beyond the routine VFS paths it aims at the properties that are
  easy to get wrong and quiet when they are: that a pin taken by a task that
  then *dies without unmapping* is still released, and that failing the Nth
  flash operation and remounting always yields a consistent volume with every
  committed file intact. The sweeps cover create, unlink, mkdir/rmdir and
  defragmentation, each with the failing operation left torn as well as
  cleanly refused.

* **`examples/nxflatxip`** — the end-to-end demonstration: write an NXFLAT
  module into the volume at run time the way a download would, confirm the
  file system hands out a real flash pointer for it, and run two instances
  concurrently. Both report the same text address, inside the flash window,
  while their stacks differ. Two subcommands cover the file system rather than
  the module: `bench` times a bulk read out of mapped flash before and after a
  write, and `defrag` fragments the volume until an allocation genuinely
  cannot be satisfied, compacts, retries, then verifies every relocated file
  byte-for-byte and again after a remount.

### A note on the demo module

The module is built in-tree exactly the way `examples/nxflat` builds its test
programs, so it needs the same two host tools, `mknxflat` and `ldnxflat`.

It deliberately has **no static data and no string constants**, and reports
through a callback into the firmware instead of formatting its own output.
That is not stylistic. The `ldnxflat` in circulation resolves the GOT entries
for `.bss` objects into the import table — so the module overwrites the
addresses of the functions it imports and branches into nothing on its next
call out — and drops the addend already in the instruction when resolving the
PC-relative references to `.rodata`, so every string pointer lands in the
middle of the code. Both were diagnosed on hardware; the comment at the top of
`module/xipmod.c` records them. The callback also exercises the other
direction of the interface: firmware code entered with the module's data base
still live in its PIC register.

## Impact

* **New applications, all off by default**, and all depending on
  `CONFIG_FS_XIPFS`, so nothing changes for existing configurations.
* `examples/nxflatxip` additionally depends on `NXFLAT` and `LIBC_EXECFUNCS`,
  and its build needs `mknxflat`/`ldnxflat` and a board `Make.defs` that names
  them — the same requirement `examples/nxflat` already has.
* `testing/fs/xipfs` writes and erases its volume continuously, the power-loss
  sections most of all. The Kconfig help and the documentation both say to
  point it at a volume whose contents do not matter.
* No existing file is modified.

## Testing

Host: macOS 15 (arm64), `arm-none-eabi-gcc 14.2.rel1`.

Targets:

* `sim:xipfs` — XIPFS on `rammtd`.
* `pimoroni-pico-2-plus:xipfs` — XIPFS on real QSPI NOR via `/dev/rpflash`.

Every hardware run was bracketed by a `uname -a` check of the git revision and
board name, before the first command and after the last, so the logs cannot be
from a different build.

### Full suite on real flash

```
[identity ok before] rev 0e9e189444, board pimoroni-pico-2-plus
-- basic VFS --                      -- directories --
-- write-once enforcement --         -- defragmentation --
-- strict mapping semantics --       -- power loss during create (clean) --
-- XIP mapping --                    -- power loss during create (torn) --
-- concurrent mappings --            -- power loss during unlink --
-- cross-task pin refcounting --     -- power loss during mkdir and rmdir --
-- pin release on task exit --       -- power loss during defrag --
    swept 110 injection points, covering every operation of the whole pass
==== 90 passed, 0 failed ====
[identity ok after] rev 0e9e189444, board pimoroni-pico-2-plus
```

Same suite on `sim:xipfs`: `==== 90 passed, 0 failed ====`.

### `nxflatxip` on hardware

```
=== NXFLAT module executed in place from xipfs ===
staged 268 bytes to /mnt/xipfs/xipmod
extent: block 97 x1, size 268, flash addr 0x10161000
running two concurrent instances...
[while running] pin count on the shared text = 2
  instance seed 1: text 0x10161070, stack 0x20006980, sum 2080 -- data private and intact
  instance seed 2: text 0x10161070, stack 0x20007588, sum 4160 -- data private and intact
[after exit]    pin count on the shared text = 2 (released when this task exits)
=== done ===
```

### `nxflatxip defrag` on hardware

```
[1] writing 55 files...
[2] deleting every other file to fragment it...
[3] requesting a 209-block (836 KB) contiguous file
    result: FAILED -ENOSPC (free space is not contiguous)
[4] compacting...
    elapsed          : 1230 ms
    extents relocated: 27
    blocks reclaimed : 27
    largest free run : 913408 bytes (223 blocks)
    stop reason      : 0 (done)
[5] retrying, sized to the reported largest run (223 blocks)
    result: OK -- allocation now fits
[6] verifying every relocated file byte-for-byte
    all survivors intact
[7] remounting to prove it is durable
    all files intact after remount
=== defrag complete ===
```

### `nxflatxip bench` on hardware

```
XIP read benchmark (256 KB)
  before any flash write : 30 ms
  after  a flash write   : 20 ms
```

### `xipfs` command on hardware

```
nsh> xipfs -n
/mnt/xipfs: 250 blocks of 4096 bytes (1000 KB)
  file                             start  blocks    bytes  pin
  bin/greet                            0       1        6    0
  bin/sub/deep                         1       1        5    0
     0 |##................................................|
  '.' free   '#' in use   'P' pinned by a live mapping
  used 2, free 248, largest free run 248 blocks (992 KB)
  1 free run, fragmentation 0%
  nothing to gain: all free space is already contiguous
```

`tools/checkpatch.sh -c -u -m -g` passes on every commit in the branch.


acassis (dismissed):

casaroli:
The failing `sim:alsa` CI job is unrelated to this PR's XIPFS/NXFLAT example, test, and command changes.

The `Linux (sim-01)` log fails while linking the bundled LAME encoder, with unresolved x86 runtime-dispatch vector symbols including `quantize_lines_xrpow_avx2`, `quantize_lines_xrpow_sse2`, and `count_bit_esc_sse2`. This is a pre-existing CMake source-integration issue: the LAME CMake target omits the vector translation units enabled by its generated configuration.

The fixes are tracked separately:

- Root-cause Apps fix: https://github.com/apache/nuttx-apps/pull/3666
- Companion NuttX codec-header integration fix: https://github.com/apache/nuttx/pull/19541

The two fixes were validated together with the CI-equivalent `sim:alsa` CMake/Ninja build. The Apps PR should merge first; subsequent NuttX PR CI runs will then use the corrected Apps `master`.

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+  while ((option = getopt(argc, argv, "nt:h")) != ERROR)
+    {
+      switch (option)
+        {
+          case 'n':
+            dryrun = true;
+            break;
+
+          case 't':
+            max_ms = (uint32_t)atoi(optarg);
```
xiaoxiang781216: remove the cast

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+  if (s->nfiles == 0)
+    {
+      printf("\n%s is empty; nothing to compact\n", mount);
+      return OK;
+    }
+
+  memset(&arg, 0, sizeof(arg));
+  arg.max_ms = max_ms;
+
+  fd = open(mount, O_RDONLY | O_DIRECTORY);
```
xiaoxiang781216: add | O_CLOEXEC

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+      return -errno;
+    }
+
+  printf("  %s: moved %lu extent%s, reclaimed %lu block%s, %lu pinned\n",
+         reason_str(arg.result.reason),
+         (unsigned long)arg.result.extents_moved,
+         arg.result.extents_moved == 1 ? "" : "s",
+         (unsigned long)arg.result.blocks_reclaimed,
+         arg.result.blocks_reclaimed == 1 ? "" : "s",
+         (unsigned long)arg.result.blocks_pinned);
```
xiaoxiang781216: remove ALL casts

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+
+  printf("  largest free run now %lu bytes\n",
+         (unsigned long)arg.result.largest_free_run);
```
xiaoxiang781216: ditto

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+                          FAR struct xipfs_survey_s *s)
+{
+  uint32_t frag;
+  uint32_t i;
+  int width;
+  int f;
+
+  printf("%s: %lu blocks of %lu bytes (%lu KB)\n",
+         mount, (unsigned long)s->nblocks, (unsigned long)s->blocksize,
+         (unsigned long)(s->nblocks * (uint64_t)s->blocksize / 1024));
```
xiaoxiang781216: remove ALL cast

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+    {
+      /* A name here is a path relative to the mountpoint, so it can be
+       * longer than one component.  Widen the column to the longest one
+       * rather than let it push the rest of the row out of alignment.
+       */
+
+      width = XIPFS_NAME_MAX;
+      for (f = 0; f < s->nfiles; f++)
+        {
+          int len = (int)strlen(s->files[f].name);
```
xiaoxiang781216, line 372: remove the cast

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+      printf("\n  %-*s %6s %7s %8s %4s\n", width, "file",
+             "start", "blocks", "bytes", "pin");
+
+      for (f = 0; f < s->nfiles; f++)
+        {
+          FAR struct xipfs_entry_s *e = &s->files[f];
+
+          printf("  %-*s %6lu %7lu %8lu %4lu\n", width, e->name,
+                 (unsigned long)e->start, (unsigned long)e->nblocks,
+                 (unsigned long)e->size, (unsigned long)e->pincount);
```
xiaoxiang781216: remove ALL casts

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+      char row[MAP_COLS + 1];
+
+      if (n > MAP_COLS)
+        {
+          n = MAP_COLS;
+        }
+
+      memcpy(row, &s->map[i], n);
+      row[n] = '\0';
+      printf("  %4lu |%s|\n", (unsigned long)i, row);
```
xiaoxiang781216: remove the cast

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+
+  /* The share of free space that a single allocation cannot reach.  This is
+   * the number that says whether compacting is worth doing: at 0% the
+   * largest possible file already fits, whatever the map looks like.
+   */
+
+  frag = 0;
+  if (s->freeblocks > 0)
+    {
+      frag = (uint32_t)(((uint64_t)(s->freeblocks - s->largestrun) * 100) /
```
xiaoxiang781216, line 420: remove the cast

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+                        s->freeblocks);
+    }
+
+  printf("\n  used %lu, free %lu, largest free run %lu blocks (%lu KB)\n",
+         (unsigned long)s->used, (unsigned long)s->freeblocks,
+         (unsigned long)s->largestrun,
+         (unsigned long)(s->largestrun * (uint64_t)s->blocksize / 1024));
+  printf("  %lu free run%s, fragmentation %lu%%\n",
+         (unsigned long)s->nruns, s->nruns == 1 ? "" : "s",
+         (unsigned long)frag);
```
xiaoxiang781216: remove ALL cast

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+
+  if (s->pinned > 0)
+    {
+      printf("  %lu block%s pinned and cannot be relocated\n",
+             (unsigned long)s->pinned, s->pinned == 1 ? "" : "s");
```
xiaoxiang781216: ditto

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+  if (frag == 0)
+    {
+      printf("  nothing to gain: all free space is already contiguous\n");
+    }
+  else
+    {
+      printf("  compacting could raise the largest run to %lu blocks"
+             " (%lu KB)\n",
+             (unsigned long)s->freeblocks,
+             (unsigned long)(s->freeblocks * (uint64_t)s->blocksize / 1024));
```
xiaoxiang781216: ditto

system/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,579 @@
...
+          if (ret < 0)
+            {
+              closedir(dirp);
+              return ret;
+            }
+
+          continue;
+        }
+
+      fd = open(path, O_RDONLY);
```
xiaoxiang781216: O_CLOEXEC

testing/fs/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,2032 @@
...
+ * Name: fill_pattern
+ ****************************************************************************/
+
+static void fill_pattern(FAR uint8_t *buf, size_t len, uint8_t seed)
+{
+  size_t i;
+
+  for (i = 0; i < len; i++)
+    {
+      buf[i] = (uint8_t)(seed + (i * 7));
```
xiaoxiang781216, line 128: remove the cast

testing/fs/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,2032 @@
...
+ *
+ ****************************************************************************/
+
+static int create_file(FAR const char *path, size_t len, uint8_t seed)
+{
+  ssize_t nwritten;
+  int fd;
+  int ret = 0;
+
+  fd = open(path, O_WRONLY | O_CREAT | O_TRUNC, 0644);
```
xiaoxiang781216: O_CLOEXEC for ALL open

testing/fs/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,2032 @@
...
+  if (ftruncate(fd, len) < 0)
+    {
+      ret = -errno;
+      goto out;
+    }
+
+  fill_pattern(g_buffer, len, seed);
+
+  nwritten = write(fd, g_buffer, len);
+  if (nwritten != (ssize_t)len)
```
xiaoxiang781216, line 163: remove ALL cast

testing/fs/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,2032 @@
...
+
+  expect = malloc(len);
+  if (expect == NULL)
+    {
+      return false;
+    }
+
+  fill_pattern(expect, len, seed);
+
+  fd = open(path, O_RDONLY);
```
xiaoxiang781216: ditto

testing/fs/xipfs/xipfs_main.c
```diff
@@ -0,0 +1,2032 @@
...
+  if (fd < 0)
+    {
+      free(expect);
+      return false;
+    }
+
+  nread = read(fd, g_buffer, len + 1);
+  if (nread == (ssize_t)len && memcmp(g_buffer, expect, len) == 0)
```
xiaoxiang781216, line 204: remove the cast

examples/nxflatxip/module/xipmod.c
```diff
@@ -0,0 +1,116 @@
...
+    {
+      if (values[i] != seed * (i + 1))
+        {
+          errors++;
+        }
+
+      sum += values[i];
+    }
+
+  xipmod_report(seed, (const void *)main, (const void *)values, sum, errors);
```
xiaoxiang781216, line 113: remove all cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+ ****************************************************************************/
+
+static int stage_module(void)
+{
+  ssize_t nwritten;
+  int fd;
+
+  unlink(MODPATH);
+
+  fd = open(MODPATH, O_WRONLY | O_CREAT | O_TRUNC, 0755);
```
xiaoxiang781216: add O_CLOEXEC

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+    {
+      syslog(LOG_INFO, "ERROR: ftruncate failed: %d\n", errno);
+      close(fd);
+      return -errno;
+    }
+
+  nwritten = write(fd, g_xipmod, g_xipmod_len);
+  close(fd);
+
+  if (nwritten != (ssize_t)g_xipmod_len)
```
xiaoxiang781216, line 162: remove ALL cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+ *
+ ****************************************************************************/
+
+static int extent_info_path(FAR const char *path,
+                            FAR struct xipfs_extent_info_s *info)
+{
+  int fd;
+  int ret;
+
+  fd = open(path, O_RDONLY);
```
xiaoxiang781216: ditto

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+  if (ret < 0)
+    {
+      syslog(LOG_INFO, "ERROR: cannot query extent: %d\n", ret);
+      return EXIT_FAILURE;
+    }
+
+  syslog(LOG_INFO, "extent: block %" PRIu32 " x%" PRIu32 ", size %" PRIu32
+         ", flash addr 0x%08lx\n",
+         info.start_block, info.nblocks, info.size,
+         (unsigned long)info.xipaddr);
```
xiaoxiang781216: remove the cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+   * sleeps, then checks it.  Shared data would show up as a mismatch.
+   */
+
+  syslog(LOG_INFO, "\nrunning two concurrent instances...\n\n");
+
+  for (i = 0; i < 2; i++)
+    {
+      snprintf(seed[i], sizeof(seed[i]), "%d", i + 1);
+
+      args[0] = (FAR char *)"xipmod";
```
xiaoxiang781216, line 838: remove the cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+             &freeblk, &largest);
+
+  /* Bigger than any single hole, but comfortably inside the total free
+   * space -- so it can only be satisfied once the holes are coalesced.
+   */
+
+  want = largest + (freeblk - largest) / 2;
+
+  syslog(LOG_INFO, "\n[3] requesting a %lu-block (%lu KB) contiguous file\n",
+         (unsigned long)want, (unsigned long)(want * erasesize / 1024));
```
xiaoxiang781216: remove ALL cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+    }
+
+  syslog(LOG_INFO, "    elapsed          : %ld ms\n", ms);
+  syslog(LOG_INFO, "    extents relocated: %lu\n",
+         (unsigned long)res.extents_moved);
+  syslog(LOG_INFO, "    blocks reclaimed : %lu\n",
+         (unsigned long)res.blocks_reclaimed);
+  syslog(LOG_INFO, "    largest free run : %lu bytes (%lu blocks)\n",
+         (unsigned long)res.largest_free_run,
+         (unsigned long)(res.largest_free_run / erasesize));
```
xiaoxiang781216: remove ALL cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+         ret < 0 ? "still failed" : "OK -- allocation now fits");
+  unlink(MOUNTPT "/bigalloc");
+
+  syslog(LOG_INFO, "\n[6] verifying every relocated file byte-for-byte\n");
+
+  ret = 0;
+  for (i = 1; i < DF_FILES; i += 2)
+    {
+      snprintf(path, sizeof(path), MOUNTPT "/m%02d", i);
+      if (!verify_file(path, 2000, (uint8_t)(i + 1)))
```
xiaoxiang781216, line 742: remove the cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+
+  unlink(MOUNTPT "/bigalloc");
+
+  fd = open(MOUNTPT "/bigalloc", O_WRONLY | O_CREAT | O_TRUNC, 0644);
+  if (fd < 0)
+    {
+      return -errno;
+    }
+
+  ret = ftruncate(fd, (off_t)nblocks * erasesize);
```
xiaoxiang781216, line 555: ditto

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+          n = DF_MAPCOLS;
+        }
+
+      memcpy(row, &g_map[i], n);
+      row[n] = '\0';
+      syslog(LOG_INFO, "    %3lu |%s|\n", (unsigned long)i, row);
+    }
+
+  syslog(LOG_INFO, "    free %lu blocks, largest contiguous run %lu\n",
+         (unsigned long)free, (unsigned long)largest);
```
xiaoxiang781216: ditto

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+ * plenty of free space, but none of it contiguous -- then compacts and
+ * retries the same allocation.  On a real board everything below runs
+ * against real NOR: each move is a page-at-a-time copy, a metadata commit,
+ * and an erase of the vacated blocks.
+ ****************************************************************************/
+
+#define DF_FILES   55
+#define DF_MAPCOLS 50
+
+static char g_map[512];
```
xiaoxiang781216: move to macro/data section

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+      return false;
+    }
+
+  fill_pattern(want, len, seed);
+
+  fd = open(path, O_RDONLY);
+  if (fd >= 0)
+    {
+      n = read(fd, got, len);
+      ok = (n == (ssize_t)len) && (memcmp(got, want, len) == 0);
```
xiaoxiang781216, line 384: remove the cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+
+  buf = malloc(len);
+  if (buf == NULL)
+    {
+      return -ENOMEM;
+    }
+
+  fill_pattern(buf, len, seed);
+
+  fd = open(path, O_WRONLY | O_CREAT | O_TRUNC, 0644);
```
xiaoxiang781216: add O_CLOEXEC for ALL open

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+  return ret < 0 ? -errno : 0;
+}
+
+static void fill_pattern(FAR uint8_t *buf, size_t len, uint8_t seed)
+{
+  size_t i;
+
+  for (i = 0; i < len; i++)
+    {
+      buf[i] = (uint8_t)(seed + (i * 7));
```
xiaoxiang781216, line 317: remove the cast

examples/nxflatxip/nxflatxip_main.c
```diff
@@ -0,0 +1,892 @@
...
+  if (buf == NULL)
+    {
+      return -1;
+    }
+
+  fd = open(BENCH_PATH, O_RDONLY);
+  if (fd < 0)
+    {
+      free(buf);
+      return -1;
```
xiaoxiang781216, line 231: all -1 to the correct error code

system/xipfs/xipfs_main.c
casaroli: i think we should keep this to make the conversion explicit

system/xipfs/xipfs_main.c
casaroli: I think these are necessary because we do the calculation with uint64_t then truncate to 32 bit explicitly

testing/fs/xipfs/xipfs_main.c
casaroli: I think this is necessary

testing/fs/xipfs/xipfs_main.c
casaroli: I think it is better to keep the truncation explicit here

testing/fs/xipfs/xipfs_main.c
casaroli: this is necessary to avoid warning

examples/nxflatxip/nxflatxip_main.c
casaroli: we need to remove the const here from the string literal.

casaroli (commented):
@xiaoxiang781216 i addressed most of the concerns, most casts are gone with the use of <inttypes.h>, some of them are necessary (signed/unsigned comparison, larger-width arithmethic), and some of them are explicit truncation, I think we should keep them to avoid modern compiler warnings.

system/xipfs/xipfs_main.c
xiaoxiang781216: why not check type to size_t

testing/fs/xipfs/xipfs_main.c
xiaoxiang781216: I suppose that the build system doesn't enable sign mismatch comparison warning?

xiaoxiang781216 (approved):

jerpelea (approved):

PR #3668, opened 2026-07-26
nshlib: add relative image path support in boot command

Junbo-Zheng:
## Summary

cmd_boot passed the image path straight to boardctl(), which resolves it in a context that does not inherit the NSH shell cwd, so relative paths failed, and only absolute paths worked. Use nsh_getfullpath() to resolve relative paths against the cwd before calling boardctl().

## Impact

- **Users**: `boot <relative-path>` now resolves against the NSH cwd;
  absolute-path behavior is unchanged.
- **Build**: None -- uses the existing `nsh_getfullpath`/`nsh_freefullpath`
  helpers, no new dependency.
- **Hardware**: None -- NSH command logic only, no board-specific code.
- **Documentation**: None.
- **Security & Compatibility**: None -- behavior extension; absolute
  paths behave exactly as before.

## Testing

Local build on Cortex-M55 (enables CONFIG_BOARDCTL_BOOT_IMAGE); the boot ELF builds and the `boot` NSH command succeeds.

And the change follows the established `nsh_getfullpath()`/`nsh_freefullpath()` convention already used by the other NSH file-path commands. Representative upstream examples (the same pattern is also used by `mv`, `ls`, `ln`, `mkdir`, and every other NSH file-path command):

- cp (src)  -- https://github.com/apache/nuttx-apps/blob/master/nshlib/nsh_fscmds.c#L1106
- cp (dest) -- https://github.com/apache/nuttx-apps/blob/master/nshlib/nsh_fscmds.c#L1113
- cat       -- https://github.com/apache/nuttx-apps/blob/master/nshlib/nsh_fscmds.c#L854
- rm        -- https://github.com/apache/nuttx-apps/blob/master/nshlib/nsh_fscmds.c#L2667

### DIFF
--- a/nshlib/nsh_syscmds.c
+++ b/nshlib/nsh_syscmds.c
@@ -405,6 +405,7 @@ int cmd_switchboot(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
 int cmd_boot(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
 {
   struct boardioc_boot_info_s info;
+  FAR char *fullpath = NULL;
 
   memset(&info, 0, sizeof(info));
 
@@ -421,7 +422,8 @@ int cmd_boot(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
         /* Go through */
 
       case 1:
-        info.path = argv[1];
+        fullpath = nsh_getfullpath(vtbl, argv[1]);
+        info.path = fullpath;
 
         /* Go through */
 
@@ -439,6 +441,7 @@ int cmd_boot(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
    */
 
   nsh_error(vtbl, g_fmtcmdfailed, argv[0], "boardctl", NSH_ERRNO);
+  nsh_freefullpath(fullpath);
   return ERROR;
 }
 #endif


xiaoxiang781216 (approved):

cederom (approved):
Thank you @Junbo-Zheng :-)

simbit18 (approved):

acassis (changes requested):
@Junbo-Zheng I think this feature needs Documentation

cederom:
I also wonder about impact on firmware size, have you compared that @Junbo-Zheng ? Maybe we want this as an option if the function is not part of the firmware by default? I understand that `cp, rm, cat` can use relative paths but these can be disabled by default. If no impact then no problem :-)

Junbo-Zheng:
> @Junbo-Zheng I think this feature needs Documentation

@acassis Done, please review https://github.com/apache/nuttx/pull/19553

Junbo-Zheng:
> I also wonder about impact on firmware size, have you compared that @Junbo-Zheng ? Maybe we want this as an option if the function is not part of the firmware by default? I understand that `cp, rm, cat` can use relative paths but these can be disabled by default. If no impact then no problem :-)

@cederom 

The `boot` nsh command is already optional — it is gated by `CONFIG_BOARDCTL_BOOT_IMAGE` (opt-in) together with `CONFIG_NSH_DISABLE_BOOT` (opt-out), see [nsh_syscmds.c#L404](https://github.com/apache/nuttx-apps/blob/master/nshlib/nsh_syscmds.c#L404). Platforms that don't need image boot simply leave CONFIG_BOARDCTL_BOOT_IMAGE unset, and `cmd_boot` (along with its nsh_getfullpath() call) is not compiled in at all.

Local build with `Arm GNU Toolchain 13.2.rel1`
```
/home/mi/xiaomi/trunk [26-07-27_13:53:56] arm-none-eabi-gcc --version
arm-none-eabi-gcc (Arm GNU Toolchain 13.2.rel1 (Build arm-13.7)) 13.2.1 20231009
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

- before
```
/home/mi/xiaomi/trunk/out [26-07-27_12:16:26] arm-none-eabi-size nuttx 
   text	   data	    bss	    dec	    hex	filename
 469784	  15352	49257896	49743032	2f704b8	nuttx
```
- after
    - CONFIG_BOARDCTL_BOOT_IMAGE=y
    - CONFIG_NSH_DISABLE_BOOT not set
```
/home/mi/xiaomi/trunk/out [26-07-27_12:18:43] arm-none-eabi-size nuttx 
   text	   data	    bss	    dec	    hex	filename
 469792	  15352	49257896	49743040	2f704c0	nuttx
```

cederom (approved):

acassis (approved):

PR #3673, opened 2026-07-27
testing, system, netutils: do not depend on fork() where it is absent or unneeded

casaroli:
## Summary

Four places in `apps` call `fork()` from code that is compiled unconditionally. That is fine only for as long as every architecture provides it, and NuttX is about to stop doing so: **apache/nuttx#19562** splits `fork()` into three primitives, after which `CONFIG_ARCH_HAVE_FORK` announces POSIX `fork()` specifically and is **off** until an architecture implements it. Those four then fail to build, and the NuttX PR's CI cannot go green, because NuttX PRs build against `apps` master.

This PR removes that blockage and nothing else. It introduces no new configuration symbols, no compatibility layer, and no test changes.

It splits along a real line, one commit each. Two of the four are tests that genuinely exercise `fork()` and so must stop being built where it is absent. The other two never wanted `fork()` at all — they reached for it only to put work in the background, and each has a NuttX-native way to do that, so they need no fork primitive and no new dependency.

### Tests that really do need `fork()`

**`system/libuv`** — `test-fork.c` and `test-pipe-close-stdout-read-stdin.c` are filtered out of the `test-*.c` glob where `CONFIG_ARCH_HAVE_FORK` is unset, in both the Makefile and the CMake build. Nothing is lost even then, because every test the two files define is *already* excluded from the task list on NuttX: `0001-libuv-port-for-nuttx.patch` widens the two `#ifndef _WIN32` guards in `test/test-list.h` to `#if !defined(_WIN32) && !defined(__NuttX__)`, and those guards cover all nine `fork_*` entries and `pipe_close_stdout_read_stdin` — the complete set both files define. They are compiled today but never run.

**`testing/ltp`** — the `open_posix_testsuite` is filtered through LTP's existing `BLACKWORDS` mechanism, which already drops tests for absent features and is already conditioned on configuration symbols. The pattern `[^v_]fork(` spares `vfork()` and `task_fork()`. Where `fork()` is not provided this drops **278 of 1943** test files; where it *is* provided — everywhere, today — nothing is dropped. That loss is the honest price of the split: those tests exercise `fork()` and cannot link without it. They return per architecture as real `fork()` lands.

### Callers that only wanted to run in the background

**`netutils/dropbear`** — the port already routes every fork-then-exec through `vfork()`, and has done since it was written: `sysoptions.h` selects `DROPBEAR_VFORK` when `HAVE_FORK` is undefined, and the port leaves it undefined, so `spawn_command()` in `dbutil.c` and both call sites in `scp.c` take the `vfork()` branch. The single exception is the `daemon()` fallback that `compat.c` compiles under `#ifndef HAVE_DAEMON`, which calls `fork()` directly and bypasses that switch. NuttX has `daemon()` in `libs/libc/unistd/lib_daemon.c` and declares it in `unistd.h`, so the fallback is redundant — this defines `HAVE_DAEMON` alongside the `HAVE_STRLCAT` and `HAVE_STRLCPY` entries that are in `nuttx_config.h` for exactly the same reason. The code was unreachable in any case: the port hands `svr_getopts()` an argv containing `-F`, so `svr_opts.forkbg` is always zero and dropbear never calls `daemon()` at all.

**`testing/drivers/nand_sim`** — forked so the parent could return to the shell while the child registered the MTD device and slept forever. Nothing from before the `fork()` is used after it, so the child is a self-contained entry point and `task_create()` expresses that directly; the emulator body moves into `nand_sim_daemon()` unchanged. `TESTING_NAND_SIM` therefore needs **no** fork dependency and its `Kconfig` is untouched, so `sim:nand` and `sim:mnemofs` keep working whatever `ARCH_HAVE_FORK` is set to — no defconfig edits are needed on either side, and no coverage is lost at any point in the merge order.

Two other `fork()` mentions need nothing: `games/NXDoom`'s is inside `#if 0 /* UNUSED */`, and `system/syslogd` already uses `posix_spawn()`.

## Impact

**Against today's master this is a no-op, with nothing left over.** `CONFIG_ARCH_HAVE_FORK` is set on every architecture, so neither the LTP filter nor the libuv filter drops anything; `nand_sim` and dropbear behave identically either way, since `task_create()` and `daemon()` do not depend on the symbol at all. Every configuration builds exactly the same set of files it does today.

**After apache/nuttx#19562**, `ARCH_HAVE_FORK` goes off until a per-architecture PR turns it back on. The two test suites stop being built rather than failing to link; `nand_sim` and dropbear are unaffected.

`interpreters/bas` is deliberately left alone. Its `SHELL` and `EDIT` statements want the same treatment, but `checkpatch.sh` checks the *whole* of any file a patch touches, and `bas_statement.c` produces 1681 pre-existing findings before this patch is applied at all — a one-newline commit against master fails CI identically. Migrating BAS has to follow a style cleanup of that file, and neither belongs here. The consequence is small: `CONFIG_EXAMPLES_BAS_SHELL` is `EXPERIMENTAL` and already `depends on ARCH_HAVE_FORK`, so it becomes unselectable rather than misbehaving.

The companion test work — splitting `ostest`'s fork test into `task_fork`, `vfork` and `fork` — is **apache/nuttx-apps#3685**, which depends on apache/nuttx#19562 and must merge after it.

## Testing

Host: macOS 15 (Darwin 25.5.0) on Apple Silicon. QEMU 11.0.3, xPack `riscv-none-elf-gcc` 14.2.0-3, Arm GNU `arm-none-eabi-gcc` 14.2.Rel1.

### Style

`../nuttx/tools/checkpatch.sh -c -u -m -g <base>..HEAD`, the exact command `.github/workflows/check.yml` runs — **✔️ All checks pass**, with `codespell`, `cvt2utf`, `cmake-format` and `nxstyle` all installed.

### Verification

This PR was restructured after review: the test work it originally carried has moved to apache/nuttx-apps#3685, leaving only the changes needed to unblock apache/nuttx#19562, and the branch has been rebased onto current master. The functional matrix is being re-run against the restructured branch and will be posted here; the previous revision's results are in this PR's history.

The three configurations CI reported failing on apache/nuttx#19562 — `sim:nand`, `sim:mnemofs` and `sim:dropbear` — were rebuilt against that PR's branch, where `CONFIG_ARCH_HAVE_FORK` is genuinely off, under Linux GCC rather than the development host's compiler.

At this PR's base the two reported diagnostics reproduce at exactly the locations CI gives:

```
dropbear/src/compat.c:162:17: implicit declaration of function 'fork'
nand_sim_main.c:145:9:        implicit declaration of function 'fork'
```

With this PR applied both are gone — zero occurrences across all three configurations — and `sim:dropbear` builds and links clean with no warnings at all. So the change is measured against the real failure rather than an approximation of it.

By inspection the change is a no-op against today's master: `CONFIG_ARCH_HAVE_FORK` is set on every architecture, so the LTP filter drops nothing, and the two libuv files removed from the glob are already absent from the task list on NuttX.


jerpelea (changes requested):
please replace 
Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
with 
Assisted-by: Claude Opus 5 (1M context) <noreply@anthropic.com>

jerpelea (dismissed):

testing/ltp/CMakeLists.txt
```diff
@@ -86,6 +86,12 @@ if(CONFIG_TESTING_LTP)
       list(APPEND BLACKWORDS "pthread_spin_init" "pthread_spin_destroy"
            "pthread_spin_trylock")
     endif()
+
+    # See testing/ltp/Makefile.
+
+    if(NOT CONFIG_ARCH_HAVE_FORK AND NOT CONFIG_FORK_IS_TASK_FORK)
```
xiaoxiang781216: why need CONFIG_FORK_IS_TASK_FORK

testing/ostest/ostest.h
```diff
@@ -282,12 +282,58 @@ void priority_inheritance(void);
...
+ * vfork_test() and fork_test() deliberately have no such mapping.  They
+ * check semantics a pre-split nuttx does not describe -- the parent
+ * suspension and the private copy -- and ARCH_HAVE_VFORK is the evidence
+ * that the split has landed.
+ *
+ * This block comes out with the split.
+ */
+
+#if defined(CONFIG_TASK_FORK) || \
+    (defined(CONFIG_ARCH_HAVE_FORK) && !defined(CONFIG_ARCH_HAVE_TASK_FORK))
```
xiaoxiang781216: why need check CONFIG_ARCH_HAVE_FORK/CONFIG_ARCH_HAVE_TASK_FORK

testing/ostest/ostest.h
```diff
@@ -282,12 +282,58 @@ void priority_inheritance(void);
...
+
+#ifdef CONFIG_ARCH_HAVE_VFORK
+#  define OSTEST_HAVE_VFORK 1
+#endif
+
+#if defined(CONFIG_ARCH_HAVE_FORK) && defined(CONFIG_ARCH_HAVE_VFORK)
+#  define OSTEST_HAVE_FORK 1
+#endif
+
+#if defined(OSTEST_HAVE_TASK_FORK) && !defined(CONFIG_TASK_FORK)
```
xiaoxiang781216: why need the above mapping?
it's better to just check CONFIG_ARCH_TASK_FORK/CONFIG_ARCH_FORK/CONFIG_ARCH_VFORK.

testing/ostest/Makefile
```diff
@@ -144,10 +144,21 @@ CSRCS += sigev_thread.c
...
 endif
 
-ifeq ($(CONFIG_ARCH_HAVE_FORK),y)
-ifeq ($(CONFIG_SCHED_WAITPID),y)
+# Each test is built where the primitive it tests exists.  See ostest.h for the
+# one transitional exception.
+
+ifeq ($(CONFIG_TASK_FORK),y)
+CSRCS += task_fork.c
+else ifeq ($(CONFIG_ARCH_HAVE_FORK)$(CONFIG_ARCH_HAVE_TASK_FORK),y)
```
xiaoxiang781216: remove

testing/ostest/Makefile
```diff
@@ -144,10 +144,21 @@ CSRCS += sigev_thread.c
...
+CSRCS += task_fork.c
+endif
+
+ifeq ($(CONFIG_ARCH_HAVE_VFORK),y)
 CSRCS += vfork.c
 endif
+
+ifeq ($(CONFIG_ARCH_HAVE_FORK)$(CONFIG_ARCH_HAVE_VFORK),yy)
```
xiaoxiang781216: why need check CONFIG_ARCH_HAVE_VFORK for fork test

testing/ostest/fork.c
```diff
@@ -0,0 +1,280 @@
...
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include "ostest.h"
+
+#ifdef OSTEST_HAVE_FORK
```
xiaoxiang781216: already done in Makefile, why need check again?

testing/ostest/vfork.c
```diff
@@ -31,59 +31,100 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
 #include "ostest.h"
 
-#if defined(CONFIG_ARCH_HAVE_FORK) && defined(CONFIG_SCHED_WAITPID)
+#ifdef OSTEST_HAVE_VFORK
```
xiaoxiang781216: remove too

acassis (dismissed):

testing/ltp/CMakeLists.txt
casaroli: This is a retro compatibility flag disabled by default. It is available to alias `fork()` to `task_fork()` in non MMU systems, in case someone depends on the exact current NuttX behavior without needing to update their program source code.

testing/ostest/Makefile
casaroli: replaced this with nested `ifeq`. This is required for this to merge before the nuttx PR. We can remove this with a follow up pr when the companion pr in nuttx gets merged

testing/ostest/Makefile
casaroli: if we do only `CONFIG_ARCH_HAVE_FORK` then all architectures will build that today, and `lm3s6965-ek:qemu-protected` is already almost over budget, it would not fit in `.text`. We can remove this with a follow up pr after the companion pr in nuttx gets merged.

In fact, we will need to decide how we deal with `lm3s6965-ek:qemu-protected` in the companion PR because I think `.text` will overflow there.

testing/ostest/fork.c
casaroli: done

testing/ostest/vfork.c
casaroli: done

testing/ltp/CMakeLists.txt
casaroli: it's config FORK_IS_TASK_FORK in arch/Kconfig in apache/nuttx#19562 — default n, depends on TASK_FORK && !ARCH_HAVE_FORK

testing/ostest/ostest.h
casaroli: This is required for this to merge before the nuttx PR. We can remove this with a follow up pr after the companion pr in nuttx gets merged

testing/ostest/ostest.h
casaroli: This is required for this to merge before the nuttx PR. We can remove this with a follow up pr after the companion pr in nuttx gets merged

casaroli:
Follow up pr https://github.com/apache/nuttx-apps/pull/3685

acassis (dismissed):

testing/ostest/Makefile
xiaoxiang781216: it's better to optimize the size of lm3s6965-ek:qemu-protected to pass the ci instead.

testing/ostest/ostest.h
xiaoxiang781216: please follow the instruction in https://github.com/apache/nuttx/pull/19075 to mark the dependence.

testing/ostest/ostest.h
xiaoxiang781216: https://github.com/apache/nuttx/pull/19075 could help you.

testing/ltp/CMakeLists.txt
xiaoxiang781216: but it's better to drop the mapping to reduce the confusion.

interpreters/python/Makefile
```diff
@@ -114,6 +114,11 @@ ifeq ($(CONFIG_ARCH_HAVE_FORK),y)
 else
 	@echo "export ac_cv_func_fork=\"no\"" >> $@
 endif
+ifneq ($(CONFIG_ARCH_HAVE_VFORK)$(CONFIG_ARCH_HAVE_FORK),)
```
xiaoxiang781216:
```suggestion
ifneq ($(CONFIG_ARCH_HAVE_VFORK),)
```

netutils/libwebsockets/lws_config_private.h
```diff
@@ -111,10 +114,14 @@
 /* #undef LWS_HAVE_VFORK_H */
 
 /* Define to 1 if `fork' works. */
+#ifdef CONFIG_ARCH_HAVE_FORK
 #define LWS_HAVE_WORKING_FORK
+#endif
 
 /* Define to 1 if `vfork' works. */
+#if defined(CONFIG_ARCH_HAVE_VFORK) || defined(CONFIG_ARCH_HAVE_FORK)
```
xiaoxiang781216:
```suggestion
#ifdef CONFIG_ARCH_HAVE_VFORK
```

testing/drivers/nand_sim/Kconfig
```diff
@@ -6,6 +6,7 @@
 config TESTING_NAND_SIM
 	boolean "NAND Flash Simulator"
 	depends on MTD_NAND_RAM && ENABLE_ALL_SIGNALS
+	depends on TASK_FORK || (ARCH_HAVE_FORK && !ARCH_HAVE_TASK_FORK)
```
xiaoxiang781216:
```suggestion
	depends on TASK_FORK || ARCH_HAVE_FORK
```

testing/fs/fdsantest/fdsantest_simple.c
```diff
@@ -96,6 +96,7 @@ static void test_case_overflow(void **state)
   assert_int_equal(open_count, close_count);
 }
 
+#if defined(CONFIG_ARCH_HAVE_VFORK) || defined(CONFIG_ARCH_HAVE_FORK)
```
xiaoxiang781216:
```suggestion
#ifdef CONFIG_ARCH_HAVE_VFORK
```

testing/fs/fdsantest/fdsantest_simple.c
```diff
@@ -129,7 +131,9 @@ int main(int argc, FAR char *argv[])
     cmocka_unit_test(test_case_unowned_tagged_close),
     cmocka_unit_test(test_case_owned_tagged_close),
     cmocka_unit_test(test_case_overflow),
+#if defined(CONFIG_ARCH_HAVE_VFORK) || defined(CONFIG_ARCH_HAVE_FORK)
```
xiaoxiang781216:
```suggestion
#ifdef CONFIG_ARCH_HAVE_VFORK
```

testing/ltp/Makefile
```diff
@@ -44,6 +44,13 @@ BLACKWORDS  += "pthread_spin_destroy"
 BLACKWORDS  += "pthread_spin_trylock"
 endif
 
+# Where NuttX does not declare fork(), a test that calls it cannot be built.
+# The pattern spares vfork() and task_fork(), which remain available.
+
+ifeq ($(CONFIG_ARCH_HAVE_FORK)$(CONFIG_FORK_IS_TASK_FORK),)
```
xiaoxiang781216: ditto

testing/ostest/CMakeLists.txt
```diff
@@ -143,10 +143,20 @@ if(CONFIG_TESTING_OSTEST)
...
 
-  if(CONFIG_ARCH_HAVE_FORK)
-    if(CONFIG_SCHED_WAITPID)
-      list(APPEND SRCS vfork.c)
-    endif()
+  # See testing/ostest/ostest.h for the one transitional exception.
+
+  if(CONFIG_TASK_FORK)
+    list(APPEND SRCS task_fork.c)
+  elseif(CONFIG_ARCH_HAVE_FORK AND NOT CONFIG_ARCH_HAVE_TASK_FORK)
```
xiaoxiang781216: remove the fallback path

testing/ostest/CMakeLists.txt
```diff
@@ -143,10 +143,20 @@ if(CONFIG_TESTING_OSTEST)
...
+    list(APPEND SRCS task_fork.c)
+  endif()
+
+  if(CONFIG_ARCH_HAVE_VFORK)
+    list(APPEND SRCS vfork.c)
+  endif()
+
+  if(CONFIG_ARCH_HAVE_FORK AND CONFIG_ARCH_HAVE_VFORK)
```
xiaoxiang781216:
```suggestion
  if(CONFIG_ARCH_HAVE_FORK)
```

testing/ostest/Makefile
```diff
@@ -144,10 +144,27 @@ CSRCS += sigev_thread.c
 endif
 endif
 
+# Each test is built where the primitive it tests exists.  See ostest.h for the
+# one transitional exception.
+
+ifeq ($(CONFIG_TASK_FORK),y)
+CSRCS += task_fork.c
+else ifeq ($(CONFIG_ARCH_HAVE_TASK_FORK),)
```
xiaoxiang781216: ditto

testing/ostest/Makefile
```diff
@@ -144,10 +144,27 @@ CSRCS += sigev_thread.c
...
+CSRCS += task_fork.c
+endif
+endif
+
+ifeq ($(CONFIG_ARCH_HAVE_VFORK),y)
 CSRCS += vfork.c
 endif
+
+ifeq ($(CONFIG_ARCH_HAVE_FORK),y)
+ifeq ($(CONFIG_ARCH_HAVE_VFORK),y)
```
xiaoxiang781216: ditto

testing/ostest/ostest.h
```diff
@@ -282,12 +282,58 @@ void priority_inheritance(void);
...
+ * check semantics a pre-split nuttx does not describe -- the parent
+ * suspension and the private copy -- and ARCH_HAVE_VFORK is the evidence
+ * that the split has landed.
+ *
+ * This block comes out with the split.
+ */
+
+#if defined(CONFIG_TASK_FORK) || \
+    (defined(CONFIG_ARCH_HAVE_FORK) && !defined(CONFIG_ARCH_HAVE_TASK_FORK))
+#  define OSTEST_HAVE_TASK_FORK 1
```
xiaoxiang781216: remove OSTEST_HAVE_XXX, check CONFIG_ARCH_HAVE_XXX directly

testing/ostest/ostest.h
```diff
@@ -282,12 +282,58 @@ void priority_inheritance(void);
...
+#ifdef CONFIG_ARCH_HAVE_VFORK
+#  define OSTEST_HAVE_VFORK 1
+#endif
+
+#if defined(CONFIG_ARCH_HAVE_FORK) && defined(CONFIG_ARCH_HAVE_VFORK)
+#  define OSTEST_HAVE_FORK 1
+#endif
+
+#if defined(OSTEST_HAVE_TASK_FORK) && !defined(CONFIG_TASK_FORK)
+#  define task_fork() fork()
```
xiaoxiang781216: remove the mapping

casaroli:
@xiaoxiang781216 I rewrote the PR so most of the complexity is moved to https://github.com/apache/nuttx-apps/pull/3685

acassis (dismissed):

system/libuv/Makefile
```diff
@@ -144,7 +144,15 @@ CSRCS += runner.c
 CSRCS += runner-unix.c
 CSRCS += echo-server.c
 
-CSRCS += $(wildcard libuv/test/test-*.c)
+# test-fork.c and test-pipe-close-stdout-read-stdin.c call fork().  Every
+# test they define is already excluded from the task list on NuttX by
+# 0001-libuv-port-for-nuttx.patch, so they are dead code here.
```
xiaoxiang781216: should we keep fork related code if the arch support the fork

casaroli: I think we should try to enable this after we have real `fork()` working. We would need to fix the patch.

xiaoxiang781216: if so, why not check CONFIG_ARCH_HAVE_FORK before removing them from build.

casaroli: Done — both the Makefile and `CMakeLists.txt` now drop the two files only when `CONFIG_ARCH_HAVE_FORK` is unset, so on every architecture today they are built exactly as before.

On the first question, though: keeping them where the arch supports `fork()` does not actually recover any coverage, because the exclusion is not keyed on fork availability. `0001-libuv-port-for-nuttx.patch` widens the two `#ifndef _WIN32` guards in `test/test-list.h` to `#if !defined(_WIN32) && !defined(__NuttX__)`, and those guards cover the complete set of tests the two files define — all nine `fork_*` entries (`fork_timer`, `fork_socketpair`, `fork_socketpair_started`, `fork_signal_to_child`, `fork_signal_to_child_closed`, `fork_fs_events_child`, `fork_fs_events_child_dir`, `fork_fs_events_file_parent_child`, `fork_threadpool_queue_work_simple`) plus `pipe_close_stdout_read_stdin`. On NuttX they compile and link but are never entered.

I have taken the suggestion anyway, because gating is the better shape for two reasons that do not depend on that argument: it keeps this PR a strict no-op on master rather than nearly one, and it matches what `testing/ltp` does a few lines away instead of asking the reader to accept a claim about a vendored patch.

Verified on `sim:citest` under Linux GCC, building the same tree twice against a NuttX with and without the symbol: with `CONFIG_ARCH_HAVE_FORK=y` 170 libuv test files are compiled including both of these, and with it unset 168 are compiled and these two are the only ones missing.

xiaoxiang781216 (approved):

acassis (approved):

PR #3693, opened 2026-08-03
examples/fdpicxip, testing/fs/xipfs: A DT_NEEDED library is one instance

casaroli:
## Summary

`examples/fdpicxip` and `testing/fs/xipfs` assert that each running instance of a module gets its own copy of a library named in `DT_NEEDED`: two instances adding their own seed each saw a total of `seed * 3`.

That was true of the loader that walked `DT_NEEDED` itself. The loader now hands the work to `dlopen()`, which returns the object already in the module registry rather than loading a second copy, so there is one library and one set of its globals, shared by every module that names it. The module's own data stays private per instance, because `exec()` loads the module afresh each time.

What an instance can still assert on its own is that every add it made landed in the library, so that is what it checks. The test additionally checks the consequences: the library is pinned once rather than once per instance, and its destructor runs once, at the last close, holding what both instances built up.

The embedded module blobs are regenerated from the sources beside them.

depends-on: [apache/nuttx/pull/19673]

## Testing

Run in full on QEMU `mps2-an500` and on a Pimoroni Pico Plus 2 (RP2350), against apache/nuttx#19673.

```
xipfs_test          131 passed, 0 failed
xipfs_test fdpic     34 passed, 0 failed
xipfs_test reject     7 passed, 0 failed
fdpicxip qsort | solib | cxx | jmprel
```


examples/fdpicxip/modules/cxxuser.cpp
```diff
@@ -87,7 +90,7 @@ static Instance g_self;
 #define USER_FAIL_OWN_CTOR  0x01
 #define USER_FAIL_LIB_CTOR  0x02
 #define USER_FAIL_ORDER     0x04
-#define USER_FAIL_PRIVATE   0x08
+#define USER_FAIL_SHARED    0x08
```
cederom, line 93: do we want to keep `#define USER_FAIL_PRIVATE 0x08` and add `#define USER_FAIL_SHARED 0x10` or this `USER_FAIL_PRIVATE` will never be used again and nowhere else?

examples/fdpicxip/modules/Makefile
```diff
@@ -46,10 +46,10 @@ CPU ?= cortex-m3
 
 FDPICDIR  = $(NUTTX_DIR)/tools/fdpic
 MODULE_MK = $(FDPICDIR)/nuttx-fdpic.mk
-EMBED     = $(FDPICDIR)/fdpic-embed
+EMBED     = $(FDPICDIR)/fdpic-embed.py
```
xiaoxiang781216, line 49: could we integrate fdpic special process into the post link after you corporate fdpic into elf binfmt?

xiaoxiang781216 (approved):

examples/fdpicxip/modules/Makefile
casaroli: for building fdpic modules, we can do with stock gcc, however for _linking_ them we need upstream binutils rebuilt with `arm-uclinuxfdpiceabi-ld`, as stock arm ld does not ship with it:

```
$ arm-none-eabi-ld -V
GNU ld (Arm GNU Toolchain 15.2.Rel1) 2.45.1
  Supported emulations:
   armelf
```

What do you suggest? can we include the "custom" binutils in the CI image?

examples/fdpicxip/modules/cxxuser.cpp
casaroli: Keeping `0x08` as `PRIVATE` and adding `SHARED` at `0x10` would imply the two can be reported independently, which would not be true. Also, `USER_FAIL_SHARED` is only used here and `xipfs_main.c`, so I just renamed the flag.

cederom: roger! thanks :-)

acassis (approved):

cederom (approved):

examples/fdpicxip/modules/Makefile
xiaoxiang781216: we can change LD in Tooldefine.def to the special version if CONFIG_FDPIC is enabled.

> What do you suggest? can we include the "custom" binutils in the CI image?

we can add the prebuilt tool to Dockfile directly:
https://github.com/apache/nuttx/blob/master/tools/ci/docker/linux/Dockerfile
if the tool doesn't have the official release binary, we can build from scratch like bloaty:
https://github.com/apache/nuttx/blob/master/tools/ci/docker/linux/Dockerfile

PR #3697, opened 2026-08-04
{bp-3554} examples/userfs: use designated initializers for struct dirent

jerpelea:
## Summary

The g_rootdir[] table initialized each entry's struct dirent member positionally as { DTYPE_FILE, "FileN" }, which assumes d_type is the first field.  This breaks when struct dirent gains the POSIX d_ino field as its first member (see apache/nuttx: "fs/dirent: add d_ino member to struct dirent"): the file-name string would be assigned to d_type, producing -Wint-conversion warnings and an "initializer element is not computable at load time" error.

Use designated initializers (.d_type / .d_name) so the table is robust against the field order of struct dirent.

Also fix a 'Initiallly' typo in a nearby comment flagged by codespell.

## Impact

RELEASE

## Testing

CI

### DIFF
--- a/examples/userfs/userfs_main.c
+++ b/examples/userfs/userfs_main.c
@@ -142,19 +142,28 @@ static char g_file3_data[UFSTEST_FILE3_MXSIZE] = "This is file 3";
 static struct ufstest_file_s g_rootdir[UFSTEST_NFILES] =
 {
     {
-      { DTYPE_FILE, "File1" },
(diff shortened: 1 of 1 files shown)


cederom:
ci failed.. still some cherry-pick needed @jerpelea ? :-)

jerpelea:
@cederom circular dependency. Please merge as is!

PR #3706, opened 2026-08-07
testing/libc/arch_libc: Add a throughput benchmark.

Fishwaldo:
## Summary

Rebased onto #3712 as requested, and reduced to the part #3712 does not already
cover.

The correctness suite is gone. #3712 tests all sixteen functions this PR used to
test, plus `stpcpy`, `strcat` and `strncpy` that it did not, so there is nothing
left for it to add.

What remains is the throughput measurement. #3712 times one call per function at
one size, 128 bytes, with both operands aligned. This sweeps four sizes and four
source/destination alignment pairs across seventeen functions, which is #3712's
sixteen plus `strlcpy`.

It sits behind `TESTING_ARCH_LIBC_BENCH`, default n. Nothing changes when it is
off.

## Why the alignment matrix

A machine implementation usually takes its wide path only when the pointers meet
some alignment condition, so a single aligned measurement reports the best case
and says nothing about the rest of the input space. On rv64:

```
  strcpy    32768 B  s+0/d+0     2938.0 MB/s
  strcpy    32768 B  s+1/d+1     2942.0 MB/s
  strcpy    32768 B  s+1/d+2      626.0 MB/s
```

Two pointers misaligned by the same amount run at the aligned rate. Misaligned by
different amounts they fall to a fifth of it. Neither number is visible from an
aligned measurement alone.

A function with no machine implementation reports the same rate at every
alignment, so the sweep also shows which functions a machine directory actually
covers.

In passing this caught a `strlcpy` regression in 931d5f50d4, where the misaligned
case runs 55x slower than the generic C it replaced. I will open a separate PR
for that.

## What it reports

Each measurement gives MB/s, which compares across machines, and cycles per byte
where `perf_gettime()` is reachable from an application, both from one timed
loop.

## Testing

EIC7700X (rv64, 1.4 GHz), `CONFIG_BUILD_KERNEL`, 248 measurements per run, in two
configurations: the generic C library, and the RISC-V assembly currently on
master. Both pass.

## Note for kernel builds

`perf_gettime()` is not in `syscall.csv`, so an application cannot reach it
unless the C library builds its own copy
(`CONFIG_ARCH_HAVE_PERF_EVENTS_USER_ACCESS`) or the build is flat. This
benchmark guards for that and reports MB/s alone where the counter is out of
reach.

### DIFF
--- a/testing/libc/arch_libc/arch_libc_bench.c
+++ b/testing/libc/arch_libc/arch_libc_bench.c
@@ -92,6 +92,8 @@
 #define OP_STPCPY    14
 #define OP_STRNCPY   15
 #define OP_STRCAT    16
+#define OP_MEMCCPY   17
+#define OP_STPNCPY   18
 
 /****************************************************************************
  * Private Types
@@ -134,6 +136,8 @@ static const struct bench_op_s g_ops[] =
   {"stpcpy",  OP_STPCPY,  true,  true},
   {"strncpy", OP_STRNCPY, true,  true},
   {"strcat",  OP_STRCAT,  true,  true},
+  {"memccpy", OP_MEMCCPY, true,  false},
+  {"stpncpy", OP_STPNCPY, true,  true},
 };
 
 static const size_t g_sizes[] =
@@ -323,6 +327,12 @@ static void bench_one(FAR const struct bench_op_s *o, size_t n,
                 break;
               case OP_STRNCPY:
                 strncpy(d, s, n);
+                break;
+              case OP_MEMCCPY:
+                g_sink += (uintptr_t)memccpy(d, s, '~', n);
+                break;
+              case OP_STPNCPY:
+                g_sink += (uintptr_t)stpncpy(d, s, n);
                 break;
               case OP_STRCAT:
 
--- a/testing/libc/arch_libc/arch_libc_bench.c
+++ b/testing/libc/arch_libc/arch_libc_bench.c
@@ -116,6 +116,7 @@ struct bench_op_s
 static FAR char *g_src;
 static FAR char *g_dst;
 static volatile unsigned long g_sink;
+static clockid_t g_clock = CLOCK_MONOTONIC;
 
 static const struct bench_op_s g_ops[] =
 {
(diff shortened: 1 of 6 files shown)


Fishwaldo:
The RISC-V series this suite was written for is now open as
apache/nuttx#19735. Every correctness claim in that PR is this suite's output,
on rv64 hardware and on rv32 under QEMU.

This PR remains independent of it: the suite is arch-neutral, passes against
the generic C library when nothing is overridden, and is useful to any machine
directory that overrides these functions.

acassis (dismissed):

cederom (dismissed):
Thank you @Fishwaldo :-)

testing/libc/arch_libc/arch_libc_test_main.c
```diff
@@ -232,6 +232,14 @@ int arch_libc_strcpy_speed(void)
 
 #endif
 
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+#ifdef CONFIG_TESTING_ARCH_LIBC_STRING
+int arch_libc_test_string(void);
```
xiaoxiang781216: @Fishwaldo I think https://github.com/apache/nuttx-apps/pull/3712/ cover more string function than this pr, could you check it.

xiaoxiang781216 (dismissed):

cederom:
Looks like #3712 is a joint merge of several test suites and should be merged instead? :-)

xiaoxiang781216:
@Fishwaldo since https://github.com/apache/nuttx-apps/pull/3712, could you try that change and close this pr? thanks.

hartmannathan:
What's the status of this PR? Did #3712 include all the new tests that are implemented here?

xiaoxiang781216:
> What's the status of this PR? Did #3712 include all the new tests that are implemented here?

@Fishwaldo could you check where #3712 contain all your test, if not, please apply your patch on top of #3712, thanks.

Fishwaldo:
Updated. Dropped my correctness tests in favor of #3712 and updated the benchmarks to cover everything in #3712. 

Note, on my box, perf_gettime isn't available for Kernel builds so also keep the original MB/s measurements.

Fishwaldo:
Sorry, found a few functions that were missing some correctness tests while working on the optimized fixes for riscv and the newlib variants. Added memccpy, stpncpy and strlcpy for completeness (including benchmarks)

testing/libc/arch_libc/arch_libc_bench.c
```diff
@@ -397,6 +445,14 @@ int arch_libc_bench(void)
       return 1;
     }
 
+  if (!bench_pick_clock())
```
xiaoxiang781216: move before line 438 and remove line 451-452

testing/libc/arch_libc/arch_libc_bench.c
```diff
@@ -0,0 +1,424 @@
...
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+#include <nuttx/clock.h>
+
+#ifdef CONFIG_TESTING_ARCH_LIBC_BENCH
```
xiaoxiang781216: remove the check

testing/libc/arch_libc/arch_libc_bench.c
Fishwaldo: Done

testing/libc/arch_libc/arch_libc_bench.c
Fishwaldo: Done

xiaoxiang781216 (approved):

acassis (approved):

PR #3713, opened 2026-08-10
nshlib/ifconfig: apply only requested settings and support "up|down"

JianyuWang0623:

## Summary

Fixes and extends `ifconfig` in nshlib. Three commits:

1. **only apply the settings that were asked for** — `cmd_ifconfig()`
   previously pushed *every* setting to the device on each call, whether or
   not it was on the command line, and stopped parsing at `mtu`. So
   `ifconfig eth0 hw <mac>` also silently cleared the address, netmask,
   gateway and resolver and kicked off DHCP; `ifconfig eth0 mtu 1500 dns ...`
   dropped the `dns`. Now each setting is written only when the user provides
   it, matching how Linux net-tools walks the argument vector (each keyword is
   an independent, idempotent operation).

2. **drop the redundant "argc > 2" check** — the function returns early for
   `argc <= 2`, so the check in the argument loop is dead code. Removed and the
   loop unindented. No functional change.

3. **support "ifconfig \<iface\> up|down"** — previously `ifconfig eth0 up`
   fell through to host-IP parsing, where `inet_addr("up")` returns
   `INADDR_NONE`, silently setting the address to 255.255.255.255. Recognize
   `up`/`down` as explicit keywords and apply them via
   `netlib_ifup()`/`netlib_ifdown()` after the rest of the config, so up/down
   is just another argument as in Linux `ifconfig`.

## Impact

- User: `ifconfig` no longer clobbers unrelated interface settings; new
  `ifconfig <iface> up|down` keyword supported (help text updated).
- Backward compatible: existing invocations that set a full config behave the
  same; only the unintended side effects are removed.
- No build, hardware, security or dependency changes. up/down works
  regardless of `CONFIG_NSH_DISABLE_IFUPDOWN`.

## Testing
```
core1> ifconfig eth0 172.16.1.201 up

core1> ifconfig eth0
eth0    Link encap:Ethernet HWaddr **:**:**:**:**:** at RUNNING mtu 1500
        inet addr:172.16.1.201 DRaddr:172.16.1.1 Mask:255.255.255.0

        RX: Received Fragment Errors   Bytes
            00000000 00000000 00000000 0
            IPv4     ARP      Dropped
            00000000 00000000 00000000
        TX: Queued   Sent     Errors   Timeouts Bytes
            00000000 00000000 00000000 00000000 0
        Total Errors: 00000000

core1>
```

### DIFF
--- a/nshlib/nsh_command.c
+++ b/nshlib/nsh_command.c
@@ -296,7 +296,7 @@ static const struct cmdmap_s g_cmdmap[] =
   CMD_MAP("ifconfig", cmd_ifconfig, 1, 12,
     "[interface [mtu <len>]|[address_family] [[add|del] <ip-address>|dhcp]]"
     "[dr|gw|gateway <dr-address>] [netmask <net-mask>|prefixlen <len>] "
-    "[dns <dns-address>] [hw <hw-mac>]"),
+    "[dns <dns-address>] [hw <hw-mac>] [up|down]"),
 #  endif
 #  if defined(CONFIG_NET_VLAN) && !defined(CONFIG_NSH_DISABLE_VCONFIG)
   CMD_MAP("vconfig", cmd_vconfig, 3, 5,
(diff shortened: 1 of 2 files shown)


cederom (approved):
Thank you @JianyuWang0623 good catch!! :-)

acassis (approved):

PR #3718, opened 2026-08-11
system/nxpkg: Validate catalog and database contents before use.

aviralgarg05:

Depends-On: apache/nuttx-apps/pull/3642

## Summary

A manifest read from a catalog, and an entry read from the installed
database, were used as they arrived. A name or a version containing a path
separator therefore reached the filesystem, and a database could name a
current version that it did not list among the installed ones.

This checks those fields are usable as a single path component, that the
current and previous versions appear in the recorded version list, and
applies a size limit to each field.

Version comparison also had two problems: it could overflow on a long
numeric prefix, and it treated versions sharing that prefix as equal, so 1a
and 1b compared the same. It now compares the numeric part without
overflowing and falls back to the remainder.

The manifest gains an optional icon, and the manifest actually recorded for
an installed version can now be read back, which matters after a rollback
has left the installed version behind the catalog.

This is the second of four parts of #3642, which was one commit covering
several unrelated changes.

## Impact

- New feature: NO, this is validation and correctness on existing paths.
- User adaptation: NO.
- Build: NO.
- Hardware: NO.
- Documentation: NO.
- Security: YES. A catalog or database that has been tampered with can no
  longer smuggle a path traversal through a name or version field.
- Compatibility: A manifest whose fields were already valid parses as
  before. One that was not is now rejected instead of being acted on.

## Testing

(`esp-14.2.0_20251107`).

  every changed file
- all `system/nxpkg` sources compiled for the target with this commit
  applied on top of #3642

This commit does not build on `master` alone because it uses declarations
added by #3642; the `Depends-On` line above lets CI apply that first.

## PR verification Self-Check

### DIFF
--- a/system/nxpkg/pkg_manifest.c
+++ b/system/nxpkg/pkg_manifest.c
@@ -78,6 +78,41 @@ static bool pkg_validate_hex(FAR const char *value)
  * Public Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: pkg_validate_path_component
+ *
+ * Description:
+ *   Check that a value is safe as one path component.
+ *
+ ****************************************************************************/
+
+bool pkg_validate_path_component(FAR const char *value)
+{
+  FAR const char *p;
+
+  if (pkg_validate_required(value) < 0)
+    {
+      return false;
+    }
+
+  /* Reject dot names and parent traversal. */
+
+  if (value[0] == '.')
+    {
+      return false;
+    }
+
+  for (p = value; *p != '\0'; p++)
+    {
+      if (*p == '/' || *p == '\\')
+        {
+          return false;
+        }
+    }
+
+  return true;
+}
+
 const char *pkg_manifest_type_str(enum pkg_payload_type_e type)
 {
   switch (type)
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (dismissed):

system/nxpkg/pkg_manifest.c
```diff
@@ -82,13 +82,7 @@ static bool pkg_validate_hex(FAR const char *value)
...
  *
  * Description:
- *   Reject any value that could escape the intended directory when spliced
- *   into a filesystem path (pkg_store.c's PKG_STORE_DIR "/%s/%s/..."
- *   formatters). This is required for "name" and "version" specifically,
- *   since both come straight from an untrusted, network-fetched
- *   index.json and are used unsanitized to build install paths - a
- *   version of "../../evil" would otherwise let a malicious index write
- *   or delete files outside the package store entirely.
+ *   Check that a value is safe as one path component.
```
xiaoxiang781216, line 85: squash into one patch

xiaoxiang781216 (approved):

acassis (approved):

PR #3719, opened 2026-08-11
system/nxpkg: Fetch the catalog and artifacts over the network.

aviralgarg05:

Depends-On: apache/nuttx-apps/pull/3718

## Summary

A package could only be installed from a catalog that was already on the
device, so getting one there meant copying it by hand.

This adds a synchronization path that reads an index from an `http://` or
`https://` source and stores it locally, along with the artifact download
that install needs. The catalog and the source it came from are written
under one lock, so the two cannot end up describing different syncs, and
the lock is renewed as data arrives so a slow download over a slow link is
not mistaken for an abandoned one and reclaimed while it is still running.

A repository is plain static files. Any host that can serve an `index.json`
and the artifacts it names works; no server-side logic is involved.

This is the third of four parts of #3642, which was one commit covering
several unrelated changes.

## Impact

- New feature: YES, packages can be fetched over the network.
- User adaptation: NO. Existing local catalogs keep working.
- Build: YES, one new source file in `system/nxpkg`.
- Hardware: NO, but a target needs networking configured to use it.
- Documentation: The companion documentation is apache/nuttx#18875.
- Security: Payloads are verified against the SHA-256 in the manifest, so a
  corrupted or substituted artifact is rejected before it is installed.
  Plain HTTP carries no confidentiality; the digest covers integrity.
- Compatibility: A source that is not a URL is still treated as a path.

## Testing

(`esp-14.2.0_20251107`).

  every changed file
- all `system/nxpkg` sources compiled for the target with this commit
  applied on top of #3718

This commit does not build on `master` alone because it uses declarations
and helpers added by #3642 and #3718; the `Depends-On` line above lets CI
apply those first.

## PR verification Self-Check

### DIFF
--- a/system/nxpkg/pkg_repo.c
+++ b/system/nxpkg/pkg_repo.c
@@ -1,0 +1,721 @@
+/****************************************************************************
+ * apps/system/nxpkg/pkg_repo.c
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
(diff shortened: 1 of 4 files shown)


xiaoxiang781216 (dismissed):

system/nxpkg/pkg_repo.c
```diff
@@ -49,16 +48,7 @@
...
- * SD-card write call; at the old 512-byte size, a sub-1MB file like
- * nxdoom's ~940KB ELF took ~1900 round trips and, in practice, close
- * to two minutes to install - easily read as "stuck" with only a bare
- * spinner for feedback.  4KB cuts that to ~230 round trips and lines
- * up with typical SD card erase-block granularity, which also reduces
- * write amplification.  Still small enough to be a safe stack-local
- * buffer against the 16KB (CLI) / 16KB (nxstore install worker) task
- * stacks that call into this.
- */
+/* Balance network throughput against the 16 KiB caller stacks. */
```
xiaoxiang781216, line 51: squash into one patch

xiaoxiang781216 (approved):

acassis (approved):

PR #3690, opened 2026-08-02
ci: Support pull request dependencies via Depends-On.

zhangning21:
## Summary

Add the `nuttx-apps` side of the pull-request dependency support already merged in [apache/nuttx#19075](https://github.com/apache/nuttx/pull/19075). This keeps the two repositories' Build workflows consistent and lets an apps PR test exact same- or cross-repository companion PR heads before either change is merged.

### Why this is needed

NuttX and nuttx-apps are built together. For a normal PR, `Fetch-Source` checks out the target branch from the changed repository and the corresponding branch from the other repository. A feature that must change both repositories can therefore fail in both PRs because neither target branch contains the companion change yet. This feature lets CI test the exact combined changes before either PR is merged. It also supports a PR depending on another PR in the same repository.

## Implementation

- Add the same standard-library `depends_on.py` parser and 40 regression tests used by NuttX.
- Gate heavy CI on dependency-changing PR-description edits.
- Fetch and cherry-pick ordered dependency PR commits once in `Fetch-Source`, before the existing matrix.
- Upload a structured report from the read-only Build workflow.
- Validate and render that untrusted report in a default-branch `workflow_run` with narrowly scoped `actions: read` and `pull-requests: write` permissions.

The trusted `depends-on-comment.yml` workflow becomes active after it is merged into the repository's default branch, because GitHub only triggers a newly added `workflow_run` workflow when that workflow file exists on the default branch.

The follow-up comment reports one of three outcomes:

- `ok`: each dependency PR and its fetched head SHA, abbreviated in the comment;
- `invalid`: a marker was found but no valid dependency could be parsed, so no dependency is applied and CI continues with normal source selection;
- `failed`: a valid dependency could not be applied, with a fixed diagnostic reason. This causes `Fetch-Source` to fail.

The accepted declaration forms, result comments, release behavior, and limitations are already documented in NuttX's central CI documentation by apache/nuttx#19075.

## Usage

A PR targeting `master` may declare one dependency per line:

```text
Depends-On: https://github.com/apache/nuttx/pull/1234
Depends-On: https://github.com/apache/nuttx-apps/pull/5678
```

or one inline list:

```text
Depends-On: [apache/nuttx/pull/1234 apache/nuttx-apps/pull/5678]
```

The marker is case-insensitive. Duplicate references are applied once in first-seen order. Invalid declarations do not apply a dependency and do not fail Build; a valid dependency that cannot be fetched or cherry-picked fails `Fetch-Source`.

## Impact and security

[...]

PR #3727, opened 2026-08-14
nshlib: apply privilege prompt markers only after real login

Abhishekmishra2808:
## Summary

- Emergency CI / NTFC is failing after https://github.com/apache/nuttx-apps/pull/3714 with `TimeoutError: device boot timeout` on `sim/citest` and `qemu-rv/citest`.
- `nsh_consolemain()` always starts the session with `NSH_LOGIN_LOCAL`, even when `CONFIG_NSH_CONSOLE_LOGIN` is disabled. PR 3714 treated any non-`NONE` session as a login and rewrote `nsh>` to `nsh#` whenever `CONFIG_SCHED_USER_IDENTITY` was enabled.
- NTFC detects boot by looking for `CONFIG_NSH_PROMPT_STRING` (`nsh>`), so those citest configs (USER_IDENTITY on, console login off) never come up.
- Apply `#`/`$` markers only after a successful console or telnet login (or `su`). No-login boot keeps `nsh>`.

## Impact

- `CONFIG_SCHED_USER_IDENTITY` boards without NSH login keep the configured prompt at boot (CI/NTFC compatible).
- Login-enabled sessions and `su` still get `nsh#` / `nsh$`.
- No change when `CONFIG_SCHED_USER_IDENTITY` is disabled.

## Testing

Build host: Linux WSL2, `sim:nsh` with `CONFIG_SCHED_USER_IDENTITY=y` and console login disabled (same prompt path as `sim/citest` / `qemu-rv/citest`).

No-login boot keeps `CONFIG_NSH_PROMPT_STRING` (`nsh>`), so NTFC boot detection can succeed. After `su`, root becomes `nsh#`:

```
NuttShell (NSH) NuttX-13.0.0
nsh> id
uid=0(root) gid=0(root)
nsh> su
nsh# id
uid=0(root) gid=0(root)
nsh# whoami
root
nsh# 
```

PR #3729, opened 2026-08-15
ci: do not trigger Build on PR description edits

raiden00pl:
## Summary
  
The Depends-On feature (commit 2aebae740) made the Build workflow
trigger on PR description edits. A gate job checks whether the edit
changed any Depends-On declaration: if yes, the build jobs run again
with the new dependencies; on any other edit the gate skips all build
jobs.
    
The gate has a side effect that breaks PR check results. Skipped jobs
still register check results on the PR, and the PR checks view shows
the newest check run of each name. So after any description edit the
PR shows "skipped" for every build check instead of the pass/fail
from the real run. Re-running that newest run only repeats the skip,
so the real results never come back. This can also hide a red X from
a failed build.
    
Fix by not triggering Build on description edits at all: remove the
"edited" event type and the gate job.
    
Depends-On keeps working: dependencies are read from the description
at the start of every run against master, as before. Fetch-Source now
re-reads the description through the API instead of using the copy
stored in the event payload, so every run uses the current Depends-On
state no matter how it was triggered.
    
After editing a Depends-On line, retrigger CI by any of:
    - pushing new or rebased commits to the PR branch
    - closing and reopening the PR
    - pressing "Re-run all jobs" on the existing Build run
    
A description edit alone no longer triggers anything, which is
exactly the behavior that corrupted the PR check results.

## Impact

try to fix https://github.com/apache/nuttx-apps/pull/3690#issuecomment-5301179168

## Testing

none, must be checked upstream

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,99 +14,24 @@ name: Build
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, edited]
   push:
     branches:
       - 'releases/*'
     tags:
 
+# pull-requests read: Fetch-Source re-reads the PR description so that a
+# manual re-run picks up Depends-On lines edited after the run was created.
 permissions:
   contents: read
+  pull-requests: read
 
 concurrency:
   group: build-${{ github.event.pull_request.number || github.ref }}
-  # Edited runs do not request cancellation of an active code build.
-  # GitHub may still replace an older pending run in this concurrency group.
-  cancel-in-progress: ${{ github.event.action != 'edited' }}
+  cancel-in-progress: true
 
 jobs:
-  # Gate heavy CI on dependency-changing edits.
-  Changes:
-    runs-on: ubuntu-latest
-    outputs:
-      should_build: ${{ steps.gate.outputs.should_build }}
-    steps:
-      # Do not let PR code control its own edit gate.
-      - name: Checkout base-branch CI scripts
-        if: ${{ github.event_name == 'pull_request' && github.event.action == 'edited' }}
-        uses: actions/checkout@v7
-        with:
-          ref: ${{ github.event.pull_request.base.sha }}
-          sparse-checkout: .github/scripts
-          sparse-checkout-cone-mode: false
-          fetch-depth: 1
-          path: base-ci
-        continue-on-error: true
-      - name: Checkout PR CI scripts (fallback)
-        if: ${{ github.event_name == 'pull_request' && github.event.action == 'edited' }}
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

Abhishekmishra2808:
Thanks. Please rebase this onto the current master (including https://github.com/apache/nuttx-apps/pull/3727) and re-run CI.

zhangning21:
Thanks for catching this and for the quick fix, @raiden00pl. Sorry for the breakage — the gate job was my addition in #3690 and I clearly missed this failure mode.

Your analysis is exactly right: skipped jobs still register check results, and since the PR checks view shows the newest run per check name, any description edit replaced the real pass/fail with "skipped". The part that worries me most is the one you called out — it can mask a red X from a failed build. That makes the gate strictly worse than not having it at all.

The fix looks right to me, and I don't think there is a smarter variant: GitHub has no workflow-level conditional, so once `on:` matches, the run is created and every job in it reports a check run, skipped ones included. Not listening for `edited` at all is the only thing that actually removes the problem.

Re-reading the description through the API in `Fetch-Source` also fixes a latent bug in my original version: a manual "Re-run all jobs" used the body from the event payload, so an edited Depends-On line was silently ignored on re-run.

I see you already opened apache/nuttx#19855 for the same fix on the OS side — thanks for doing both. Looks good to me. Merging both should get CI back to normal.

Two things worth writing down, since the behaviour changes for authors:

1. Editing the description no longer triggers anything, and nothing reports that. `depends-on-comment.yml` runs on Build completion, so a changed Depends-On line now produces no comment and no warning at all. Retriggering needs a push, a close/reopen, or "Re-run all jobs" — and "Re-run failed jobs" is not enough, because it does not rerun a `Fetch-Source` job that already succeeded. For contributors pushing from a fork, only push and close/reopen are available; re-running needs write access.
2. Changing the base branch no longer triggers a Build either. That matters here because a release/backport base ignores Depends-On, so a green result produced with dependencies applied can survive a base change to a release branch.

One follow-up that is missing: `Documentation/testing/nuttx-ci.rst` still describes the old behaviour ("Editing the pull request description triggers the CI dependency gate..."), and apache/nuttx#19855 does not update it. I am happy to send that documentation patch — either as a commit on your PR or as a follow-up, whichever you prefer.

jerpelea (approved):

cederom (approved):
Thank you @raiden00pl :-)

cederom:
Ugh, CI failed :-(

PR #3735, opened 2026-08-19
nshlib,testing: add chroot command and ostest

Abhishekmishra2808:
## Summary

Adds an NSH `chroot` helper that `chdir`s into the new root, calls `chroot(".")`, then `chdir("/")` so relative paths stay in the jail. The command form `chroot <newroot> <command>` closes non-stdio, non-`O_CLOEXEC` descriptors before `execvp` so inherited host fds cannot bypass the jail. ostest covers jail visibility, host-path isolation, child inheritance, and leftover pre-opened fds.

## Impact

The NSH command is gated by `CONFIG_FS_CHROOT` and can be disabled with `CONFIG_NSH_DISABLE_CHROOT`. ostest runs the new case only when `CONFIG_FS_CHROOT` and `CONFIG_SCHED_WAITPID` are set and the build is not kernel. Companion kernel change is apache/nuttx#19900.

## Testing

Host: WSL2 x86_64. Board: `sim` (`CONFIG_FS_CHROOT=y`, `CONFIG_TESTING_OSTEST=y`), with nuttx `feature/fs-chroot`.

```
$ ./tools/checkpatch.sh -f apps/nshlib/nsh_envcmds.c apps/testing/ostest/chroot.c
✔️ All checks pass.
```

```
login: root
User Logged-in!
nsh> ostest
...
user_main: chroot test
chroot_test: Starting test
chroot_test: /marker is visible inside the jail
chroot_test: host paths are not visible inside the jail
chroot_test: pre-opened host fd still usable after chroot
chroot_test: grandchild still sees the jail
chroot_test: PASSED
```

### DIFF
--- a/nshlib/nsh_envcmds.c
+++ b/nshlib/nsh_envcmds.c
@@ -30,6 +30,7 @@
 #include <unistd.h>
 #include <string.h>
 #include <ctype.h>
+#include <fcntl.h>
 #include <libgen.h>
 #include <errno.h>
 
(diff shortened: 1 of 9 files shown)


jerpelea (approved):

acassis (approved):

nshlib/nsh_envcmds.c
```diff
@@ -324,6 +324,64 @@ int cmd_cd(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
 }
 #endif
 
+/****************************************************************************
+ * Name: cmd_chroot
+ ****************************************************************************/
+
+#if defined(CONFIG_FS_CHROOT) && !defined(CONFIG_NSH_DISABLE_CHROOT)
+int cmd_chroot(FAR struct nsh_vtbl_s *vtbl, int argc, FAR char **argv)
```
xiaoxiang781216, line 332: @Abhishekmishra2808 it's better to implement chroot to a standalone command since nsh doesn't has the root permission normally.

PR #3742, opened 2026-08-20
examples/lvglterm: replace PIPES with PSEUDOTERM

fdcavalcanti:
## Summary
The lvglterm terminal now runs the shell on a pseudo-terminal instead of three plain pipes.
Reason for this change is that I'm trying to run Python on LVGL terminal and we need to use pseudo terminal for it to render the interactive mode.

- apps/examples/lvglterm/lvglterm.c — create_terminal() now calls openpty() and gives the shell the slave end as stdin/stdout/stderr; the single master fd replaces the three pipe pairs
- New `lvglterm_send_input()` in the core owns the write path. It only ever writes as much as the terminal reports free via FIONSPACE and queues the rest for the timer.
- Both input variants no longer echo locally (lvglterm_touch.c, lvglterm_kbd.c) — the terminal driver echoes the characters as the shell reads them, exactly as on a UART console, and echoing again would double everything.
- Kconfig gained select PSEUDOTERM, so no defconfig edits are needed.

### AI Disclaimer
Used Cursor to execute the changes and evaluate the impact.

## Impact

Should affect the following boards that have `lvgl_term` example:
- esp32p4-tab5
- esp32s3-m5-cardputer
- linum-stm32h753bi

## Testing
Works the same as before.
Tested on `esp32p4-tab5` and it works great. Need feedback for the other board.
WhatsApp Image 2026-08-20 at 15 35 29

@JorgeGzm could you do a test run with keyboard and touch? Which I think are the ones affected on linum board.

### DIFF
--- a/examples/lvglterm/Kconfig
+++ b/examples/lvglterm/Kconfig
@@ -7,8 +7,14 @@ menuconfig EXAMPLES_LVGLTERM
 	tristate "LVGL Terminal"
 	default n
 	depends on GRAPHICS_LVGL
+	select PSEUDOTERM
 	---help---
 		Enable LVGL Terminal
+
+		The shell runs on a pseudo-terminal so that it, and the programs
+		it starts, see a console on their standard streams:  isatty() is
+		what tells an interactive program to prompt, echo and flush its
+		output line by line.
 
 if EXAMPLES_LVGLTERM
 
--- a/examples/lvglterm/lvglterm.c
+++ b/examples/lvglterm/lvglterm.c
@@ -24,11 +24,10 @@
  * "NuttX RTOS for PinePhone: LVGL Terminal for NSH Shell"
  * https://lupyuen.github.io/articles/terminal
  *
- * Code shared by both input variants: it starts the NSH shell with its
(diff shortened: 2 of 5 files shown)


examples/lvglterm/Kconfig
```diff
@@ -7,9 +7,15 @@ menuconfig EXAMPLES_LVGLTERM
 	tristate "LVGL Terminal"
 	default n
 	depends on GRAPHICS_LVGL
+	select PSEUDOTERM
```
xiaoxiang781216, line 10:
```suggestion
	depends on PSEUDOTERM
```

acassis (approved):

fdcavalcanti:
@JorgeGzm are you able to test it?
Lets wait one more day if not we can proceed.

cederom (approved):
Thank you @fdcavalcanti :-)

I have restarted failed CI jobs there was local CI break which now should be fixed, but lets wait for the CI to complete :-)

cederom:
@fdcavalcanti would you be able to share some boards so I could help in testing on a real hardware? :-)

PR #3751, opened 2026-08-24
system/nxinit: support compound command and resetcause-based triggers

JianyuWang0623:
## Summary
- Add support for compound command execution in init.rc action bodies (`&&` / `||` short-circuit semantics), e.g.:
  ```
  echo "start" && hello && echo "done"
  ls /missing || echo "not found"
  ```
- Add built-in property `sys.boot.reason` so action triggers can fire based on the board's reset cause, e.g.:
  ```
  on property:sys.boot.reason=cpu_soft_reset(bootloader)
      echo "bootloader mode ..."
      start fastboot
  ```
- Add subreason support for `sys.boot.reason` to allow finer-grained matching, e.g.:
  ```
  on init && property:sys.boot.reason=watchdog,4
  on init && property:sys.boot.reason=bootloader|recovery|thermal
  ```
- Fix a missing `"assert"` entry in the `resetflag[]` designated-initializer array that could lead to a NULL pointer dereference when accessing `reset.flag`.

## Impact
- No impact on existing behavior when these features are unused.
- `action.c`/`action.h`: `init_action_reap_command()` signature gained a second parameter to propagate the command's return value for `&&`/`||` short-circuiting.

## Testing
Build Host: Linux x86_64 (Ubuntu, kernel 6.8.0), gcc (Ubuntu 13.4.0-6ubuntu1~22~ppa2) 13.4.0
Target: sim:nsh (apps/system/nxinit is architecture-independent; verified via the host simulator since no esp32p4 riscv32 cross toolchain is available on this machine)

Build (excerpt, `CONFIG_SYSTEM_NXINIT=y`):
```
Register: init
CC:  action.c
CC:  init.c
CC:  parser.c
CC:  builtin.c
CC:  import.c
CC:  service.c
CC:  property_simple.c
IN: /.../apps/libapps.a -> staging/libapps.a
LD:  nuttx
Pac SIM with dynamic libs..
SIM elf with dynamic libs archive in nuttx.tgz
```

Runtime (`./nuttx`, actual console output):
```
NuttShell (NSH) NuttX-13.0.1-RC0
nsh> ps
    0     0     0   0 FIFO     Kthread   - Ready              0000000000000000 0069584 Idle_Task
    1     0     0 224 FIFO     Kthread   - Waiting  Semaphore 0000000000000000 0067448 sim_loop_wq 0x72f7d6c003f0 0x72f7d6c00470
    2     0     0 224 FIFO     Kthread   - Waiting  Semaphore 0000000000000000 0067472 hpwork 0x40167dc0 0x40167e40
    4     4     0 100 FIFO     Task      - Running            0000000000000000 0067504 nsh_main
nsh> exit
```

Style check (`tools/checkpatch.sh -g apache/master..nxinit-action-trigger-enhancements`, actual output):
```
/tmp/.../apps/system/nxinit/action.c:117:4: error: Bad left brace alignment
```
The single reported issue (`action.c:117`) is pre-existing on `apache/master` — reproduced by running `nxstyle` directly against the unmodified `apache/master` copy of `action.c`, which reports the identical error at the same line. It is not introduced by this change.

Not verified: esp32p4/esp32s3 board-level flashing and serial output — this change is architecture-independent apps-layer code, and this machine has no esp32p4 riscv32 cross toolchain installed.

### DIFF
--- a/system/nxinit/action.c
+++ b/system/nxinit/action.c
@@ -223,6 +223,83 @@ static int event_callback(FAR struct action_manager_s *am,
   return event->pending;
 }
 
+static FAR char *parse_operator(FAR char *pos,
+                                FAR enum action_cmd_op_e *op)
+{
+  bool in_quotes = false;
+
+  for (; *pos != '\0'; pos++)
+    {
+      if (*pos == '"')
+        {
+          in_quotes = !in_quotes;
+          continue;
+        }
+
+      if (in_quotes)
+        {
+          continue;
+        }
+
+      if (*pos == '&' && *(pos + 1) == '&')
+        {
+          *op = CMD_OP_AND;
+          *pos = '\0';
+          return pos + 2;
+        }
+
+      if (*pos == '|' && *(pos + 1) == '|')
+        {
+          *op = CMD_OP_OR;
+          *pos = '\0';
+          return pos + 2;
+        }
+    }
+
+  *op = CMD_OP_NONE;
+  return pos;
+}
+
+static int parse_command(FAR char *pos,
+                         FAR struct list_node *cmds)
+{
+  FAR enum action_cmd_op_e op = CMD_OP_NONE;
+  FAR enum action_cmd_op_e next_op;
+  FAR struct action_cmd_s *cmd;
+  FAR char *next_pos;
+
+  while (*pos)
+    {
+      next_pos = parse_operator(pos, &next_op);
+
+      cmd = calloc(1, sizeof(*cmd));
+      if (cmd == NULL)
+        {
+          return -errno;
+        }
+
+      cmd->argc = init_parse_arguments(pos, true,
+                                       nitems(cmd->argv) - 1,
+                                       cmd->argv);
+      if (cmd->argc < 1)
+        {
+          free(cmd);
+          init_err("invalid argument");
+          return -EINVAL;
+        }
+
+      cmd->op = op;
+      list_add_tail(cmds, &cmd->node);
+      init_debug("add command %p(%s) to list %p", cmd, cmd->argv[0], cmds);
+      init_dump_args(cmd->argc, cmd->argv);
+
+      op = next_op;
+      pos = next_pos;
+    }
+
+  return 0;
+}
+
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
@@ -315,6 +392,15 @@ int init_action_run_command(FAR struct action_manager_s *am)
                                         struct action_cmd_s, node);
     }
 
+  if ((am->running->op == CMD_OP_AND && ready->prev_ret != 0) ||
+      (am->running->op == CMD_OP_OR && ready->prev_ret == 0))
+    {
+      init_debug("Skipping command due to op (prev_ret %d)",
+                 ready->prev_ret);
+      init_action_reap_command(am, ready->prev_ret);
+      return 0;
+    }
+
 #if defined(CONFIG_SYSTEM_NXINIT_ACTION_WARN_SLOW) && \
     CONFIG_SYSTEM_NXINIT_ACTION_WARN_SLOW > 0
   clock_gettime(CLOCK_MONOTONIC, &am->time_run);
@@ -326,13 +412,13 @@ int init_action_run_command(FAR struct action_manager_s *am)
     }
   else
     {
-      init_action_reap_command(am);
+      init_action_reap_command(am, ret);
     }
 
   return 0;
 }
 
-void init_action_reap_command(FAR struct action_manager_s *am)
+void init_action_reap_command(FAR struct action_manager_s *am, int ret)
 {
   FAR struct action_s *ready = list_peek_head_type(&am->ready_actions,
                                                    struct action_s,
@@ -361,6 +447,7 @@ void init_action_reap_command(FAR struct action_manager_s *am)
 #endif
 
   am->pid_running = -1;
+  ready->prev_ret = ret;
   if (list_is_tail(&ready->cmds, &am->running->node))
     {
       am->running = NULL;
@@ -377,7 +464,6 @@ int init_action_parse(FAR const struct parser_s *parser,
 {
   FAR struct action_manager_s *am = parser->priv;
   FAR char *argv[2 * CONFIG_SYSTEM_NXINIT_ACTION_EVENTS_MAX];
-  FAR struct action_cmd_s *cmd;
   FAR struct action_s *a;
   int ret;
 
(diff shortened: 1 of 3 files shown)


JianyuWang0623:
## CI failure investigation

The `check` job failure (`action.c:117:4: error: Bad left brace alignment`) is **not caused by this PR's commits**. It is a false positive from a regression in `apache/nuttx`'s `tools/nxstyle` (commit `bdeb262b8d3`, "tools/nxstyle: indent code against its enclosing brace, not modulo 4", merged 2026-07-27).

Evidence:

1. The flagged line (`system/nxinit/action.c:117`) is outside the diff of all 4 commits in this PR — the hunks in this PR only touch lines 223+, 315+, 326+, 361+, 377+, 435+.
2. Rebuilding `nxstyle` from current `apache/master` and running it against the **pre-PR baseline** (commit `0923948d80`, before any commit in this PR) reproduces the identical error on the identical line — the code this PR is based on already fails under the new tool.
3. The new rule's exception for iterator/critical-section macros (`prevlastcode == ')' && indent == prevcodeindent + 2`) only correctly handles single-line macro invocations. For multi-line `list_for_every_entry(...)` calls where the closing `)` is on a continuation line, it computes the wrong expected indent and flags a false "Bad left brace alignment". This pattern is used throughout the codebase, e.g. `apache/nuttx`'s own `drivers/vhost/vhost.c` (lines 138-140, 237-239, 278-280, 346-348, 388-390, 408-410) hits the same false positive with the same nxstyle build.

Since `tools/checkpatch.sh -g` (as invoked by CI, without `-r`) lints the whole file rather than just the diff range, this pre-existing false positive surfaces on any PR touching `action.c`, unrelated to what the PR actually changes.

No code change is needed in this PR. Marking ready for review; CI should pass once the upstream `nxstyle` regression is fixed (or if maintainers choose to bypass/merge despite the known tool issue).

xiaoxiang781216 (dismissed):

fdcavalcanti:
Very nice! Please fix CI error

cederom (changes requested):
Very cool thank you @JianyuWang0623 the change is good, but also please fix caught syntax issues as we are here and we have them fixed for future :-)

JianyuWang0623:
@cederom @fdcavalcanti 
A fix is already available: https://github.com/apache/nuttx/pull/19555#issuecomment-5397889289

cederom (dismissed):
Thank you @JianyuWang0623 :-)

We may want to wait for fixes from https://github.com/apache/nuttx/pull/19555 or merge if other CI checks pass when in hurry in that case :-)

JianyuWang0623:
> Thank you @JianyuWang0623 :-)
> 
> We may want to wait for fixes from https://github.com/apache/nuttx/pull/19555 or merge if other CI checks pass when in hurry in that case :-)

@cederom Would be nice if we could ignore this false‑positive. Some in‑flight PRs need this PR(#3751) merged first.

xiaoxiang781216:
@JianyuWang0623 fix typo error:
```
/home/runner/work/nuttx-apps/nuttx-apps/apps/system/nxinit/init.c:119: unkown ==> unknown
/home/runner/work/nuttx-apps/nuttx-apps/apps/system/nxinit/init.c:130: unkown ==> unknown
```

xiaoxiang781216 (approved):

jerpelea (approved):

JianyuWang0623:
Root cause of the risc-v-03 CI failure identified and fixed upstream, unrelated to this PR's changes.

**What happened:** `nuttx#19952` (boards: update LVGLTERM requirements) and `nuttx-apps#3742` (examples/lvglterm: replace PIPES with PSEUDOTERM) were merged 4 seconds apart:
- `#19952` explicitly added `CONFIG_PSEUDOTERM=y` to three board defconfigs (`esp32p4-tab5/lvgl_term`, `esp32s3-m5-cardputer/lvglterm`, `linum-stm32h753bi/lvglterm{,_kbda}`) so lvglterm would build, since at that point the app still relied on plain pipes.
- `#3742` (merged seconds later) switched lvglterm to run the shell on a pseudo-terminal instead of pipes, and made `EXAMPLES_LVGLTERM` do `select PSEUDOTERM`, making the explicit line redundant. It also left `CONFIG_PIPES=y` as dead weight in those same defconfigs from the old pipe-based implementation.

Neither PR's author was aware of the other at merge time, so the redundant lines were never cleaned up. `testbuild.sh` runs `refresh.sh` (savedefconfig normalization) and diffs it against the committed defconfig — since both lines are now implied rather than needed, the diff comes back non-empty and the tree-cleanliness check fails. This PR's CI run happened to land on a nuttx SHA right after `#19952` merged, so it caught the window.

**Fix:** opened https://github.com/apache/nuttx/pull/19963 to drop the two now-redundant options from all four affected defconfigs. Verified `refresh.sh --silent` now reports a clean diff and all four configs (esp32p4-tab5, esp32s3-m5-cardputer, linum-stm32h753bi x2) still build successfully across their respective toolchains, plus `checkpatch.sh` passes on all four files.

This is unrelated to `system/nxinit` and doesn't block this PR's merge.

linguini1:
So do applications need to adapt and use the BOARDIOC command to set reset cause for the reason field to work?

What happens if no reset cause is given, is the default behaviour to not trigger?

This new functionality should be documented, please!

linguini1 (dismissed):
Please provide documentation about the new features.

JianyuWang0623:
@linguini1 @cederom Documentation for the new features is up: apache/nuttx#19971 (draft, `Documentation/applications/system/nxinit/index.rst`).

It adds:
- **Built-in Properties** section: explains `sys.boot.reason` — it's set automatically by NXInit at boot from `BOARDIOC_RESET_CAUSE` (no app action needed), documents both value forms (`<cause>,<subreason>` for hardware resets, or a single reason string for software resets like `bootloader`/`recovery`/etc.), and covers the two "nothing happens" cases: `CONFIG_BOARDCTL_RESET_CAUSE` disabled (property never set, so matching triggers just never fire) vs. the `boardctl()` call itself failing (NXInit aborts startup).
- **Compound Commands** section: documents the `&&`/`||` short-circuit chaining on a single action line, with the quoting caveat.

@linguini1 to directly answer your questions from the review:
- Apps don't need to call `BOARDIOC_RESET_CAUSE` themselves — NXInit queries it once at startup and sets `sys.boot.reason` before the `boot` event fires.
- If `CONFIG_BOARDCTL_RESET_CAUSE` isn't enabled, the property is simply never set, so any trigger referencing it never matches (there's no default/unset value it falls back to).

acassis (approved):

PR #3753, opened 2026-08-25
fs: rename PSEUDOFS_SOFTLINKS to FS_LINKS

Kaben123:
## Summary

Update all nuttx-apps references to the kernel filesystem-link config
option, which is renamed from PSEUDOFS_SOFTLINKS to FS_LINKS in the
companion NuttX PR. The link support is no longer limited to the pseudo
filesystem and now covers both soft (symbolic) and hard links across the
VFS, so the more accurate name FS_LINKS is used.

Affected areas: nshlib (ln/readlink/ls), interpreters (luv, wamr), adb,
tlpi example, and the testsuites — 18 references across 9 files.

## Impact

- Build-time only. No runtime/ABI/behavior change.
- Must be merged together with the NuttX kernel-side rename; on its own
  each repository's build would break because the config symbol only
  exists under one name at a time.

## Testing

- tools/checkpatch.sh passes on all changed lines.

Note: The kernel-side dependency (apache/nuttx#19949) has already been merged into master.

PR #3755, opened 2026-08-25
system/nxinit: add cmocka unit tests for parser/action/service

JianyuWang0623:
## Summary
- Fix a real bug in `init_parse_config_lines()`: a dead early `continue` for a truly-empty line (`buf == "\0"`) skipped the `memmove()` bookkeeping its sibling whitespace-only-line branch performs, corrupting the remaining-length tracking and **silently dropping every subsequent line in that refill chunk** (not just one bookkeeping entry). Since "blank line between sections" is one of the most common patterns in an `init.rc` file, this is a real, reachable bug, not just dead-code cleanup. The whitespace-skip loop right below already handles the empty-string case correctly, so the buggy early exit is simply redundant and removed.
- Fix a second real bug found during review, in `init_parse_config_buffer()` (the buffer-based counterpart of `init_parse_config_file()`, used to parse the builtin "preset" rc content): it computed the per-refill copy length as `MIN(len - off, sizeof(tmp))` without subtracting the `n` leftover bytes already held at the front of `tmp` from a previous refill, so `memcpy(&tmp[n], ..., r)` could write past `tmp[]`. Reproduced locally with an ASan host harness at `RC_LINE_MAX=32/48`; fixed to `MIN(len - off, sizeof(tmp) - n)` and reverified clean at 32/48/64/128 (the current default). `SYSTEM_NXINIT_RC_LINE_MAX` also had no lower bound, so it now has `range 64 4096` to stop this from being reachable by lowering that Kconfig value.
- Add a `test/` subdirectory (mirroring `apps/system/uorb/test/`) with cmocka-based unit tests covering the logic most prone to regression in NxInit:
  - `init_parse_arguments()`: plain/quoted arguments, `--` separator vs. `--option` long options (regression coverage for a previously fixed bug), argv-capacity truncation (now also asserting the exact folded contents of the last slot, not just that it exists).
  - `init_parse_config_file()`/`init_parse_config_lines()`/`init_parse_config_buffer()`: section routing, blank/whitespace-only line skipping, unknown-section rejection, over-length line rejection, and a line straddling two read-buffer refills — exercised through both the file-based and buffer-based entry points (the latter added specifically to catch the buffer-overflow bug above; `init_parse_config_buffer()` is exposed in `parser.h` behind `CONFIG_SYSTEM_NXINIT_TEST` for this purpose only).
  - Action event matching (`init_action_parse()`/`init_action_trigger_event()`): exact match, invert (`!=`), fnmatch wildcards, AND semantics across multiple events per action. `CONFIG_SYSTEM_NXINIT_ACTION_EVENTS_MAX` now defaults to 2 under `CONFIG_SYSTEM_NXINIT_TEST` specifically so this case actually runs instead of being permanently skipped (production default is unchanged at 1). All action test cases now also release the events/commands they allocate.
  - Service conflict detection (`init_service_parse()`/`init_service_check()`): duplicate service name rejection, override replacing an earlier duplicate, `SERVICE_ARGS_MAX` boundary (built dynamically from the configured `CONFIG_SYSTEM_NXINIT_SERVICE_ARGS_MAX` value rather than hardcoding the default of 8, so the test still exercises the real boundary under a different config).

  Test sources compile `action.c`/`parser.c`/`service.c` a second time into a separate `nxinit_unit_test` program, gated behind new `CONFIG_SYSTEM_NXINIT_TEST` (depends on `TESTING_CMOCKA`); the default `init` program is unaffected. `CONFIG_SYSTEM_NXINIT_TEST_STACKSIZE` now defaults to 8192 (was `DEFAULT_TASK_STACKSIZE` = 2048): several test cases build multi-hundred-byte stack buffers on top of cmocka's own overhead, and on real hardware the smaller default caused the test task to silently overflow its stack and stop producing output (no crash dump, no watchdog reset) partway through the suite.

  The two build systems handle the "compile a second time" differently: **CMake** actually rebuilds `action.c`/`parser.c`/`service.c` into a second, separate `nxinit_unit_test` target (verified below); **Make** instead appends the test sources into the same shared `CSRCS` list used by `init`, so `parser`/`action`/`service` are compiled once and linked into both `init` and `nxinit_unit_test` from the same objects. Both are correct, just not equivalent in what gets compiled where.

Opened as **draft** for early review/CI feedback.

## Impact
- Users: NO (test-only addition; the two parser fixes only remove dead/buggy code paths not reachable at default configuration, no behavior change for well-formed configs at default Kconfig values).
- Build: adds one new optional Kconfig-gated test target; default build unaffected.
- Hardware: NO.
- Documentation: NO.
- Security: Fixes a stack-buffer-overflow in `init_parse_config_buffer()`, but it is unreachable at the default `SYSTEM_NXINIT_RC_LINE_MAX=128` (the builtin preset content is 95 bytes); only reachable if that Kconfig value is lowered below ~90, which is now bounded to a 64 minimum by this PR.
- Compatibility: NO.

## Testing
Build Host: Ubuntu 22.04 LTS x86_64, gcc 13.4.0 (Ubuntu 13.4.0-6ubuntu1~22~ppa2)
Target: sim:citest / sim:nsh (nuttx `sim` config, no cross toolchain needed; NxInit is pure apps-layer logic) plus a real esp32p4-pico-wifi-wareshare hardware target for the latest commit (see below).

`nxstyle`/`checkpatch.sh` pass on all changed files.

**Note on CI coverage**: no upstream `defconfig` currently enables `CONFIG_SYSTEM_NXINIT_TEST` (checked via a full search of `boards/`), so this test target is not yet exercised by the existing CI test matrix. It has been verified locally (sim `Make`/`CMake` builds below) and on real esp32p4 hardware. Happy to wire it into an existing cmocka-enabled defconfig (e.g. `sim:citest`) if that's preferred — it would need `CONFIG_EXPERIMENTAL`, `CONFIG_LIBC_EXECFUNCS`, `CONFIG_SCHED_CHILD_STATUS`, `CONFIG_SYSTEM_NXINIT`, and `CONFIG_SYSTEM_NXINIT_TEST` added, none of which `sim:citest` currently has.

### Make build (sim:citest, default Kconfig, `ACTION_EVENTS_MAX=1`)

```
nsh> nxinit_unit_test
[==========] nxinit_tests: Running 17 test(s).
[ RUN      ] test_nxinit_parser_arguments_spaces
[       OK ] test_nxinit_parser_arguments_spaces
[ RUN      ] test_nxinit_parser_arguments_quoted
[       OK ] test_nxinit_parser_arguments_quoted
[ RUN      ] test_nxinit_parser_arguments_dashdash_separator
[       OK ] test_nxinit_parser_arguments_dashdash_separator
[ RUN      ] test_nxinit_parser_arguments_long_option
[       OK ] test_nxinit_parser_arguments_long_option
[ RUN      ] test_nxinit_parser_arguments_truncate
[       OK ] test_nxinit_parser_arguments_truncate
[ RUN      ] test_nxinit_parser_config_sections
[       OK ] test_nxinit_parser_config_sections
[ RUN      ] test_nxinit_parser_config_skip_blank_lines
[       OK ] test_nxinit_parser_config_skip_blank_lines
[ RUN      ] test_nxinit_parser_config_unknown_section
[       OK ] test_nxinit_parser_config_unknown_section
[ RUN      ] test_nxinit_parser_config_line_too_long
[       OK ] test_nxinit_parser_config_line_too_long
[ RUN      ] test_nxinit_parser_config_line_crosses_boundary
[       OK ] test_nxinit_parser_config_line_crosses_boundary
[ RUN      ] test_nxinit_action_event_match_exact
[       OK ] test_nxinit_action_event_match_exact
[ RUN      ] test_nxinit_action_event_match_invert
[       OK ] test_nxinit_action_event_match_invert
[ RUN      ] test_nxinit_action_event_match_fnmatch
[       OK ] test_nxinit_action_event_match_fnmatch
[ RUN      ] test_nxinit_action_event_and_semantics
[  SKIPPED ] test_nxinit_action_event_and_semantics
[ RUN      ] test_nxinit_service_duplicate_conflict
[       OK ] test_nxinit_service_duplicate_conflict
[ RUN      ] test_nxinit_service_override_replaces_duplicate
[       OK ] test_nxinit_service_override_replaces_duplicate
[ RUN      ] test_nxinit_service_args_max_boundary
[       OK ] test_nxinit_service_args_max_boundary
[==========] nxinit_tests: 17 test(s) run.
[  PASSED  ] 16 test(s).
[  SKIPPED ] nxinit_tests: 1 test(s), listed below:
[  SKIPPED ] test_nxinit_action_event_and_semantics
```

With `CONFIG_SYSTEM_NXINIT_ACTION_EVENTS_MAX=4`, `test_nxinit_action_event_and_semantics` runs and passes (17/17 PASSED), confirming the AND-semantics test is not dead code.

With `CONFIG_SYSTEM_NXINIT_SERVICE_ARGS_MAX=10` (instead of the default 8), `test_nxinit_service_args_max_boundary` still passes (16/17, same as default), confirming the args-max test now tracks the configured value instead of a hardcoded 8.

### CMake build (sim:nsh, default Kconfig)

```
$ cmake -B build -DBOARD_CONFIG=sim:nsh -GNinja
...
-- Build files have been written to: .../build
$ kconfig-tweak --file build/.config --enable CONFIG_EXPERIMENTAL \
    --enable CONFIG_SCHED_CHILD_STATUS --enable CONFIG_SYSTEM_NXINIT \
    --enable CONFIG_SYSTEM_NXINIT_TEST --enable CONFIG_TESTING_CMOCKA
$ cmake --build build -j$(nproc)
...
[1246/1248] Linking C executable nuttx
[1247/1248] Generating System.map
[1248/1248] Pac SIM with dynamic libs in nuttx.tgz
$ echo -e "nxinit_unit_test\nexit\n" | ./build/nuttx
NuttShell (NSH)
nsh> nxinit_unit_test
[==========] nxinit_tests: Running 17 test(s).
...
[  PASSED  ] 16 test(s).
[  SKIPPED ] nxinit_tests: 1 test(s), listed below:
[  SKIPPED ] test_nxinit_action_event_and_semantics
```

This CMake build previously failed to link (`undefined reference to 'nxinit_unit_test_main'`) because `nuttx_add_application()` only renames the *first* `SRCS` entry's `main()`, and `test/test_nxinit.c` was not first in `TEST_SRCS`; fixed by reordering `TEST_SRCS` in `CMakeLists.txt`.

### Real hardware build (esp32p4-pico-wifi-wareshare:nsh, latest commit)

Built via `make CROSSDEV=riscv-none-elf-` with `CONFIG_SYSTEM_NXINIT_TEST=y` (which now defaults `ACTION_EVENTS_MAX=2` and `TEST_STACKSIZE=8192`), flashed and ran `nxinit_unit_test` over UART on real esp32p4-pico-wifi-wareshare hardware:

```
nsh> nxinit_unit_test
[==========] nxinit_tests: Running 18 test(s).
[ RUN      ] test_nxinit_parser_arguments_spaces
[       OK ] test_nxinit_parser_arguments_spaces
[ RUN      ] test_nxinit_parser_arguments_quoted
[       OK ] test_nxinit_parser_arguments_quoted
[ RUN      ] test_nxinit_parser_arguments_dashdash_separator
[       OK ] test_nxinit_parser_arguments_dashdash_separator
[ RUN      ] test_nxinit_parser_arguments_long_option
[       OK ] test_nxinit_parser_arguments_long_option
[ RUN      ] test_nxinit_parser_arguments_truncate
[       OK ] test_nxinit_parser_arguments_truncate
[ RUN      ] test_nxinit_parser_config_sections
[       OK ] test_nxinit_parser_config_sections
[ RUN      ] test_nxinit_parser_config_skip_blank_lines
[       OK ] test_nxinit_parser_config_skip_blank_lines
[ RUN      ] test_nxinit_parser_config_unknown_section
[       OK ] test_nxinit_parser_config_unknown_section
[ RUN      ] test_nxinit_parser_config_line_too_long
[       OK ] test_nxinit_parser_config_line_too_long
[ RUN      ] test_nxinit_parser_config_line_crosses_boundary
[       OK ] test_nxinit_parser_config_line_crosses_boundary
[ RUN      ] test_nxinit_parser_config_buffer_crosses_boundary
[       OK ] test_nxinit_parser_config_buffer_crosses_boundary
[ RUN      ] test_nxinit_action_event_match_exact
[       OK ] test_nxinit_action_event_match_exact
[ RUN      ] test_nxinit_action_event_match_invert
[       OK ] test_nxinit_action_event_match_invert
[ RUN      ] test_nxinit_action_event_match_fnmatch
[       OK ] test_nxinit_action_event_match_fnmatch
[ RUN      ] test_nxinit_action_event_and_semantics
[       OK ] test_nxinit_action_event_and_semantics
[ RUN      ] test_nxinit_service_duplicate_conflict
[       OK ] test_nxinit_service_duplicate_conflict
[ RUN      ] test_nxinit_service_override_replaces_duplicate
[       OK ] test_nxinit_service_override_replaces_duplicate
[ RUN      ] test_nxinit_service_args_max_boundary
[       OK ] test_nxinit_service_args_max_boundary
[==========] nxinit_tests: 18 test(s) run.
[  PASSED  ] 18 test(s).
```

All commits carry `Assisted-by:` trailers indicating AI usage, and the branch has been rebased onto current `apache/master` (no conflicts).

### DIFF
--- a/system/nxinit/Kconfig
+++ b/system/nxinit/Kconfig
@@ -70,7 +70,7 @@ config SYSTEM_NXINIT_ACTION_WARN_SLOW
 
 config SYSTEM_NXINIT_ACTION_EVENTS_MAX
 	int "Max number of events"
-	default 1
+	default 2
 	range 1 64
 	---help---
 		Maximum number of event and action events.
@@ -110,6 +110,45 @@ config SYSTEM_NXINIT_SERVICE_ARGS_MAX
 config SYSTEM_NXINIT_SERVICE_RESTART_PERIOD
 	int "Service restart period in ms"
 	default 5000
+
+comment "NXInit Testing"
+
+config SYSTEM_NXINIT_TEST
+	bool "NxInit unit tests"
+	default n
+	depends on TESTING_CMOCKA
+	---help---
+		Enable cmocka-based unit tests covering NxInit's argument/config
+		parser, action event matching, and service conflict detection.
+		Builds a separate "nxinit_unit_test" program. Under the CMake
+		build the test sources compile into a dedicated target and the
+		"init" program is unaffected; under the Make build the test
+		sources are appended to the shared CSRCS list, so they are also
+		linked into "init", enlarging it and pulling in the cmocka
+		dependency. Keep this disabled for production Make builds.
+
+if SYSTEM_NXINIT_TEST
+
+config SYSTEM_NXINIT_TEST_PRIORITY
+	int "Test task priority"
+	default 100
+
+config SYSTEM_NXINIT_TEST_STACKSIZE
+	int "Test task stack size"
+	default 8192
+	---help---
+		Several of the parser test cases build multi-hundred-byte
+		buffers on the stack (e.g. lines several times
+		CONFIG_SYSTEM_NXINIT_RC_LINE_MAX long) on top of cmocka's own
+		framework overhead. On at least one real embedded target this
+		overflowed the default DEFAULT_TASK_STACKSIZE (2048) silently
+		(no crash dump, no watchdog reset, the test task just stopped
+		producing output), which is far harder to diagnose than an
+		outright test failure. 8192 has been verified to run all test
+		cases cleanly on that target; lower it back down only if you
+		have confirmed your target's stack usage stays within bounds.
+
+endif # SYSTEM_NXINIT_TEST
 
 comment "NXInit Log level"
 
--- a/system/nxinit/Kconfig
+++ b/system/nxinit/Kconfig
@@ -40,6 +40,7 @@ config SYSTEM_NXINIT_RC_FILE_PATH
 config SYSTEM_NXINIT_RC_LINE_MAX
 	int "Max line length of RC file"
 	default 128
+	range 64 4096
 	---help---
 		Maximum line length of RC file.
 		More details: https://android.googlesource.com/platform/system/core/+/master/init/README.md
--- a/system/nxinit/CMakeLists.txt
+++ b/system/nxinit/CMakeLists.txt
@@ -38,4 +38,38 @@ if(CONFIG_SYSTEM_NXINIT)
     SRCS
     ${CSRCS})
 
+  if(CONFIG_SYSTEM_NXINIT_TEST)
+
+    # NxInit unit tests (cmocka). Unlike the Makefile build (which folds the
+    # test sources into the same CSRCS used by the "init" program above, so
+    # parser/action/service are compiled once and shared), CMake compiles
+    # action.c/parser.c/service.c a second time into this separate
+    # nxinit_unit_test target.
+
+    set(TEST_SRCS
+        test/test_nxinit.c
+        action.c
+        parser.c
+        service.c
+        test/test_nxinit_common.c
+        test/test_nxinit_parser.c
+        test/test_nxinit_action.c
+        test/test_nxinit_service.c)
+
+    nuttx_add_application(
+      NAME
+      nxinit_unit_test
+      PRIORITY
+      ${CONFIG_SYSTEM_NXINIT_TEST_PRIORITY}
+      STACKSIZE
+      ${CONFIG_SYSTEM_NXINIT_TEST_STACKSIZE}
+      MODULE
+      ${CONFIG_SYSTEM_NXINIT}
+      DEPENDS
+      cmocka
+      SRCS
+      ${TEST_SRCS})
+
+  endif()
+
 endif()
--- a/system/nxinit/Makefile
+++ b/system/nxinit/Makefile
@@ -24,7 +24,7 @@ include $(APPDIR)/Make.defs
 
 # NuttX Init
 
-MAINSRC = init.c
+MAINSRC += init.c
 CSRCS += builtin.c
 CSRCS += parser.c
 CSRCS += action.c
@@ -32,9 +32,24 @@ CSRCS += service.c
 CSRCS += import.c
 CSRCS += property_simple.c
 
-PROGNAME = $(CONFIG_SYSTEM_NXINIT_PROGNAME)
-PRIORITY = $(CONFIG_SYSTEM_NXINIT_PRIORITY)
-STACKSIZE = $(CONFIG_SYSTEM_NXINIT_STACKSIZE)
+PROGNAME += $(CONFIG_SYSTEM_NXINIT_PROGNAME)
+PRIORITY += $(CONFIG_SYSTEM_NXINIT_PRIORITY)
+STACKSIZE += $(CONFIG_SYSTEM_NXINIT_STACKSIZE)
 MODULE = $(CONFIG_SYSTEM_NXINIT)
+
+# NxInit unit tests (cmocka), reusing the same parser/action/service CSRCS
+# built above.
+
+ifneq ($(CONFIG_SYSTEM_NXINIT_TEST),)
+CSRCS += test/test_nxinit_common.c
+CSRCS += test/test_nxinit_parser.c
+CSRCS += test/test_nxinit_action.c
+CSRCS += test/test_nxinit_service.c
+MAINSRC += test/test_nxinit.c
+
+PROGNAME += nxinit_unit_test
+PRIORITY += $(CONFIG_SYSTEM_NXINIT_TEST_PRIORITY)
+STACKSIZE += $(CONFIG_SYSTEM_NXINIT_TEST_STACKSIZE)
+endif
 
 include $(APPDIR)/Application.mk
--- a/system/nxinit/parser.c
+++ b/system/nxinit/parser.c
@@ -56,10 +56,6 @@ static int init_parse_config_lines(FAR const struct parser_s *parser,
       *(nl++) = '\0';
       *len -= nl - buf;
       init_debug("Line %-3zu '%s'", ++*line, buf);
-      if (*buf == '\0')
-        {
-          continue;
-        }
 
       /* Skip empty lines and lines containing only whitespace */
 
--- a/system/nxinit/parser.c
+++ b/system/nxinit/parser.c
@@ -96,8 +96,8 @@ static int init_parse_config_lines(FAR const struct parser_s *parser,
   return 0;
 }
 
-static int init_parse_config_buffer(FAR const struct parser_s *parser,
-                                    FAR const char *buf, size_t len)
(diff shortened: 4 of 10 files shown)


linguini1 (commented):
Please add the `Assisted-by` field to your commits to indicate AI usage.

JianyuWang0623:
The `sim/lua` build failure on `Linux (sim-02)` in CI is unrelated to this PR. Root cause: the kernel renamed the Kconfig option `PSEUDOFS_SOFTLINKS` → `FS_LINKS`, but `interpreters/luamodules/luv/Kconfig` in apps still uses the old name (`depends on PSEUDOFS_SOFTLINKS`). Since that symbol no longer exists, `CONFIG_LUA_LUV_MODULE` becomes unselectable and gets silently dropped by `savedefconfig`, leaving the `sim/lua` defconfig dirty and failing the build.

The fix is already up as #3753 (`fs: rename PSEUDOFS_SOFTLINKS to FS_LINKS`), which renames the dependency in `luv/Kconfig` (and other apps files). Once that merges, re-running CI here should be green. This PR itself only touches `system/nxinit/`.

system/nxinit/parser.h
```diff
@@ -56,4 +57,17 @@ int init_parse_arguments(FAR char *buf, bool dup, int argc, FAR char **argv);
 int init_parse_configs(FAR const struct parser_s *parser);
 int init_parse_config_file(FAR const struct parser_s *parser,
                            FAR const char *file);
+
+#ifdef CONFIG_SYSTEM_NXINIT_TEST
```
xiaoxiang781216: let's expose init_parse_config_buffer directly

system/nxinit/Kconfig
```diff
@@ -70,6 +70,7 @@ config SYSTEM_NXINIT_ACTION_WARN_SLOW
 
 config SYSTEM_NXINIT_ACTION_EVENTS_MAX
 	int "Max number of events"
+	default 2 if SYSTEM_NXINIT_TEST
```
xiaoxiang781216: let's change the default value directly

system/nxinit/parser.h
JianyuWang0623: done

system/nxinit/Kconfig
JianyuWang0623: done

xiaoxiang781216 (approved):

acassis (approved):

PR #3761, opened 2026-08-28
system/nxinit: add rptun/unlink builtins and service fallback option

JianyuWang0623:
## Summary

1. **rptun / unlink builtin commands.** Some products drive early bring-up
   entirely from `init.rc` rather than NSH, so the NSH `rptun` command is
   not available there. Add `rptun` as an init builtin (guarded by
   `CONFIG_RPTUN`) supporting `start`/`stop`, plus a generic `unlink`
   builtin (always available) for removing device nodes, e.g.:

   ```
   rptun stop /dev/rptun/proc
   unlink /dev/rptun/proc
   rptun start /dev/rptun/proc
   ```

   `RPTUNIOC_START` returns the pid of the rptun kernel thread in async
   mode (`CONFIG_RPTUN_START_SYNC` unset). That thread is a detached
   kthread, never a child of init, so returning a positive value would
   make the action engine `waitpid()` on it and block the whole action
   queue (and `on init` / console would never run). A successful start is
   normalized to 0.

2. **`fallback` service option.** Add an `SVC_FALLBACK` flag and a
   `fallback` service option — the semantic opposite of `override` — to
   resolve same-name service conflicts. `override` makes the new
   definition replace the old; `fallback` makes the definition marked
   fallback yield to the other one, so a board-level `init.rc` can provide
   a default service that is silently dropped when another `init.rc`
   defines a service with the same name (and vice versa). If neither flag
   is set, duplicate service names still produce `-EEXIST` as before.

## Impact

- Only touches `system/nxinit/{builtin.c,service.c,service.h}`. No change
  to existing builtins/options; new behavior is opt-in via the new
  `rptun`/`unlink` commands and the `fallback` option.
- `rptun` builtin is compiled only when `CONFIG_RPTUN` is enabled;
  `unlink` and `fallback` are always available.
- No new Kconfig symbols.

## Testing

Built and verified locally with the `sim` target (host gcc). `nxstyle`
clean on all three touched files.

`fallback` verified at runtime on `sim` + nxinit with an `init.rc` that
declares two services named `console` (A/B against the same config
without the flag):

```
# without "fallback":
Error Redefined service 'console'          # parse fails (-EEXIST)

# with "fallback" on the second definition:
nsh> ps
  TID  PID PPID ... COMMAND
    4    4    0 ... init_main               # nxinit is init
    5    5    4 ... sh                       # the non-fallback console starts;
                                             # the fallback one is silently ignored, no error
```

`unlink` is always compiled in and `rptun` compiles under `CONFIG_RPTUN`;
both build cleanly. The `sim` target has no rptun device, so the rptun
runtime path is not exercised here — the concurrency fix (normalizing a
successful async `RPTUNIOC_START` to 0 so the action engine does not
`waitpid()` on the detached rptun kthread) is a targeted one-line change
verified by build + code review.

cc @wyr-7

### DIFF
--- a/system/nxinit/builtin.c
+++ b/system/nxinit/builtin.c
@@ -323,7 +323,7 @@ static int cmd_rptun(FAR struct action_manager_s *am,
     }
 
   close(fd);
-  return ret;
+  return ret < 0 ? ret : 0;
 }
 #endif
 
--- a/system/nxinit/builtin.c
+++ b/system/nxinit/builtin.c
@@ -25,15 +25,22 @@
  ****************************************************************************/
 
 #include <errno.h>
+#include <fcntl.h>
 #include <inttypes.h>
 #include <stdlib.h>
 #include <string.h>
 #include <spawn.h>
+#include <unistd.h>
 #include <sys/boardctl.h>
+#include <sys/ioctl.h>
 #include <sys/param.h>
 #include <sys/wait.h>
 
 #include <netutils/netinit.h>
+
+#ifdef CONFIG_RPTUN
+#include <nuttx/rptun/rptun.h>
+#endif
 
 #include "builtin.h"
 #include "init.h"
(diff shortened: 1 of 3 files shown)


JianyuWang0623:
Note: The `Assisted-by` trailer appears only in commits that were originally authored with the help of AI coding tools. Some commits in this series (e.g., the rptun/unlink builtin and the fallback service option) were originally created on our internal Gerrit at an earlier date, before AI-assisted workflows were adopted, so they do not carry the `Assisted-by` tag. The commit message content is otherwise preserved as-is from the original internal commits.

xiaoxiang781216 (approved):

GUIDINGLI (approved):

PR #3762, opened 2026-08-28
examples/fdpicxip: Build the module fixtures without nuttx/tools/fdpic.

casaroli:
## Summary

The FDPIC modules this example and `testing/fs/xipfs` carry are built by an explicit `make regen`, which reached into `nuttx/tools/fdpic` for a makefile, a script that turned a module into a header, and two more that checked its imports.

Review of apache/nuttx#19940 asked that NuttX not carry a module build of its own, and it no longer does: with `CONFIG_FDPIC` an ordinary module is built by `apps/Application.mk` like any other. What these fixtures need comes here instead, and it is much less.

They keep a build at all because they are loader edge cases, not ordinary modules: a library with a `SONAME`, a module with more `DT_NEEDED` entries than the loader will follow, one whose imports stay in the lazy binding table, and one naming a symbol the firmware does not export, which exists to be refused. `Application.mk` cannot say `-soname`, or "no entry point", or "link against that other module".

What arrives is 145 changed lines in the makefile that was already here, plus a 24-line license template. What does not arrive is four files and 536 lines:

- the generic module makefile, which existed to be included by anything and was only ever included here
- `fdpic-embed.py`, because `xxd -i` does that and `examples/elf` already uses it that way
- `fdpic-verify.sh` and `nuttx-exports.sh`, with no replacement: they checked at build time what the xipfs suite already asserts at run time

The fixtures also stop carrying a `crt0` and a linker script. Both come from the tree named by `NUTTX_DIR`, which is where the in-tree build takes them. The `crt0` source is compiled here rather than the built object taken, because these are deliberately built for `cortex-m3` while the firmware is not: a v7-M module runs on both v7-M and v8-M targets, so one set of headers serves the RP2350 and mps2-an500 alike.

## Impact

None on any build. `make regen` is not part of the application build and nothing else runs these files. The `*_bin.h` headers stay committed, so both apps still build with a plain toolchain and no FDPIC linker.

`NUTTX_DIR` now has to be configured with `CONFIG_FDPIC`, because that is what puts the FDPIC segments in the generated `gnu-elf.ld`.

The committed headers are unchanged here. They will change when next regenerated, because a fixture now carries the tree's `crt0` rather than one of its own, and that is a change the xipfs suite should be run against rather than made blind.

## Testing

Rebuilt three fixtures from clean against a tree configured with `CONFIG_FDPIC`, covering both kinds and both languages: `qsorter` (C module), `libshape` (C++ library) and `cxxuser` (C++ module linked against it).

```
$ readelf -h qsorter.fdpic
  OS/ABI:      ARM FDPIC
  Entry point: 0x1              <- _start, from the tree's crt0
$ readelf -A qsorter.fdpic
  Tag_CPU_arch: v7
$ readelf -d libshape.fdpic
  0x0e (SONAME)      Library soname: [libshape.so]
  0x19 (INIT_ARRAY)  0x1000
```

Two `PT_LOAD` segments, and v7-M as intended rather than the firmware's v8-M.

`callback` is the one fixture not rebuilt here: it uses `SIGEV_THREAD`, so it needs a tree with `CONFIG_SIG_EVTHREAD`.

`tools/checkpatch.sh -c -u -m -g` passes.


examples/fdpicxip/modules/Makefile
```diff
@@ -44,19 +44,70 @@
...
+
+ifeq ($(origin LD),default)
+  LD := arm-uclinuxfdpiceabi-ld
+endif
+
+# -mfdpic -fPIC is the whole of what makes an FDPIC object; the rest is what
+# a module needs anywhere.  -fno-use-cxa-atexit puts a static object's
+# destructor in .fini_array, which is where the loader looks for it.
+
+MODCFLAGS   = -mcpu=$(CPU) -mthumb -mfdpic -fPIC -Os -fno-builtin -Wall \
```
xiaoxiang781216: don't need any more with the new kernel side change

examples/fdpicxip/modules/Makefile
```diff
@@ -93,50 +144,45 @@ endif
...
-	$(BUILD) MODULE=cxxuser CXXSRCS=cxxuser.cpp LIBS=libshape.so
+cxxuser.fdpic: cxxuser.o crt0.o libshape.so
+	$(LINKMOD) cxxuser.o libshape.so
 
 # BINDNOW is emptied so the imported descriptors stay in the lazy binding
 # table, which is the case this module exists to cover.
 
-lazymod.fdpic: lazymod.c
-	$(BUILD) MODULE=lazymod SRCS=lazymod.c BINDNOW=
+lazymod.fdpic: lazymod.o crt0.o
```
xiaoxiang781216: remove fdpic suffix, could we generate FDPIC elf just like the ordinary elf by setting MODULE=m

acassis (dismissed):

acassis (changes requested):
@casaroli please include the CMakefile.txt too

casaroli:
> @casaroli please include the CMakefile.txt too

Done

acassis (approved):

testing/fs/xipfs/missingsym_bin.h
```diff
@@ -20,181 +20,871 @@
  *
  ****************************************************************************/
 
-/* Generated from missingsym.fdpic -- do not edit.
```
xiaoxiang781216, line 23: could we remove the generated code from git, and reference the generated code directly?

casaroli: I think it is not good idea because not every developer will have the supported binutils like we have in the CI image. So anyone wanting to run the tests will need it. If we want, we could check that the files match the binaries, in case anyone change the sources and forgets to update the binaries

examples/fdpicxip/modules/Makefile
```diff
@@ -44,19 +45,57 @@
...
+MODCXXFLAGS = $(filter-out $(DROPFLAGS),$(CXXELFFLAGS)) -mcpu=$(CPU) -g0
+MODLDFLAGS  = $(filter-out %crt0.o -e _start,$(LDELFFLAGS))
 
 DEMODIR = ..
 DEMOREL = apps/examples/fdpicxip
 TESTDIR = ../../../testing/fs/xipfs
 TESTREL = apps/testing/fs/xipfs
 
-BUILD = $(MAKE) -f $(MODULE_MK) NUTTX_DIR=$(NUTTX_DIR) CPU=$(CPU)
+%.o: %.c
```
xiaoxiang781216, line 72: can we build fdpic so like the normal so by `DYNLIB = y` like examples/sotest, examples/module?

PR #3769, opened 2026-09-03
graphics/microwindows: nanoxcalc fix and upstream update to fix warnings

ppisa:
## Summary

graphics/microwindows: nanoxcalc fix and upstream update to fix warnings

## Impact

This silences warnings in the build and makes integration of example configuration
including Microwindows build possible.

## Testing

Tested on qemu-intel64 target for now

### DIFF
--- a/examples/nanoxcalc/nxcalc.c
+++ b/examples/nanoxcalc/nxcalc.c
@@ -176,6 +176,7 @@ int main(int argc, char **argv)
         {
           int x = grid_x + c * cell_w;
           int y = grid_y + r * cell_h;
+
           btn_rects[r][c].x = x;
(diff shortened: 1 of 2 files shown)


ppisa:
There is no way how to resolve
```
error: Mixed case identifier found
```
because these identifiers are mandated by Nano-X API and has equivalents in X11.

See the file `XtoNX.h` for standard X11 identifiers mapping to Nano-X ones

https://github.com/ghaerr/microwindows/blob/master/src/nx11/XtoNX.h

which enables to build X11 applications against Nano-X.

`XOpenDisplay` maps to `GrOpen`
 `XCreateSimpleWindow` to `GrNewWindow`
etc.

PR #3771, opened 2026-09-04
testing/nettest: add checksum zero-length regression test

imeghar2408-max:
## Summary

Add a CMocka regression test for the zero-length fragment handling issue in
`checksum()` reported in Apache NuttX issue #20010.

When `checksum()` is called with `len == 0` while an odd byte is pending, the
previous implementation could read `data[0]` even though the current fragment
contained no valid bytes. It could also clear the pending odd-byte state.

The regression test covers this through the public `chksum_iob()` path using:

`AA BB CC | empty fragment | DD EE`

and verifies that the chained checksum matches the checksum of the equivalent
contiguous data:

`AA BB CC DD EE`

This verifies that an empty fragment does not affect the checksum and that
the pending odd-byte state is preserved across the empty fragment.

## Changes

- Add `test_others_chksum()` regression test.
- Register the test in `testing/nettest/others/test_others.c`.
- Add its declaration to `testing/nettest/others/test_others.h`.
- Add the test source to `testing/nettest/Makefile`.
- Add the test source to `testing/nettest/CMakeLists.txt`.

## Impact

This change only adds a regression test to `nuttx-apps`.
It does not change runtime behavior or production code.
### Regression test with the buggy implementation

The regression test was verified against the implementation without the
zero-length guard and failed as expected:

```text
[ RUN      ] test_others_chksum
[ ERROR    ] --- 30617 != 30396
[ LINE     ] --- others/test_others_chksum.c:52
[  FAILED  ] test_others_chksum
Regression test with the fix applied
````
With the len == 0 handling fix restored, the test passed:

```text
nsh: mount: mount failed: 20
nsh> cmocka_net_others
[==========] others_tests: Running 2 test(s).
[ RUN      ] test_others_bufpool
[       OK ] test_others_bufpool
[ RUN      ] test_others_chksum
[       OK ] test_others_chksum
[==========] others_tests: 2 test(s) run.
[  PASSED  ] 2 test(s).
````
The mount failed: 20 message occurs during simulator startup and did not
prevent cmocka_net_others from running successfully.

Checkpatch

The corresponding NuttX fix was checked with the CI-style command:

Used config files:
````text
    1: .codespellrc
✔️ All checks pass.
````

### DIFF
--- a/testing/nettest/CMakeLists.txt
+++ b/testing/nettest/CMakeLists.txt
@@ -77,7 +77,8 @@ if(CONFIG_TESTING_NET_TEST)
     set(SRCS
         ${CMAKE_CURRENT_LIST_DIR}/others/test_others.c
         ${CMAKE_CURRENT_LIST_DIR}/others/test_others_common.c
-        ${CMAKE_CURRENT_LIST_DIR}/others/test_others_bufpool.c)
+        ${CMAKE_CURRENT_LIST_DIR}/others/test_others_bufpool.c
+        ${CMAKE_CURRENT_LIST_DIR}/others/test_others_chksum.c)
 
     nuttx_add_application(
       NAME
--- a/testing/nettest/Makefile
+++ b/testing/nettest/Makefile
@@ -54,7 +54,8 @@ endif
 ifeq ($(CONFIG_TESTING_NET_OTHERS),y)
 MAINSRC  += others/test_others.c
 PROGNAME += cmocka_net_others
-CSRCS    += others/test_others_common.c others/test_others_bufpool.c
+CSRCS    += others/test_others_common.c others/test_others_bufpool.c \
+            others/test_others_chksum.c
 endif
 
 endif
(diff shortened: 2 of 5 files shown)


acassis:
@imeghar2408-max please fix the issues found by the CI,
before submitting the PR you should verify it locally using the checkpatch.sh, please read the contributing guy

imeghar2408-max:
Hi, thanks for pointing that out. I fixed the CI issues and verified the changes locally with checkpatch.sh. The check now passes with all checks passing. I’ve also force-pushed the updated commit.

> @imeghar2408-max please fix the issues found by the CI, before submitting the PR you should verify it locally using the checkpatch.sh, please read the contributing guy

xiaoxiang781216 (dismissed):

jerpelea (dismissed):

xiaoxiang781216 (approved):

jerpelea (approved):

PR #3776, opened 2026-09-08
system/nxinit: add a per-service "console" option

JianyuWang0623:
## Summary

Adds a per-service `"console [<device>]"` option to nxinit's init.rc
service syntax, plus two small robustness fixes uncovered while adding
it.

Services started by nxinit (e.g. plain `"sh"`) do not open a console
device on their own, unlike `nsh_main`, which explicitly does so via
`nsh_consolemain()`/`nsh_waitusbready()` for USB gadget consoles
(CDC-ACM/PL2303). When a board switches its top-level init from
`nsh_main` to nxinit, no code path ever registers/connects a USB
console gadget, and a service that just execs a plain `"sh"` inherits
whatever (invalid, for a gadget console not yet opened) stdio nxinit
itself has. This was the actual root cause of a real-hardware
regression reported on esp32s3-xiao (board hangs, no USB console) —
see the companion apache/nuttx PR for the board-side fix that uses
this option.

A service declared with `console [<device>]` gets the given device
(`CONFIG_SYSTEM_NXINIT_CONSOLE_DEV`, `"/dev/console"` by default, if no
device is given) opened and dup'd onto its stdin, stdout and stderr via
`posix_spawn_file_actions` before it is spawned. This does not depend
on nsh being enabled at all — for a USB gadget console, boards are
expected to bring the gadget up themselves before any service using
`"console"` starts (e.g. via an `exec -- sercon` init.rc action, since
`apps/system/cdcacm` already implements exactly that and depends on
neither nxinit nor nsh either).

Two commits unrelated to the option itself, split out for review:

- `system/nxinit: fix missing blank line after declaration in
  service.c` — pre-existing nxstyle nit in `option_reboot_on_failure()`,
  unrelated to this change but in a file this PR already touches.
- `system/nxinit: retry a service whose spawn failed` —
  `init_service_refresh()` ignored the return value of
  `init_service_start()`; a service whose spawn fails stayed
  `SVC_RESTARTING` with nothing left to arm this function's own poll
  timeout, and (being a failed spawn) no `SIGCHLD` either to wake the
  event loop some other way. If it were the only pending timer, the
  loop would block indefinitely and the service would never be
  attempted again. Also moves the `time_started` update in
  `init_service_start()` from after a successful spawn to before the
  spawn is even attempted, so a failure doesn't leave a stale timestamp
  behind (which would otherwise make a repeatedly failing service look
  permanently "overdue" and get retried on every unrelated wakeup,
  bypassing its own `restart_period`).

## Impact

- New `"console"` service option in nxinit's init.rc syntax; existing
  init.rc files using services without it are unaffected.
- `init_service_refresh()`/`init_service_start()` behavior change is a
  pure bug fix (previously-ignored failure case); no change to the
  successful-spawn path.

## Testing

Covered by a new unit test, `test_nxinit_service_console_option`
(`system/nxinit/test/test_nxinit_service.c`), exercising the option
with and without an explicit device, and confirming services without
the option are left untouched.

Build-verified against `esp32s3-xiao:usbnsh`/`combo` (apache/nuttx)
with the option actually wired up via the companion board-side PR:

```
$ ./tools/configure.sh -l esp32s3-xiao:usbnsh && make -j8
...
MKIMAGE: ESP32-S3 binary
Successfully created esp32s3 image.
Generated: nuttx.bin

$ tools/checkpatch.sh -g <each of the 3 commits>
✔️ All checks pass.
```

End-to-end hardware verification on real Seeed XIAO ESP32-S3 Sense
(`usbnsh`): a real physical reset (`dmesg` shows a clean USB
disconnect/reconnect, not a software-triggered one) now enumerates the
NuttX CDC-ACM console gadget instead of leaving USB dark:

```
usb 1-2.3: USB disconnect, device number 97
usb 1-2.3: New USB device found, idVendor=0525, idProduct=a4a7
usb 1-2.3: Product: CDC/ACM Serial
usb 1-2.3: Manufacturer: NuttX
```

Serial console over that device, showing `init_main` (nxinit) as PID 2
and the `console` service's `sh` as its child (PID 4), i.e. the
`"console"` option actually connected `sh`'s stdio to the gadget:

```
nsh> uname -a
NuttX 13.0.1-RC0 7cc6707a76b-dirty Sep  8 2026 10:53:53 xtensa esp32s3-xiao
nsh> ps
    0     0     0   0 FIFO     Kthread   - Ready              0000000000000000 0003040 Idle_Task
    2     2     0 100 RR       Task      - Waiting  Semaphore 0000000000000000 0008120 init_main
    4     4     2 100 RR       Task      - Running            0000000000000000 0008144 sh
nsh> ls /etc
/etc:
 init.d/
```

Before this option existed (plain `service console sh`, no `console`
line, no `exec -- sercon`), the same hardware reproduced the original
regression exactly: after a real reset the USB device disappears
entirely and never re-enumerates.

More detail on the board-side setup (`init.rc`/defconfig changes, the
`exec -- sercon` step, and a from-scratch repro of the original hang)
is in the companion apache/nuttx#20088.

### DIFF
--- a/system/nxinit/Kconfig
+++ b/system/nxinit/Kconfig
@@ -111,6 +111,20 @@ config SYSTEM_NXINIT_SERVICE_RESTART_PERIOD
 	int "Service restart period in ms"
 	default 5000
 
+config SYSTEM_NXINIT_CONSOLE_DEV
+	string "Default console device of a service"
+	default "/dev/console"
+	---help---
+		The device used as stdin/stdout/stderr of a service declared with the
+		option "console" and no explicit device, i.e. "console" rather than
+		"console <device>".
+
+		This is needed by services which expect a controlling terminal but do
+		not open one themselves (e.g. plain "sh"), and is the only way to get
+		a console when the console device is not available at the time the
+		file descriptors of the idle task are set up, as is the case for a USB
+		gadget console.
+
 comment "NXInit Testing"
 
 config SYSTEM_NXINIT_TEST
--- a/system/nxinit/service.c
+++ b/system/nxinit/service.c
@@ -323,7 +323,11 @@ int init_service_refresh(FAR struct service_manager_s *sm)
           ms = TIMESPEC2MS(diff);
           if (ms >= service->restart_period)
             {
-              init_service_start(service);
+              if (init_service_start(service) < 0)
+                {
+                  min = MIN(min, service->restart_period);
+                }
+
               continue;
             }
 
@@ -466,6 +470,8 @@ int init_service_start(FAR struct service_s *service)
       return -ret;
     }
 
+  clock_gettime(CLOCK_MONOTONIC, &service->time_started);
+
   ret = posix_spawnp(&pid, service->argv[2], NULL, &attr, &service->argv[2],
                      environ);
   posix_spawnattr_destroy(&attr);
@@ -477,7 +483,6 @@ int init_service_start(FAR struct service_s *service)
     }
 
   service->pid = pid;
-  clock_gettime(CLOCK_MONOTONIC, &service->time_started);
   add_flags(service, SVC_RUNNING);
   remove_flags(service, SVC_RESTARTING);
   remove_flags(service, SVC_DISABLED);
--- a/system/nxinit/service.c
+++ b/system/nxinit/service.c
@@ -279,6 +279,7 @@ static int option_reboot_on_failure(FAR struct service_manager_s *sm,
 {
   FAR struct service_s *s = list_last_entry(&sm->services, struct service_s,
                                             node);
+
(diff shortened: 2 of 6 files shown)


acassis (approved):
@JianyuWang0623 please update the Documentation to include info about this "console" option

JianyuWang0623:
> @JianyuWang0623 please update the Documentation to include info about this "console" option

@acassis Docs are up in https://github.com/apache/nuttx/pull/20098  (currently draft)

xiaoxiang781216 (approved):

simbit18 (approved):

simbit18 (approved):

simbit18 (approved):

PR #3782, opened 2026-09-14
netutils/ptpd: IEEE 1588 P2P delay mechanism, hardware clock phase-lock and outlier rejection

daniel-p-carvalho:
## Summary

`apps/netutils/ptpd` only implemented the end-to-end (E2E) delay mechanism. The IEC/IEEE 61850-9-3 profile uses the peer-to-peer (P2P) delay mechanism instead, so a Grandmaster running that profile (`linuxptp`'s `ptp4l`) could not be synchronized against without switching it to E2E. This PR adds P2P, lets the daemon discipline a hardware PTP clock (`/dev/ptp0`) in phase as well as in frequency, and adds an optional filter that keeps a single disturbed Sync sample from moving the clock.

The series was rewritten after review: the custom socket ioctls (`SIOCG_TX_HW_TIMESTAMP`, `SIOCS_PTP_ADJFREQ`, `SIOCS_PTP_ADJPHASE`) that an earlier revision consumed are gone, replaced by the POSIX `clock_adjtime()` interface on a PTP hardware clock device, as suggested in the review of the kernel side. It is rebased on current `master` (which already contains #3779).

### Commits

1. `Fix Delay_Resp consumption by sendmsg.` `ptp_sendmsg()` did a blocking `recvmsg()` on the transmit socket right after sending a `Delay_Req` when hardware timestamping was selected. That socket shares the connection with the event socket, so the call swallowed the next PTP packet on the wire, normally the `Delay_Resp`, into a buffer that was then discarded, and `path_delay_ns` stayed at 0. The path delay is also computed with the IEEE 1588-2008 clause 11.3 formula instead of an approximation that only held once the clock had converged. This bug is still present in `master`.
2. `Implement IEEE 1588 peer-to-peer (P2P) delay mechanism.` Adds the `Pdelay_Req`, `Pdelay_Resp` and `Pdelay_Resp_Follow_Up` messages, the `-P` option (exclusive with `-E`), requester and responder logic, and replaces `bool delay_e2e` with `enum ptp_delay_mechanism_e` in `struct ptpd_config_s`. The messages carry PTP version 2.0 and `controlField` 0x05, which `linuxptp` requires to accept them. An orphaned `Pdelay_Resp_Follow_Up` from an abandoned cycle is rejected instead of being paired with stale timestamps, P2P without `CONFIG_SCHED_TICKLESS` prints a warning at startup, and IP multicast handling is skipped for `AF_PACKET`.
3. `Discard implausible drift-rate samples before averaging.` A single drift sample taken over an unusually short or long interval could pass the slew-rate check and corrupt the long-term average. Adds `CONFIG_NETUTILS_PTPD_MAX_DRIFT_PPB` (default 500000) as a dedicated plausibility bound.
4. `Phase-lock hardware clock via POSIX clock_adjtime.` When the clock is a PTP hardware clock, the residual phase offset is now converted to a frequency correction over `CONFIG_CLOCK_ADJTIME_PERIOD_MS` and applied with `ADJ_FREQUENCY`, so the counter, and any PPS output driven by it, is pulled into phase instead of only tracking the frequency.
5. `Add ingress latency compensation for RX timestamps.` Adds `CONFIG_NETUTILS_PTPD_INGRESS_LATENCY_NS` (default 0) and the `-I` option. The value is subtracted from every hardware receive timestamp, which is the `ingressLatency` port parameter of IEEE 1588.
6. `Retain in-memory IPC for CONFIG_BUILD_FLAT in ptpd_status().` The file based IPC introduced for the protected and kernel builds made `ptpd -t` time out on flat builds without a mounted `/tmp`. The in-memory queue is kept for `CONFIG_BUILD_FLAT` and the file is used otherwise.
7. `Discard outlier Sync phase error samples.` A Sync sample whose receive timestamp was taken late, for example because the task was scheduled late with software timestamping, went straight into the phase correction and the drift estimate. Adds `CONFIG_NETUTILS_PTPD_OUTLIER_THRESHOLD_NS` (default 0, which disables the check). A phase error that differs by more than this many nanoseconds from the median of the last five accepted samples is discarded with a warning. After eight consecutive rejections the sample is accepted and the history restarts from it, so a real step of the master is still followed while a short burst of disturbed samples is ridden out. The history is also restarted whenever the clock is stepped.

## Impact

- New options: `-P` and `-I`, `CONFIG_NETUTILS_PTPD_MAX_DRIFT_PPB`, `CONFIG_NETUTILS_PTPD_INGRESS_LATENCY_NS`, `CONFIG_NETUTILS_PTPD_OUTLIER_THRESHOLD_NS`. The defaults leave the previous behaviour unchanged (E2E remains available with `-E`, ingress compensation is 0, outlier rejection is off).
- `struct ptpd_config_s` changes: `delay_e2e` is replaced by `delay_mechanism`, and `ingress_latency_ns` is added. The only user in the tree is `system/ptpd`, which is updated here.
- The phase-lock in commit 4 needs a PTP hardware clock device. On STM32 that is `/dev/ptp0` from apache/nuttx#20180. Users that keep the default `CLOCK_REALTIME` are not affected.
- Hardware transmit timestamps are not part of this PR. Until they land (a follow-up series, together with apache/nuttx#20148) the transmit side of `Delay_Req` and `Pdelay_Req` is timestamped in software.

## Testing

Built with `stm32f4discovery:netnsh` plus `NETUTILS_PTPD`, in two configurations (with and without `NET_PKT` and `NET_TIMESTAMP`), without errors or warnings from `ptpd`. The last commit was also built with `CONFIG_NETUTILS_PTPD_OUTLIER_THRESHOLD_NS` at 0 and at 100000. `./tools/checkpatch.sh -g <base>..HEAD` passes.

On hardware: a custom STM32F407 board with the STM32 Ethernet MAC and a DP83848 PHY, over `AF_PACKET` (`ptpd -2 -s -p /dev/ptp0 -B -H -P -i eth0`), against a GNSS-referenced `ptp4l` Grandmaster using the IEC/IEEE 61850-9-3 profile. These runs were done with this series together with the hardware transmit timestamp series mentioned above:

- Direct cable: `path_delay` of about 1.2 µs.
- Through a Fast Ethernet switch, 20 minute run, 159 samples, no failed queries: `path_delay` 9265.3 ns (σ = 2.5 ns), `drift` about -79.95 ppm, phase offset about -2.2 µs.
- Several 5 minute runs after rebasing: `path_delay` between 9.26 and 9.33 µs and no failed queries.

This exact series, without the transmit timestamp series, was also run on the same hardware (5 minutes). Its transmit timestamps are taken in software, after `sendmsg()` returns, and the measured peer delay came out negative (about -18 µs), which the path delay filter rejects, so `path_delay_ns` stayed at 0 and the phase offset was worse (about -6.7 µs against -2.4 µs with hardware transmit timestamps). In other words P2P on this hardware needs accurate transmit timestamps. The ingress compensation is implemented but its value has not been calibrated, so the default 0 applies no correction.

Outlier rejection (commit 7):

- The selection logic was checked on the host with the function extracted from the source: a burst of four bad samples is discarded and the next normal one is accepted, and a real 500 µs step has seven samples discarded and the eighth accepted, after which the new level keeps being accepted.
- With hardware timestamps and the threshold at 100000 ns, a 220 sample run discarded none: the deviation from the median had a median of 2.4 µs, a 95th percentile of 2.8 µs and a maximum of 81 µs, so the filter does not interfere in the normal case.
- With software timestamps (`-S`) the deviations are hundreds of microseconds and come in bursts; a limit of three consecutive rejections let one of those bursts through, which is why the limit is eight. The accuracy of `-S` is still bounded by the receive timestamp being taken when the task wakes up, and this filter does not change that.


acassis (dismissed):

jerpelea (dismissed):

xiaoxiang781216:
please fix the conflict. @daniel-p-carvalho

netutils/ptpd/ptpd.c
```diff
@@ -861,7 +861,25 @@ static int ptp_sendmsg(FAR struct ptp_state_s *state, FAR const void *buf,
...
+      struct ifreq req;
+      int ret_ts;
+
+      memset(&req, 0, sizeof(req));
+      strlcpy(req.ifr_name, state->config->interface, sizeof(req.ifr_name));
+      req.ifr_data = sendts;
+
+      ret_ts = ioctl(state->tx_socket, SIOCG_TX_HW_TIMESTAMP,
+                     (unsigned long)&req);
+      if (ret_ts < 0)
```
xiaoxiang781216: `if (ioct(...) < 0)` and remove line 866

netutils/ptpd/ptpd.c
```diff
@@ -1349,14 +1474,53 @@ static int ptp_process_delay_req(FAR struct ptp_state_s *state,
...
 }
 
+/* Record and filter measured path delay (used by both E2E and P2P) */
+
+static void ptp_record_path_delay(FAR struct ptp_state_s *state,
+                                  int64_t path_delay)
+{
+  int64_t max_path_delay;
+
+  max_path_delay = CONFIG_NETUTILS_PTPD_MAX_PATH_DELAY_NS;
```
xiaoxiang781216, line 1543: merge with line 1482

netutils/ptpd/ptpd.c
```diff
@@ -1349,14 +1474,53 @@ static int ptp_process_delay_req(FAR struct ptp_state_s *state,
...
+       * timestamping is available.
+       */
+
+      max_path_delay = 10 * (int64_t)NSEC_PER_MSEC;
+    }
+
+  if (path_delay >= 0 && path_delay < max_path_delay)
+    {
+      if (state->path_delay_avgcount <
+          CONFIG_NETUTILS_PTPD_DELAYREQ_AVGCOUNT)
```
xiaoxiang781216, line 1558: merge into one line

netutils/ptpd/ptpd.c
```diff
@@ -1349,14 +1474,53 @@ static int ptp_process_delay_req(FAR struct ptp_state_s *state,
...
+        {
+          state->path_delay_avgcount++;
+        }
+
+      state->path_delay_ns += (path_delay - state->path_delay_ns)
+                              / state->path_delay_avgcount;
+
+      ptpinfo("Path delay: %ld ns (avg: %ld ns)\n",
+              (long)path_delay, (long)state->path_delay_ns);
```
xiaoxiang781216: remove all cast

netutils/ptpd/ptpd.c
```diff
@@ -1349,14 +1474,53 @@ static int ptp_process_delay_req(FAR struct ptp_state_s *state,
...
+    }
+  else
+    {
+      ptpwarn("Path delay out of range: %lld ns\n",
+              (long long)path_delay);
```
xiaoxiang781216, line 1572: remove too

netutils/ptpd/ptpd.c
```diff
@@ -1395,52 +1559,218 @@ static int ptp_process_delay_resp(FAR struct ptp_state_s *state,
...
+      return OK; /* Not for us */
+    }
 
-  if (msg->header.logmessageinterval <= 12)
+  sequence = ptp_get_sequence(&msg->header);
+  if (sequence != state->pdelay_req_seq)
     {
-      interval = (1 << msg->header.logmessageinterval);
+      ptpwarn("Ignoring out-of-sequence Pdelay_Resp (%d vs. expected %d)\n",
+              (int)sequence, (int)state->pdelay_req_seq);
```
xiaoxiang781216: remove the cast

netutils/ptpd/ptpd.c
```diff
@@ -1395,52 +1559,218 @@ static int ptp_process_delay_resp(FAR struct ptp_state_s *state,
...
+  ptp_format_to_timespec(msg->requestreceipttimestamp,
+                         &state->pdelayreq_rx_time);
+  ptp_add_correction_time(msg->header.correction,
+                          &state->pdelayreq_rx_time);
+
+  if (msg->header.flags[0] & PTP_FLAGS0_TWOSTEP)
+    {
+      state->pdelay_waiting_followup = true;
+      ptpinfo("Waiting for Pdelay_Resp_Follow_Up, seq %d\n",
+              (int)sequence);
```
xiaoxiang781216: ditto

netutils/ptpd/ptpd.c
```diff
@@ -1395,52 +1559,218 @@ static int ptp_process_delay_resp(FAR struct ptp_state_s *state,
...
+      correction_time = (((uint64_t)msg->header.correction[0]) << 40)
+                      | (((uint64_t)msg->header.correction[1]) << 32)
+                      | (((uint64_t)msg->header.correction[2]) << 24)
+                      | (((uint64_t)msg->header.correction[3]) << 16)
+                      | (((uint64_t)msg->header.correction[4]) <<  8)
+                      | (((uint64_t)msg->header.correction[5]) <<  0);
+
+      t4_t1_ns = timespec_delta_ns(&state->pdelayresp_rx_time,
+                                   &state->pdelayreq_tx_time);
+      t3_t2_ns = (int64_t)correction_time;
```
xiaoxiang781216, line 1778: remove the cast

netutils/ptpd/ptpd.c
```diff
@@ -1395,52 +1559,218 @@ static int ptp_process_delay_resp(FAR struct ptp_state_s *state,
...
+    {
+      return OK;
+    }
+
+  sequence = ptp_get_sequence(&msg->header);
+  if (sequence != state->pdelay_req_seq)
+    {
+      ptpwarn("Ignoring out-of-sequence Pdelay_Resp_Follow_Up "
+              "(%d vs. expected %d)\n",
+              (int)sequence, (int)state->pdelay_req_seq);
```
xiaoxiang781216: remove too

netutils/ptpd/ptpd.c
```diff
@@ -1076,119 +1079,118 @@ static int ptp_update_local_clock(FAR struct ptp_state_s *state,
...
-          drift_ppb < -CONFIG_CLOCK_ADJTIME_SLEWLIMIT_PPM * 1000)
-        {
-          ptpwarn("Drift estimate out of range: %lld\n",
-                  (long long)drift_ppb);
-          drift_ppb = state->drift_ppb;
-        }
+          if (drift_ppb > slew_limit_ppb || drift_ppb < -slew_limit_ppb)
+            {
+              ptpwarn("Drift estimate out of range: %lld\n",
+                      (long long)drift_ppb);
```
xiaoxiang781216, line 1312: remove the cast

netutils/ptpd/ptpd.c
```diff
@@ -861,7 +861,25 @@ static int ptp_sendmsg(FAR struct ptp_state_s *state, FAR const void *buf,
 
   if (sendts != NULL)
     {
+#ifdef SIOCG_TX_HW_TIMESTAMP
+      struct ifreq req;
```
wenquan2015: better to use SO_TIMESTAMPING TX HARDWEAR with MSG_ERRQUEUE based on https://github.com/apache/nuttx/pull/20161

daniel-p-carvalho: Agreed, SO_TIMESTAMPING + MSG_ERRQUEUE is the right long-term direction here - it matches the discussion on the companion kernel PR (apache/nuttx#20148), where we're holding off migrating SIOCG_TX_HW_TIMESTAMP until apache/nuttx#20161 merges. Once that lands we'll migrate this ioctl to the same mechanism in both PRs for consistency.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Fixed, along with all the other cast-removal / one-line-merge comments below - just pushed. One exception: the `if (state->path_delay_avgcount < CONFIG_NETUTILS_PTPD_DELAYREQ_AVGCOUNT)` merge (the comment on line 1499 in the original diff) would exceed nxstyle's 78-column limit as a single line, so I kept it wrapped as it was. Also rebased onto current master to resolve the merge conflict.

daniel-p-carvalho:
Rebased onto latest `upstream/master` and updated the hardware clock disciplining approach:

1. **POSIX `clock_adjtime` instead of proprietary ioctls:**
   - Removed all custom ioctl calls (`SIOCS_PTP_ADJFREQ`, `SIOCS_PTP_ADJPHASE`, `SIOCG_TX_HW_TIMESTAMP`).
   - When a hardware PTP clock device (e.g. `/dev/ptp0`) is used via `-p /dev/ptp0`, `ptpd` now acts as a PI servo driving the hardware clock's frequency and phase alignment directly via standard POSIX `clock_adjtime(state->clockid, &buf)` with `ADJ_FREQUENCY`.
   - This aligns with the new POSIX PTP hardware clock driver introduced in apache/nuttx#20180.

2. **Clean P2P delay & bugfix series:**
   - Fixed Delay_Resp consumption by `sendmsg()` and implemented canonical IEEE 1588 §11.3 mean path delay calculation.
   - Implemented IEEE 1588 Peer-to-Peer (P2P) delay mechanism (`-P`).
   - Added warning when P2P is used without `CONFIG_SCHED_TICKLESS`.
   - Reset `pdelay_waiting_followup` when initiating a new P2P cycle to prevent stale timestamp pairing.
   - Discard implausible crystal oscillator drift rate spikes before averaging to prevent drift divergence.

All 6 commits have been verified with `checkpatch.sh` (all checks pass) and validated on physical hardware (STM32F4 Discovery with STM32 Ethernet PTP MAC against a Meinberg PTP Grandmaster).

acassis (dismissed):

xiaoxiang781216:
@daniel-p-carvalho please rebase and fix the conflict.

daniel-p-carvalho:
Hi @wenquan1,

Following your request to test PR #3789 on physical hardware, we identified a regression when running on flat builds (`CONFIG_BUILD_FLAT=y`) without a filesystem mounted on `/tmp` (e.g. bare-metal microcontrollers like STM32): `ptpd -t <pid>` timed out with error 110 because the status file in `/tmp` could not be created or read.

To address this cleanly without breaking your improvements for Protected/Kernel modes, we updated this PR with a hybrid approach in commit `35073e5e0` (`netutils/ptpd: retain in-memory IPC for CONFIG_BUILD_FLAT in ptpd_status()`):

- On **`CONFIG_BUILD_FLAT`**: retains the original fast in-memory `sigqueue` + semaphore IPC directly in address space, removing the filesystem requirement for lightweight embedded boards.
- On **`!CONFIG_BUILD_FLAT`** (Protected / Kernel builds): retains your file-based IPC via `/tmp` to safely cross user/kernel address space boundaries.

We have validated this on real hardware (running NuttX flat build). Could you please take a look and share your review?

Thanks!

netutils/ptpd/ptpd.c
```diff
@@ -2191,6 +2246,9 @@ int ptpd_start(FAR const struct ptpd_config_s *config)
       ptp_periodic_send(state);
 
       state->selected_source_valid = is_selected_source_valid(state);
+#ifdef CONFIG_BUILD_FLAT
+      ptp_process_statusreq(state);
+#endif
       ptp_dump_status_file(state);
```
wenquan2015, line 2366: #else
          ptp_dump_status_file(state);
#endif

wenquan2015:
> Hi @wenquan1,
> 
> Following your request to test PR #3789 on physical hardware, we identified a regression when running on flat builds (`CONFIG_BUILD_FLAT=y`) without a filesystem mounted on `/tmp` (e.g. bare-metal microcontrollers like STM32): `ptpd -t <pid>` timed out with error 110 because the status file in `/tmp` could not be created or read.
> 
> To address this cleanly without breaking your improvements for Protected/Kernel modes, we updated this PR with a hybrid approach in commit `35073e5e0` (`netutils/ptpd: retain in-memory IPC for CONFIG_BUILD_FLAT in ptpd_status()`):
> 
> * On **`CONFIG_BUILD_FLAT`**: retains the original fast in-memory `sigqueue` + semaphore IPC directly in address space, removing the filesystem requirement for lightweight embedded boards.
> * On **`!CONFIG_BUILD_FLAT`** (Protected / Kernel builds): retains your file-based IPC via `/tmp` to safely cross user/kernel address space boundaries.
> 
> We have validated this on real hardware (running NuttX flat build). Could you please take a look and share your review?
> 
> Thanks!

Sorry I overlooked this scenario. If the target has a writable filesystem, we can set NETUTILS_PTPD_STATUSFILE to a suitable path and let ptpd write its status there directly. the hybrid patch have some comments to fix.
Thanks!

netutils/ptpd/ptpd.c
```diff
@@ -42,6 +42,9 @@
 #include <sched.h>
 #include <assert.h>
 #include <errno.h>
+#ifdef CONFIG_BUILD_FLAT
```
xiaoxiang781216: remove the check?

netutils/ptpd/ptpd.c
```diff
@@ -67,6 +70,16 @@
  * Private Types
  ****************************************************************************/
 
+#ifdef CONFIG_BUILD_FLAT
+/* Carrier structure for querying PTPD status in flat build mode */
+
+struct ptpd_statusreq_s
+{
+  FAR sem_t *done;
+  FAR struct ptpd_status_s *dest;
```
xiaoxiang781216: remove pointer from done/dest

netutils/ptpd/ptpd.c
```diff
@@ -75,6 +88,9 @@ struct ptp_state_s
 
   bool stop;
   bool dump;                     /* Set by SIGUSR1, checked in main loop */
+#ifdef CONFIG_BUILD_FLAT
+  struct ptpd_statusreq_s status_req;
```
xiaoxiang781216: but add pointer here

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done, `<semaphore.h>` is included unconditionally now.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done. `done` and `dest` are members by value now (`sem_t done; struct ptpd_status_s dest;`). `ptpd_status()` copies `dest` to the buffer of the caller after the semaphore is posted.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done. `status_req` in `struct ptp_state_s` is a pointer now. The signal handler only stores the pointer it gets in `si_value.sival_ptr`, it no longer copies the structure, and `ptp_process_statusreq()` fills the request it points to, clears the pointer and posts the semaphore.

One more change that comes with it: `ptpd` keeps that pointer until it answers, which can be after `ptpd_status()` gave up waiting after one second and returned, so the request cannot be on the stack of the caller. It is a static object now, and a mutex lets only one caller use it at a time. A late answer to a request that timed out is dropped before the next one is sent. This also means that two callers at the same time are served one after the other instead of the second overwriting the first.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done. In flat builds only `ptp_process_statusreq()` is called and in the other builds only `ptp_dump_status_file()`. The dump function and the `dump` flag are built only for the non-flat case, so there is no unused function in flat builds.

daniel-p-carvalho:
Thanks for the review. I have answered each comment in its thread; the changes are in the updated commit `netutils/ptpd: retain in-memory IPC for CONFIG_BUILD_FLAT in ptpd_status()`.

About using the status file on flat builds too: the in-memory request does not need a mounted filesystem, which is the case on the boards where I hit the timeout (bare STM32 without `/tmp`), so flat builds keep it and the two ways are exclusive now, as you suggested in the main loop. Builds that are not flat still use `CONFIG_NETUTILS_PTPD_STATUSFILE`.

I built the flat variant for `stm32f4discovery` and ran `ptpd -t` repeatedly on the board (34 queries in two runs, no failure). The variant that is not flat was only compiled, not run, since the board cannot run it.

netutils/ptpd/ptpd.c
```diff
@@ -181,6 +197,25 @@ struct ptp_state_s
...
+ ****************************************************************************/
+
+#ifdef CONFIG_BUILD_FLAT
+/* The status request of ptpd_status(). The daemon keeps its address until it
+ * answers, which can be after ptpd_status() gave up waiting and returned, so
+ * it lives in static memory and never on the stack of the caller. The lock
+ * lets only one caller use it at a time.
+ */
+
+static struct ptpd_statusreq_s g_statusreq =
```
xiaoxiang781216, line 211: why not change to stack variable and remove g_statusreq_lock

daniel-p-carvalho: The request cannot live on the stack of the caller: `ptpd_status()` gives up after one second if the daemon does not answer and returns, and the daemon keeps the pointer it got in the signal until it answers, which can be later. With a stack variable it would write in a frame that no longer exists. The lock is there because there is one shared request: without it two callers at the same time would overwrite each other.

If you prefer no static object and no lock, I can allocate the request with a reference count, so that the caller and the daemon each drop their reference and the last one frees it. It is more code, so I kept the static one as the smaller change. Tell me which one you prefer and I will change it.

xiaoxiang781216: > The request cannot live on the stack of the caller: `ptpd_status()` gives up after one second if the daemon does not answer and returns, and the daemon keeps the pointer it got in the signal until it answers, which can be later. With a stack variable it would write in a frame that no longer exists. The lock is there because there is one shared request: without it two callers at the same time would overwrite each other.
> 

Thanks for the clear explanation.

> If you prefer no static object and no lock, I can allocate the request with a reference count, so that the caller and the daemon each drop their reference and the last one frees it. It is more code, so I kept the static one as the smaller change. Tell me which one you prefer and I will change it.

look like the origin implementation is better if we need handle the timeout, sorry for the wrong comment.

acassis (approved):

acassis:
@xiaoxiang781216 PTAL

xiaoxiang781216:
> @xiaoxiang781216 PTAL

not address this comment yet: https://github.com/apache/nuttx-apps/pull/3782#discussion_r4069106759

xiaoxiang781216:
@daniel-p-carvalho please fix the conflcit.

daniel-p-carvalho:
This PR is superseded by #3791, now merged: #3791's branch was built on top of this one locally, so all 7 commits here were already included in that squash-merge (`c78e4d69e`). Diffing `upstream/master` against this branch confirms there is nothing left to contribute - every difference is this branch's older code that #3791 already replaced (older `ptpd_status()` IPC, no outlier rejection, no ingress/egress latency compensation, no TX hardware-timestamp failure tracking, `PTP_VERSION_2_0` instead of the corrected `2.1`).

Closing this one.

PR #3786, opened 2026-09-15
fsutils/libtinycbor: fix build on picolibc/newlib

FelipeMdeO:
## Summary

`CONFIG_FSUTILS_TINYCBOR_LIB` fails to build on toolchains with their own `<memory.h>` (picolibc, newlib): `cbortojson.c` does `#include <memory.h>` meaning TinyCBOR's own (where `cbor_malloc`/`cbor_realloc`/`cbor_free` are declared), but only the parent directory was on the include path, so the toolchain's own `<memory.h>` wins instead, leaving those symbols implicitly declared.

## Impact

Only affects `CONFIG_FSUTILS_TINYCBOR_LIB` builds on toolchains that ship their own `<memory.h>` (picolibc, newlib). No effect otherwise.

## Testing

Built clean end to end as part of an out-of-tree application linking `libtinycbor` on `xtensa-esp-elf-gcc` (picolibc). `tools/checkpatch.sh -c -u -m -g` (from the `nuttx` repo, pointed at this repo) passes clean.


fsutils/libtinycbor/Make.defs
```diff
@@ -23,4 +23,9 @@
 ifneq ($(CONFIG_FSUTILS_TINYCBOR_LIB),)
 CONFIGURED_APPS += $(APPDIR)/fsutils/libtinycbor
 CFLAGS += ${INCDIR_PREFIX}$(APPDIR)/fsutils/libtinycbor
```
xiaoxiang781216, line 25: why not remove this line

xiaoxiang781216 (approved):

acassis (approved):

PR #3790, opened 2026-09-17
testing/ostest: accept -ENOENT from work_cancel() for unqueued work

raiden00pl:
## Summary

Since nuttx commit 5a209a853e ("sched/wqueue: restore -ENOENT from work_cancel() for unqueued work") work_cancel() and work_cancel_wq() return -ENOENT when the work is not queued. The tester thread queues work with zero delay, so the worker may already have consumed it by the time it is cancelled and the ASSERT(ret == OK) fires;

Accept -ENOENT there and expect it when cancelling idle work in the API validation test.

## Impact

fix CI

## Testing

CI


xiaoxiang781216 (approved):

xiaoxiang781216:
@acassis please merge this patch to unblock ci.

acassis (approved):

PR #3791, opened 2026-09-19
netutils/ptpd: hardware TX timestamps via MSG_ERRQUEUE and egress latency compensation

daniel-p-carvalho:
> **Depends on apache/nuttx-apps#3782 and on apache/nuttx#20148, which in turn depends on apache/nuttx#20180. Please do not merge this PR before them.**
>
> The first seven commits below are the commits of #3782, and are reviewed there. Until #3782 is merged they show up in this diff; only the last two commits are new here. This PR stays a draft until then, and I will rebase it onto `master` once #3782 is in, so that it is reduced to those two commits. The transmit timestamps it reads only exist with the driver support of apache/nuttx#20148; without it the daemon falls back to software timestamps, as described below.

## Summary

With hardware timestamping, `ptpd` took the transmit time of its event messages (`Sync`, `Delay_Req`, `Pdelay_Req`) from the clock after `sendmsg()` returned. That is later than the frame really left, so a fast peer's reply could appear to arrive before the request had gone, and the peer delay measured with the peer-to-peer mechanism came out negative (about -18 µs on the board I tested with) and was rejected. #3782 says so: on this hardware P2P needs accurate transmit timestamps.

This PR reads the transmit timestamp from the hardware the way Linux does. When hardware timestamping is selected on an `AF_PACKET` socket, `ptpd` requests `SOF_TIMESTAMPING_TX_HARDWARE` for each event message, waits for the looped-back packet on the error queue with `MSG_ERRQUEUE` and takes the timestamp from its `SO_TIMESTAMPING` control message. The driver side of that is apache/nuttx#20148.

### Commits

The first seven are the commits of #3782, unchanged.

8. `Retrieve hardware TX timestamps via MSG_ERRQUEUE.` Timestamps transmitted event messages with the hardware clock instead of reading the clock after `sendmsg()` returns.
   - A transmit socket of its own, separate from the event socket, so the error queue is not shared with received packets.
   - `POLLERR` is handled separately from `POLLIN` in the main loop, and all pending packets are drained on each wakeup.
   - If the timestamp does not arrive, a software timestamp taken before the frame is sent is used. After three consecutive failures the driver is assumed not to provide hardware transmit timestamps, a warning is printed once and only software timestamps are used, so a driver without support does not stall the daemon.
   - The software timestamp is taken before sending in every mode.
   - A measured path delay down to -100 µs is accepted and clamped to zero, since hardware timestamps on both ends can make a short link measure slightly negative.
9. `Add egress latency compensation for TX timestamps.` The frame leaves the MAC later than the moment its hardware transmit timestamp is latched, because of the clock domain crossing and the PHY. That fixed delay is the `egressLatency` port parameter of IEEE 1588. Adds `CONFIG_NETUTILS_PTPD_EGRESS_LATENCY_NS` (default 0, no compensation), the `-O` option to override it at run time and `egress_latency_ns` in `struct ptpd_config_s`. It is the counterpart of the ingress compensation of #3782, and software timestamps are not affected.

## Impact

- New option `-O` and `CONFIG_NETUTILS_PTPD_EGRESS_LATENCY_NS`, default 0. `struct ptpd_config_s` gets `egress_latency_ns`; the only user in the tree is `system/ptpd`, which is updated here.
- Only relevant with hardware timestamping (`-H`). A driver that does not return transmit timestamps makes the daemon switch to software timestamps after three failures.
- The egress compensation value has not been calibrated, so the default 0 applies no correction.

## Testing

Built for `stm32f4discovery:ethraw` with `CONFIG_STM32_ETH_TIMESTAMP_TX=y`, without errors or warnings from `ptpd`. `./tools/checkpatch.sh -g <base>..HEAD` passes.

On hardware: a custom STM32F407 board with the STM32 Ethernet MAC and a DP83848 PHY, over `AF_PACKET` and through a Fast Ethernet switch (`ptpd -2 -s -p /dev/ptp0 -B -H -P -i eth0`), against a GNSS-referenced `ptp4l` Grandmaster using the IEC/IEEE 61850-9-3 profile, with the driver support of apache/nuttx#20148:

- `path_delay` 9265.3 ns with σ = 2.5 ns over a 20 minute run (159 samples, no failed queries), phase offset about -2.2 µs, drift about -79.95 ppm.
- Several 5 minute runs: `path_delay` between 9.26 and 9.33 µs.
- The PPS output of the MAC, measured on an oscilloscope against the Grandmaster's PPS, stayed within about 20 µs.
- The same series without this PR, with software transmit timestamps, gave a negative peer delay (about -18 µs) that was rejected, so `path_delay_ns` stayed at 0 and the phase offset was worse (about -6.7 µs against -2.4 µs).

The end-to-end mechanism (`-E`), master mode, the best master clock algorithm, a 24 hour run and a calibrated `-O` value were not tested.


netutils/ptpd/ptpd.c
```diff
@@ -116,6 +122,13 @@ struct ptp_state_s
 
   int tx_socket;
 
+  /* Hardware TX timestamp retrieval: consecutive failures, and whether it
+   * was given up on because the driver does not provide the timestamps.
+   */
+
+  unsigned int hwts_tx_failures;
```
xiaoxiang781216, line 129: why need dynamically detect whether driver support hw timestamp in a close system? is it enough to check CONFIG_NET_TIMESTAMP

netutils/ptpd/ptpd.c
```diff
@@ -880,11 +899,103 @@ static int ptp_check_multicast_status(FAR struct ptp_state_s *state)
   return OK;
 }
 
+#if defined(CONFIG_NET_TIMESTAMP) && defined(SO_TIMESTAMPING)
```
xiaoxiang781216: ifdef CONFIG_NET_TIMESTAMP

netutils/ptpd/ptpd.c
```diff
@@ -880,11 +899,103 @@ static int ptp_check_multicast_status(FAR struct ptp_state_s *state)
...
+}
+#endif
+
 static int ptp_sendmsg(FAR struct ptp_state_s *state, FAR const void *buf,
                        size_t buflen, FAR const void *addr,
                        socklen_t addrlen, FAR struct timespec *sendts)
 {
   int ret;
+  struct timespec sw_ts;
+#if defined(CONFIG_NET_TIMESTAMP) && defined(SO_TIMESTAMPING)
```
xiaoxiang781216: #ifdef CONFIG_NET_TIMESTAMP

netutils/ptpd/ptpd.c
```diff
@@ -938,9 +1049,48 @@ static int ptp_sendmsg(FAR struct ptp_state_s *state, FAR const void *buf,
       msg.msg_control = NULL;
       msg.msg_controllen = 0;
 
+#if defined(CONFIG_NET_TIMESTAMP) && defined(SO_TIMESTAMPING)
```
xiaoxiang781216: ditto

netutils/ptpd/ptpd.c
```diff
@@ -951,7 +1101,38 @@ static int ptp_sendmsg(FAR struct ptp_state_s *state, FAR const void *buf,
 
   if (sendts != NULL)
     {
-      ptp_gettime(state, sendts);
+#if defined(CONFIG_NET_TIMESTAMP) && defined(SO_TIMESTAMPING)
```
xiaoxiang781216: ditto

netutils/ptpd/ptpd.c
```diff
@@ -2311,15 +2497,47 @@ int ptpd_start(FAR const struct ptpd_config_s *config)
 
       if (pollfds[0].revents)
         {
-          /* Receive time-critical packet, potentially with cmsg
-           * indicating the timestamp.
+#if defined(CONFIG_NET_TIMESTAMP) && defined(SO_TIMESTAMPING)
```
xiaoxiang781216: ditto

netutils/ptpd/ptpd.c
daniel-p-carvalho: The detection is there because `CONFIG_NET_TIMESTAMP` only says that the network stack can carry timestamps, not that the driver returns one for transmitted frames. There is no way to know that at build time: the stack has a flag for drivers that stamp received packets (`NETDEV_RX_STAMP`) but none for transmitted ones, and only the sim driver stamps TX in software. So with `CONFIG_NET_TIMESTAMP=y` and a driver without TX timestamps (for example the STM32 driver without `CONFIG_STM32_ETH_TIMESTAMP_TX`) the error queue stays empty and every event message waits for the whole poll timeout (500 ms) in the main loop. That is what I saw on the board, and the reason for giving up after a few failures. `ptpd` cannot look at the driver options instead, since they belong to the architecture.

You are right that in a closed system the integrator knows which driver is used. The capability itself also does not change while the system runs, so the detection only decides once whether to use the hardware timestamps. The alternatives I see:

1. Keep the detection as it is: three consecutive failures, then a warning and software timestamps.
2. A Kconfig option of `ptpd` that says the driver returns TX timestamps, without any detection. Less code, but a wrong setting brings the waits back and nothing recovers from them.
3. Give up at the first failure. The least code. The price is that a failure that is not permanent would switch hardware timestamps off until restart: the driver can miss a timestamp while working (no IOB for the clone under load, or a frame that ends without the timestamp bit set), and the software timestamp is much less accurate (tens of microseconds on the board I tested).

I kept 1 because it is the safest, but I can change to 2 or 3 if you prefer.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done, the condition is `#ifdef CONFIG_NET_TIMESTAMP` now (`SO_TIMESTAMPING` is always defined). I changed the same condition in the other places of the file too.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done, the condition is `#ifdef CONFIG_NET_TIMESTAMP` now (`SO_TIMESTAMPING` is always defined). I changed the same condition in the other places of the file too.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done, the condition is `#ifdef CONFIG_NET_TIMESTAMP` now (`SO_TIMESTAMPING` is always defined). I changed the same condition in the other places of the file too.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done, the condition is `#ifdef CONFIG_NET_TIMESTAMP` now (`SO_TIMESTAMPING` is always defined). I changed the same condition in the other places of the file too.

netutils/ptpd/ptpd.c
daniel-p-carvalho: Done, the condition is `#ifdef CONFIG_NET_TIMESTAMP` now (`SO_TIMESTAMPING` is always defined). I changed the same condition in the other places of the file too.

acassis (approved):

acassis:
@daniel-p-carvalho is this PR ready or should keep as Draft?

daniel-p-carvalho:
Yes, ready for review. It is out of draft; the dependency note at the top now just says it should not be merged before #3782 (their first seven commits are the same, reviewed there), since apache/nuttx#20148 that it also needed is already merged.

netutils/ptpd/ptpd.c
xiaoxiang781216: another option is controlled hw/sw timestamp through command line option, which is better since:

1. avoid the first several failure
2. support the multiple NIC with the different capability in one device

daniel-p-carvalho: Good point, and it made me look at how Linux itself solves this: `ethtool -T <iface>` / `SIOCETHTOOL` with `ETHTOOL_GET_TS_INFO` lets a driver statically declare `SOF_TIMESTAMPING_TX_HARDWARE` support, so `ptp4l` never has to probe by trial and error - it reads the capability once at startup, the same way it reads `SOF_TIMESTAMPING_RX_HARDWARE` for RX. NuttX already has the RX half of that symmetry (`NETDEV_RX_STAMP` in `d_features`), just not the TX half.

So instead of a command-line option (which still leaves the daemon guessing at runtime, just with an extra knob to get wrong), I'd rather add a `NETDEV_TX_STAMP` flag mirroring `NETDEV_RX_STAMP`, have drivers that support it (starting with the STM32 one this PR exercises) declare it, and have `ptpd` read that once instead of counting failures. That also matches how a *genuine* runtime failure on an interface that already declared support should be handled: not a silent fallback, but a fault - the same way `ptp4l` takes a port to `FAULTY` when a TX timestamp does not show up on the error queue in time, rather than quietly switching timestamping mode underneath.

That is a `nuttx` core change (`d_features` is shared by every driver), not an `nuttx-apps` one, so I would rather not block this PR and #3782 on it - both have been open and reviewed for a while now, and this one has been validated on real hardware overnight without issues. I will open a follow-up PR against `apache/nuttx` for `NETDEV_TX_STAMP`, and a second one here to switch `ptpd` to read it once that lands. Does that sequencing work for you?

xiaoxiang781216 (approved):

netutils/ptpd/ptpd.c
xiaoxiang781216: OK, let's fix in the upcoming patch

PR #3792, opened 2026-09-20
system/uorb: Select the printf extension the format strings need.

Fishwaldo:
## Summary

`UORB_FORMAT` compiles in the per-topic format strings, which reach their data through `%pB`. That is a NuttX printf extension, not a standard conversion, and `UORB_FORMAT` does not select it.

A configuration that enables the listener but not the debug output therefore builds cleanly, runs, and prints the pointer instead of the reading:

```
sensor_voltage(now:4294968998000):0xc0203c98B
```

The address is the `va_format` the extension was meant to expand, and the trailing `B` is the conversion character being taken as ordinary text. Nothing warns, because to the compiler and to printf this is a valid format string that means something else.

`DEBUG_UORB` already selects the extension, so the fault is invisible to anyone who has that on. The two are independent options and only one of them declared what it needed. Move the `select` onto `UORB_FORMAT`, the option that decides whether the format strings exist at all.

## Impact

One line of `system/uorb/Kconfig`. Affects configurations with `UORB_FORMAT=y` and `DEBUG_UORB=n`, which gain `LIBC_PRINT_EXTENSION`. Configurations that already set `DEBUG_UORB` are unchanged, since it selects the extension today.

## Testing

Reproduced on an ESWIN EIC7700 EVB with `UORB_FORMAT=y` and `DEBUG_UORB=n`: `uorb_listener` printed `0xc0203c98B` for every sample. With this change the same build prints the value:

```
sensor_voltage(now:115415014):timestamp:115414993,voltage:12.037500
```


cederom (approved):
Thank you @Fishwaldo :-)

acassis (approved):

xiaoxiang781216 (approved):

linguini1:
Shouldn't we use "depends on"?

PR #3793, opened 2026-09-20
system/sensortest: Add rows for the voltage, current and power topics.

Fishwaldo:
## Summary

sensortest matches a node name against a table that carries each type's structure size, so a type with no row is rejected before the node is ever opened:

```
nsh> sensortest -n 2 voltage0
The sensor node name:voltage0 is invalid
```

voltage, current and power publish like any other topic. Add their rows. All three are a timestamp and a single value, the shape `print_valf` already handles.

The second commit puts a blank line after the declarations in the ten print helpers. `nxstyle` reports an error on every one of them, so the file could not be checked cleanly and any change to it had to sort its own errors out of the existing ones. Whitespace only; `git diff -w` is empty.

## Impact

sensortest only. No existing type changes behaviour. Three types that could not be read at all now can.

## Testing

ESWIN EIC7700 EVB, reading an MPQ8785 regulator and four INA226 rails:

```
nsh> sensortest -n 2 voltage0
voltage0: timestamp:71949895 value:0.72
voltage0: timestamp:72950890 value:0.72
nsh> sensortest -n 1 voltage5
voltage5: timestamp:93633993 value:12.03
nsh> sensortest -n 1 current5
current5: timestamp:94560994 value:0.73
nsh> sensortest -n 1 power5
power5: timestamp:95488993 value:8.79
```

12.03 V x 0.73 A = 8.78 W against the 8.79 W reported, so the values are consistent rather than merely present. `nxstyle system/sensortest/sensortest.c` is clean after the second commit.


cederom (approved):
Thank you @Fishwaldo :-)

acassis (approved):

xiaoxiang781216 (approved):

linguini1 (approved):

